PR #187, opened 2024-10-24
Refactor Remote Querying to Support Older Versions of SQLite

GregHattJr:


aidemsined:
this is likely too outdated and no point trying to get this in

PR #319, opened 2025-01-17
Feature/Update tensor list 

dcblundell:
Adds additional information and reformats existing information.

Screenshot 2025-01-17 at 9 18 55 AM


aidemsined:
we may want to have shape and type parsed as well and close together and at the top
image

src/components/ListItem.tsx
```diff
@@ -51,6 +53,15 @@ const ListItem: React.FC<ListItemProps> = ({
             {/* <GoldenTensorComparisonIndicator value={op.goldenLocal} /> */}
```
aidemsined, line 53: do we still need those?

dcblundell: Time to go! :hocho:

dcblundell:
> we may want to have shape and type parsed as well and close together and at the top image

Added.

aidemsined (approved):

PR #320, opened 2025-01-17
Fix/Local storage persistence

dcblundell:
Now checking session for active reports or profiles and updating the relevant atoms.

Screenshot 2025-01-17 at 11 27 10 AM


src/components/OperationList.tsx
```diff
@@ -307,7 +308,7 @@ const OperationList = () => {
                                         >
                                             <div className='arguments-wrapper'>
                                                 <p className='monospace'>
-                                                    Python execution time: {operation.duration}s
+                                                    Python execution time: {formatSize(operation.duration)} s
```
aidemsined, line 311: could you please confirm correct conversion? does the data come in milli or micro or nano seconds?

aidemsined: we might also want to rename `formatSize` to something more specific

dcblundell: Added math function

aidemsined (approved):

PR #329, opened 2025-01-21
improved tensor information rendering

aidemsined:
image


src/components/operation-details/OperationDetailsComponent.tsx
```diff
@@ -264,7 +264,7 @@ const OperationDetailsComponent: React.FC<OperationDetailsProps> = ({ operationI
                                     </Tooltip>
                                     {tensixIOVisualisationOpen && (
                                         <TensorVisualisationComponent
-                                            title={`Operation ${operationId} inputs/outputs`}
+                                            title={`${operation.id} ${operation.name}  (${operation.operationFileIdentifier}) inputs/outputs`}
```
dcblundell, line 267: Curious about this change because we already have `operation.id` saved as `operationId`.

PR #331, opened 2025-01-21
Improve perf and profiler binding, detect errors and render status

aidemsined:
image

### DIFF
--- a/src/components/Layout.tsx
+++ b/src/components/Layout.tsx
@@ -3,14 +3,16 @@
 // SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
 
 import { Outlet } from 'react-router-dom';
-import { Classes, Tooltip } from '@blueprintjs/core';
+import { Classes, Icon, Tooltip } from '@blueprintjs/core';
 import { Helmet } from 'react-helmet-async';
 import { useAtomValue } from 'jotai';
 import { ToastContainer, cssTransition } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.min.css';
 import classNames from 'classnames';
+import { IconNames } from '@blueprintjs/icons';
 import { activePerformanceTraceAtom, activeReportAtom } from '../store/app';
 import MainNavigation from './MainNavigation';
+import { useGetDeviceOperationListPerf } from '../hooks/useAPI';
 
 const BounceIn = cssTransition({
     enter: `Toastify--animate Toastify__bounce-enter`,
@@ -26,7 +28,8 @@ function Layout() {
     const appVersion = import.meta.env.APP_VERSION;
     const activeReport = useAtomValue(activeReportAtom);
     const activePerformanceTrace = useAtomValue(activePerformanceTraceAtom);
-
+    const useGetDeviceOperationListPerfResult = useGetDeviceOperationListPerf();
+    const isInSync = useGetDeviceOperationListPerfResult.length > 0;
     return (
         <div className={Classes.DARK}>
             <Helmet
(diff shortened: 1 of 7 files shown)


src/components/Layout.tsx
```diff
@@ -71,6 +74,28 @@ function Layout() {
...
                         </Tooltip>
                     )}
+
+                    {activeReport && activePerformanceTrace && (
+                        <span>
+                            {isInSync ? (
+                                <strong>
+                                    <Icon
+                                        icon={IconNames.TickCircle}
+                                        color='#32a467'
```
dcblundell, line 84: Can't we use the Intent colours here and for the error state?

src/components/Layout.tsx
```diff
@@ -71,6 +74,28 @@ function Layout() {
...
+                                    />{' '}
+                                    Profiler and perf reports syncronized
+                                </strong>
+                            ) : (
+                                <strong>
+                                    <Icon
+                                        icon={IconNames.ISSUE}
+                                        color='#fa512e'
+                                    />{' '}
+                                    Profiler and perf reports cant be synchronized
```
dcblundell, line 94: `can't` or `cannot` be synchronised

src/hooks/useAPI.tsx
```diff
@@ -412,15 +426,17 @@ export const useGetDeviceOperationListPerf = () => {
...
             const perfData = df[index];
             if (perfData && perfData['OP CODE'] === deviceOperation.name) {
                 deviceOperation.perfData = df[index];
+            } else {
+                isInvalid = true;
             }
         });
-
-        return deviceOperations;
+        // console.log('isInvalid', isInvalid);
```
dcblundell, line 438: Can remove

src/hooks/useAPI.tsx
```diff
@@ -412,15 +426,17 @@ export const useGetDeviceOperationListPerf = () => {
         }
 
         df = df.filter((r) => !r['OP CODE']?.includes('(torch)') && !(r['OP CODE']?.toString() === ''));
-
+        let isInvalid = false;
         deviceOperations.forEach((deviceOperation, index) => {
```
dcblundell, line 430: Wonder if it makes more sense to use a breakable for loop, as once we find `isInvalid` is true we don't need to look at the remaining data.

src/hooks/useAPI.tsx
```diff
@@ -354,28 +354,42 @@ export const useGetDeviceOperationsListByOp = () => {
...
+                        !op.params.name.includes('ttnn.'),
+                )
+                .map((deviceOperation) => ({
+                    name: deviceOperation.params.name,
+                    id: operation.id,
+                    operationName: operation.name,
+                })),
         );
-    }, [operations]);
+        // return result;
```
dcblundell, line 390: Can remove

src/hooks/useAPI.tsx
```diff
@@ -354,28 +354,42 @@ export const useGetDeviceOperationsListByOp = () => {
...
+        }
+        const result = operations.flatMap((operation) =>
+            operation.device_operations
+                .filter(
+                    (op) =>
+                        op.node_type === NodeType.function_start &&
+                        op.params.name &&
+                        !op.params.name.includes('(torch)') &&
+                        !op.params.name.includes('::') &&
+                        !op.params.name.includes('ttnn.'),
```
dcblundell, line 382: We're using this type of filtering in a few places now. We should use a utility function to cover the param name part.

dcblundell (changes requested):

src/components/Layout.tsx
aidemsined: icons dont support intent

src/scss/_blueprintjs.scss
```diff
@@ -84,3 +84,11 @@ $dark-danger-background-base: $tt-red-accent;
...
     }
 }
+
+.intent-ok {
+    color: $tt-ok;
+}
+
+.intent-not-ok {
+    color: $tt-not-ok;
+}
```
dcblundell, line 94: Award winning naming here

dcblundell (approved):

src/scss/_blueprintjs.scss
aidemsined: very self explanatory, yes?

PR #339, opened 2025-01-27
Feature/add-env-defaults

dcblundell:
Adds default env value capability in our Vite config.

Also adds dotenv support to Python.

### DIFF
--- a/backend/ttnn_visualizer/requirements.txt
+++ b/backend/ttnn_visualizer/requirements.txt
@@ -16,6 +16,7 @@ pandas
 wheel
 build
 PyYAML==6.0.2
+python-dotenv==1.0.1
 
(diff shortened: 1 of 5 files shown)


backend/ttnn_visualizer/requirements.txt
```diff
@@ -16,6 +16,7 @@ pandas
 wheel
 build
 PyYAML==6.0.2
+python-dotenv==1.0.1
```
aidemsined, line 19: do we need this in `.toml`? if this is a dev dependency only maybe move it further down?

dcblundell: Moved

aidemsined (approved):

PR #353, opened 2025-01-31
Feature/range slider

dcblundell:
Screenshot 2025-02-04 at 12 58 16 PM

### DIFF
--- a/src/components/RangeSlider.tsx
+++ b/src/components/RangeSlider.tsx
@@ -1,9 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+
 import { Button, InputGroup, NumberRange, RangeSlider } from '@blueprintjs/core';
 import { useAtom, useAtomValue } from 'jotai';
 import { IconNames } from '@blueprintjs/icons';
 import { useLocation } from 'react-router';
 import { useEffect } from 'react';
-import { operationRangeAtom, selectedRangeAtom } from '../store/app';
+import { operationRangeAtom, performanceRangeAtom, selectedRangeAtom } from '../store/app';
 import ROUTES from '../definitions/Routes';
 import 'styles/components/RangeSlider.scss';
 
@@ -11,7 +15,7 @@ const RANGE_STEP = 25;
 
 function Range() {
     const operationRange = useAtomValue(operationRangeAtom);
-    // const performanceRange = useAtomValue(performanceRangeAtom);
+    const performanceRange = useAtomValue(performanceRangeAtom);
     const [selectedRange, setSelectedRange] = useAtom(selectedRangeAtom);
     const location = useLocation();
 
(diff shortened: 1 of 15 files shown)


src/components/RangeSlider.tsx
```diff
@@ -0,0 +1,74 @@
+import { Button, InputGroup, NumberRange, RangeSlider } from '@blueprintjs/core';
```
aidemsined, line 1: spdx please

src/components/report-selection/LocalFolderSelector.tsx
```diff
@@ -280,4 +297,36 @@ const LocalFolderOptions: FC = () => {
 
 const getReportName = (files: FileList) => files[0].webkitRelativePath.split('/')[0];
 
+// const getPerformanceRange = async (operationRange: NumberRange | null): Promise<NumberRange> => {
```
aidemsined, line 300: do we need this?

src/components/report-selection/LocalFolderSelector.tsx
```diff
@@ -115,6 +126,12 @@ const LocalFolderOptions: FC = () => {
             setSelectedDevice(DEFAULT_DEVICE_ID);
             setActiveReport(fileName);
             createToastNotification('Active report', fileName);
+
+            if (operations) {
```
aidemsined, line 130: i dont like this. this feels like a broken encapsulation. while this works in a pinch useapi should just respond with the calculations you want, not the data that is used to do thith math

src/store/app.ts
```diff
@@ -20,6 +21,10 @@ export const activeToastAtom = atom<number | null>(null);
 export const selectedAddressAtom = atom<number | null>(null);
 export const selectedTensorAtom = atom<number | null>(null);
 
+export const operationRangeAtom = atomWithStorage<NumberRange>('operationRange', [0, 0]);
```
aidemsined, line 24: perisistent across senssion?

aidemsined (changes requested):
comments + notes in slack

src/components/RangeSlider.tsx
dcblundell: Done.

src/components/report-selection/LocalFolderSelector.tsx
dcblundell: Will remove.

src/store/app.ts
dcblundell: Removed persistence.

src/hooks/useAPI.tsx
```diff
@@ -589,11 +618,22 @@ export const useNextBuffer = (address: number | null, consumers: number[], query
...
-    return useQuery({
+export const useBuffers = (bufferType: BufferType, useRange?: boolean) => {
+    const range = useAtomValue(selectedRangeAtom);
+
+    const response = useQuery({
         queryFn: () => fetchAllBuffers(bufferType),
         queryKey: ['fetch-all-buffers', bufferType],
     });
+
+    return useMemo(() => {
```
aidemsined, line 629: i see this block repeated a few times. shoudl we make a function?

src/components/RangeSlider.tsx
```diff
@@ -0,0 +1,130 @@
...
+        </div>
+    ) : null;
+}
+
+const getStepSize = (max: number) => (RANGE_STEP > max ? 1 : max / RANGE_STEP);
+
+const getOperationLabel = (selectedId: number, operations?: OperationDescription[], isTooltip?: boolean): string => {
+    const matchingOperation = operations?.find((operation) => operation.id === selectedId);
+
+    // TODO: Not sure why but the id is not always a round number here, hence the Math.round. Figure out why.
```
aidemsined, line 126: there should really be a correction on a value, not on a label, can we please doublecheck the value is being correct? there is also a stepSize prop we shoudl use

src/hooks/useAPI.tsx
dcblundell: It would be a pain to type it appropriately. All these blocks are altering different API responses.

src/components/RangeSlider.tsx
dcblundell: Changed some of the maths to avoid the non-rounded numbers.

`stepSize` is the size of each movement of the slider, so we don't want that to be anything other than 1.

aidemsined (approved):

PR #382, opened 2025-02-20
feature/use-pnpm

dcblundell:
Switch package manager to [PNPM](https://pnpm.io/).


package.json
```diff
@@ -1,33 +1,40 @@
 {
-  "name": "ttnn-visualzer",
+  "name": "ttnn-visualizer",
```
aidemsined, line 2: ha

PR #383, opened 2025-02-20
Feature/Use deviceBuffers for memory legend

dcblundell:
Screenshot 2025-02-20 at 3 08 46 PM


src/functions/getGroupedMemoryReport.ts
```diff
@@ -0,0 +1,25 @@
+import { BufferData, FragmentationEntry } from '../model/APIData';
+import { BufferType } from '../model/BufferType';
+
+const getGroupedMemoryReport = (data: BufferData[], bufferType: BufferType) => {
+    const groupedMap = new Map<number, FragmentationEntry[]>();
+
+    data.filter((buffer) => buffer.buffer_type === bufferType).forEach((buffer, index) => {
```
aidemsined, line 7: can/should this not be a part of the model?

dcblundell: Moved this.

PR #384, opened 2025-02-24
chip legend markup + initial blackhole support

aidemsined:
image


package-lock.json
```diff
@@ -0,0 +1,14209 @@
+{
```
dcblundell, line 1: Don't want this file at all.

package.json
```diff
@@ -38,6 +38,7 @@
     "normalize.css": "^8.0.1",
     "papaparse": "^5.4.1",
     "plotly.js": "^2.33.0",
+    "pnpn": "^0.0.1",
```
dcblundell, line 41: We don't need this listed as a dependency here.

src/definitions/DeviceArchicture.ts
```diff
@@ -0,0 +1,6 @@
+export enum DeviceArchicture {
```
dcblundell, line 1: We have a lesser version of this enum in APIData. We can probably just use this one but then please replace the use of the APIData one.

dcblundell (changes requested):

src/hooks/useAPI.tsx
```diff
@@ -702,3 +705,53 @@ export const useSession = (reportName: string | null, profileName: string | null
         initialData: null,
     });
 };
+export const useArchitecture = (arch: DeviceArchicture) => {
```
dcblundell, line 708: In Performance.tsx we compute the architecture. Probably makes sense to use this new hook and if you are going to remove the DeviceArchitecture enum as I mentioned in another comment, you need to touch that part of the code as well anyway.

dcblundell (changes requested):

dcblundell (approved):

PR #386, opened 2025-02-24
Fix/npe/congestion map

aidemsined:
image


src/components/npe/NPEViewComponent.tsx
```diff
@@ -212,15 +221,24 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
...
-                    stepSize={1}
-                    labelStepSize={npeData.timestep_data.length > 20 ? npeData.timestep_data.length / 20 : 1}
-                    value={selectedTimestep}
-                    onChange={(value: number) => handleScrubberChange(value)}
+                <div style={{ position: 'relative', width: `${switchwidth}px` }}>
+                    <Slider
+                        min={0}
+                        max={npeData.timestep_data.length - 1}
+                        stepSize={1}
+                        labelStepSize={npeData.timestep_data.length > 25 ? npeData.timestep_data.length / 20 : 1}
```
dcblundell, line 229: I'm a little concerned about the magic numbers here. I think they could benefit from having some named consts just to explain what they are.

src/scss/components/NPEComponent.scss
```diff
@@ -7,6 +7,17 @@
...
+    .bp5-slider-progress {
+        background-color: $tt-black;
+    }
+
+    .bp5-slider-progress.duplicate {
+        left: 0;
+        bottom: 0;
+        position: absolute;
+        z-index: -1;
+    }
```
dcblundell, line 19: Any reason you wrote it like this instead of `&.duplicate`?

dcblundell (changes requested):

dcblundell (approved):

PR #389, opened 2025-02-26
Fix/Updated lockfile

dcblundell:


aidemsined:
what is the purpose of this?

dcblundell:
> what is the purpose of this?

If Slack were working you'd know, but I think it's the cause of the failure to build the wheel with the github action. I ran the action locally with this updated lockfile and it worked.

PR #394, opened 2025-03-04
Feature/Package updates

dcblundell:
Updates package minor/patch versions, and changed various deprecated blueprintjs props to their new equivalents.


aidemsined:
tensor core allocation popup close button lost x color
image

aidemsined (changes requested):
as per comments. only a few small things

aidemsined (approved):

PR #399, opened 2025-03-05
Update README to include installation from pypi instructions

smountenay-tt:
This PR updates the README to have instructions on installing from PyPI.

I removed the previous "Getting Started" section, and replaced it with "Quick Start" at the top, with a link to the getting started page.

I've put more detailed instructions in a new section of the getting started guide. I've left the original instructions on downloading the wheel file from GitHub releases intact, although it may be redundant to have both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ A tool for visualizing the Tenstorrent Neural Network model (TT-NN)
 
 ## Quick Start
 
-TTNN-Visualizer can be installed from PyPI:
+TT-NN Visualizer can be installed from PyPI:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,6 +21,19 @@ A tool for visualizing the Tenstorrent Neural Network model (TT-NN)
 
 </div>
 
+## Quick Start
+
+TTNN-Visualizer can be installed from PyPI:
```
aidemsined, line 26: same

aidemsined (changes requested):

PR #412, opened 2025-03-12
restoring peak memory calculations with correct values per bank

aidemsined:
image


src/hooks/useAPI.tsx
```diff
@@ -740,6 +740,23 @@ export const useArchitecture = (arch: DeviceArchitecture) => {
             throw new Error(`Unknown architecture: ${arch}`);
     }
 };
+
+export const useGetTensorSizesById = (tensorIdList: number[]): { id: number; size: number }[] => {
```
aidemsined, line 744: this should eventually move to BE

dcblundell (approved):

PR #421, opened 2025-03-14
Chore/Remove docker related files and references

dcblundell:


.env.sample
```diff
@@ -9,16 +9,6 @@ DEV_SERVER_PORT=5173
 DEV_SERVER_HOST=localhost
 LAUNCH_BROWSER_ON_START=true
 
-# Docker Configuration
-DOCKER_WEB_CPUS=1
-DOCKER_WEB_MEMORY=1024m
-DOCKER_WEB_HEALTHCHECK_TEST="curl localhost:8000/api/up"
-WEB_CONCURRENCY=1
```
smountenay-tt, line 16: `WEB_CONCURRENCY` is a gunicorn var. It's not docker specific.

smountenay-tt (approved):

.env.sample
dcblundell: Restored.

PR #422, opened 2025-03-17
merge device operations by device

aidemsined:
image


src/components/operation-details/DeviceOperationsFullRender.tsx
```diff
@@ -227,14 +228,52 @@ const DeviceOperationsFullRender: React.FC<{
         return operations;
     }, []);
 
+    const mergeDevices = (operations: Node[]) => {
+        const multiDeviceOps: Node[] = [];
+        const operationsByDevice: Map<string | number | undefined, Node[]> = new Map();
+        let currentDeviceId: string | number | undefined;
```
dcblundell, line 234: `currentDeviceId` gets set to `deviceId` or nothing, and `deviceId` is converted to a number so I don't think this can be a string can it?

dcblundell (approved):

PR #433, opened 2025-03-24
Rename tabs to instances and allow the instance paths to be passed as cli args

smountenay-tt:
This PR does two main things:

* Refactors all the code that refers to "tabs" to use "instance" instead
* Adds a feature for the instance paths to be passed with `--report-path` and `--profiler-path` CLI args

When the `--report-path` or `--profiler-path` options are used when running the `ttnn-visualizer` command, a record will be created in the ttnn db containing the path(s) given as cli args, and when the browser is opened, a new query var will be appended to the URL: `http://localhost:5173?instanceId=8ilh6lky4w`. The `instanceId` is created randomly from Python and put in the db, and then passed in the URL when the browser is opened. The frontend has been modified to look for this query var when loading, and if it is present, it will use that one rather than creating a new random one.

Prior to this PR, the frontend was always responsible for creating the `tabId`. It was created in the frontend as a random string, and passed in the API calls to flask. If no `instanceId` is passed to the React app, the frontend will still create a new one as it did before.

With the refactor of "tabs" to "instances", the `instanceId` here is effectively what used to be the `tabId`.


aidemsined:
@smountenay-tt and @dcblundell please check if DB version needs to be updated

src/libs/axiosInstance.ts
```diff
@@ -9,23 +9,28 @@ const axiosInstance = axios.create({
...
+    return instanceId;
 };
 
 axiosInstance.interceptors.request.use(
     (config) => {
-        const tabId = getOrCreateTabId();
-        if (tabId) {
+        const instanceId = getOrCreateInstanceId();
+        if (instanceId) {
             // Add the tabId to the query params
```
dcblundell, line 30: tabId -> instanceId

dcblundell (approved):

src/libs/axiosInstance.ts
smountenay-tt: I updated the comment

smountenay-tt:
It seemed to work with both tables in the same db during dev, but I later deleted the old `ttnn_*.sqlite` dbs in my backend directory to have a fresh one without the old table.

I think `DefaultConfig.VERSION` should be getting updated on every release, and that will create a new ttnn db for the next release. But I think updating the `DefaultConfig.VERSION` constant is outside the scope of this PR but it should be getting done every release.

PR #437, opened 2025-03-26
Feature/Optimise queries

dcblundell:
Adds an infinite `staletime` to various queries so they don't refetch unless you reupload report data.


src/hooks/useAPI.tsx
```diff
@@ -324,19 +320,14 @@ export const useOperationDetails = (operationId: number | null) => {
...
-        ['get-operation-detail', operationId, deviceId],
-        fetchDetails,
-        {
-            retry: 2,
-            retryDelay: (retryAttempt) => Math.min(retryAttempt * 100, 500),
-        },
-    );
+    const operationDetails = useQuery<OperationDetailsData>(['get-operation-detail', operationId], fetchDetails, {
+        retry: 2,
+        retryDelay: (retryAttempt) => Math.min(retryAttempt * 100, 500),
```
aidemsined, line 328: is this too much of a delay?

aidemsined (approved):
looks good except for that one comment

src/hooks/useAPI.tsx
dcblundell: This wasn't added as part of the PR and it hasn't changed for seven months. Will investigate after this PR.

PR #440, opened 2025-03-26
Update getting started guide to document CLI args

smountenay-tt:
This PR updates the getting started guide to document the new `--report-path` and `--profiler-path` cli args.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -117,6 +117,3 @@ The `ttnn-visualizer` command supports two CLI arguments for passing custom data
 
 * `--report-path` - specify the local path to `db.sqlite` file containing the report
 * `--profiler-path` - specify the local path to the folder containing the profiler data
-
-If either of these CLI args are passed, an instance record in the ttnn db will be created with the path(s)
-given, and the browser will opened with a custom `instanceId` parameter in the URL to load the data from the paths given. 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -110,3 +110,13 @@ Loading data remotely requires you to have SSH access to the relevant machine. Y
 You can have multiple sets of profiler data on the remote paths, but they must be separated into their own folders.
 
 The default behaviour is to sync the files to your local machine, but you may also enable [remote querying](https://github.com/tenstorrent/ttnn-visualizer/blob/main/docs/remote-querying.md) which queries the files directly on the remote machine.
+
+### Custom report and profiler data paths
+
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -110,3 +110,13 @@ Loading data remotely requires you to have SSH access to the relevant machine. Y
...
+
+### Custom report and profiler data paths
+
+The `ttnn-visualizer` command supports two CLI arguments for passing custom data paths:
+
+* `--report-path` - specify the local path to `db.sqlite` file containing the report
+* `--profiler-path` - specify the local path to the folder containing the profiler data
+
+If either of these CLI args are passed, an instance record in the ttnn db will be created with the path(s)
+given, and the browser will opened with a custom `instanceId` parameter in the URL to load the data from the paths given.
```
aidemsined, line 122: im not sure we need implementation details here

aidemsined: specifying how to use should suffice

aidemsined (changes requested):

docs/getting-started.md
smountenay-tt: I removed the "implementation detail". My motivation for adding that was because it's not obvious that anything happened when you use these args, since the UI just loads the same way as it always does. Or there may be cases where someone runs Flask on a server and tries to access from a local machine, and doesn't get why it's not using the paths.

That was my motivation but I have just removed it.

docs/getting-started.md
```diff
@@ -110,3 +110,13 @@ Loading data remotely requires you to have SSH access to the relevant machine. Y
 You can have multiple sets of profiler data on the remote paths, but they must be separated into their own folders.
 
 The default behaviour is to sync the files to your local machine, but you may also enable [remote querying](https://github.com/tenstorrent/ttnn-visualizer/blob/main/docs/remote-querying.md) which queries the files directly on the remote machine.
+
+### Custom report and profiler data paths
+
+The `ttnn-visualizer` command supports two CLI arguments for passing custom data paths:
+
+* `--report-path` - specify the local path to `db.sqlite` file containing the report
```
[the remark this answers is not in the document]
smountenay-tt: The `--report-path` argument directly corresponds to the `report_path` column in the db. Passing the argument on the CLI saves the value in the `report_path` column of the `instances` table (formerly `tab_sessions`). That column takes the path to the `db.sqlite` file so that's what the CLI arg takes. The `config.json` file isn't saved to the db.

PR #453, opened 2025-04-01
Feature/l1 small demarcation

aidemsined:
image
image

### DIFF
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -10,14 +10,13 @@ from http import HTTPStatus
 from pathlib import Path
 from typing import List
 
-import yaml
-from flask import Blueprint, Response, jsonify
+from flask import Blueprint
 from flask import request, current_app
 
 from ttnn_visualizer.csv_queries import DeviceLogProfilerQueries, OpsPerformanceQueries, OpsPerformanceReportQueries
 from ttnn_visualizer.decorators import with_session
-from ttnn_visualizer.exceptions import DataFormatError
 from ttnn_visualizer.enums import ConnectionTestStates
+from ttnn_visualizer.exceptions import DataFormatError
 from ttnn_visualizer.exceptions import RemoteConnectionException
 from ttnn_visualizer.file_uploads import (
     extract_report_name,
@@ -40,7 +39,7 @@ from ttnn_visualizer.serializers import (
     serialize_buffer_pages,
     serialize_operation_buffers,
     serialize_operations_buffers,
-    serialize_devices,
+    serialize_devices, serialize_buffer,
 )
 from ttnn_visualizer.sessions import (
     update_instance,
@@ -306,6 +305,26 @@ def tensor_detail(tensor_id, session: Instance):
         return dataclasses.asdict(tensors[0])
 
 
+@api.route("/buffers", methods=["GET"])
+@with_session
+def get_all_buffers(session: Instance):
+    buffer_type = request.args.get("buffer_type", "")
+    device_id = request.args.get("device_id", None)
+    if buffer_type and str.isdigit(buffer_type):
+        buffer_type = int(buffer_type)
+    else:
+        buffer_type = None
+
+    with DatabaseQueries(session) as db:
+        buffers = list(
+            db.query_buffers(
+                filters={"buffer_type": buffer_type, "device_id": device_id}
+            )
+        )
+        serialized = [serialize_buffer(b) for b in buffers]
+        return jsonify(serialized)
+
+
 @api.route("/operation-buffers", methods=["GET"])
 @with_session
 def get_operations_buffers(session: Instance):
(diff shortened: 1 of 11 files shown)


backend/ttnn_visualizer/views.py
```diff
@@ -306,6 +305,26 @@ def tensor_detail(tensor_id, session: Instance):
         return dataclasses.asdict(tensors[0])
 
 
+@api.route("/buffers", methods=["GET"])
+@with_session
+def get_all_buffers(session: Instance):
+    buffer_type = request.args.get("buffer_type", "")
+    device_id = request.args.get("device_id", None)
+    if buffer_type and str.isdigit(buffer_type):
```
smountenay-tt, line 313: The `str.is_digit` method is a bit wonky, it's got it's own nuances that may not be immediately obvious coming from other languages. For example:

`str.isdigit('²')`

A more Pythonic way would be just to do a try/catch or maybe even regex match.

smountenay-tt: Both ways work, but you could also do:

```
try:
    buffer_type = int(buffer_type)
except:
    pass
````

But it might be pedantic.

**Review by smountenay-tt — APPROVED** [review]

File: src/components/operation-details/MemoryLegendElement.tsx [hunk 2]
```diff
@@ -62,15 +63,15 @@ export const MemoryLegendElement: React.FC<{
...
+                    button: !chunk.empty && chunk.bufferType !== 'L1_SMALL',
                     active: selectedTensorAddress === chunk.address,
                     dimmed: selectedTensorAddress !== null && selectedTensorAddress !== chunk.address,
                     'extra-info': bufferType || layout,
                 },
                 className,
             )}
             // eslint-disable-next-line react/jsx-props-no-spreading
-            {...(!chunk.empty
+            {...(!chunk.empty && chunk.bufferType !== 'L1_SMALL'
```
dcblundell, line 74: We do have a BufferType enum, but I see chunk does not use it. We should tighten up the typing here at some point.

src/components/operation-details/MemoryLegendElement.tsx
```diff
@@ -90,13 +91,23 @@ export const MemoryLegendElement: React.FC<{
...
+                    }),
                 }}
             />
             <div className='format-numbers monospace'>{prettyPrintAddress(chunk.address, memSize)}</div>
             <div className='format-numbers monospace keep-left'>({toHex(chunk.address)})</div>
             <div className='format-numbers monospace nowrap'>
-                {formatSize(chunk.size)}
-                {numCoresLabel}
+                {chunk.bufferType === 'L1_SMALL' ? (
+                    'L1_SMALL region'
```
dcblundell, line 104: I think we can called this `L1 small region`.

src/components/buffer-summary/BufferSummaryPlotRenderer.tsx
```diff
@@ -89,6 +97,10 @@ function BufferSummaryPlotRenderer({ buffersByOperation, tensorListByOperation }
         setHasScrolledToBottom(el.scrollTop + el.offsetHeight >= el.scrollHeight);
     };
 
+    const memoryRegionsMarkers = showMemoryRegions
+        ? [{ color: L1_SMALL_MARKER_COLOR, address: l1SmallMarker, label: 'L1_SMALL' }]
```
dcblundell, line 101: I think we can call the label `L1 small`.

src/components/operation-details/MemoryPlotRenderer.tsx
```diff
@@ -44,6 +46,37 @@ const MemoryPlotRenderer: React.FC<MemoryPlotRendererProps> = ({
...
+            },
+
+            label: {
+                // textangle: 0,
+                // position: 'bottom',
+                text: marker.label,
+                // align: 'start',
+                // placement: 'start',
+                // xanchor: 'right',
+                // yanchor: 'bottom',
```
dcblundell, line 71: Need any of these commented out attributes?

src/components/operation-details/MemoryLegendElement.tsx
```diff
@@ -90,13 +91,23 @@ export const MemoryLegendElement: React.FC<{
                                       ? getTensorColor(chunk.tensorId) || getTensorColor(derivedTensor?.id)
                                       : getBufferColor(chunk.address + (colorVariance || 0)),
                           }),
+                    ...(chunk.bufferType === 'L1_SMALL' && {
+                        // backgroundColor: 'var(--l1-small-color)',
```
dcblundell, line 95: Need this?

dcblundell (changes requested):

src/components/operation-details/MemoryLegendElement.tsx
aidemsined: this is really a silly convention. chunk has CB and L1_SMALL distinction where as BufferType is a DB model mapping. they are unrelated and one shoudl be renamed

src/components/operation-details/MemoryLegendElement.tsx
aidemsined: yes

src/components/buffer-summary/BufferSummaryPlotRenderer.tsx
aidemsined: 👍

PR #462, opened 2025-04-03
Update cluster descriptor api to work with local cluster descriptor files

smountenay-tt:
This PR renames the `/api/cluster_desc` API to `/api/cluster-descriptor`, and updates it to work with local `cluster_descriptor.yaml` files, not just remote ones.


aidemsined (approved):
seems to be working fine in local mode

PR #469, opened 2025-04-07
Feature/Allow selection of uploaded local reports

dcblundell:
- Adds UI to switch between existing reports for local uploads
- Renames reports => profiler in code and endpoints 
- Renames profiler => performance in code and endpoints

Screenshot 2025-04-11 at 3 40 12 PM
Screenshot 2025-04-11 at 3 38 23 PM


src/components/report-selection/RemoteConnectionDialog.tsx
```diff
@@ -31,12 +31,12 @@ const RemoteConnectionDialog: FC<RemoteConnectionDialogProps> = ({
...
         host: '',
         port: 22,
-        reportPath: '',
+        profilerPath: '',
         username: '',
     };
     const defaultConnectionTests: ConnectionStatus[] = [
         { status: ConnectionTestStates.IDLE, message: 'Test connection' },
-        { status: ConnectionTestStates.IDLE, message: 'Test report folder path' },
+        { status: ConnectionTestStates.IDLE, message: 'Test profiler folder path' },
```
aidemsined, line 39: profiler _report_?

aidemsined: as in "Test profiler report folder path" ?

aidemsined (approved):

src/components/report-selection/RemoteConnectionDialog.tsx
dcblundell: Changed as discussed.

README.md
```diff
@@ -95,13 +95,13 @@ You may test the application using the following sample reports.
...
-[report](https://github.com/user-attachments/files/17996493/segformer_encoder.zip)
+[profiler report](https://github.com/user-attachments/files/17996493/segformer_encoder.zip)
 
 **Segformer decoder**
-[report](https://github.com/user-attachments/files/17996491/segformer_decoder_good.zip)
+[profiler](https://github.com/user-attachments/files/17996491/segformer_decoder_good.zip)
 
 **Llama mlp**
-[report + performance trace](https://github.com/user-attachments/files/18770763/llama_attn_32l_10iter_30jan.zip)
+[profiler + performance report](https://github.com/user-attachments/files/18770763/llama_attn_32l_10iter_30jan.zip)
```
smountenay-tt, line 104: I think reports should be plural since the .zip file contains two separate reports, and this wording suggests "profiler + performance report" is one thing.

backend/ttnn_visualizer/app.py
```diff
@@ -154,8 +154,8 @@ def open_browser(host, port, instance_id=None):
 
 def parse_args():
     parser = argparse.ArgumentParser(description="A tool for visualizing the Tenstorrent Neural Network model (TT-NN)")
-    parser.add_argument("--report-path", type=str, help="Specify a report path", default=None)
-    parser.add_argument("--profiler-path", help="Specify a profiler path", default=None)
+    parser.add_argument("--profiler-path", type=str, help="Specify a profiler path", default=None)
+    parser.add_argument("--performance-path", help="Specify a performance path", default=None)
```
smountenay-tt, line 158: It's ok like this, but arguably we should include "report" in these CLI arg names and help text.

`parser.add_argument("--profiler-path", type=str, help="Specify a profiler path", default=None)`

It almost looks like this arg is the path to tracy or something, not the output files.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -544,6 +543,22 @@ def get_all_entries(
...
+    def get_all_folders(directory: str) -> List[str]:
+        """
+        Get a list of all folder names in the specified directory.
+
+        :param directory: Path to the /profiles directory.
+        :return: List of folder names.
+        """
+        try:
+            return [
+                folder.name
```
smountenay-tt, line 555: This list comprehension is looping over paths, not folders. I think it would be more clear if we used the name `path` instead of `folder`, because the `folder` object might actually point to a file.

```
return [
                path.name
                for path in Path(directory).iterdir()
                if path.is_dir()
            ]
```

backend/ttnn_visualizer/models.py
```diff
@@ -194,73 +194,73 @@ class RemoteReportFolder(SerializeableModel):
...
     profiler_path: Optional[str] = None
+    performance_path: Optional[str] = None
     npe_path: Optional[str] = None
-    active_report: Optional[ActiveReport] = None
+    active_report: Optional[ActiveReports] = None
     remote_connection: Optional[RemoteConnection] = None
-    remote_folder: Optional[RemoteReportFolder] = None
-    remote_profile_folder: Optional[RemoteReportFolder] = None
+    remote_profiler_folder: Optional[RemoteReportFolder] = None
+    remote_performance_folder: Optional[RemoteReportFolder] = None
```
smountenay-tt, line 203: I'm not sure why both "path" and "folder" are being used.

I would rather not use the term "folder" in the backend. If we have `performance_path` for the local path, the remote one should be called `remote_performance_path`, not `remote_performance_folder`, IMHO.

PR #470, opened 2025-04-07
Add list endpoint

dcblundell:
Adds a new endpoint to list the folders in `LOCAL_DATA_DIRECTORY`.

### DIFF
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -395,6 +395,30 @@ def get_profiler_performance_data(session: Instance):
         result = csv.get_all_entries(as_dict=True, limit=100)
         return jsonify(result)
 
+@api.route("/profiler/perf-results/list", methods=["GET"])
+@with_session
+def get_profiler_performance_data_list(session: Instance):
+    report_directory = Path(current_app.config["LOCAL_DATA_DIRECTORY"])
+    path = f"{report_directory}/profiles"
+    folder_names = [folder.name for folder in Path(path).iterdir() if folder.is_dir()]
+
+    valid_folders = []
+
+    for folder_name in folder_names:
+        folder_path = Path(path) / folder_name
+        files = list(folder_path.glob("**/*"))
+
+        # Would like to use the existing validate_files function but there's a type difference I'm not sure how to handle
+        if not any(file.name == "profile_log_device.csv" for file in files):
+            continue
+        if not any(file.name == "tracy_profile_log_host.tracy" for file in files):
+            continue
+        if not any(file.name.startswith("ops_perf_results") for file in files):
+            continue
+
+        valid_folders.append(folder_name)
+
+    return jsonify(valid_folders)
 
 @api.route("/profiler/perf-results/raw", methods=["GET"])
 @with_session
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -403,14 +403,14 @@ def get_profiler_performance_data_list():
     else:
         report_directory = Path(current_app.config["LOCAL_DATA_DIRECTORY"])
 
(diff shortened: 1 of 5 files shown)


backend/ttnn_visualizer/views.py
```diff
@@ -395,6 +395,30 @@ def get_profiler_performance_data(session: Instance):
         result = csv.get_all_entries(as_dict=True, limit=100)
         return jsonify(result)
 
+@api.route("/profiler/perf-results/list", methods=["GET"])
+@with_session
+def get_profiler_performance_data_list(session: Instance):
+    report_directory = Path(current_app.config["LOCAL_DATA_DIRECTORY"])
```
dcblundell, line 401: This needs to support remote as well.

smountenay-tt:
I would rather standardize on either "directory" or "folder", but not use both terms in the same function.

```
    path = f"{report_directory}/profiles"
    folder_names = [folder.name for folder in Path(path).iterdir() if folder.is_dir()]
```
In Python and backend in general, "directory" or "dir" are more common. Personally I would suggest standardizing on that.

backend/ttnn_visualizer/views.py
```diff
@@ -395,6 +395,34 @@ def get_profiler_performance_data(session: Instance):
...
 
+@api.route("/profiler/perf-results/list", methods=["POST"])
+def get_profiler_performance_data_list():
+    location = request.json.get("location")
+    if location == "remote":
+        report_directory = Path(current_app.config["REMOTE_DATA_DIRECTORY"])
+    else:
+        report_directory = Path(current_app.config["LOCAL_DATA_DIRECTORY"])
+
+    path = f"{report_directory}/profiles"
```
smountenay-tt, line 406: Since you are using Pathlib, the Pythonic way with a Path option would be to use the `/` operator. Then you wouldn't have to instantiate another Path object below with `Path(path)`.

```suggestion
    path = report_directory / "profiles"
    folder_names = [folder.name for folder in path.iterdir() if folder.is_dir()]
```

backend/ttnn_visualizer/views.py
```diff
@@ -395,6 +395,34 @@ def get_profiler_performance_data(session: Instance):
         result = csv.get_all_entries(as_dict=True, limit=100)
         return jsonify(result)
 
+@api.route("/profiler/perf-results/list", methods=["POST"])
+def get_profiler_performance_data_list():
+    location = request.json.get("location")
+    if location == "remote":
```
smountenay-tt, line 401: An alternative to the if/else block could be:

```suggestion
    config_key = "REMOTE_DATA_DIRECTORY" if location == "remote" else "LOCAL_DATA_DIRECTORY"
    report_directory = Path(current_app.config[config_key])
```

backend/ttnn_visualizer/views.py
dcblundell: Changed.

backend/ttnn_visualizer/views.py
```diff
@@ -395,6 +395,34 @@ def get_profiler_performance_data(session: Instance):
         result = csv.get_all_entries(as_dict=True, limit=100)
         return jsonify(result)
```
smountenay-tt, line 397: Python PEP8 style is to have two blank lines. We need to get linting and/or autoformatting set up.

dcblundell: Added space.

smountenay-tt (approved):

backend/ttnn_visualizer/views.py
aidemsined: not really. remote stuff gets copied over

dcblundell: Directories are different. `/data/local/*` and `data/remote/*.

PR #471, opened 2025-04-08
Add delete endpoint for perf reports

dcblundell:


backend/ttnn_visualizer/csv_queries.py
```diff
@@ -261,7 +261,6 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 class DeviceLogProfilerQueries:
     DEVICE_LOG_FILE = "profile_log_device.csv"
-    LOCAL_PROFILER_DIRECTORY = "profiler"
```
dcblundell, line 264: This didn't seem to be used anywhere.

backend/ttnn_visualizer/utils.py
```diff
@@ -66,10 +66,7 @@ def get_profiler_path(profile_name, current_app, remote_connection=None):
         # Default to local directory if no remote connection is present
         base_dir = local_dir
 
-    if not remote_connection:
-        profile_dir = base_dir / "profiles"
-    else:
-        profile_dir = base_dir / "profiler"
+    profile_dir = base_dir / current_app.config["PROFILER_DIRECTORY_NAME"]
```
smountenay-tt, line 69: Nitpick but the var name could be `profiler_dir` to maintain consistency in terminology.

smountenay-tt:
@dcblundell I would suggest using a more restful pattern and changing `perf-results/list` to just `/perf-results` and the `POST perf-results/delete` to `DELETE /perf-results/{id}`.

backend/ttnn_visualizer/views.py
```diff
@@ -396,12 +397,45 @@ def get_profiler_performance_data(session: Instance):
...
 
 
-@api.route("/profiler/perf-results/list", methods=["POST"])
-def get_profiler_performance_data_list():
-    location = request.json.get("location")
-    config_key = "REMOTE_DATA_DIRECTORY" if location == "remote" else "LOCAL_DATA_DIRECTORY"
+@api.route("/profiler/perf-results/delete", methods=["POST"])
+@with_session
+def delete_performance_report(session: Instance):
+    is_remote = True if session.remote_connection else False
```
smountenay-tt, line 403:
```suggestion
    is_remote = bool(session.remote_connection)
```

backend/ttnn_visualizer/views.py
```diff
@@ -396,12 +397,45 @@ def get_profiler_performance_data(session: Instance):
...
+        path = report_directory / connection.host / current_app.config["PROFILER_DIRECTORY_NAME"]
+    else:
+        path = report_directory / current_app.config["PROFILER_DIRECTORY_NAME"] / report
+
+    if path.exists() and path.is_dir():
+        shutil.rmtree(path)
+    else:
+        return Response(status=HTTPStatus.NOT_FOUND, response=f"Directory does not exist: {path}")
+
+    return Response(status=HTTPStatus.OK, response=f"Directory deleted successfully: {path}")
```
smountenay-tt, line 423: If you do change it to use HTTP DELETE instead of POST then a restful response code would be 204 with no response message like this necessary. But I guess this project isn't doing strict rest.

backend/ttnn_visualizer/views.py
aidemsined: i find original more readable tbh

dcblundell: I'm gonna go with the Python adult suggestion here.

backend/ttnn_visualizer/views.py
dcblundell: This project hasn't been able to do strict much of anything due to its nature, but we try. I'm going to go with this idea.

PR #472, opened 2025-04-08
NPE playback speed options

aidemsined:
updates as per user feedback
image

### DIFF
--- a/src/components/npe/NPEViewComponent.tsx
+++ b/src/components/npe/NPEViewComponent.tsx
@@ -134,7 +134,7 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
         return groups;
     }, [selectedTransferList, selectedNode]);
 
-    const startAnimation = () => {
+    const startAnimation = (speed: number = 1) => {
         setIsPlaying(true);
         clearInterval(animationInterval as number);
         const range = npeData.timestep_data.length;
@@ -143,7 +143,7 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
             setSelectedTimestep((prev) => {
                 return prev < range - 1 ? prev + 1 : 0;
             });
-        }, 100);
+        }, 100 / speed);
         setAnimationInterval(interval as unknown as number);
     };
     const stopAnimation = () => {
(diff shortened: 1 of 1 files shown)


dcblundell:
There's a duplicate key warning when I push play.
Screenshot 2025-04-09 at 9 09 47 AM

src/components/npe/NPEViewComponent.tsx
```diff
@@ -154,9 +154,30 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
     const onPlay = () => {
         startAnimation();
     };
+    const onPlay2x = () => {
+        startAnimation(2);
+    };
```
dcblundell, line 159: From a UX perspective, I'd like to have a "slow down" button. If you hit play2x you can then only stop the playback, you can't go back down to 1x without stopping and starting again.

dcblundell: I'd like it to be more clear when 2x is engaged. You can see that the timeline is playing faster but there's no other visual indication. Maybe we can highlight the play and 2x buttons when they're active with a style, or use `Intent.Primary` or something.

dcblundell (changes requested):

dcblundell (approved):
Approving with duplicate key conflict.

src/components/npe/NPEViewComponent.tsx
```diff
@@ -134,29 +136,52 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
         return groups;
     }, [selectedTransferList, selectedNode]);
 
-    const startAnimation = () => {
-        setIsPlaying(true);
+    const startAnimation = (speed: number = PLAYBACK_SPEED) => {
+        setPlaybackSpeed(speed);
+        // setIsPlaying(true);
```
dcblundell, line 141: Remove.

src/components/npe/NPEViewComponent.tsx
```diff
@@ -134,29 +136,52 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
...
                 return prev < range - 1 ? prev + 1 : 0;
             });
-        }, 100);
+        }, 100 / speed);
         setAnimationInterval(interval as unknown as number);
     };
     const stopAnimation = () => {
-        setIsPlaying(false);
+        setPlaybackSpeed(0);
+        // setIsPlaying(false);
```
dcblundell, line 154: Here.

dcblundell (approved):

PR #487, opened 2025-04-16
Preliminary cluster file loading, updates to routing and navigaiton. Routed modal support

aidemsined:


src/components/cluster/ClusterRenderer.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return (
+        <div
+            style={{
+                position: 'absolute',
+                top: '80px',
+                backgroundColor: '#000000aa',
+                width: '100%',
+                bottom: 0,
+                zIndex: 10,
+            }}
```
dcblundell, line 24: Boo inline

src/libs/ModalAwareOutlet.tsx
```diff
@@ -0,0 +1,9 @@
+import { RouteObject, useLocation, useRoutes } from 'react-router-dom';
+
+export function ModalAwareOutlet({ routes }: { routes: RouteObject[] }) {
+    const location = useLocation();
+    const state = location.state as { background?: Location };
```
dcblundell, line 5: Seems like this should be a reusable type? It's used in two places. I'm surprised React Query doesn't have one.

src/hooks/useAPI.tsx
```diff
@@ -342,6 +342,21 @@ const fetchDeviceLogRaw = async (): Promise<FetchDeviceLogRawResult> => {
...
+    return data;
+};
+
+export const useGetClusterDescription = () => {
+    const activeProfilerReport = useAtomValue(activeProfilerReportAtom);
+
+    return useQuery({
+        queryFn: () => fetchClusterDescription(),
+        queryKey: ['get-cluster-description', activeProfilerReport],
+        initialData: null,
```
dcblundell, line 356: Could add `staleTime: Infinity` here because unless the queryKey changes the returned data should be the same.

dcblundell (changes requested):

src/components/MainNavigation.tsx
```diff
@@ -93,21 +100,44 @@ function MainNavigation() {
...
                     icon={IconNames.Random}
                     variant='minimal'
                     size='large'
                     className='npe-button'
                 >
                     <small>beta</small>
                 </Button>
+                <Button
+                    text='Topology'
+                    onClick={() => handleOpenModal(ROUTES.CLUSTER)}
```
dcblundell, line 113: Should this toggle the modal as well?

dcblundell (changes requested):

src/components/cluster/ClusterRenderer.tsx
aidemsined: yes

src/components/MainNavigation.tsx
aidemsined: arguable. its a navigation. can it a hve togle buttons? niot sure

src/libs/ModalAwareOutlet.tsx
aidemsined: yeah, not quite sure what to call it

PR #499, opened 2025-05-01
Feature/Apply ranges to performance charts

dcblundell:
Screenshot 2025-05-01 at 5 00 51 PM


aidemsined:
when loading a comparison report the range while disabled is still showing reduced which doesnt match the charts rendered

image

dcblundell:
> when loading a comparison report the range while disabled is still showing reduced which doesnt match the charts rendered

Have fixed this now.

PR #501, opened 2025-05-05
Fix/Grouped bar data is cut off

dcblundell:
Due to the way plotly plots grouped data it's possible for some data to be cut off by the current logic determining x-axis length.

Added an offset to avoid data being cut off when needed.

### DIFF
--- a/src/functions/getMaxArrayLength.ts
+++ b/src/functions/getMaxArrayLength.ts
@@ -1,8 +1,0 @@
-// SPDX-License-Identifier: Apache-2.0
-//
-// SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
-
-export default function getMaxArrayLength(arrays: Array<unknown[]>, comparisonReport: string | null): number {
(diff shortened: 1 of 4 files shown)


src/functions/getMaxArrayLength.ts
```diff
@@ -2,6 +2,7 @@
 //
 // SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
 
-export default function getMaxArrayLength(arrays: Array<unknown[]>): number {
-    return Math.max(...arrays.map((arr) => arr.length), 0);
+export default function getMaxArrayLength(arrays: Array<unknown[]>, comparisonReport: string | null): number {
```
aidemsined, line 5: i dont see why this is called `getMaxArrayLength` and why comparison report is a string

dcblundell: It's comparing n arrays and finding the one with the biggest size. 

`comparisonReport` here should probably be a boolean, so I'll change that.

aidemsined (approved):

PR #504, opened 2025-05-06
Feature/Adds scroll position tracking

dcblundell:
This can probably be turned into a hook at some point if we start adding it to other lists.

Screenshot 2025-05-06 at 10 29 47 AM


aidemsined (approved):
it works reasonably well most of the time, i think this is definitely an improvement and is better than resetting the position

PR #508, opened 2025-05-09
fix: support perf result file name without datetime suffix

sott0n:
The perf result file name was previously assumed to follow the format `ops_perf_results_[datetime].csv`. However, files generated by `ttrt perf` use the format `ops_perf_results.csv`, without the `_[datetime]` suffix, causing glob pattern matching to fail.

This commit adds support for the `ops_perf_results.csv` format by including an additional glob pattern, ensuring compatibility with both formats.


smountenay-tt (approved):

dcblundell:
Thanks for the contribution!

PR #519, opened 2025-05-14
tensor deallocation reporting part 1

aidemsined:
image

### DIFF
--- a/src/components/buffer-summary/BufferSummaryPlotRenderer.tsx
+++ b/src/components/buffer-summary/BufferSummaryPlotRenderer.tsx
@@ -53,7 +53,7 @@ function BufferSummaryPlotRenderer({ buffersByOperation, tensorListByOperation }
     const [hasScrolledFromTop, setHasScrolledFromTop] = useState(false);
     const [hasScrolledToBottom, setHasScrolledToBottom] = useState(false);
     const [activeRow, setActiveRow] = useState<number | null>(null);
-    const [showDeallocationReport, setShowDeallocationReport] = useState(false);
+    const [showDeallocationReport, setShowDeallocationReport] = useState(true);
     const [showHex, setShowHex] = useAtom(showHexAtom);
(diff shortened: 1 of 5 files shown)


src/components/buffer-summary/BufferSummaryPlotRenderer.tsx
```diff
@@ -138,6 +166,13 @@ function BufferSummaryPlotRenderer({ buffersByOperation, tensorListByOperation }
                         setIsZoomedIn(!isZoomedIn);
                     }}
                 />
+                <Switch
+                    label='Mark tensors that are not deallocated'
```
dcblundell, line 170: Suggest something like "Flag tensors with delayed deallocation" or "Highlight delayed tensor deallocation" maybe?

src/components/buffer-summary/BufferSummaryPlotRendererDRAM.tsx
```diff
@@ -191,6 +191,7 @@ function BufferSummaryPlotRendererDRAM({
                                                 memoryEnd={isZoomedIn ? zoomedMemoryOptions[index].end : MEMORY_SIZE}
                                                 memoryPadding={zoomedMemoryOptions[index].padding}
                                                 tensorList={tensorListByOperation.get(operation.id)!}
+                                                tensorDeallocationReport={[]}
```
dcblundell, line 194: Shouldn't we make it optional, or is this something we could support in the future?

dcblundell (commented):
I'm not sure about the UI change itself, but the feature is cool.

src/components/buffer-summary/BufferSummaryPlotRenderer.tsx
dcblundell: Going with "Mark late tensor deallocations".

dcblundell (approved):

PR #521, opened 2025-05-16
Feature/Restore memory overlay to buffers

dcblundell:
Feature was never reimplemented when we moved to canvas for buffer rendering.

Screenshot 2025-05-16 at 9 23 10 AM

Screenshot 2025-05-16 at 9 22 55 AM


src/components/buffer-summary/BufferSummaryRow.tsx
```diff
@@ -239,4 +251,90 @@ const BufferSummaryRow = ({
     );
 };
 
+const FG_COLOUR = 'rgba(0, 0, 0, 0.7)';
+
+function getBackgroundPattern(ctx, layout, position, size): string | null {
+    const pattern: string | null = null;
```
aidemsined, line 257: this entire block of code can be and maybe even should be optimized and cleaned up

aidemsined (changes requested):
so tweaks needed

src/components/buffer-summary/BufferSummaryRow.tsx
```diff
@@ -239,4 +251,90 @@ const BufferSummaryRow = ({
     );
 };
 
+const FG_COLOUR = 'rgba(0, 0, 0, 0.7)';
+
+function getBackgroundPattern(ctx, layout, position, size): string | null {
```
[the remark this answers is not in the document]
dcblundell: Added.

src/components/buffer-summary/BufferSummaryRow.tsx
dcblundell: Moved to its own file and optimised.

aidemsined (approved):

PR #529, opened 2025-05-21
Feature/Support disabling remote sync

dcblundell:
To test, set your .env to include `SERVER_MODE=true` or change the default value for SERVER_MODE in `settings.py`.

Screenshot 2025-05-21 at 3 28 30 PM


vite.config.ts
```diff
@@ -23,6 +23,7 @@ export default defineConfig(({ mode }) => {
         define: {
             'import.meta.env.APP_VERSION': JSON.stringify(version),
             'import.meta.env.VITE_API_ROOT': JSON.stringify(env.VITE_API_ROOT) ?? '"http://localhost:8000/api"',
+            'import.meta.env.VITE_DISABLE_REMOTE_SYNC': JSON.stringify(env.VITE_DISABLE_REMOTE_SYNC) ?? null,
```
aidemsined, line 26: im not thrilled about feature specific flags. we will have a plethora of things that will need to work differently between desktop and online

aidemsined (changes requested):

vite.config.ts
dcblundell: Renamed to `VITE_SERVER_MODE`.

aidemsined (approved):

PR #552, opened 2025-06-11
Fix/update dependency licenses

aidemsined:


dcblundell (approved):
I trust this is correct.

PR #554, opened 2025-06-12
Feature/Upgrade normalisation process

dcblundell:
Supports 'n' comparison reports.


src/components/performance/PerfReport.tsx
```diff
@@ -345,4 +381,29 @@ const enrichRowData = (rows: PerfTableRow[], opIdsMap: { perfId?: string; opId:
     });
 };
 
+const getTotalDataLength = (
+    useNormalisedData: boolean,
```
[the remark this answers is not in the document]
dcblundell: Simplified it.

src/functions/normalisePerformanceData.ts
```diff
@@ -32,96 +32,96 @@ const PLACEHOLDER: TypedPerfTableRow = {
...
+    const otherIndexes = new Array(otherDatasets.length).fill(0);
+
+    while (refIndex < refData.length) {
+        const refRow = refData[refIndex];
+        const refOp = refRow.raw_op_code;
+        let allMatched = true;
+
+        aligned[0].push({ ...refRow });
+
+        for (let d = 0; d < otherDatasets.length; d++) {
```
aidemsined, line 57: i really dislike the use of j, k, d and such for the itterators. is there a chance they can be a little descriptive?

src/functions/normalisePerformanceData.ts
```diff
@@ -32,96 +32,96 @@ const PLACEHOLDER: TypedPerfTableRow = {
...
+        const ratio = missingCounts[d] / maxLen;
+        if (ratio > maxMissingRatio) {
+            aligned[d + 1] = [];
+        }
+    }
 
-    return [result1, result2];
+    return {
+        data: aligned,
+        missingRows: Array.from(missingRows.values()),
```
aidemsined, line 123: this could be a nitpick but on larger datasets or frequent updates spread [...] will perform faster and have less overhead, this is where it might be important

PR #555, opened 2025-06-13
Feature/npe performance updates

aidemsined:


src/components/npe/NPEViewComponent.tsx
```diff
@@ -55,6 +54,14 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
     const [isAnnotatingCores, setIsAnnotatingCores] = useState<boolean>(true);
     const [nocFilter, setNocFilter] = useState<NoCType | null>(null);
 
+    const links = useMemo(() => {
+        return npeData.timestep_data[selectedTimestep];
+    }, [npeData, selectedTimestep]);
+
+    const transfers = useMemo(() => {
+        return npeData.noc_transfers.filter((tr) => links?.active_transfers.includes(tr.id));
+    }, [npeData.noc_transfers, links]);
```
dcblundell, line 63: Just noticed that the first useMemo has `npeData` as a dependency, but the second one has `npeData.noc_transfers`. Can probably either both use `npeData` or both reference the specific objects.

src/components/npe/useNPEHandlers.tsx
```diff
@@ -0,0 +1,154 @@
...
+                setSelectedTransferList([]);
+                setSelectedNode(null);
+                return;
+            }
+
+            if (selectedNode?.index === index) {
+                setSelectedNode(null);
+                setSelectedTransferList([]);
+                return;
+            }
```
dcblundell, line 42: Could combine these if conditions.

dcblundell (changes requested):

PR #557, opened 2025-06-13
Feature/Disable compare for server mode

dcblundell:
Screenshot 2025-06-13 at 3 57 44 PM


src/components/report-selection/LocalFolderPicker.tsx
```diff
@@ -39,7 +39,7 @@ const LocalFolderPicker = ({
         return (
             <div
                 className='folder-picker-menu-item'
-                key={folder.reportName}
+                key={folder.path}
```
dcblundell, line 42: Changed to avoid key duplication issue in server mode.

PR #561, opened 2025-06-17
Feature/Disable report deletion on server

dcblundell:
Disables the UI and adds safe guard in the endpoint.

Screenshot 2025-06-17 at 10 59 19 AM

### DIFF
--- a/backend/ttnn_visualizer/decorators.py
+++ b/backend/ttnn_visualizer/decorators.py
@@ -117,12 +117,14 @@ def remote_exception_handler(func):
     return remote_handler
 
 
-def with_disable_deletion_in_server_mode(f):
+def local_only(f):
     from flask import current_app
 
     @wraps(f)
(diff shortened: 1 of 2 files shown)


backend/ttnn_visualizer/decorators.py
```diff
@@ -115,3 +115,14 @@ def remote_handler(*args, **kwargs):
             )
 
     return remote_handler
+
+
+def with_disable_deletion_in_server_mode(f):
```
smountenay-tt, line 120: I'm not sure that a decorator is necessary for this. If we keep it though, I'd suggest making the name and description more generic, so it can be used on other views also. There's likely other views we don't want accessible in server mode also, but not necessarily deleting a report.

I'd probably choose a short and concise name like `@local_only`.

dcblundell: Good with me 👍

PR #582, opened 2025-06-20
Prefix uploaded filesystem paths

smountenay-tt:
This PR modifies the app to prefix uploaded files with a Unix timestamp before saving them. This will happen when the app is running in server mode, not when running the app locally. The reason for prefixing the filenames is to prevent uploads from different users, with the same report name, from overwriting each other.

When running locally it will continue to overwrite any existing report with the same filenames, but on the server it will create a new directory for each upload.

It modifies how uploaded reports are tracked in the user's session. Prior to this PR, we tracked what instance ids belonged to the user, in their session. With this PR, it's being modified to also track every file uploaded by the user, and records the path in the session. Only paths in the session vars will be available in the select report dropdowns.

### DIFF
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -600,13 +600,10 @@ def get_performance_results_report(instance: Instance):
 
     name = request.args.get("name", None)
 
-    if name:
-        if current_app.config["SERVER_MODE"]:
-            return Response(status=HTTPStatus.FORBIDDEN)
-        else:
-            performance_path = Path(instance.performance_path).parent / name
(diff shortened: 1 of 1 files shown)


backend/ttnn_visualizer/views.py
```diff
@@ -583,13 +601,6 @@ def get_performance_results_report(instance: Instance):
     if not instance.performance_path:
         return Response(status=HTTPStatus.NOT_FOUND)
 
-    name = request.args.get("name", None)
```
dcblundell, line 586: This is needed.

PR #585, opened 2025-06-23
Feature/npe demo reports

aidemsined:
this feature is only available in a `server mode`
image


dcblundell (approved):
Code looks good, would just suggest that we don't hide the dropdown if you upload your own data and we move the dropdown to be in the same row as the upload.

PR #586, opened 2025-06-23
Feature/Cluster description hook call

dcblundell:
Moved hook call to RangeSlider and removed retrying to avoid multiple 404 calls.


aidemsined:
other then retry false im not clear of the purpose of this. what problem are we solving?

aidemsined (approved):

PR #598, opened 2025-06-24
Fix setting report paths when switching instances

smountenay-tt:
This PR fixes a bug that was recently introduced with the changes to support running the app on the server, where the report paths did not get updated correctly in the DB when updating the reports with the dropdowns.


dcblundell (approved):
Looks fixed to me.

PR #630, opened 2025-07-11
Gracefully handle SSH authentication errors

### LINKED ISSUE #629: Gracefully handle error when SSH fails due to password being required
The remote sync feature currently works with SSH keys only. There is no form field or anywhere else to put in the SSH password. It's required for the user to have their public key in the `authorized_keys` file on the server. When testing the SSH connection, however, the user is now being prompted to enter the password in the Flask console, because of the refactoring to use CLI ssh tools instead of Paramiko. We need to make sure that the user is not prompted for a password and ideally that a more graceful error message is provided to the user.

Even if the user were to enter the password in the terminal and it connects for the test, it would not work on subsequent requests. This is not supported.

smountenay-tt:
After the refactor from Paramiko to CLI ssh, there was a traceback showing up in the Flask console when there was an SSH authentication error. Additionally, it was prompting the user in the Flask terminal to enter the password. This was happening due to the Python back calling `ssh` with the `subprocess` module, and since the `ssh` command prompts for a password by default if there is not a valid key set up, this prompt from `ssh` was getting passed through to the Flask terminal.

This PR cleans this up so that every invocation of `ssh` or `sftp` commands passes an option to never prompt for a password, it replaces the traceback and instead shows a single line error message in the Flask logs, and instead of a 500 error it returns 422, with a json error message.

[Closes #629]


dcblundell (approved):
Seems good.

PR #641, opened 2025-07-16
Feat/Add responsive styles to main nav

dcblundell:
Adds some responsive styles to the main nav which trigger at two breakpoints.

Screenshot 2025-07-16 at 1 01 15 PM

Screenshot 2025-07-16 at 1 01 21 PM

https://github.com/tenstorrent/ttnn-visualizer/issues/659


src/scss/components/MainNavigation.scss
```diff
@@ -113,4 +114,29 @@
         transition: transform 300ms ease-out 50ms;
         gap: 7px;
     }
+
+    @media screen and (width <= 1300px) {
```
aidemsined, line 118: not a fan of random numbers

dcblundell: Made variables.

PR #684, opened 2025-07-25
Update Node.js version to 22.1.0 across all workflows

aidemsined:
Bump Node.js version from v20.11.1 to v22.1.0 in .nvmrc and GitHub Actions workflows. Regenerate package-lock.json and update pnpm-lock.yaml to reflect dependency changes for the new Node.js version.


package.json
```diff
@@ -47,6 +47,7 @@
     "normalize.css": "^8.0.1",
     "papaparse": "^5.5.3",
     "plotly.js": "^2.35.3",
+    "pnpm": "^10.13.1",
```
dcblundell, line 50: This shouldn't be a package.

dcblundell (changes requested):

Issue #628, opened 2025-07-09
Remote performance path being set on instance before sync


smountenay-tt:
After creating a new remote sync server, fetching the remote folders list, then syncing the memory report, the frontend updates the "instance" with both profile report and performance report paths - even though only memory report had been synced. This leads to numerous errors showing up in the Flask logs, and failing API requests in browser dev tools, due to the performance report not existing.

## Steps to reproduce:

* Delete the data/remote/localhost directory and open the app in a new incognito window, to start fresh
* Click the "Add New Connection" button and put in valid SSH credentials and report paths
* Click the "Fetch remote folders list" button
* Observe that both the memory report and performance report dropdowns show one or more folders
* Press the sync icon for the selected memory report
* Observe that it makes an API call to the sync API, which only contains details for the memory report
* Observe that the next API call is a POST to /api/remote/use with the details for both the memory report and perf report, causing these values to get saved to the respective `profiler_path` and `memory_path` columns in the `instances` table in the db
* Observe that there are errors in the Flask console, and that the performance tab in the UI becomes active, before any perf report has been synced

## Expected Behaviour

The POST to `/api/remote/use` should not contain the perf report details if that folder has not been synced, and viceversa. If the sync button is pressed for memory reports, it should not be sending/changing the perf report details, and if the perf report is changed, the frontend shouldn't be sending the profile report details.

There's other ways this could break, not just on first run. For example, if the user changed the dropdown value without syncing. The call to /api/remote/use would presumably send the value for whatever was currently selected in the other dropdown, even if that folder has not (ever) been synced yet.

PR #693, opened 2025-07-29
Allow syncing a remote performance report without a profiler report

smountenay-tt:
This PR resolves an issue where 500 errors were being returned when you put in the details for a report SSH connection, but tried to sync the performance report before the profiler report.

The remote sync wasn't set up to allow syncing performance report independent of profiler report. After fetching the remote folder list, if you pressed sync for the performance report, it would download the remote files for just the performance report, but the post to `/api/remote/use` didn't handle this scenario right. First of all the frontend was sending the profiler path in the post to `/api/remote/use` even though only the performance path had been synced by the user. The frontend is being updated to not do that anymore.

With the frontend sending the profiler path even though it hadn't been synced, the `/api/remote/use` code would save the path to the `instance` record in the db, and this would activate the corresponding tabs in the UI, but since the report had not been synced and saved locally to the path in the `instance.profiler_path` field in the db, all of the other APIs would return 500 errors, since the path did not exist locally.

I have updated the backend for `/api/remote/use` to independently save the profiler fields and the performance fields. They are only saved if they were sent in the request, and made it not required in the backend to have the profiler data sent.

Additionally, we are renaming the fields sent in the post to `/api/remote/use` and `/api/remote/sync`. The field names it was using, `folder` to refer to what is otherwise called the "profiler report", and `profile` to refer to what is otherwise called the performance report, was super confusing.

I am also removing code in the view for `/api/remote/use` that made a `connection_directory` var, had error handling to check that it existed, then constructed a `report_path` var, because all it was used for was a log statement. Way to much code for a not meaningfully useful log statement.

Closes [#628](https://github.com/tenstorrent/ttnn-visualizer/issues/628)

### DIFF
--- a/backend/ttnn_visualizer/app.py
+++ b/backend/ttnn_visualizer/app.py
@@ -229,8 +229,6 @@ def main():
 
     gunicorn_args = [
         "gunicorn",
-        "-t",
-        "300",
         "-k",
         config.GUNICORN_WORKER_CLASS,
         "-w",
--- a/backend/ttnn_visualizer/app.py
+++ b/backend/ttnn_visualizer/app.py
@@ -229,6 +229,8 @@ def main():
 
     gunicorn_args = [
         "gunicorn",
+        "-t",
+        "300",
         "-k",
         config.GUNICORN_WORKER_CLASS,
         "-w",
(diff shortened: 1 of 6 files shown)


smountenay-tt:
I have pushed two more changes to this PR:

* Modify the `/api/performance` view to check if the performance-reports path exists before trying to iterate over the folders, and if it doesn't, return an empty list. It had code which created the `data/remote/{host}/performance-reports` directory if it didn't exist, but it only did this, specifically, if it was not a remote report.
* Change the `/api/performance/perf-results/report` API to return a `400 Bad Request` with the response body "No performance data found for instance." if you do a GET to that API when there is no `performance_path` stored in the db for that instance. Previously it returned `404 Not Found` with no error message in this case.

dcblundell (approved):
Looking good 👍

PR #702, opened 2025-07-30
Show SSH error detail when testing the connection

### LINKED ISSUE #703: Show raw ssh errors in add remote connection dialog on failure
Show the raw ssh error returned from the ssh command in the add remote connection dialog. There is currently a "user-friendly" but somewhat generic message being shown. Showing the actual error returned by the SSH command will assist in debugging and helping to get people up and running with the remote sync feature.

smountenay-tt:
This PR modifies the Add Remote Connection dialog to display the raw ssh error that came back from the ssh process. It keeps the user-friendly message in place, but shows the ssh error underneath in monospace with a darker background.

In doing this, I have also refactored the SSH invocations into a common `SSHClient` class. We had something similar with Paramiko, but when Paramiko was removed we originally just called the system SSH everywhere instead. This creates a SSHClient like we had with Paramiko and consolidates SSH operations into one spot.

The post to /api/remote/test gets a new `detail` field that is read by the frontend:

```
[
  {
    "detail": "ssh: connect to host localhost port 22222: Connection refused",
    "message": "Unable to establish SSH connection to localhost. Please check the hostname, port, and network connectivity. Ensure SSH key-based authentication is properly configured.",
    "status": 2
  }
]
```

The frontend code is modified to recognize the detail field, and if it is set, it shows it under the message.

In my testing I noticed that sometimes the detail was being shown and in some other cases it was not, and that seems to have been caused by different HTTP status codes coming back, depending on what the error was. There is a change to the error handling to ensure that the `detail` message is actually shown. In my local testing I had seen the cases where correctly formatted responses with a `detail` value were coming back but not being displayed, but other times it was. That was due to the differing HTTP status codes leading to different behaviour in the frontend.

Screenshot 2025-07-30 at 12 03 16 PM

[Closes #703]


src/components/report-selection/ConnectionTestMessage.tsx
```diff
@@ -2,13 +2,14 @@
...
-import { IconName, IconNames } from '@blueprintjs/icons';
 import { Icon, Intent } from '@blueprintjs/core';
+import { IconName, IconNames } from '@blueprintjs/icons';
 import { ConnectionTestStates } from '../../definitions/ConnectionStatus';
 
 interface ConnectionTestMessageProps {
     status: ConnectionTestStates;
     message: string;
+    detail?: string;
 }
```
dcblundell, line 13: This is actually the same as our `ConnectionStatus` interface so we could probably use that here instead of duplicating it.

src/scss/components/RemoteConnectionDialog.scss
```diff
@@ -42,6 +43,32 @@ $tt-pending: $tt-yellow;
...
+        flex-direction: column;
+        gap: 4px;
+    }
+
+    .connection-status-text {
+        line-height: 1.4;
+    }
+
+    .connection-status-detail {
+        font-family: 'Courier New', Consolas, 'Lucida Console', monospace;
```
dcblundell, line 60: Can use existing general `monospace` class we have to apply a monospace font.

src/components/report-selection/ConnectionTestMessage.tsx
```diff
@@ -34,7 +35,10 @@ function ConnectionTestMessage({ status, message }: ConnectionTestMessageProps)
                 size={20}
                 intent={INTENT_MAP[status]}
             />
-            <span className='connection-status-text'>{message}</span>
+            <div className='connection-status-content'>
+                <span className='connection-status-text'>{message}</span>
```
dcblundell, line 39: I think we can use the `<code>` tag here to get some of the formatting for "free" rather than adding our own CSS.

PR #779, opened 2025-08-20
Add notify endpoint

### LINKED ISSUE #755: Send websocket messages to notify frontend when reports are updated
When running the visualizer directly on a Tenstorrent host with the `TT_METAL_HOME` env var set, TTNN-Visualizer should send a websocket message to the frontend when the reports are updated after a `pytest` run.

The `pytest` config and/or fixtures will be updated so that after a `pytest` run finishes, it notifies the TTNN-Visualizer server (on the tenstorrent host with tt-metal) that it finished a run and that the reports should be updated.

PyTest could either call a webhook in Flask, such as `POST /api/notify`, call the `ttnn-visualizer` CLI command with an option to tell it about a new report, or an IPC mechanism, whichever is best. When the ttnn-visualizer server gets this message, it should forward it to all connected websockets, so the frontend knows to refresh the data automatically or prompt the user to refresh.

Ideally the websocket message would contain the report name that was updated, maybe a timestamp and pass/fail status.

smountenay-tt:
This PR adds a new API endpoint `/api/notify`. You post to this API, and a websocket message is sent to connected browsers, telling the frontend about the new report that is available.

The frontend will be updated separately to handle the notification.

```
curl -X POST http://localhost:8000/api/notify -H "Content-Type: application/json" -d '{"report_name": "llama_test", "exit_status": "PASS", "profiler_path": "/path/to/profiler/report", "performance_path": "/path/to/performance_report"}'
{
  "exit_status": "PASS",
  "message": "Report update notification sent",
  "performance_path": "/path/to/performance_report",
  "profiler_path": "/path/to/profiler/report",
  "report_name": "llama_test",
  "success": true,
  "timestamp": "2025-08-20T17:54:53.861188"
}
```

And the websocket message to the browser looks like this:

Screenshot 2025-08-20 at 1 57 36 PM

This will allow IDE plugins or other dev tools to let the visualizer UI know that there is updated data, such as when `pytest` is done running in tt-metal.

This will close #755 but separately we will need to update the dev tools in question to perform the POST to the new endpoint.

[Closes #755]


aidemsined:
{
  ~"exit_status": "PASS"~,  "status": "SUCCESS",
  "message": "Report update notification sent",
  "performance_path": "/path/to/performance_report",
  "profiler_path": "/path/to/profiler/report",
  "report_name": "llama_test",
  "success": true,
  "timestamp": "2025-08-20T17:54:53.861188"
}

PR #786, opened 2025-08-25
Chore/Remove Plotyjs package

dcblundell:
We use the React package, not the vanilla js one.


aidemsined (changes requested):
lets keep it, we relay on data types

aidemsined (approved):

Issue #798, opened 2025-08-27
Initial fetch of operations list returns malformed JSON


dcblundell:
You can see errors in the console where a method is expecting a JSON parsed array. This is mostly noticable in the reports route because that's when the first hit to `/operations` tends to be, but you can also encounter this elsewhere.

<img width="1500" height="1046" alt="Image" src="https://github.com/user-attachments/assets/41e12a71-18da-4a3f-8f35-e878ab5314a7" />

<img width="1952" height="1046" alt="Image" src="https://github.com/user-attachments/assets/0c82f6dc-d7ac-444e-aea9-03d8b70f359b" />

<img width="610" height="436" alt="Image" src="https://github.com/user-attachments/assets/fc2acea1-5d5a-4514-bcd9-b51b215475ae" />

PR #804, opened 2025-08-28
Standardize on using ojson in all views

smountenay-tt:
We previously introduced the orjson library to replace `flask.jsonify`, but only one view was updated at the time, to address a specific bug. In this PR I am standardizing on using orjson for all non-error responses. There are other views that could return at least several megabytes of JSON data, which could also benefit from orjson.

I have kept jsonify in place for error responses that return empty objects and lists and such, because the shorthand is nicer that returning a full response object, and they would not benefit from orjson. But anything returning report data could.

This is being prompted by the error reported in issue #798. I found that the view in question was not using jsonify, and just returning a regular Python object. Flask does support doing that now with certain limitations. I'd rather standardize on being explicit about it and using orjson to encode the JSON, except in the case of simple error responses.

### DIFF
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -102,8 +102,7 @@ def operation_list(instance: Instance):
         inputs = list(db.query_input_tensors())
         devices = list(db.query_devices())
         producers_consumers = list(db.query_producers_consumers())
-
-        return serialize_operations(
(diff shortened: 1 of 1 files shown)


smountenay-tt:
I tested locally with the report linked in issue #798, through the remote sync feature, and it worked fine for me. Can you please try this branch out @dcblundell and see if it resolves your issue?

PR #812, opened 2025-09-03
Fix/Handle perf data processing error 

dcblundell:
When a user tries to use the old style perf reports, we now return a 422 from the API (when tt-perf-report throws an error) and then show the user a message in the UI.

**Show an error on Reports when an invalid report is chosen**
Screenshot 2025-09-03 at 3 09 57 PM

**Print a nice error when on Performance**
Screenshot 2025-09-03 at 3 09 45 PM

### DIFF
--- a/src/routes/Performance.tsx
+++ b/src/routes/Performance.tsx
@@ -142,8 +142,9 @@ export default function Performance() {
                     Data format is not supported, try using{' '}
                     <a href='https://github.com/tenstorrent/ttnn-visualizer/releases/tag/v0.49.0'>
                         TT-NN Visualizer v0.49.0
-                    </a>
-                    .
+                    </a>{' '}
+                    or earlier, or regenerate performance report using a newer version of{' '}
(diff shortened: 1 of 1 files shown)


src/routes/Performance.tsx
```diff
@@ -125,10 +130,25 @@ export default function Performance() {
...
 
+    if (perfDataError?.status === HttpStatusCode.UnprocessableEntity) {
+        return (
+            <>
+                <h2>Unable to process performance data</h2>
+                <p>
+                    Data format is not supported, try using{' '}
+                    <a href='https://github.com/tenstorrent/ttnn-visualizer/releases/tag/v0.49.0'>
+                        TT-NN Visualizer v0.49.0
+                    </a>
```
aidemsined, line 145: `...or earlier`

aidemsined: and maybe linke to this issue
https://github.com/tenstorrent/tt-metal/issues/25606

aidemsined (approved):

PR #813, opened 2025-09-03
Add route highlighting to ActiveTransferDetails

aidemsined:
image


src/scss/components/ActiveTransferDetails.scss
```diff
@@ -49,6 +49,10 @@ $outer-padding: 10px;
                 background-color: variables.$tt-dark-minimal-button-background-color-hover;
             }
         }
+
+        .bp6-switch {
+            margin-bottom: 0;
+        }
```
dcblundell, line 55: Ideally we don't change the bp6 classes directly.

dcblundell (approved):

PR #850, opened 2025-09-16
Add event type filter for fabric and local events in NPE view

### LINKED ISSUE #751: NPE: expand fabric filtering
control local and fabric event data rendering separately. user will have an option to view either or both

aidemsined:
Introduces EVENT_TYPE_FILTER enum and replaces the boolean fabricEventsOnlyFilter with a more flexible filter supporting all, fabric, and local events. Updates UI switches and filtering logic in NPEViewComponent to allow toggling between event types for transfers and link demand display.

closes #751

### DIFF
--- a/src/components/npe/NPEViewComponent.tsx
+++ b/src/components/npe/NPEViewComponent.tsx
@@ -10,7 +10,15 @@ import { Button, ButtonGroup, ButtonVariant, Intent, Size, Slider, Switch } from
 import { IconNames } from '@blueprintjs/icons';
 import classNames from 'classnames';
 import { Fragment } from 'react/jsx-runtime';
-import { NPEData, NPE_COORDINATES, NPE_LINK, NoCFlowBase, NoCTransfer, NoCType } from '../../model/NPEModel';
+import {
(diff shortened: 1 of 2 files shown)


src/components/npe/NPEViewComponent.tsx
```diff
@@ -79,23 +93,32 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
...
+    }, [npeData.noc_transfers, npeData.timestep_data, selectedTimestep, fabricEventsFilter]);
 
     const transfers = useMemo(() => {
         return npeData.noc_transfers
             .filter((tr) => links?.active_transfers.includes(tr.id))
             .filter((tr) => {
-                return fabricEventsOnlyFilter ? tr.fabric_event_type : true;
+                if (fabricEventsFilter === EVENT_TYPE_FILTER.ALL_EVENTS) {
+                    return true;
+                }
```
dcblundell, line 112: Probably don't need to explicitly say this as the default state is `return true`.

dcblundell (approved):

Issue #706, opened 2025-07-30
Graph deallocate ops ignores filtering values


dcblundell:
When toggling deallocate ops, any value for filtering ops is not respected until you interact with the filter again.

PR #854, opened 2025-09-17
Fix/Restore Graph Focus When Using Deallocate Ops Switch

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/706.


aidemsined (changes requested):
~address update is broken, everything else seems ok~ nm, it has been for a while
image

aidemsined (approved):

PR #856, opened 2025-09-18
Feature/Add Stack By Input 0 Support

dcblundell:
Allows users to toggle stacking operations by input 0 and adds op code filtering to the stacked table.

Signposts (rows which are not operations but markers in the code) are going to be properly supported in https://github.com/tenstorrent/ttnn-visualizer/pull/858 but I'm adding some code to handle their appearance in the data here. Most of this is required anyway for that PR, or will be adjusted slightly when we fully support signposts as a future kwarg. 

**Default (current behaviour)**
Screenshot 2025-09-18 at 12 47 54 PM

**Disabled stacking**
Screenshot 2025-09-18 at 12 47 48 PM

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/860.

### DIFF
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -68,6 +68,7 @@ from ttnn_visualizer.utils import (
     create_path_resolver,
     get_cluster_descriptor_path,
     read_last_synced_file,
+    str_to_bool,
     timer,
 )
 
@@ -687,7 +688,7 @@ def get_performance_results_report(instance: Instance):
         )
 
     name = request.args.get("name", None)
-    stackByIn0 = request.args.get("stackByIn0", "true").lower() == "true"
+    stackByIn0 = str_to_bool(request.args.get("stackByIn0", "true"))
 
     if name and not current_app.config["SERVER_MODE"]:
         performance_path = Path(instance.performance_path).parent / name


backend/ttnn_visualizer/csv_queries.py
```diff
@@ -452,6 +455,8 @@ def generate_report(cls, instance):
         stacked_png_file = os.path.splitext(csv_output_file)[0] + ".png"
 
         try:
+            print("$$$$$$$$$$$$$$$ generating report $$$$$$$$$$$$$$$")
```
aidemsined, line 458: does this need to stay?

dcblundell: Forgot to push commit, done.

src/components/performance/PerfReport.tsx
```diff
@@ -127,11 +146,20 @@ const PerformanceReport: FC<PerformanceReportProps> = ({
...
+
     const updateColumnFilter = (key: TableKeys, value: string) => {
-        setFilters({
+        const updatedFilters = {
             ...filters,
             [key]: value ?? '',
-        } as Record<TableKeys, string>);
+        };
+
+        // TODO: Sort this madness out
```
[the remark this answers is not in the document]
dcblundell: This is the path I've taken so far. It needs refactoring at some point but that's why I added the TODO.

backend/ttnn_visualizer/views.py
```diff
@@ -687,13 +687,16 @@ def get_performance_results_report(instance: Instance):
...
+    stackByIn0 = request.args.get("stackByIn0", "false").lower() == "true"
 
     if name and not current_app.config["SERVER_MODE"]:
         performance_path = Path(instance.performance_path).parent / name
         instance.performance_path = str(performance_path)
         logger.info(f"************ Performance path set to {instance.performance_path}")
 
     try:
+        # TODO: Scott to check
+        OpsPerformanceReportQueries.DEFAULT_NO_STACK_BY_IN0 = stackByIn0
```
smountenay-tt, line 699: @dcblundell We should pass the value another way. This way is not ideal, because the value will persist on the `OpsPerformanceReportQueries` class after this request - because the class is not being instantiated in the view. Also, `OpsPerformanceReportQueries.DEFAULT_NO_STACK_BY_IN0` looks like a constant, due to being uppercase, so you wouldn't expect that value to be getting changed in a view. I would expect `OpsPerformanceReportQueries.generate_report` to take this value as a kwarg with a default value.

smountenay-tt (approved):
I'd suggest changing that one line to use the `str_to_bool` util function, but otherwise looks ok.

PR #858, opened 2025-09-19
Feature/Support Signposts

dcblundell:


backend/ttnn_visualizer/csv_queries.py
```diff
@@ -452,14 +451,21 @@ def generate_report(cls, instance, **kwargs):
         csv_file = StringIO(raw_csv)
         csv_output_file = tempfile.mktemp(suffix=".csv")
         csv_stacked_output_file = tempfile.mktemp(suffix=".csv")
+        use_signpost = kwargs.get("signpost", cls.DEFAULT_SIGNPOST)
+        ignore_signposts = cls.DEFAULT_IGNORE_SIGNPOSTS
+        stack_by_in0 = kwargs.get("stackByIn0", cls.DEFAULT_NO_STACK_BY_IN0)
```
smountenay-tt, line 456: I would suggest `stackByIn0` by passed snake case also. It's the convention in Python.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -452,14 +451,21 @@ def generate_report(cls, instance, **kwargs):
         csv_file = StringIO(raw_csv)
         csv_output_file = tempfile.mktemp(suffix=".csv")
         csv_stacked_output_file = tempfile.mktemp(suffix=".csv")
+        use_signpost = kwargs.get("signpost", cls.DEFAULT_SIGNPOST)
```
smountenay-tt, line 454: I'd suggest standardizing the name - `use_signpost` or `signpost`. It would be more intuitive if the local var name matched the kwarg.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -481,12 +487,25 @@ def generate_report(cls, instance, **kwargs):
         for row in ops_perf_results_reader:
             ops_perf_results.append(row)
 
+        # Returns a list of unique signposts in the order they appear
+        # TODO: Signpost names are not unique but tt-perf-report treats them as such
```
smountenay-tt, line 491: These lines seem to be a duplicate of the lines below it. For example the comment with the TODO message appears twice.

dcblundell: Ugh this snuck back in. Apologies.

backend/ttnn_visualizer/csv_queries.py
dcblundell: Good point.

backend/ttnn_visualizer/csv_queries.py
dcblundell: Will do.

smountenay-tt (approved):

PR #867, opened 2025-09-24
transfer visualization mode (fabric/local/mixed transfer events)

### LINKED ISSUE #835: NPE: add new visualization mode
In this mode:
- links will be purple if all active_transfers have `fabric_event_type=false`
- links will be yellow if all active_transfers have `fabric_event_type=true`
- links will be red if all some active_transfers have `fabric_event_type=false` and some have `fabric_event_type=true`

Users should be switch between this mode (which we can call `event mode`) and the previous mode (which we can call `congestion mode`)

aidemsined:
image

closes #835

### DIFF
--- a/src/components/npe/ActiveTransferDetails.tsx
+++ b/src/components/npe/ActiveTransferDetails.tsx
@@ -3,7 +3,7 @@
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
 import { useState } from 'react';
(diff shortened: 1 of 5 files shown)


aidemsined:
[sample data](https://github.com/user-attachments/files/22519800/_ID41988-fabric.npeviz.zst.zip)

dcblundell (approved):

PR #869, opened 2025-09-24
Fix/Filter Signpost From Perf Charts

dcblundell:
Fixes https://github.com/tenstorrent/ttnn-visualizer/issues/870.


src/components/BufferDetails.tsx
```diff
@@ -41,7 +41,7 @@ function BufferDetails({ tensor, operations, className }: BufferDetailsProps) {
                         <th>Producer</th>
                         <td>
                             {isValidNumber(firstOperationId)
-                                ? getFirstOperation(firstOperationId, operations, tensor)
+                                ? getFirstOperation(firstOperationId, operations)
```
aidemsined, line 44: i think this is a bad name for the function. feel free to not do it in this PR

PR #871, opened 2025-09-26
Chore/reorganise perf types and atoms

dcblundell:


src/components/performance/PerfReport.tsx
```diff
@@ -64,7 +64,7 @@ interface PerformanceReportProps {
     signposts?: Signpost[];
 }
 
-const INITIAL_TAB_ID = 'perf-table-0';
+const INITIAL_TAB_ID = 'perf-table-0'; // `perf-table-${index}`
```
aidemsined, line 67: do we need this comment?

src/definitions/PerfTable.ts
```diff
@@ -75,9 +76,13 @@ export interface TypedPerfTableRow
     flops_percent: number | null;
 }
 
-export type MathFidelity = 'HiFi4' | 'HiFi2' | 'LoFi';
+export enum MathFidelity {
```
aidemsined, line 79: nit: while i get the convention - uppercase strings like taht are REALLY hard to read

src/functions/getChartData.ts
```diff
@@ -14,8 +14,8 @@ export default function getChartData(
     overrides?: { color?: string; colorVariance?: number; hovertemplate?: string },
     options?: { renderPattern?: boolean },
 ): Partial<PlotDataCustom>[] {
-    // TODO: we may want to double check this eventually
-    // @ts-expect-error - PlotDataCustom works just fine
+    // TODO: we may want to double check this eventually (there does't seem to be an error here anymore)
```
aidemsined, line 17: lets remove then

src/components/performance/PerfReport.tsx
dcblundell: It's just explaining what it is what it is.

src/definitions/PerfTable.ts
dcblundell: Fair point, changed casing.

PR #873, opened 2025-09-30
Feature/Add Multi Op Code Filter

dcblundell:
Adds a dropdown filter for the raw op codes which can work in combination with the existing string filter.

Did some refactoring relating to the types used and the naming of things.

Screenshot 2025-09-30 at 3 11 13 PM

Screenshot 2025-09-30 at 3 11 38 PM

### DIFF
--- a/src/components/performance/PerfReport.tsx
+++ b/src/components/performance/PerfReport.tsx
@@ -105,12 +105,14 @@ const PerformanceReport: FC<PerformanceReportProps> = ({
         activeMultiSelectFilters: activeRawOpCodeFilters,
         OptionComponent: RawOpCodeOption,
     } = useMultiSelectFilter('raw_op_code', data || []);
+
     const {
         getMultiSelectOptions: getMathFilterOptions,
         updateMultiSelect: updateMathFilters,
         activeMultiSelectFilters: activeMathFilters,
         OptionComponent: MathOption,
     } = useMultiSelectFilter('math_fidelity', data || []);
+
     const opIdsMap = useOpToPerfIdFiltered();
 
     const isSignpostsDisabled = !signposts || signposts.length === 0;


src/functions/getChartData.ts
```diff
@@ -15,7 +15,7 @@ export default function getChartData(
     options?: { renderPattern?: boolean },
 ): Partial<PlotDataCustom>[] {
     // TODO: we may want to double check this eventually
-    // @ts-expect-error - PlotDataCustom works just fine
+    // // @ts-expect-error - PlotDataCustom works just fine
```
dcblundell, line 18: Might not need this because if I remove it there's no TS error.

src/hooks/useMultiSelectFilter.tsx
```diff
@@ -3,40 +3,46 @@
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
 import { useState } from 'react';
-import TableFilterItem from '../components/TableFilterItem';
+import { Checkbox } from '@blueprintjs/core';
+import 'styles/components/TableFilterItem.scss'; // Bit weird having this in a hook
```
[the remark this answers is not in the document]
dcblundell: Removed this.

src/functions/sortAndFilterStackedPerfTableData.ts
```diff
@@ -40,6 +46,17 @@ const sortAndFilterStackedPerfTableData = (
         });
     }
 
+    // In the stacked data the op_code field is named just "op_code" not "raw_op_code"
+    if (rawOpCodeFilter?.length > 0) {
+        filteredRows = filteredRows.filter(
+            (row) =>
+                row?.op_code !== null &&
+                rawOpCodeFilter.some((filterValue) =>
+                    row.op_code.toLowerCase().includes(String(filterValue).toLowerCase()),
```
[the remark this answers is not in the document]
dcblundell: This needs to be includes because the stacked op_code value can include other stuff.

src/components/performance/PerfReport.tsx
```diff
@@ -73,7 +72,18 @@ const PerformanceReport: FC<PerformanceReportProps> = ({
     comparisonStackedData,
     signposts,
 }) => {
-    const { getFilterOptions, updateFilters, activeFilters, FilterItem } = useTableFilter('math_fidelity', data || []);
+    const {
```
aidemsined, line 75: granted with will change as per dsicussion but line breaks for readability please

aidemsined, line 75: also a nitpick, you want useState to be first things in a component (together with atoms)

src/functions/getChartData.ts
aidemsined: if this is not needed remove it please

src/components/performance/PerfReport.tsx
aidemsined: formatter doesnt dictate line breaks between code blocks, this is not python

PR #883, opened 2025-10-03
sticky header + fixed active transfer

### LINKED ISSUE #872: NPE: Pin top bar as we scroll down
Image

The bar should stay on the top of the window as we scroll down instead of going away. This makes it easier using the bar on large meshes like the 6U.

aidemsined:
Moved the ActiveTransferDetails component outside the main content flex row and updated its CSS to use fixed positioning with full viewport height. Also made the NPE header sticky for improved UI consistency.

image

closes #872

### DIFF
--- a/src/components/npe/NPEViewComponent.tsx
+++ b/src/components/npe/NPEViewComponent.tsx
@@ -742,23 +742,23 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
                         );
                     })}
                 </div>
-
-                <ActiveTransferDetails
-                    groupedTransfersByNoCID={groupedTransfersByNoCID}
-                    selectedNode={selectedNode}
-                    congestionData={links?.link_demand.filter(
-                        (route) =>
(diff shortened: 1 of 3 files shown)


dcblundell:
Probably an existing bug, but the text here does not wrap.

Frame 632 with "Show Individual route" enabled  in report `selection-ID65566.npeviz.zst`.

Screenshot 2025-10-06 at 9 31 52 AM

src/scss/definitions/_variables.scss
```diff
@@ -21,3 +21,7 @@ $tt-dark-minimal-button-background-color-hover: color.adjust($dark-button-backgr
 // NPE
 $npe-highlight-src: #ff0;
 $npe-highlight-dst: #ff4500;
+
+// TODO: we should refactor ALL z-indexes across the code base. none use vars
```
dcblundell, line 25: Good point.

dcblundell (approved):

PR #894, opened 2025-10-07
Improve buffer error handling and messaging

### LINKED ISSUE #801: Maximum call stack size exceeded error
My goal was to see memory usage of the SDXL model. I ran into Out of Memory problem so only a part of ops got executed. Until that point, 7000 operation got executed so there's just a ballpark number of ops in the entire model. When I generate data from TTNN and put it into visualizer application, I get a "Sorry, and unexpected error has occurred. Maximum call stack size exceeded" screen when I try to switch to the Buffers tab or any other. When waiting for ~10 seconds, I can switch to the Buffers tab but when clicking on a concrete op, this error screen pops up. I believe this is some frontend rendering problem because maybe I try to click into some component that hasn't been loaded yet so I get an error and waiting for that component to load for some time allows me to click on it later on. Currently, I have my report [here](https://drive.google.com/drive/folders/1Ho_HCpJ8aaSyF8_YTOtI3nsM6R0XvbrU?usp=sharing).
I also noticed that when I was able to access the Buffers tab, L1 memory was empty and I can see an op was being executed at that time with some large arguments that should definitely appear but I'm not sure if this is an issue with the report or visualizer but wanted to mention it here.
Here is the screen I see:

Image

aidemsined:
Enhances error handling in useBuffers by validating API responses and throwing errors for invalid or oversized data. Updates BufferSummary to display detailed error messages from the API, providing more context to users when buffer data fails to load.

image

closes #801

### DIFF
--- a/src/hooks/useAPI.tsx
+++ b/src/hooks/useAPI.tsx
@@ -816,11 +816,25 @@ export const useBuffers = (bufferType: BufferType, useRange?: boolean) => {
     const activeProfilerReport = useAtomValue(activeProfilerReportAtom);
 
     const response = useQuery<BuffersByOperationData[], AxiosError>({
-        queryFn: () => (activeProfilerReport !== null ? fetchBuffersByOperation(bufferType) : Promise.resolve([])),
         queryKey: ['fetch-all-buffers', bufferType, activeProfilerReport],
         enabled: activeProfilerReport !== null,
         retry: false,
         staleTime: Infinity,
+        queryFn: async () => {
+            const data = await fetchBuffersByOperation(bufferType);
+            if (activeProfilerReport === null) {
+                await Promise.resolve([]);
+            }
+            // @ts-expect-error will happen with extra large data sets where we get a string instead of an array
+            if (data === '' || !Array.isArray(data)) {
+                throw new AxiosError(
+                    `Invalid response: data is invalid or too large to render."`,
+                    'ERR_INVALID_RESPONSE',
+                );
+            }
+
+            return data;
+        },
     });
 
     return useMemo(() => {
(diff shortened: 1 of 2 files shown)


src/hooks/useAPI.tsx
```diff
@@ -816,11 +816,25 @@ export const useBuffers = (bufferType: BufferType, useRange?: boolean) => {
...
-        queryFn: () => (activeProfilerReport !== null ? fetchBuffersByOperation(bufferType) : Promise.resolve([])),
         queryKey: ['fetch-all-buffers', bufferType, activeProfilerReport],
         enabled: activeProfilerReport !== null,
         retry: false,
         staleTime: Infinity,
+        queryFn: async () => {
+            const data = await fetchBuffersByOperation(bufferType);
+            if (activeProfilerReport === null) {
+                await Promise.resolve([]);
+            }
```
dcblundell, line 827: This seems funny, it fetches before there's an activeProfilerReport. I don't think that can happen in reality but is there a reason you changed the logic?

src/hooks/useAPI.tsx
```diff
@@ -816,11 +816,25 @@ export const useBuffers = (bufferType: BufferType, useRange?: boolean) => {
...
+            // @ts-expect-error will happen with extra large data sets where we get a string instead of an array
```
dcblundell, line 828: If you put this in a try/catch would you get the typescript error?

dcblundell (changes requested):

src/routes/BufferSummary.tsx
```diff
@@ -118,7 +118,7 @@ function BufferSummary() {
                                 title='Error loading buffer data'
                                 compact
                             >
-                                <p>{`We've been unable to load the L1 buffer data for /${activeProfilerReport}.`}</p>
+                                <p>{`We've been unable to load the L1 buffer data for /${activeProfilerReport}. ${buffersError.message}`}</p>
```
dcblundell, line 121: I wonder if the error from the response should go on another line because you've got our nicely written error message and then a more technical one.

> We've been unable to load the L1 buffer data for /${activeProfilerReport}
> Invalid response: data is invalid or too large to render

dcblundell (changes requested):

dcblundell (approved):

PR #916, opened 2025-10-21
Feature/Format Operation List Errors

dcblundell:
Toggling the stack trace will scroll to the current virtual item index if needed. It's not precise, but it avoids weird issues with the virtualised list.

Screenshot 2025-10-21 at 11 34 51 AM

### DIFF
--- a/src/components/Collapsible.tsx
+++ b/src/components/Collapsible.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
-import { Button, Collapse } from '@blueprintjs/core';
+import { Button, ButtonVariant, Collapse, Size } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import React, { useEffect } from 'react';
 import { JSX } from 'react/jsx-runtime';
@@ -45,8 +45,8 @@ const Collapsible: React.FC<React.PropsWithChildren<CollapsibleProps>> = ({
                 {children && (
                     <Button
                         className='collapsible-button'
-                        size='small'
-                        variant='minimal'
+                        size={Size.SMALL}
+                        variant={ButtonVariant.MINIMAL}
                         onClick={
                             !isDisabled
                                 ? () => {
--- a/src/components/FooterInfobar.tsx
+++ b/src/components/FooterInfobar.tsx
@@ -13,17 +13,13 @@ import {
     activeProfilerReportAtom,
     operationRangeAtom,
     performanceRangeAtom,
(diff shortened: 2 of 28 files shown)


src/components/operation-details/StackTrace.tsx
```diff
@@ -4,49 +4,60 @@
...
 }
 
-function StackTrace({ stackTrace }: StackTraceProps) {
-    const [isFullStackTrace, setIsFullStackTrace] = useAtom(isFullStackTraceAtom);
+function StackTrace({
+    stackTrace,
+    language,
+    hideSourceButton,
+    isInline,
+    isExpanded,
```
aidemsined: this implies the component state is controlled externally. and it is not. i dont have a better suggestion

src/components/operation-details/StackTrace.tsx
```diff
@@ -83,7 +94,9 @@ function StackTrace({ stackTrace }: StackTraceProps) {
         }
 
         return '';
-    }, [fileContents, stackTrace]);
+    }, [fileContents, stackTrace, language]);
+
+    const toggleViewingFile = useCallback(() => setIsViewingFile((open) => !open), [setIsViewingFile]);
```
aidemsined: not clear what we are toggling here

src/components/operation-details/StackTrace.tsx
```diff
@@ -4,49 +4,60 @@
...
 
-function StackTrace({ stackTrace }: StackTraceProps) {
-    const [isFullStackTrace, setIsFullStackTrace] = useAtom(isFullStackTraceAtom);
+function StackTrace({
+    stackTrace,
+    language,
+    hideSourceButton,
+    isInline,
+    isExpanded,
+    onToggleExpanded,
```
aidemsined: onExpanded? toggle means you keep calling it to flip the state

src/components/OperationDetailsNavigation.tsx
```diff
@@ -129,7 +129,7 @@ function OperationDetailsNavigation({ operationId, isLoading }: OperationDetails
 
                         {operation?.error && (
                             <Tooltip
-                                content='Error detected in this operation'
```
aidemsined: nit: Error recorded in ~this~ operation? this feels unnecessary

src/components/operation-details/StackTrace.tsx
dcblundell: `isViewingSourceFile`

src/components/operation-details/StackTrace.tsx
dcblundell: If a function is supplied here it is called when you toggle expand.

PR #935, opened 2025-11-03
Enable zone navigation in NPE components

### LINKED ISSUE #934: NPE: make zones clickable
clicking a zone should navigate timeline to zone start cycle

aidemsined:
Adds interactive zone selection to the NPE zone filter, allowing users to click on a zone to navigate directly to its corresponding timestep in the main NPE view. It also improves the user interface by making zones visually interactive and adjusts the filter button label for clarity.

closes #935
closes #934

### DIFF
--- a/src/components/npe/NPEViewComponent.tsx
+++ b/src/components/npe/NPEViewComponent.tsx
@@ -278,7 +278,7 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
             return prev < range - 1 ? prev + 1 : 0;
         });
     };
-    const onNavigateToZone = (zone: NPEZone) => {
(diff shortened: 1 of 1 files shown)


src/components/npe/NPEViewComponent.tsx
```diff
@@ -821,6 +826,7 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
                         [key]: state,
                     }));
                 }}
+                onZoneClick={onNavigateToZone}
```
dcblundell: The current naming sounds like it's chaining two events, e.g. `onZoneClick do onNavigateToZone` which is not quite right. I think the function should be called something like `updateSelectedTimestep` or `handleZoneClick`.

PR #939, opened 2025-11-03
Highlight current position in congestion heatmap and fixes inaccurate tooltip

### LINKED ISSUE #936: NPE: mark current timestep on heatmap
render current timeline position on timeline

### LINKED ISSUE #938: NPE: inaccurate timestep rendering on heatmap
there is a slight misalignment on tooltip indicating current mouse position

aidemsined:
image

closes #936 
closes #938

### DIFF
--- a/src/components/npe/NPEViewComponent.tsx
+++ b/src/components/npe/NPEViewComponent.tsx
@@ -548,7 +548,7 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
                 <NPECongestionHeatMap
                     timestepList={npeData.timestep_data}
                     canvasWidth={canvasWidth}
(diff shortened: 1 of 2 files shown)


src/components/npe/NPEViewComponent.tsx
```diff
@@ -548,8 +548,9 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
                 <NPECongestionHeatMap
                     timestepList={npeData.timestep_data}
                     canvasWidth={canvasWidth}
+                    currentPosition={selectedTimestep}
```
dcblundell: Wonder if this should be `currentTimestep` because position implies coordinates, but those are calculated within the function.

PR #972, opened 2025-11-11
Feature/Include Device IDs In Perf Table

dcblundell:
Adds a Device ID column to the Perf Table and adds a note explaining the multi device merge logic we inherit from tt-perf-report.

Screenshot 2025-11-11 at 12 34 35 PM

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/968.


aidemsined (approved):
maybe reduce the margins on the callout slightly

dcblundell:
> maybe reduce the margins on the callout slightly

Added `compact` prop which reduces padding.

PR #977, opened 2025-11-12
Fix/Restore PM Ideal Data

dcblundell:
Restores the parsing of the `pm_ideal_ns` field which can then be used to calculate Core Utilization.

**Before**
Screenshot 2025-11-12 at 12 26 02 PM

**After**
Screenshot 2025-11-12 at 12 25 28 PM

Fixes https://github.com/tenstorrent/ttnn-visualizer/issues/978.

### DIFF
--- a/backend/ttnn_visualizer/csv_queries.py
+++ b/backend/ttnn_visualizer/csv_queries.py
@@ -508,6 +508,9 @@ class OpsPerformanceReportQueries:
                     next(reader, None)
                     for row in reader:
                         op_id = int(row[0])
+                        # IDs in result column one correspond to row numbers in ops perf results csv
+                        idx = op_id - 2
+
                         processed_row = {
                             column: row[index]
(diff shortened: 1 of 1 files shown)


backend/ttnn_visualizer/csv_queries.py
```diff
@@ -507,27 +507,34 @@ def generate_report(cls, instance, **kwargs):
...
-                            raw_idx = op_id - 2
-                            if 0 <= raw_idx < len(ops_perf_results):
-                                processed_row["op_type"] = ops_perf_results[
-                                    raw_idx
-                                ].get("OP TYPE")
-                            else:
-                                processed_row["op_type"] = None
+                        for key, value in cls.PASSTHROUGH_COLUMNS.items():
+                            idx = (
+                                op_id - 2
```
aidemsined: please move outisde of the inner loop

Issue #980, opened 2025-11-12
Performance: Core Utilizatation Scale


dcblundell:
With small values, the y-axis scale is unhelpful.

<img width="934" height="539" alt="Image" src="https://github.com/user-attachments/assets/f42c876d-ad23-4142-85db-37b69c155e90" />

PR #981, opened 2025-11-12
Fix/Core Utilization Scale

dcblundell:
Max utilization under 1% causes the y axis scale to display all `0%`.

I had a rethink about this and set the Utilization scale to be a 0-100% at a minimum and then scale up if needed (multi device). I've also adjusted the scale to always show the top end axis label, e.g. `100%` because sometimes it was getting cut off.

Screenshot 2025-11-18 at 10 29 02 AM

Screenshot 2025-11-18 at 10 29 22 AM

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/980.


aidemsined (approved):

aidemsined (changes requested):
Image

something is weird

PR #983, opened 2025-11-12
Refactor/NPE Processing Status

dcblundell:
Refactors component to reduce complexity.


dcblundell:
> imo: `getNpeDataErrorType` and`isValidNpeData` should be combined. there should be a single point to validate the data and respond with: valid or invalid and here is the error type. error rendering component should not have the logic to determine error rendering, it shoudl just render it based on validation result

Have refactored.

PR #1014, opened 2025-11-26
Prioritize gunicorn from same directory as ttnn-visualizer script

### LINKED ISSUE #1012: Prioritize running gunicorn from the same directory as the ttnn-visualizer script
There is a long-standing issue with running the `ttnn-visualizer` boot script, in which it sometimes runs the wrong `gunicorn`. This has manifested itself as thinking you are starting one version of TTNN Visualizer, but actually seeing another in your browser. This has tripped a few people up, and the recommendation has been to ensure you do not have multiple copies of TTNN Visualizer installed at the same time (i.e. in a venv and outside a venv).

In other cases, it's possible that a user runs the `ttnn-visualizer` command, but there can be an error about `gunicorn` not being found.

The reason this happens, is because the start-up code in `app.py` runs a separate `gunicorn` process, to start the Flask webserver. In other words, when you run the `ttnn-visualizer` wrapper script (when installed from pip or wheel, not local dev), it loads code from `app.py`, which after doing some basic initialization, launches a `gunicorn` process to start the Flask web server.

The problem is that it just runs whatever `gunicorn` is found in the `PATH`. So if you have `gunicorn` or `ttnn-vizualizer` installed globally (i.e. not in a venv), then you run `.venv/bin/ttnn-visualizer` for example, it's possible that the `gunicorn` command it finds is the global one. In theory it could even be `gunicorn` installed by a different package.

Another case where this problem happens is if you run `pip install ttnn-visualizer` on an Ubuntu container, and then try to run `ttnn-visualizer` command, but you don't have `.local/bin` in your PATH.  On newer and / or full Ubuntu distributions, `.local/bin` may be getting added to `PATH` automatically, but on containers I'm running TT-Metal in, using [ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-22.04-dev-amd64:latest](http://ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-22.04-dev-amd64:latest), I don't have `.local/bin` in my PATH automatically and running `ttnn-visualizer` in the shell gives a not found error.

If I run `.local/bin/ttnn-visualizer` without `.local/bin` in the PATH, it gives an error about `gunicorn` not being found:

```
FileNotFoundError: [Errno 2] No such file or directory: 'gunicorn'
```

This is easy to solve with `export PATH=~/.local/bin/:$PATH` but the file not found error is not very friendly and doesn't really give any indication to the user what is wrong.

## Proposed solution

Modify the bootstrap code in `app.py` so that it looks for `gunicorn` in the same directory as where the `ttnn-visualizer` command is run from. In other words, if you run `ttnn-visualizer` CLI command and it runs `.local/bin/ttnn-visualizer`, then it should run `.local/bin/ttnn-gunicorn` if it exists, not another gunicorn such as `/usr/local/bin/gunicorn` or something that else is found in the `PATH` first.

Instead of leaving it up to whatever `gunicorn` is found in the `PATH` first, it should look actually look for `gunicorn` in the local bin dir first, and if its not found, it should probably display an error rather than running another gunicorn script from another dir. There is no situation in which we'd actually want to run a different `gunicorn` script other than one in the same `bin` directory as the `ttnn-visualizer` command.

smountenay-tt:
This PR changes how `gunicorn` is run from `app.py`. Instead of depending on `gunicorn` being found in the `PATH`, it explicitly tries to run `gunicorn` from the same directory as the `ttnn-visualizer` entry point. In other words, if you run `.local/bin/ttnn-visualizer`, it looks for `.local/bin/gunicorn`, rather than just running whatever `gunicorn` script is found in the `PATH` first.

If `gunicorn` is not found or there are mulitple `gunicorn` scripts found in the `PATH`, it will show a warning message, but still tries to run it.

[Closes #1012]

### DIFF
--- a/backend/ttnn_visualizer/tests/test_utils.py
+++ b/backend/ttnn_visualizer/tests/test_utils.py
@@ -16,7 +16,6 @@ def test_find_gunicorn_in_same_directory(
     mock_which, mock_is_file, mock_exists, mock_access
 ):
     """Test finding gunicorn in the same directory as ttnn-visualizer."""
-    # Setup: gunicorn exists in same dir and is executable
     mock_exists.return_value = True
     mock_is_file.return_value = True
     mock_access.return_value = True
@@ -24,7 +23,6 @@ def test_find_gunicorn_in_same_directory(
 
     gunicorn_path, warning = find_gunicorn_path()
 
-    # Check path ends with expected value (resolve() may add prefix on some systems)
     assert gunicorn_path.endswith("/home/user/.local/bin/gunicorn")
     assert warning is None
 
@@ -38,7 +36,6 @@ def test_find_multiple_gunicorn_installations(
     mock_which, mock_is_file, mock_exists, mock_access
 ):
     """Test warning when multiple gunicorn installations are detected."""
-    # Setup: gunicorn in same dir and also in PATH
     mock_exists.return_value = True
     mock_is_file.return_value = True
     mock_access.return_value = True
@@ -46,7 +43,6 @@ def test_find_multiple_gunicorn_installations(
 
     gunicorn_path, warning = find_gunicorn_path()
 
-    # Check path ends with expected value (resolve() may add prefix on some systems)
     assert gunicorn_path.endswith("/home/user/.local/bin/gunicorn")
     assert warning is not None
     assert "Multiple gunicorn installations detected" in warning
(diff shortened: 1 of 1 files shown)


dcblundell (approved):
Looks good 👍

dcblundell (approved):

PR #1031, opened 2025-12-01
Feature/NPE Accessibility Pass

dcblundell:
Doesn't resolve the chip id low contrast text or the small text in the grid, but this is an improvement.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/785.

### DIFF
--- a/src/components/npe/NPEViewComponent.tsx
+++ b/src/components/npe/NPEViewComponent.tsx
@@ -361,24 +361,24 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
                         <Button
                             icon={IconNames.StepBackward}
                             onClick={onBackward}
-                            aria-label='Step backward'
+                            title='Step backward'
                         />
                         <Button
                             icon={isTimelinePlaying ? IconNames.Pause : IconNames.Play}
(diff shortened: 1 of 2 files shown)


src/components/npe/NPEViewComponent.tsx
```diff
@@ -361,20 +361,24 @@ const NPEView: React.FC<NPEViewProps> = ({ npeData }) => {
                         <Button
                             icon={IconNames.StepBackward}
                             onClick={onBackward}
+                            aria-label='Step backward'
```
aidemsined: might as well use title attribute

src/components/npe/NPEZoneFilterComponent.tsx
```diff
@@ -125,6 +125,7 @@ const NPEZoneFilterComponent: React.FC<NPEZoneFilterComponentProps> = ({
                         variant={ButtonVariant.MINIMAL}
                         onClick={onClose}
                         icon={IconNames.CHEVRON_LEFT}
+                        aria-label='Close zones panel'
```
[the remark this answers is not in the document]
dcblundell: Good point.

src/components/npe/NPEViewComponent.tsx
dcblundell: Good point.

PR #1042, opened 2025-12-04
Feature/Support no merge rows

dcblundell:
Adds a toggle for skipping multi device row merging in Perf data. 

When enabled, it also hides the informational banners explaining the merge logic.

Screenshot 2025-12-04 at 4 05 08 PM

Screenshot 2025-12-04 at 4 11 12 PM

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1056.


smountenay-tt:
Personally I would prefer not to use "no" in the `noMergeDevices` and `no_merge_devices` variable names. I would suggest just using `mergeDevices` and `merge_devices` and have it default to true. Then we could avoid stuff like `{!noMergeDevices && (`. It would be cleaner if it was just `{mergeDevices && (`.

I guess at this point you may not want to change it everywhere and it's not blocking for approval.

dcblundell:
> `noMergeDevices`

I agree so I've changed it everywhere.

PR #1043, opened 2025-12-05
Feature/1039 move arch and frequency to python

### LINKED ISSUE #1039: Performance: move reading arch to python
UPD: its the non performant call that should be solved on a BE side
~inconsistent loading experience~
Image
while

```
Arch: wormhole_b0
Cores: 64
```

is loading the rest of the ui is disabled functionally but looks normal

aidemsined:
This pull request refactors how device architecture and frequency metadata are retrieved and used in the perf frontend and backend. Instead of extracting this metadata from the raw device log file in the frontend, a new dedicated backend API endpoint is introduced to provide just the metadata, and the frontend is updated to use this endpoint. This simplifies data handling and improves loading speed 

* Added a new API endpoint `/performance/device-log/meta` that reads only the metadata (architecture and frequency) from the device log, making frontend data access more efficient and robust.
* Changed FE to use the above

closes #1039

### DIFF
--- a/tests/data/mockDeviceMeta.json
+++ b/tests/data/mockDeviceMeta.json
@@ -1,4 +1,4 @@
-{"deviceMeta":{
-    "architecture":"wormhole_b0",
-    "frequency":1000
-}}
+{
+    "architecture": "wormhole_b0",
+    "frequency": 1000
+}


backend/ttnn_visualizer/views.py
```diff
@@ -751,6 +751,7 @@ def get_performance_results_report(instance: Instance):
     return Response(orjson.dumps(report), mimetype="application/json")
 
 
+# this is no longer used atm. keeping for now until confirmed "not needed"
 @api.route("/performance/device-log/raw", methods=["GET"])
```
dcblundell: Maybe comment the whole route out?

PR #1067, opened 2025-12-12
Improve NPE error handling and query logic

### LINKED ISSUE #1065: NPE: perf integration is broken
Image

aidemsined:
Refactored NPE route to handle timeline file errors and distinguish between invalid JSON and other HTTP errors. Updated useNpe hook to enable queries only when a filename is provided, and adjusted PerfTable to use -1 as a default for missing global_call_count values.

this addresses NPE attempts to always load an npe file instead of the timeline if one is provided

closes #1065

### DIFF
--- a/src/hooks/useAPI.tsx
+++ b/src/hooks/useAPI.tsx
@@ -431,13 +431,15 @@ const fetchNpeOpTrace = async () => {
     return response?.data;
 };
 
-export const useNpe = (fileName: string | null) =>
-    useQuery<NPEData, AxiosError>({
+export const useNpe = (fileName: string | null) => {
+    return useQuery<NPEData, AxiosError>({
         queryFn: () => fetchNpeOpTrace(),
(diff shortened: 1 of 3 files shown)


src/hooks/useAPI.tsx
```diff
@@ -431,13 +431,15 @@ const fetchNpeOpTrace = async () => {
     return response?.data;
 };
 
-export const useNpe = (fileName: string | null) =>
-    useQuery<NPEData, AxiosError>({
+export const useNpe = (fileName: string | null) => {
+    return useQuery<NPEData, AxiosError>({
         queryFn: () => fetchNpeOpTrace(),
         queryKey: ['fetch-npe', fileName],
```
dcblundell: Not related to your PR, but just wondering if `fileName` is unique enough. If you compare similar reports for example, couldn't `fileName` be the same?

PR #1096, opened 2025-12-18
Enable portal for feedback dialog and add custom style

### LINKED ISSUE #1072: Feedback widget
The feedback widget could use a rethink. The half off screen positioning is not ideal and the popup can sometimes not be in view depending on the page scroll height.

aidemsined:
image

closes #1072

### DIFF
--- a/src/components/FeedbackButton.tsx
+++ b/src/components/FeedbackButton.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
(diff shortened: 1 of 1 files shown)


src/components/FeedbackButton.tsx
```diff
@@ -57,7 +57,8 @@ const FeedbackButton = () => {
                 title='Improving TT-NN Visualizer'
                 icon={IconNames.INFO_SIGN}
                 isOpen={isDialogOpen}
-                usePortal={false}
+                usePortal
+                className='bp6-dark'
```
dcblundell: There's an enum for this.

PR #1111, opened 2025-12-30
Add indexes to report db

smountenay-tt:
This PR modifies the upload report and remote sync features, to automatically add SQL indexes into the report db when it is saved.

The indexes are intended to optimize fetching the tensor list, when it joins on `input_tensors`, `output_tensors` and `buffers` tables, to get the `max_size_per_bank` value.

Index creation happens in a separate thread, so as not to block the current thread if it takes a while.

### DIFF
--- a/backend/ttnn_visualizer/queries.py
+++ b/backend/ttnn_visualizer/queries.py
@@ -337,6 +337,29 @@ class DatabaseQueries:
         rows = self.query_runner.execute_query(query, [address, operation_id])
         return Buffer(*rows[0]) if rows else None
 
+    def create_indexes(self):
+        """Create indexes on the database for improved query performance."""
+        # SQLite execute() only handles one statement, so we use executescript() for multiple statements
+        query = """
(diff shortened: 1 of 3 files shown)


smountenay-tt:
I changed this PR to draft. I am experiencing a lot of issues loading the 1.3 GB Marko DB and this PR isn't fixing "critical worker timeouts".

The basic index creation is OK, and with small DBs things seem to keep working with this PR. But with the large DB, I am getting critical worker timeouts, even with the indexes created in this DB.

I thought the async code might be the culprit, and in local I removed the async feature in this PR, and just left the index creation. The indexes get created no problem, and it's actually pretty quick. 1-2 seconds with the 1.3 GB DB. But I still get the critical worker timeouts on subsequent API requests.

It seemed OK when I opened the PR, but I had just been testing with small DBs. But I'm not convinced at this point that this PR is helping to load the 1.3 GB Marko DB.

The indexes help 100% when created outside the Flask app, but something is not working right when the indexes are created inside the app. It requires further debugging to get to the source of the critical worker timeouts before we can merge this.

PR #1117, opened 2026-01-06
Refactor device operation handling in OperationList

aidemsined:
Simplifies device operation mapping by using deviceOperationNameList directly instead of filtering device_operations. Updates OperationListPerfData to display device operation names when no performance data is available. Removes unused activePerformanceReportAtom from OperationList.
image

### DIFF
--- a/src/scss/components/Collapsible.scss
+++ b/src/scss/components/Collapsible.scss
@@ -67,6 +67,10 @@
         gap: 5px;
         margin-bottom: 10px;
 
+        &:empty {
+            margin-bottom: 0;
+        }
+
         @include perf.perf-report-colours;
     }
 }


src/components/OperationList.tsx
```diff
@@ -477,25 +471,8 @@ const OperationList = () => {
...
-                                                {/*                }, */}
-                                                {/*            )} */}
-                                                {/*        </ul> */}
-                                                {/*    )} */}
-
-                                                {activePerformanceReport && (
-                                                    <OperationListPerfData operation={operation} />
-                                                )}
+
+                                                <OperationListPerfData operation={operation} />
```
dcblundell: The problem with this change is it will now output the following code if there are no device ops

```
<div class="perf-data">
    <div>
    </div>
</div>
```

dcblundell (changes requested):

Issue #1118, opened 2026-01-06
Remote sync: needs better error logging


aidemsined:
`syncSelectedPerfReportFolder` and `syncSelectedReportFolder` shows alert when exception is thrown. need better error messages

https://tenstorrent.slack.com/archives/C075DUP2ZUM/p1767737051158629

PR #1123, opened 2026-01-08
Fix/SSH Folder Validation

dcblundell:
Checks that `root_folder` is a valid string and bails early if not, and outputs endpoint errors in the UI instead of a generic error.

Screenshot 2026-01-08 at 1 52 16 PM

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1118.

### DIFF
--- a/backend/ttnn_visualizer/sftp_operations.py
+++ b/backend/ttnn_visualizer/sftp_operations.py
@@ -781,7 +781,7 @@ def get_remote_performance_folders(
     else:
         error = "Performance path is not configured for this connection"
         logger.error(error)
-        # raise NoProjectsException(status=ConnectionTestStates.FAILED, message=error)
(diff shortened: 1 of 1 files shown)


backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -786,19 +801,24 @@ def get_remote_profiler_folders(
...
+
+    if remote_connection.profilerPath:
+        remote_config_paths = find_folders_by_files(
+            remote_connection, remote_connection.profilerPath, [TEST_CONFIG_FILE]
+        )
+    else:
+        error = f"No profiler reports found at {remote_connection.profilerPath}"
         logger.info(error)
-        raise NoProjectsException(status=ConnectionTestStates.FAILED, message=error)
+        # raise NoProjectsException(status=ConnectionTestStates.FAILED, message=error)
```
aidemsined: ditto

PR #1155, opened 2026-01-21
Fix/Conflicting peer dependency

### LINKED ISSUE #1156: Graph: Peer dependency conflict
Image

dcblundell:
`vis-data` had a conflicting peer dependency with `vis-network`. I've updated `vis-data` to the latest, major version where they claim there should be no breaking changes for how we're using it. I had to fix a type error after updating, so please check that @aidemsined.

Updated other packages to more recent minor/patch versions at the same time.

Fixes https://github.com/tenstorrent/ttnn-visualizer/issues/1156.


src/components/OperationGraphComponent.tsx
```diff
@@ -79,8 +79,11 @@ const OperationGraph: React.FC<{
 
     if (currentOperationId !== null && !connectedNodeIds.has(currentOperationId)) {
         const val = connectedNodeIds.values().next().value;
-        focusNodeId = val;
```
aidemsined: under which circumstance do you see this happening?

src/definitions/PerfTable.ts
```diff
@@ -62,23 +62,22 @@ export interface PerfTableRow {
     missing?: boolean;
 }
```
aidemsined: extaneous chanage?

PR #1179, opened 2026-01-29
Feature/Indicate tensor tooltip exists

### LINKED ISSUE #1170: Operation details: Unclear there's a tooltip - needs a visual cue
I didn't even know this was a feature until I accidently hovered over it. We should add a visual difference.

Image

dcblundell:
Adds a dotted underline when there's a tooltip.

**Before**
Screenshot 2026-01-29 at 11 44 28

**After**
Screenshot 2026-02-11 at 12 14 35

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1170.


aidemsined:
i think this is useful indeed but would like a little more visibility/visual improvement

aidemsined:
image

```css
text-decoration: underline #74c5df 1.5px;
text-decoration-style: wavy;
```

maybe something like that?

PR #1201, opened 2026-02-04
Change default paths for app db and reports

### LINKED ISSUE #1033: Do not save app and uploaded data to site-packages directory
By default, when you run `ttnn-visualizer` from the pip package or wheel, it saves the "app data" and the uploaded "report data" to the Python `site-packages` directory. This is an anti-pattern and Python packages should not assume they have write access to the `site-packages` directory.

It works if you install it in a virtual env, or in the `~/.local` directory on Ubuntu, but when you install it system-wide, and then run `ttnn-visualizer` as a non-privileged user, you will get an error about not being able to create the SQLite3 DB.

There are cases where you might want to install the app system wide, not in a virtual env or `.local` directory. For example, automated installation in a Docker container. It can make sense to install it system-wide in the container.

In any case, even without the permission issue that comes up when installing system wide, it's just not a great place to be saving the app's data and uploaded user data.

It's currently possible to change the app data directory by setting a `APP_DATA_DIRECTORY` environment variable, but I would like to reduce friction for running the app, especially when installed automatically in containers and server environments, and have saner default `APP_DATA_DIRECTORY` and `REPORT_DATA_DIRECTORY` values.

I think it would be more appropriate to save the data to `~/.ttnn-visualizer` or `~/ttnn-visualizer` when not run in server mode, and maybe somewhere else like `/var/lib/data/ttnn-visualizer` when run as a privileged user.

In the immediate use case prompting creation of this issue, the home directory is a shared network volume, and it may not be the best place to be saving uploaded reports that can be 500 MB or larger. It may even make sense to choose `/tmp/ttnn-visualizer` as the default directory for app data and uploaded reports when running in a container.

smountenay-tt:
This PR changes the default paths used by TTNN Visualizer for its own app database (`ttnn_0.29.0.db`) and for storing uploaded and synced reports.

The reason this is being done is because the existing paths were ok when working from the cloned repo, but less than ideal when installing from `pip`.

Prior to this PR, the `ttnn-visualizer` command, installed from `pip` and run as root (say it was set up in a Dockerfile like this), would save the data to:

`/usr/local/lib/python3.10/dist-packages/ttnn_0.29.0.db`

That's not an appropriate spot at all. The reports would go to:

`/usr/local/lib/python3.10/dist-packages/ttnn_visualizer/data`

That could be ok as root, but the problem is, if you installed `ttnn-visualizer` system wide instead of in a venv, then ran it as a regular user, the user would not have access to do writes to that directory and it would fail.

The solution implemented in this PR two different directories:

`/var/lib/ttnn-visualizer/` - when run as root (perhaps as a service on a server)
`~/.ttnn-visualizer/` - when run as a non-privileged user

There is code that does a migration if it finds the data in the old location, and moves it to the new location.

[Closes #1033]

### DIFF
--- a/backend/ttnn_visualizer/utils.py
+++ b/backend/ttnn_visualizer/utils.py
@@ -9,6 +9,7 @@ import logging
 import os
 import re
 import shutil
+import sqlite3
 import sys
 import time
 from functools import wraps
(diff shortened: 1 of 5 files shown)


backend/ttnn_visualizer/utils.py
```diff
@@ -23,18 +24,271 @@
...
 
     Returns:
         Path to the app data directory
     """
+    # Priority 1: TT_METAL_HOME mode
     if tt_metal_home and tt_metal_home.strip():
         return str(Path(tt_metal_home).expanduser() / "generated" / "ttnn-visualizer")
-    return application_dir
+
+    # Priority 2: Explicit environment variable
```
dcblundell, line 46: Shouldn't an explicit env var be the first priority?

smountenay-tt: They aren't really compatible. When it's sees `TT_METAL_HOME` is set, it gets the data straight from the `generated` directory, and doesn't use `APP_DATA_DIRECTORY` at all.

I don't see a problem with `TT_METAL_HOME` having priority over `APP_DATA_DIRECTORY` or checking if its in a container.

Although we could add checks to ensure they are not both set, and abort with an error if they are, since that is an incompatible configuration and `APP_DATA_DIRECTORY` would be silently ignored right now.

dcblundell:
Mostly for us, but we have to copy new reports to `/Users/<your_user>/.ttnn-visualizer/reports/`.

dcblundell (approved):

PR #1212, opened 2026-02-10
Fix edge case error when reloading the page

### LINKED ISSUE #1204: Can't access lexical declaration 'calculatePlotZoomRange' before initialization
Refreshing Operation Details with a loaded report can throw this error.

Image

Image

aidemsined:
Introduce a top-level mutable `details` variable and use it when initializing the plot zoom range. The effect now checks `details !== null` and includes `details` in its dependency array to avoid running with stale/undefined data. The local `details` construction was changed from a `const` to an assignment to populate that outer variable, and a previously-commented zoom-reset snippet was left commented out in the toggle handler.

closes #1204


dcblundell:
I don't know if this is a side effect but when you reload Operation Details it will say "Not Found Operation 109" in the page title (109 because I'm using http://localhost:5173/operations/109)

dcblundell (approved):

PR #1216, opened 2026-02-10
Use unique keys for op input/output lists

### LINKED ISSUE #1215: DeviceOperations component - react keys
```
installHook.js:1 Warning: Encountered two children with the same key, `2 7`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version. Error Component Stack
    at h4 (<anonymous>)
    at span (<anonymous>)
    at Collapsible (Collapsible.tsx:25:5)
    at div (<anonymous>)
    at div (<anonymous>)
    at DeviceOperationsFullRender (DeviceOperationsFullRender.tsx:385:9)
```

aidemsined:
closes #1215

### DIFF
--- a/src/components/operation-details/DeviceOperationsFullRender.tsx
+++ b/src/components/operation-details/DeviceOperationsFullRender.tsx
@@ -192,14 +192,14 @@ function useDeviceOperationsFullRenderModel(args: {
         [details],
     );
 
-    const renderOperations = useCallback(
-        (ops: Node[]) => {
+    const renderNodes = useCallback(
+        (nodes: Node[]) => {
             const deviceOpList: Node[] = [];
             const stack: JSX.Element[][] = [];
             const output: JSX.Element[] = [];
             let consecutiveCBsOutput = false;
 
-            ops.forEach((node, index) => {
+            nodes.forEach((node, index) => {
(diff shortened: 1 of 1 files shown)


src/components/operation-details/DeviceOperationsFullRender.tsx
```diff
@@ -224,19 +224,19 @@ function useDeviceOperationsFullRenderModel(args: {
                                 icon={IconNames.CUBE_ADD}
                             />
                             {opName} <DeviceID _deviceId={node.operation?.params.device_id} /> (
-                            {node.operation?.inputs.map((arg) => (
+                            {node.operation?.inputs.map((arg, ind) => (
```
dcblundell: I find the naming weird here. If you're mapping inputs then I feel like `arg` should be `input`, and `ind` is a shorthand for `index` I've never seen.

dcblundell (approved):

src/components/operation-details/DeviceOperationsFullRender.tsx
aidemsined: agree and updated

PR #1218, opened 2026-02-11
Improve SSH key handling

### LINKED ISSUE #1189: Remote: SSH connection UX
“it is not possible to select password or key to use for SSH connection. And I have to write passcode for key 5 times in the Putty window where I run ttnn-visualizer. So using ""Remote sync"" is quite inconvenient now.”

smountenay-tt:
This PR improves SSH invocation in the app, to ensure it is called with consistent parameters. It also adds the ability to specific a custom private key file, and improves the documentation to emphasize the need to be able to connect to the remote host without a password or passphrase.

Prior to this PR, there was not a common method to call the `ssh` command. In most of the spots where `ssh` was called, it was being called with parameters to ensure it did not prompt for a password, but there were a few places that were missed and where it did prompt for a password in the terminal. It was unintended that users could ever be prompted for a password in the Flask terminal, and with this PR, it should prevent the user ever being prompted in the terminal for a password. Moving to a command function for calling `ssh` helps to ensure this.

A new feature has been added to allow the user to set the key to use when connecting to the remote host. Previously it was assumed that it was already configured in the `~/.ssh/config` file, but now the user can specify the key in the UI.

The docs and error messages were updated to reflect that the private key must be added to `ssh-agent` also. The reason this is necessary is because if the private key file is locked, there is no way to enter the passphrase in the web ui. We don't want to handle passwords and passphrases in the Flask app, but it wasn't obvious that the key also needs to be added to `ssh-agent` to prevent being prompted for a passphrase. It was never intended for a user to enter a passphrase, but it was an side-effect of the paramiko to cli ssh refactor that the user could get prompted in the Flask terminal if the key had a passphrase. This should not happen now.

If a key file is set when configuring the remote connection, it will exclusively use that key, and ignore the `~/.ssh/config` file. I did that because I was getting confusing errors during dev, which made it look like it wasn't using the key I was specifying, when really it was failing due to other issues from the user's SSH config.

[Closes #1189]

### DIFF
--- a/backend/ttnn_visualizer/sftp_operations.py
+++ b/backend/ttnn_visualizer/sftp_operations.py
@@ -4,6 +4,7 @@
 
 import json
 import logging
+import os
 import subprocess
 import time
 from pathlib import Path
@@ -36,7 +37,7 @@ def _ssh_cmd_prefix(remote_connection: RemoteConnection) -> List[str]:
     cmd = ["ssh"]
     identity = (getattr(remote_connection, "identityFile", None) or "").strip()
     if identity:
-        cmd.extend(["-F", "/dev/null"])
+        cmd.extend(["-F", os.devnull])
     cmd.extend(["-o", "BatchMode=yes", "-o", "PasswordAuthentication=no"])
     if identity:
         cmd.extend(["-o", "IdentitiesOnly=yes", "-i", identity])
@@ -51,7 +52,7 @@ def _sftp_cmd_prefix(remote_connection: RemoteConnection) -> List[str]:
     cmd = ["sftp"]
     identity = (getattr(remote_connection, "identityFile", None) or "").strip()
     if identity:
-        cmd.extend(["-F", "/dev/null"])
+        cmd.extend(["-F", os.devnull])
     cmd.extend(["-o", "BatchMode=yes", "-o", "PasswordAuthentication=no"])
     if identity:
         cmd.extend(["-o", "IdentitiesOnly=yes", "-i", identity])
--- a/backend/ttnn_visualizer/sftp_operations.py
+++ b/backend/ttnn_visualizer/sftp_operations.py
@@ -127,7 +127,7 @@ def resolve_file_path(remote_connection, file_path: str) -> str:
     """
     if "*" in file_path:
         # Build SSH command to list files matching the pattern (never prompts for password)
-        ssh_cmd = _ssh_cmd_prefix(remote_connection) + [f"ls -1 {file_path}"]
+        ssh_cmd = _ssh_cmd_prefix(remote_connection) + [f"ls -1 '{file_path}'"]
 
         try:
             result = subprocess.run(ssh_cmd, capture_output=True, text=True, check=True)
--- a/backend/ttnn_visualizer/sftp_operations.py
+++ b/backend/ttnn_visualizer/sftp_operations.py
@@ -14,19 +14,96 @@ from flask import current_app
 from ttnn_visualizer.decorators import remote_exception_handler
 from ttnn_visualizer.enums import ConnectionTestStates
 from ttnn_visualizer.exceptions import (
+    AuthenticationException,
     NoProjectsException,
+    NoValidConnectionsError,
     RemoteConnectionException,
     SSHException,
 )
 from ttnn_visualizer.models import RemoteConnection, RemoteReportFolder
 from ttnn_visualizer.sockets import FileProgress, FileStatus, emit_file_status
-from ttnn_visualizer.ssh_client import SSHClient
+from ttnn_visualizer.ssh_client import _SSH_AUTH_MESSAGE, SSHClient
 from ttnn_visualizer.utils import update_last_synced
 
 logger = logging.getLogger(__name__)
 
 TEST_CONFIG_FILE = "config.json"
 TEST_PROFILER_FILE = "profile_log_device.csv"
+
+
+def _ssh_cmd_prefix(remote_connection: RemoteConnection) -> List[str]:
+    """Build SSH command prefix (never prompts for password). Includes BatchMode=yes and optional identity file."""
+    cmd = [
+        "ssh",
+        "-o",
+        "BatchMode=yes",
+        "-o",
+        "PasswordAuthentication=no",
+    ]
+    if getattr(remote_connection, "identityFile", None):
+        cmd.extend(["-i", remote_connection.identityFile])
+    if remote_connection.port != 22:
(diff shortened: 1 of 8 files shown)


backend/ttnn_visualizer/ssh_client.py
```diff
@@ -19,6 +19,14 @@
 
 logger = logging.getLogger(__name__)
 
+# User-facing message for SSH auth failures (key-based auth required, no password).
+_SSH_AUTH_MESSAGE = (
```
aidemsined: i find `_SSH_AUTH_MESSAGE` a deceiving name. `_SSH_AUTH_FAILED_MESSAGE` or something along those line would be more appropriate

backend/ttnn_visualizer/ssh_client.py
```diff
@@ -28,26 +36,40 @@ class SSHClient:
 
     def __init__(self, connection: RemoteConnection):
         self.connection = connection
+        identity = getattr(connection, "identityFile", None)
+        logger.info(
```
dcblundell: I tested this on a remote machine and it was fine, but I have a local docker container I use for remote feature testing and I noticed this and I'm unsure why it doesn't return an error with my fake path.

```
INFO:ttnn_visualizer.views:test_remote_folder request identityFile='/not/a/real/path', connection.identityFile='/not/a/real/path'
INFO:ttnn_visualizer.ssh_client:SSHClient connection.identityFile='/not/a/real/path' (will use only this key if set)
INFO:ttnn_visualizer.ssh_client:Testing SSH connection to ctr-dblundell@localhost on port 2222
DEBUG:ttnn_visualizer.ssh_client:Executing SSH command on localhost: echo 'SSH connection test'
INFO:ttnn_visualizer.ssh_client:SSHClient connection.identityFile='/not/a/real/path' (will use only this key if set)
DEBUG:ttnn_visualizer.ssh_client:Checking if remote path exists: /home/ctr-dblundell/reports
DEBUG:ttnn_visualizer.ssh_client:Executing SSH command on localhost: test -e '/home/ctr-dblundell/reports'
```

src/components/report-selection/RemoteConnectionDialog.tsx
```diff
@@ -171,6 +171,24 @@ const RemoteConnectionDialog: FC<RemoteConnectionDialogProps> = ({
                     />
                 </FormGroup>
 
+                <FormGroup
+                    label='SSH identity file (optional)'
+                    subLabel='Path to your private key on this machine (e.g. ~/.ssh/id_ed25519). Leave empty for default. The matching public key must be in ~/.ssh/authorized_keys on the server. If the key has a passphrase, run: ssh-add /path/to/key (once, in the same environment where you start this app).'
```
dcblundell: This is quite long, do we need to explain all of this here? It feels like we're hand holding to an excessive degree. I think the docs should cover additional technical information.

Perhaps just the first part? 

> "Path to your private key on this machine (e.g. ~/.ssh/id_ed25519). Leave empty for default."

dcblundell (changes requested):

backend/ttnn_visualizer/ssh_client.py
dcblundell: This is probably not very important, I'm just curious.

docs/src/remote-sync.md
```diff
@@ -11,41 +11,81 @@ TT-NN Visualizer supports syncing data from remote servers via SSH. This feature
 
 ## SSH Setup
 
+### How SSH key authentication works (overview)
```
dcblundell, line 14: Great docs update! 👍

src/components/report-selection/RemoteConnectionDialog.tsx
smountenay-tt: ok, i'll change it. My intent was to make it more obvious what you need to do since some people don't know the process and may not go look at the docs.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -14,13 +14,15 @@
...
     NoProjectsException,
+    NoValidConnectionsError,
     RemoteConnectionException,
     SSHException,
 )
 from ttnn_visualizer.models import RemoteConnection, RemoteReportFolder
 from ttnn_visualizer.sockets import FileProgress, FileStatus, emit_file_status
-from ttnn_visualizer.ssh_client import SSHClient
+from ttnn_visualizer.ssh_client import _SSH_AUTH_MESSAGE, SSHClient
 from ttnn_visualizer.utils import update_last_synced
```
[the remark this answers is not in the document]
smountenay-tt: name changed

src/components/report-selection/RemoteConnectionDialog.tsx
smountenay-tt: I shortened it to just the first part.

backend/ttnn_visualizer/ssh_client.py
smountenay-tt: @dcblundell I tried it out with CLI `ssh` command outside of the app, and it does in fact work with an invalid path. The CLI `ssh` command treats it as a warning, not an error. If we want to catch this scenario, we could explicitly check the path is valid ourselves in the Python code before passing it to the `ssh` command.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -29,6 +31,77 @@
...
+    """Build SFTP command prefix (never prompts for password). Includes BatchMode=yes and optional identity file."""
+    cmd = ["sftp"]
+    identity = (getattr(remote_connection, "identityFile", None) or "").strip()
+    if identity:
+        cmd.extend(["-F", "/dev/null"])
+    cmd.extend(["-o", "BatchMode=yes", "-o", "PasswordAuthentication=no"])
+    if identity:
+        cmd.extend(["-o", "IdentitiesOnly=yes", "-i", identity])
+    if remote_connection.port != 22:
+        cmd.extend(["-P", str(remote_connection.port)])
```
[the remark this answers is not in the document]
smountenay-tt: I'm not going to do this one right now.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -29,6 +31,77 @@
...
+    cmd.extend(["-b", "-", f"{remote_connection.username}@{remote_connection.host}"])
+    return cmd
```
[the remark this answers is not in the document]
smountenay-tt: I will not do this one right now, but we can look at it after.

dcblundell (approved):

PR #1234, opened 2026-02-17
Feature/Check for cluster description locally

### LINKED ISSUE #1190: Missing cluster description is logged as a console error
Creates additional noise by logging this as an error when it is optional.

```
"No problems. ttnn-visualizer printed some exceptions to stderr ("ERROR:ttnn_visualizer.app:Exception on /api/cluster-descriptor [GET]", ..., "TypeError: Dict key must be str"), but it did not affect viewing the report.”
```

dcblundell:
Discussed that providing more error details is not useful, but I have added a toast warning if there is not cluster description at all (very unlikely).

Most of this PR refactors where we look for the cluster description. Instead of checking a `/tmp/umd_*` folder we always look at the synced folder on the local machine now.

**Python changes**
- Removes get_cluster_desc_path
- Simplified get_cluster_desc
- /cluster-descriptor now uses the simplified get_cluster_desc to look in the active report folder for cluster_description.yaml

**New warning toast**
Screenshot 2026-02-17 at 15 12 55

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1190.

### DIFF
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -1155,7 +1155,7 @@ def get_cluster_descriptor(instance: Instance):
     except Exception as e:
         return (
             jsonify({"error": f"An unexpected error occurred: {str(e)}"}),
-            HTTPStatus.NOT_FOUND,
+            HTTPStatus.INTERNAL_SERVER_ERROR,
         )
 
 
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -1133,24 +1133,21 @@ def get_remote_folders_performance():
 @with_instance
 def get_cluster_descriptor(instance: Instance):
     try:
-        cluster_desc_file = get_cluster_desc(instance)
+        cluster_desc = get_cluster_desc(instance)
 
-        if not cluster_desc_file:
+        if not cluster_desc:
             return (
                 jsonify({"error": "cluster_descriptor.yaml not found"}),
                 HTTPStatus.NOT_FOUND,
             )
 
-        return jsonify(cluster_desc_file), HTTPStatus.OK
+        return jsonify(cluster_desc), HTTPStatus.OK
 
     except yaml.YAMLError as e:
         return (
             jsonify({"error": f"Failed to parse YAML: {str(e)}"}),
             HTTPStatus.BAD_REQUEST,
         )
-
-    except RemoteConnectionException as e:
-        return jsonify({"error": e.message}), e.http_status
 
     except Exception as e:
         return (
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -1132,36 +1132,31 @@ def get_remote_folders_performance():
 @api.route("/cluster-descriptor", methods=["GET"])
 @with_instance
 def get_cluster_descriptor(instance: Instance):
-    if instance.remote_connection:
-        try:
-            cluster_desc_file = get_cluster_desc(instance.remote_connection)
-            if not cluster_desc_file:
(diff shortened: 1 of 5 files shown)


backend/ttnn_visualizer/views.py
```diff
@@ -1132,36 +1132,31 @@ def get_remote_folders_performance():
...
-        except yaml.YAMLError as e:
-            return jsonify({"error": f"Failed to parse YAML: {str(e)}"}), 400
+    except RemoteConnectionException as e:
+        return jsonify({"error": e.message}), e.http_status
 
-    return jsonify({"error": "Cluster descriptor not found"}), 404
+    except Exception as e:
+        return (
+            jsonify({"error": f"An unexpected error occurred: {str(e)}"}),
+            HTTPStatus.NOT_FOUND,
```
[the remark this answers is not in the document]
dcblundell: Ok.

backend/ttnn_visualizer/views.py
```diff
@@ -1132,36 +1132,31 @@ def get_remote_folders_performance():
...
+            return (
+                jsonify({"error": "cluster_descriptor.yaml not found"}),
+                HTTPStatus.NOT_FOUND,
+            )
 
-        except Exception as e:
-            return jsonify({"error": f"An unexpected error occurred: {str(e)}"}), 500
-    else:
-        local_path = get_cluster_descriptor_path(instance)
+        return jsonify(cluster_desc_file), HTTPStatus.OK
```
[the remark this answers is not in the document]
dcblundell: This is the pattern we're using elsewhere.

backend/ttnn_visualizer/views.py
```diff
@@ -1132,36 +1132,31 @@ def get_remote_folders_performance():
...
-            cluster_desc_file = get_cluster_desc(instance.remote_connection)
-            if not cluster_desc_file:
-                return jsonify({"error": "cluster_descriptor.yaml not found"}), 404
-            yaml_data = yaml.safe_load(cluster_desc_file.decode("utf-8"))
-            return jsonify(yaml_data), 200
-
-        except yaml.YAMLError as e:
-            return jsonify({"error": f"Failed to parse YAML: {str(e)}"}), 400
+    try:
+        cluster_desc_file = get_cluster_desc(instance)
```
smountenay-tt: Nit: The `_file` suffix is misleading because I'd expect something with that name to be a file or a path, and in this case it's already loaded into a Python object. Personally I'd just call it `cluster_desc`.

backend/ttnn_visualizer/views.py
```diff
@@ -1132,36 +1132,31 @@ def get_remote_folders_performance():
...
+            HTTPStatus.BAD_REQUEST,
+        )
 
-        try:
-            with open(local_path) as cluster_desc_file:
-                yaml_data = yaml.safe_load(cluster_desc_file)
-                return jsonify(yaml_data)  # yaml_data is not compatible with orjson
-        except yaml.YAMLError as e:
-            return jsonify({"error": f"Failed to parse YAML: {str(e)}"}), 400
+    except RemoteConnectionException as e:
```
smountenay-tt: `RemoteConnectionException` can't be raised anymore in `get_cluster_desc` so we don't need to catch it here now.

backend/ttnn_visualizer/views.py
dcblundell: Fair.

backend/ttnn_visualizer/views.py
dcblundell: Removed.

smountenay-tt (approved):

aidemsined (approved):
we shoudl update the docs to "how to generate this cluster descriptor if its missing" as a separate PR

PR #1249, opened 2026-02-19
Interim cpp output support - add defensive checks for invalid inputs

aidemsined:
Add early-return guards to prevent runtime errors when unexpected or missing data is encountered. Changes:

- src/hooks/useAPI.tsx: return empty string if parseFileOperationIdentifier receives a falsy stackTrace; ensure getDeviceOperationNameList returns [] when operation.device_operations is not an array.
- src/model/OperationDetails.ts: ensure preprocessConnections returns [] when ops is not an array.

These safeguards avoid crashes when backend responses or stack traces are empty or malformed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ttnn-visualizer",
   "private": true,
-  "version": "0.73.0",
+  "version": "0.73.1",
   "type": "module",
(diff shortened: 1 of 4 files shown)


dcblundell (approved):
Seems fine.

PR #1273, opened 2026-02-25
Feature/Simplified formatCell performance functions 

### LINKED ISSUE #1267: Performance: formatCell function can be simplified now that we used typed data

dcblundell:
- Fixed formatting inconsistencies
- Moved high dispatch threshold definition to a shared const and used it in various new places
- Simplified formatting code

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1267.

### DIFF
--- a/src/definitions/PerfTable.ts
+++ b/src/definitions/PerfTable.ts
@@ -177,8 +177,8 @@ export const tableColumns: TableColumn[] = [
     { label: 'Cores', key: ColumnHeaders.cores, colour: 'green', sortable: true },
     { label: 'DRAM', key: ColumnHeaders.dram, colour: 'yellow', unit: 'GB/s', decimals: 1, sortable: true },
     { label: 'DRAM %', key: ColumnHeaders.dram_percent, colour: 'yellow', unit: '%', decimals: 1, sortable: true },
-    { label: 'FLOPs', key: ColumnHeaders.flops, unit: 'TFLOPs', decimals: 1, sortable: true },
-    { label: 'FLOPs %', key: ColumnHeaders.flops_percent, unit: '%', decimals: 1, sortable: true },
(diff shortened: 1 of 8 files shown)


src/definitions/PerfTable.ts
```diff
@@ -177,8 +177,8 @@ export const tableColumns: TableColumn[] = [
     { label: 'Cores', key: ColumnHeaders.cores, colour: 'green', sortable: true },
     { label: 'DRAM', key: ColumnHeaders.dram, colour: 'yellow', unit: 'GB/s', decimals: 1, sortable: true },
     { label: 'DRAM %', key: ColumnHeaders.dram_percent, colour: 'yellow', unit: '%', decimals: 1, sortable: true },
-    { label: 'FLOPs', key: ColumnHeaders.flops, unit: 'TFLOPs', decimals: 1, sortable: true },
-    { label: 'FLOPs %', key: ColumnHeaders.flops_percent, unit: '%', decimals: 1, sortable: true },
+    { label: 'FLOPS', key: ColumnHeaders.flops, unit: 'TFLOPS', decimals: 1, sortable: true },
+    { label: 'FLOPS %', key: ColumnHeaders.flops_percent, unit: '%', decimals: 1, sortable: true },
```
aidemsined, line 181: maybe remove the space here as well

aidemsined, line 181: maybe remove space here as well

aidemsined (approved):
approved with a comment to check

src/definitions/PerfTable.ts
dcblundell: I think the label reads better with a space as we're not giving a percentage value, we're describing the column values.

dcblundell: I think the label reads better with a space as we're not giving a percentage value, we're describing the column values.

PR #1277, opened 2026-02-26
Feature/Update buffer table columns

### LINKED ISSUE #1270: Buffers: Update table columns
Combine address columns and perhaps include layout information.

dcblundell:
- Added Shape, Data Type and Buffer Layout columns
- Removed Address (hex) column and folded it into Address
- Adjusted column sizing and filter/sort options

**Before**
Screenshot 2026-02-26 at 14 16 25

**After**
Screenshot 2026-02-26 at 14 16 16

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1270.

### DIFF
--- a/src/definitions/BufferSummary.ts
+++ b/src/definitions/BufferSummary.ts
@@ -12,6 +12,7 @@ export enum TAB_IDS {
     DRAM = 'DRAM',
 }
 
+// Examine current data model https://github.com/tenstorrent/ttnn-visualizer/issues/1286
 export enum ColumnKeys {
     operation_id = 'operation_id',
     tensor_id = 'tensor_id',
--- a/src/definitions/BufferSummary.ts
+++ b/src/definitions/BufferSummary.ts
@@ -12,6 +12,18 @@ export enum TAB_IDS {
     DRAM = 'DRAM',
 }
 
+export enum ColumnKeys {
+    operation_id = 'operation_id',
+    tensor_id = 'tensor_id',
+    address = 'address',
+    size = 'size',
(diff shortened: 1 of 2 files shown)


src/definitions/BufferSummary.ts
```diff
@@ -12,6 +12,18 @@ export enum TAB_IDS {
     DRAM = 'DRAM',
 }
 
+export enum ColumnKeys {
```
aidemsined, line 16: it feels like the property names should have been TS friendly and not py based, could be wrong

src/definitions/BufferSummary.ts
```diff
@@ -21,62 +33,77 @@ export interface ColumnDefinition {
 }
 
 export enum ColumnHeaders {
-    operation_id = 'Operation',
-    tensor_id = 'Tensor',
     address = 'Address',
```
aidemsined, line 37: ditto

aidemsined (changes requested):

src/definitions/BufferSummary.ts
dcblundell: Comes from `BufferData` so 🤷 
https://github.com/tenstorrent/ttnn-visualizer/blob/bec346e7a24b60d73117aeb85f8eb2e7d271325a/src/components/buffer-summary/BufferSummaryTable.tsx#L33

aidemsined: the values - sure, keys - not so much

dcblundell: Added an issue for this.

PR #1280, opened 2026-02-27
Show consumer count and high-consumer warning

### LINKED ISSUE #1274: Tensors: when marking as a high consumed tensor shoud show number of consumers
this view requires a few clicks to get the information that is already available here but not rendered
Image

aidemsined:
image

Include the actual consumer count in the tooltip and add an inline warning paragraph when a tensor's consumers exceed MAX_NUM_CONSUMERS.

closes #1274

### DIFF
--- a/src/components/TensorList.tsx
+++ b/src/components/TensorList.tsx
@@ -449,7 +449,7 @@ const TensorList = () => {
                                                 >
                                                     {tensor.consumers.length > MAX_NUM_CONSUMERS ? (
                                                         <Tooltip
-                                                            content={`Unusually high number of consumers - ${tensor.consumers.length}`}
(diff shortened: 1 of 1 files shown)


src/components/TensorList.tsx
```diff
@@ -449,7 +449,7 @@ const TensorList = () => {
                                                 >
                                                     {tensor.consumers.length > MAX_NUM_CONSUMERS ? (
                                                         <Tooltip
-                                                            content='Unusually high number of consumers'
+                                                            content={`Unusually high number of consumers - ${tensor.consumers.length}`}
```
dcblundell: Maybe `Unusually high number of consumers (33)`?

dcblundell (approved):

PR #1296, opened 2026-03-06
Feature/Align disabled tooltip logic

### LINKED ISSUE #1281: Tooltips: Align disabled behaviours
Later versions of Blueprint introduced a `disabled` prop so instead of having something like `content={isLoading ? 'Loading...' : ''}` you can use `disabled` to control the tooltip which is cleaner.

dcblundell:
Converts conditional message logic to use `disabled` prop unless there's an alternative tooltip message.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1281.

### DIFF
--- a/src/components/OperationGraphComponent.tsx
+++ b/src/components/OperationGraphComponent.tsx
@@ -426,12 +426,8 @@ const OperationGraph: React.FC<{
                 <div className='operation-graph-nav'>
                     <Tooltip
                         placement={PopoverPosition.TOP}
-                        content={
-                            previousOperation
-                                ? `Go to previous operation ${previousOperation}`
-                                : 'No previous operation'
-                        }
-                        disabled={isLoading}
+                        content={`Go to previous operation ${previousOperation}`}
+                        disabled={isLoading || !previousOperation}
                     >
                         <Button
                             icon={IconNames.ArrowLeft}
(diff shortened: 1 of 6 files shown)


src/components/OperationGraphComponent.tsx
```diff
@@ -426,12 +426,8 @@ const OperationGraph: React.FC<{
                 <div className='operation-graph-nav'>
                     <Tooltip
                         placement={PopoverPosition.TOP}
-                        content={
-                            previousOperation
-                                ? `Go to previous operation ${previousOperation}`
-                                : 'No previous operation'
-                        }
-                        disabled={isLoading}
+                        content={`Go to previous operation ${previousOperation}`}
```
dcblundell, line 429: You might argue against this change.

aidemsined: i will. graphs dont loop over

aidemsined (changes requested):

src/components/OperationGraphComponent.tsx
dcblundell: The tooltip is disabled when there is no previous operation.

aidemsined: right

aidemsined (approved):

PR #1299, opened 2026-03-07
Add support for tensors table size column

### LINKED ISSUE #1298: Add support for reading tensor size from tensors table
The change to move report generation down to C++ in TTNN, allows us to put the tensor size directly into the tensors table, since that information is now directly available.

The TTNN frontend should be modified to work with report databases that have a `tensors.size` column going forward, and maintain backwards compatibility with report databases that do not have it.

smountenay-tt:
Add support for a `size` column on `tensors` table. Backwards compatibility is being maintained for report dbs that do not have a `tesors.size` column.

This is being done in support of moving TT-Metal reporting from Python layer to C++. With reports generated from C++, the tensor size is available to put into the report db, without having to relying on the queries that were previously being used in TTNN Visualizer to derive the value.

[Closes #1298]

### DIFF
--- a/backend/ttnn_visualizer/queries.py
+++ b/backend/ttnn_visualizer/queries.py
@@ -196,39 +196,68 @@ class DatabaseQueries:
     def query_tensors(
         self, filters: Optional[Dict[str, Any]] = None
     ) -> Generator[Tensor, None, None]:
-        # Check if device_tensors table exists
+        # Check if tensors table has a size column (new schema) or we need to infer from joins
+        tensor_columns = self._get_table_columns("tensors")
+        size_on_tensors = "size" in tensor_columns
+
         device_tensors_exists = self._check_table_exists("device_tensors")
 
-        # Build the base query with joins to get size and optionally device_tensors
-        if device_tensors_exists:
-            query = """
-                SELECT 
-                    t.*, 
-                    b.max_size_per_bank as size,
-                    GROUP_CONCAT(dt.device_id || ':' || dt.address, ',') as device_tensors_data
-                FROM tensors t
-                LEFT JOIN input_tensors it ON it.tensor_id = t.tensor_id
-                LEFT JOIN output_tensors ot ON ot.tensor_id = t.tensor_id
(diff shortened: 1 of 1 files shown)


aidemsined (approved):

backend/ttnn_visualizer/queries.py
```diff
@@ -196,39 +196,68 @@ def query_buffer_pages(
     def query_tensors(
         self, filters: Optional[Dict[str, Any]] = None
     ) -> Generator[Tensor, None, None]:
-        # Check if device_tensors table exists
+        # Check if tensors table has a size column (new schema) or we need to infer from joins
+        tensor_columns = self._get_table_columns("tensors")
+        size_on_tensors = "size" in tensor_columns
```
dcblundell, line 201: Maybe this is more of a JS convention, but naming booleans with a verb like "has_size_on_tensors" gives you the ability to understand this is a true/false statement without reading the rest of the code. The current name could indicate a boolean but could also be something else.

backend/ttnn_visualizer/queries.py
```diff
@@ -196,39 +196,68 @@ def query_buffer_pages(
...
-                LEFT JOIN output_tensors ot ON ot.tensor_id = t.tensor_id
-                LEFT JOIN buffers b ON b.operation_id = COALESCE(it.operation_id, ot.operation_id)
-                    AND t.address = b.address
-                    AND t.device_id = b.device_id
-                LEFT JOIN device_tensors dt ON dt.tensor_id = t.tensor_id
-                WHERE 1=1
-            """
+        if size_on_tensors:
+            # New schema: size is on tensors table — simple select without buffer/op joins
+            if device_tensors_exists:
```
dcblundell, line 207: Actually this is a good example of the comment above, I can understand this is a boolean immediately. Again, might be a JS convention but the verb tends to come first, e.g. "onClick", "hasFlag", "isProcessed".

dcblundell (approved):
Made some non-blocking observations that I would make if this were JS.

PR #1310, opened 2026-03-11
fix memory zoom slider causing horizontal page scroll

### LINKED ISSUE #1309: Memory zoom range controls causes page horizontal scroll
Image

aidemsined:
Wrap the RangeSlider in a .zoom-range-wrap div and add CSS to contain its overflow

closes #1309


dcblundell (approved):
Screenshot or something would be nice.

PR #1319, opened 2026-03-13
Support multiple deallocate operation names

### LINKED ISSUE #1318: Expand deallocate filter
deallocate op name string matching needs to be expanded to support various ways it can be reported

aidemsined:
Add a canonical list of deallocate operation names and use it everywhere instead of a single hardcoded string. Created src/definitions/Deallocate.ts exporting DEALLOCATE_OP_NAME_LIST, updated OperationGraphComponent and useAPI hooks to check includes(...) (case-insensitive) so both 'ttnn.deallocate' and 'ttnn::deallocate' are recognized.

closes #1318


dcblundell (approved):
Seems good.

PR #1333, opened 2026-03-19
Chore/Vite upgrade

dcblundell:
The one bump was that `react-plotly-js` has some kind of flaw in its exporting syntax, hence the new PlotComponent introduced here.


aidemsined (approved):

aidemsined:
lets get this merged right after today's release

PR #1403, opened 2026-04-16
Fix/Horizontal scrollbar always on performance table

### LINKED ISSUE #1402: Performance: Table always shows horizontal scrollbar
## Summary
The Performance table always shows a horizontal scrollbar, even when the visible columns should fit without horizontal overflow.

## Steps to reproduce
1. Open a performance report.
2. Go to the table view.
3. Observe the horizontal scrollbar at the bottom of the table.

## Expected behavior
The horizontal scrollbar only appears when table content actually overflows horizontally.

## Actual behavior
A horizontal scrollbar is always visible.

## Screenshot
_Attaching screenshot in follow-up comment from local capture._

dcblundell:
Screenshot 2026-04-16 at 15 23 15

Fixes https://github.com/tenstorrent/ttnn-visualizer/issues/1402.

### DIFF
--- a/src/scss/components/BufferSummaryTable.scss
+++ b/src/scss/components/BufferSummaryTable.scss
@@ -10,11 +10,11 @@
 
 .buffer-summary-table {
     display: flex;
-    height: 600px;
     flex-direction: column;
+    height: 600px;
     width: 100%;
-    max-width: 1120px;
-    position: relative;
+    max-width: 1200px;
+    overflow-y: auto;
 
     .aside-container {
         display: flex;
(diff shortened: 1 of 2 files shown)


src/scss/components/BufferSummaryTable.scss
```diff
@@ -9,11 +9,12 @@
...
 
 .buffer-summary-table {
-    display: inline-flex;
+    display: flex;
     height: 600px;
     flex-direction: column;
-    overflow: auto;
-    max-width: 100%;
+    width: 100%;
+    max-width: 1120px;
```
aidemsined: do we have to limit max width?

aidemsined (approved):

PR #1405, opened 2026-04-16
Make memory zoom options null-safe - prevent blocking error - support for Quasar buffer summary zoom

### LINKED ISSUE #1404: Quasar support: buffer summary broken

aidemsined:
Handle empty chart segments and avoid undefined accesses for zoomed memory options. The segmentedChartData mapping now returns null for empty segments and filters them out. Consume zoomedMemoryOptions with optional chaining and default values (??) for keys, memorySize, plotZoomRange, memoryStart/memoryEnd, and memoryPadding to prevent runtime errors when segments or zoom options are missing.

closes #1404

### DIFF
--- a/src/components/buffer-summary/BufferSummaryPlotRendererDRAM.tsx
+++ b/src/components/buffer-summary/BufferSummaryPlotRendererDRAM.tsx
@@ -81,17 +81,22 @@ function BufferSummaryPlotRendererDRAM({
 
     const zoomedMemoryOptions = useMemo(
         () =>
-            segmentedChartData.map((segment) => {
-                const buffers = segment.flatMap((op) => op.buffers);
-                const zoomStart = buffers[0].address;
-                const zoomEnd = buffers[buffers.length - 1].address + buffers[buffers.length - 1].size;
+            segmentedChartData
+                .map((segment) => {
+                    if (segment.length > 0) {
+                        const buffers = segment.flatMap((op) => op.buffers);
+                        const zoomStart = buffers[0].address;
+                        const zoomEnd = buffers[buffers.length - 1].address + buffers[buffers.length - 1].size;
 
-                return {
-                    start: zoomStart,
-                    end: zoomEnd,
-                    padding: (zoomEnd - zoomStart) * MEMORY_ZOOM_PADDING_RATIO,
-                };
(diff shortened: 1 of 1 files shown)


src/components/buffer-summary/BufferSummaryPlotRendererDRAM.tsx
```diff
@@ -81,17 +81,22 @@ function BufferSummaryPlotRendererDRAM({
...
-
-                return {
-                    start: zoomStart,
-                    end: zoomEnd,
-                    padding: (zoomEnd - zoomStart) * MEMORY_ZOOM_PADDING_RATIO,
-                };
-            }),
+            segmentedChartData
+                .map((segment) => {
+                    if (segment.length > 0) {
```
dcblundell: Maybe check that buffers has a length of 2 here because we assume it does when we compute `zoomEnd`.

src/components/buffer-summary/BufferSummaryPlotRendererDRAM.tsx
```diff
@@ -154,18 +159,21 @@ function BufferSummaryPlotRendererDRAM({
...
                             plotZoomRange={
                                 isZoomedIn
                                     ? [
-                                          zoomedMemoryOptions[index].start - zoomedMemoryOptions[index].padding,
-                                          zoomedMemoryOptions[index].end + zoomedMemoryOptions[index].padding,
+                                          (zoomedMemoryOptions[index]?.start ?? 0) -
+                                              (zoomedMemoryOptions[index]?.padding ?? 0),
+
+                                          (zoomedMemoryOptions[index]?.end ?? MEMORY_SIZE) -
+                                              (zoomedMemoryOptions[index]?.padding ?? 0),
```
dcblundell, line 176: Looks better to remove padding vs add padding?

dcblundell (changes requested):

src/components/buffer-summary/BufferSummaryPlotRendererDRAM.tsx
aidemsined: actually the other way around

dcblundell (approved):

PR #1409, opened 2026-04-17
Fix non connected graph rendering errors

### LINKED ISSUE #1407: Quasar support - graph rendering is broken
likely due to graph being only 2 non connected elements

aidemsined:
Avoid updating state during render by moving the fallback logic that sets currentOperationId into a useEffect. Add guards for empty connectedNodeIds and undefined fallback values, and use a safe fallback for focusNodeId (using ??). The effect only calls setCurrentOperationId when a valid, different fallback id exists, preventing render-time state updates and related React warnings.

closes #1407

### DIFF
--- a/src/components/OperationGraphComponent.tsx
+++ b/src/components/OperationGraphComponent.tsx
@@ -80,12 +80,21 @@ const OperationGraph: React.FC<{
         return ids;
     }, [edges]);
 
-    if (currentOperationId !== null && !connectedNodeIds.has(currentOperationId)) {
-        const val = connectedNodeIds.values().next().value;
-
-        focusNodeId = val;
-        setCurrentOperationId(val);
+    if (currentOperationId !== null && connectedNodeIds.size > 0 && !connectedNodeIds.has(currentOperationId)) {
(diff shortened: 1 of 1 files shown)


src/components/OperationGraphComponent.tsx
```diff
@@ -80,13 +80,22 @@ const OperationGraph: React.FC<{
...
-        setCurrentOperationId(val);
+    if (currentOperationId !== null && connectedNodeIds.size > 0 && !connectedNodeIds.has(currentOperationId)) {
+        focusNodeId = connectedNodeIds.values().next().value ?? focusNodeId;
     }
 
+    useEffect(() => {
+        if (connectedNodeIds.size === 0) {
+            return;
+        }
+        if (currentOperationId !== null && !connectedNodeIds.has(currentOperationId)) {
```
dcblundell, line 91: Maybe combine these conditions so we return early if there are not `connectedNodeIds` or `currentOperationId`.

dcblundell (approved):

src/components/OperationGraphComponent.tsx
aidemsined: i think its easier to read separated

PR #1410, opened 2026-04-17
Use device L1 size in L1 small marker

### LINKED ISSUE #1408: Quasar support: incorrect L1 size

aidemsined:
closes #1408 
Add useGetL1Size hook to read worker_l1_size from devices and use it as the fallback/default for useGetL1SmallMarker. Replace hardcoded L1_DEFAULT_MEMORY_SIZE fallbacks with l1Size ?? L1_DEFAULT_MEMORY_SIZE, and include l1Size in the useMemo dependencies to ensure correct recalculation.

### DIFF
--- a/src/components/OperationGraphComponent.tsx
+++ b/src/components/OperationGraphComponent.tsx
@@ -80,12 +80,21 @@ const OperationGraph: React.FC<{
         return ids;
     }, [edges]);
 
-    if (currentOperationId !== null && !connectedNodeIds.has(currentOperationId)) {
-        const val = connectedNodeIds.values().next().value;
-
-        focusNodeId = val;
-        setCurrentOperationId(val);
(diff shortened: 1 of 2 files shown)


src/components/OperationGraphComponent.tsx
```diff
@@ -80,13 +80,22 @@ const OperationGraph: React.FC<{
...
-        setCurrentOperationId(val);
+    if (currentOperationId !== null && connectedNodeIds.size > 0 && !connectedNodeIds.has(currentOperationId)) {
+        focusNodeId = connectedNodeIds.values().next().value ?? focusNodeId;
     }
 
+    useEffect(() => {
+        if (connectedNodeIds.size === 0) {
+            return;
+        }
+        if (currentOperationId !== null && !connectedNodeIds.has(currentOperationId)) {
```
dcblundell, line 91: Maybe combine these so that we always return early when there are no `connectedNodeIds` or `currentOperationId`.

dcblundell (approved):
One small suggestion.

PR #1416, opened 2026-04-21
Refactor/Control z-index scale

### LINKED ISSUE #1265: CSS: z-index values are uncontrolled
We should refactor all z-index values across the code base because none use vars and there isn't a system.

dcblundell:
Adds a set of variables for z-index values and consolidates current usages into more a consistent state.

The only real change has been to `.navbar .server-mode-banner` which now has a z-index of `1`. 

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1265.

### DIFF
--- a/src/scss/components/NPEComponent.scss
+++ b/src/scss/components/NPEComponent.scss
@@ -19,7 +19,6 @@
         position: sticky;
         top: 0;
         background-color: $tt-grey-2;
-        z-index: variables.$z-index-2;
     }
 
     .npe-controls {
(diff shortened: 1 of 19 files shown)


src/scss/components/NPEComponent.scss
```diff
@@ -19,7 +19,7 @@
         position: sticky;
         top: 0;
         background-color: $tt-grey-2;
-        z-index: 5;
```
aidemsined, line 22: i dont think z-index# is solving anything. we should give each z-index usecase a name. IE z-index-npe-controls

aidemsined (changes requested):

aidemsined (approved):

PR #1457, opened 2026-05-06
Add support for multihost config.json files

smountenay-tt:
This PR modifies the Flask backend to look for both `config.json` files, and now also `config_x_of_y.json` files.

When reports are run with world size of 1 on the tt-metal size, they'll still generate `config.json` files. But if world size is > 1, once the PR to track rank in TTNN reports is merged, there will instead be `config_x_of_y.json` files. X represents rank + 1, and y is world size.

The `/api/config` is modified here so that by default it still returns the data from `config.json` if it is present. If it is not, by default it will return the data from `config_1_of_1.json` if it is found. A new `rank` query var is introduced also, so that if the frontend wants the config file for a different rank, it could get it by passing `rank` like this: `GET /api/config?rank=2`, and it would give you the config file from rank 3 (i.e. `config_4_of_<world_size>.json`).

When testing the feature I ran into an a problem with SSH timeouts, and this PR is slipping in a change to make SSH timeout values configurable.

In testing, I was also getting terminal errors from Flask that looked like this:

```
ERROR:ttnn_visualizer.sftp_operations:stderr: bash: -c: line 0: syntax error near unexpected token `do'
bash: -c: line 0: `bash -lc for f in /localdev/ctr-smountenay/tt-metal/generated/ttnn/reports/resnet50_cpp_mar02_1748/config_*_of_*.json; do [ -f "$f" ] && printf "%s\n" "$f"; done'
```

There is a change in this PR to change how the command is escaped when running over ssh.

The reason why SSH features are touched in this PR is because there is a feature which reads the remote config files in, so that had to be modified to look for the multihost config file paths with the rank and world size.


dcblundell (approved):

smountenay-tt:
The TTNN Visualizer side of this feature was merged here:

https://github.com/tenstorrent/ttnn-visualizer/pull/1457

But the TT-Metal part of the change is awaiting PR review.

PR #1473, opened 2026-05-13
Improve ops graph navigation. locate input/output and not lose focus

### LINKED ISSUE #1474: Operation graph usability. Make input/output lookup not lose main op focus

aidemsined:
image

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1474

### DIFF
--- a/src/components/OperationGraphComponent.tsx
+++ b/src/components/OperationGraphComponent.tsx
@@ -45,8 +45,6 @@ const OperationGraph: React.FC<{
     const blinkTimeoutRef = useRef<number | null>(null);
     const [compactView, setCompactView] = useState<boolean>(false);
 
-    const originalOperationId: number | null = operationId ?? operationList[0]?.id ?? null;
-
     const edges = useMemo((): Edge[] => {
         const edgeMap = new Map<string, Edge>();
         return operationList.flatMap((op) =>
@@ -243,14 +241,6 @@ const OperationGraph: React.FC<{
         [scale],
     );
 
-    const selectConnectedNode = useCallback(
-        (nodeId: number) => {
-            focusOnNode(nodeId);
-            colorHighlightIO(nodeId);
-        },
-        [focusOnNode, colorHighlightIO],
-    );
-
     const getNodeRelationToFocused = useCallback((nodeId: IdType): 'input' | 'output' | null => {
         const selectedNodeId = currentOpIdRef.current;
         if (selectedNodeId === null || selectedNodeId === undefined || nodeId === selectedNodeId) {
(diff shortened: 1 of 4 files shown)


src/components/OperationGraphComponent.tsx
```diff
@@ -239,14 +242,115 @@ const OperationGraph: React.FC<{
         [scale],
     );
 
-    const selectConnectedNode = useCallback(
+    const getNodeRelationToFocused = useCallback((nodeId: IdType): 'input' | 'output' | null => {
```
dcblundell: Enum for node relation?

src/components/OperationGraphComponent.tsx
```diff
@@ -239,14 +242,115 @@ const OperationGraph: React.FC<{
...
+    const getBlinkOnColorForNode = useCallback(
+        (nodeId: IdType): { background: string } => {
+            const relation = getNodeRelationToFocused(nodeId);
+            if (relation === 'input') {
+                return { background: tinycolor(GRAPH_COLORS.inputNode).lighten(20).toString() };
+            }
+            if (relation === 'output') {
+                return { background: tinycolor(GRAPH_COLORS.outputNode).lighten(20).toString() };
+            }
+            return { background: '#f6bc42' };
```
dcblundell: Can `#f6bc42` not also be part of `GRAPH_COLORS`?

src/components/OperationGraphComponent.tsx
```diff
@@ -708,14 +829,30 @@ const OperationGraphInfoComponent: React.FC<{
...
+                    <Button
+                        className='recenter-button'
+                        icon={IconNames.UNDO}
+                        intent={Intent.PRIMARY}
+                        onClick={onRecenterOnCurrent}
+                        aria-label={`Recenter on operation ${currentOperationId}`}
+                    >
+                        Back to {currentOperationId}
+                    </Button>
+                </Tooltip>
```
dcblundell: I wonder if we just call this "Locate op" or something and get rid of the tooltip. "Back to xxx" implies the user has moved away from it which is not true when you first click on the node. I would expect "Back" to mean you refocus on the previously highlighted node.

dcblundell (changes requested):

src/components/OperationGraphComponent.tsx
```diff
@@ -239,14 +242,115 @@ const OperationGraph: React.FC<{
...
+            }
+            try {
+                networkRef.current.focus(nodeId, {
+                    scale,
+                    animation: { duration: 500, easingFunction: 'easeInOutCubic' },
+                });
+            } catch (e) {
+                // eslint-disable-next-line no-console
+                console.warn('Error focusing viewport on node', e);
+                return;
```
dcblundell, line 351: Should this be a user error?

aidemsined: no, it can theoretically happen with certain graphs under certain conditions

dcblundell (approved):

PR #1478, opened 2026-05-13
Use explicit column names when querying tables

### LINKED ISSUE #1458: Ensure report data is loaded in a way that does not make assumptions about number of columns in a table
In order to support semantic versioning for the database schema, we need to make sure TTNN Visualizer loads the data from the database in a way that does not make any assumptions about the number of columns in the database tables.

TTNN Visualizer should support new columns being added to database tables, without there being any errors in the Python code related to an unexpected number of arguments being passed to another function.

As an example, we want to make sure that the queries are selecting columns by name, in most cases column names that match fields on the pydantic models, and we are not effectively doing "SELECT * FROM ..." and then initializing the pydantic models with whatever number of columns came back.

A new and unknown column on any table in the TTNN report database already being used by TTNN Visualizer, should not result in an error anywhere in TTNN Visualizer.

smountenay-tt:
This PR modifies the report DB queries to use explicit column names, rather than selecting whole rows and assuming things about the number of columns and / or order of columns.

This is intended to make the app more resilient when changes are made to the report db schema on the TTNN side.

[Closes #1458]


dcblundell (approved):
Looks like a great improvement!

PR #1480, opened 2026-05-13
Fix focus for initial graph load

### LINKED ISSUE #1479: Operation graph focus node is not in the viewport on load

aidemsined:
closes #1479


dcblundell (approved):
Seems to work.

PR #1577, opened 2026-05-21
Return 400 from `with_instance` when `instanceId` is missing

### LINKED ISSUE #1530: Inconsistencies: with_instance returns 404
- **`@with_instance` returns 404 on missing `instanceId`.** `backend/ttnn_visualizer/decorators.py` aborts with 404 when the query param is absent; 400 ("Bad Request") would be more semantically accurate. Pre-existing; flag if you're rewriting the decorator, otherwise leave it.

dcblundell:
Closes #1530.

## Summary

`with_instance` (`backend/ttnn_visualizer/decorators.py`) previously aborted with **404** when the `instanceId` query parameter was absent. That misrepresents the failure: the named instance isn't unknown, the client just failed to identify which one it meant. **400 ("Bad Request")** is the right semantic.

- Swap the `abort(404)` for `abort(400, description="Missing required query parameter: instanceId")`.
- Update the log line so the recorded status matches what we actually return.
- Add a regression test (`backend/ttnn_visualizer/tests/test_decorators.py`) that hits a `with_instance`-guarded endpoint (`/api/tensors`) without the parameter and pins the status code at 400.

No behavioural change for any request that already supplied `instanceId`.

## Test plan

- [x] `PYTHONPATH=backend python3 -m pytest backend/ttnn_visualizer/tests/test_decorators.py -v` passes
- [x] `pnpm flask:lint` clean
- [x] `pnpm flask:mypy` clean


smountenay-tt:
I checked out your branch and tested it out manually. I wanted to see if the response with `abort` was consistent with our custom response functions.

When I was testing it I tried it with an invalid instance id, instead of a missing query var, and I see that it is returning a 500 error with an invalid `instanceId` query var.

smountenay-tt (approved):

PR #1578, opened 2026-05-21
Detach `fileTransferProgress` handler on `SocketProvider` cleanup

### LINKED ISSUE #1529: Inconsistencies: SocketProvider cleanup
- **`SocketProvider` cleanup is incomplete.** `src/libs/SocketProvider.tsx` `off()`s `connect`/`disconnect`/`connect_error`/`reconnect` but not the `fileTransferProgress` handler registered alongside them. Pre-existing; the listener list will grow on every remount of the provider (rare in production, common under StrictMode in dev). New listeners follow the pairing rule; the existing miss is tracked tech debt.

dcblundell:
Closes #1529.

## Summary

`src/libs/SocketProvider.tsx` registered five socket listeners (`connect`, `disconnect`, `connect_error`, `reconnect`, `fileTransferProgress`) but only `off()`d four of them on effect teardown. The `fileTransferProgress` handler was leaked on every remount.

Provider remounts are rare in production but routine under React StrictMode in dev, so each dev-mode mount silently doubled the number of times the file-progress atom setter ran per server event.

- Add `socket.off('fileTransferProgress')` to the cleanup return so the on/off list is exhaustive.

## Test plan

- [x] `pnpm lint src/libs/SocketProvider.tsx` clean
- [ ] Manual: in dev with StrictMode on, upload a file and verify `fileTransferProgress` only updates the atom once per server event (where it previously fired twice after the StrictMode double-mount).


dcblundell:
Closing because changes were made in https://github.com/tenstorrent/ttnn-visualizer/pull/1581.

PR #1590, opened 2026-05-22
Improve sync status feeback to user

### LINKED ISSUE #1511: Remote: Show progress of sync
There is a loading state but it's unclear what the speed or progress of the transfer is. You get some information in the terminal but the UI doesn't expose anything.

Image

dcblundell:
Gives more information to the user about current remote sync status using the existing FileStatus mechanisms, and makes similar changes to the local file upload overlay to keep their layouts similar.

**New remote sync overlay**
Screenshot 2026-05-25 at 11 52 01

**Changes to local file upload**
Screenshot 2026-05-25 at 11 54 17

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1511.


backend/ttnn_visualizer/tests/test_sftp_operations.py
```diff
@@ -0,0 +1,411 @@
...
+from ttnn_visualizer.sftp_operations import (
+    _get_remote_file_list_without_sizes,
+    get_remote_directory_list,
+    get_remote_file_list,
+    sync_files_and_directories,
+)
+from ttnn_visualizer.sockets import FileStatus
+
+
+def _connection() -> RemoteConnection:
```
smountenay-tt: In my opinion this should be a pytest fixture, not a "private" function like.

dcblundell: Done.

smountenay-tt (approved):
I've only looked at the backend changes. They look good.

PR #1592, opened 2026-05-26
Handle instance ids which don't exist

### LINKED ISSUE #1591: Inconsistencies: Unknown instance id
Discovered after https://github.com/tenstorrent/ttnn-visualizer/issues/1530.

That issue was resolved which standardised a missing `instance_id` value, but there are still inconsistencies with handling an unknown `instance_id`, e.g. `instance_id=fake-instance-id`.

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1591.


smountenay-tt:
I tried it out and it works, but I think the appropriate error code is 422 or 400, not 404.

smountenay-tt (approved):

PR #1603, opened 2026-05-28
Add named const and enum preferences to conventions

### LINKED ISSUE #1601: Conventions: Named consts over magic values
We don't have anything about this at the moment and it is a preference we have.

dcblundell:
Adds new rules for named const and enum usage, and prunes existing rules for length.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1601.


AGENTS.md
```diff
@@ -18,9 +18,7 @@ When changing ingestion, sync, or path logic, keep these flows and paths in mind
 
 ## Deployment and security posture
 
-In practice the app is usually **run by the engineer on their own machine** while developing or profiling. The instance at **ttnn-visualizer.tenstorrent.com** is a **demo**, not the primary product shape.
-
-Do **not** assume a multi-tenant hosted SaaS model for defaults, threat model, or feature tradeoffs: treat “runs locally for one user” as the common case unless a change explicitly targets the demo deployment.
+The app usually runs on the engineer's own machine; **ttnn-visualizer.tenstorrent.com** is a demo, not the primary product shape. Default to "one local user" for threat model, defaults, and feature tradeoffs — don't assume multi-tenant SaaS unless a change explicitly targets the demo deployment.
```
smountenay-tt: I think this guidance is problematic because by default any change we make runs on the server, unless you go out of your way for it not to, and that can open up unintentional security holes if agents are being instructed to assume "one local user". For example, when a new upload feature was added recently, it wasn't initially locked down to local only, until it was caught in review. Agents need to be aware of security implications since the app does run in a hosted/multi-user environment also, even if that's not the "usual" way.

dcblundell: Fair point, will amend.

smountenay-tt (approved):

aidemsined (approved):

PR #1606, opened 2026-05-29
Show L1 fragmentation pressure in performance table

### LINKED ISSUE #1520: Perf table: L1 pressure columns (fullness at op end, fragmentation index)
## Context

The perf table reports time and a handful of perf signals per op, but no memory-side context. Today, a user investigating a slow op has to cross-reference the memory report to see whether L1 was near full or heavily fragmented at that point in the run.

## Ask

When a memory report is loaded alongside the perf report, add memory-pressure columns to the perf table:

- **L1 fullness at op end** — % of L1 in use after this op completes
- **Fragmentation index** — a simple metric (e.g. largest-free-block / total-free, or count of free segments) at op end

Surfaces ops that are slow *because* memory is tight, not because the op itself is heavy.

## Open questions

- Fragmentation metric — which formulation reads cleanest? (Largest-free / total-free is intuitive; segment count is simpler.) Maybe show both behind a tooltip.
- Per-bank vs aggregate L1 fullness — aggregate is enough for v1; per-bank is a stretch goal.
- Behaviour when memory report isn't loaded (hide columns gracefully)
- DRAM equivalents — worth surfacing the same columns for DRAM, or perf-side it's rarely the bottleneck?

## Related

- #1513 Perf table: render input/output tensor details per operation
- #1519 Perf table: heuristic flags per op (DRAM-bound, low L1 utilization, etc.)
- #680 Memory allocation failures
- #1186 Recommending improvements based on the data

dcblundell:
Adds L1 fragmentation when a memory report is linked.

Screenshot 2026-06-03 at 12 08 23

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1520.


src/routes/Performance.tsx
```diff
@@ -113,10 +117,13 @@ export default function Performance() {
...
+    const enrichedData = useMemo(
+        () => enrichRowData(rangedData, opIdsMap, l1PressureMap),
+        [rangedData, opIdsMap, l1PressureMap],
+    );
     const enrichedComparisonData = useMemo(
-        () => comparisonPerfData?.map((dataset) => enrichRowData(dataset, opIdsMap)) || [],
-        [comparisonPerfData, opIdsMap],
+        () => comparisonPerfData?.map((dataset) => enrichRowData(dataset, opIdsMap, l1PressureMap)) || [],
+        [comparisonPerfData, opIdsMap, l1PressureMap],
     );
```
aidemsined, line 129: **B1 — comparison rows show the active report's L1 metrics, not their own**

`opIdsMap` (from `useOpToPerfIdFiltered()`) and `l1PressureMap` (from `useL1PressureByOperation()`) are both derived from the **active** profiler report — `useOpToPerfIdFiltered` reads from `useProxyPerformanceReport()`, and `useL1PressureByOperation` pulls buffers via `useBuffers(BufferType.L1)` keyed on `activeProfilerReportAtom`. When we pass them into `enrichRowData(dataset, opIdsMap, l1PressureMap)` for each comparison dataset, the `op` we attach to a comparison row is whichever active-report row happens to share the same `row.id` string, and the L1 metrics we attach are therefore the active report's metrics — not the comparison report's. That's coincidence, not correctness.

The pre-existing `op` column already inherits this caveat, so we're not introducing it here, but the L1 column makes the wrong attribution very visible (per-row numbers that look authoritative).

Two options:
- **Quickest + safest for this PR**: drop `ColumnKeys.L1Fullness` from `comparisonKeys` in `src/definitions/PerfTable.ts` so the L1 cell renders empty on comparison sub-rows. Add a one-line note ("L1 metrics only available for the active profiler report") near the column injection.
- **Right answer, follow-up PR**: build a per-comparison-report `l1PressureMap` (load buffers + markers per profiler report) and thread the correct one through `enrichRowData` for each dataset.

If you go with the quick fix, B2 below also evaporates.

src/components/performance/PerfTable.tsx
```diff
@@ -325,7 +350,13 @@ const PerformanceTable = ({
...
                                                 })}
                                             >
-                                                {cellFormattingProxy(row, h, operationsList, filters?.[h.key])}
+                                                {cellFormattingProxy(
+                                                    row,
+                                                    h,
+                                                    operationsList,
+                                                    filters?.[h.key],
+                                                    isFirstOfOpRun,
+                                                )}
```
aidemsined, line 362: **B2 — "first-of-op" suppression is bypassed on comparison rows**

Primary rows now thread `isFirstOfOpRun` through `cellFormattingProxy` (here), so the L1 cell shows once per TTNN-op group. The comparison-row branch around line ~398 still calls `formatCell(subRow, h, operationsList, filters?.[h.key])` without `isFirstOfOpRun`, so it defaults to `true` and every comparison row with a non-null `l1_fullness_percent` shows the value. Result: the two halves of the same table look inconsistent — collapsed on top, expanded on the bottom.

If B1 is addressed by dropping `L1Fullness` from `comparisonKeys`, this disappears. Otherwise: compute a per-dataset `firstRowOfOpRun` (the current `firstRowOfOpRun` memo is keyed off `data` and so only covers the primary rows), then thread `isFirstOfOpRun` through the comparison `formatCell` call too.

src/components/performance/PerfTable.tsx
```diff
@@ -93,10 +97,29 @@ const PerformanceTable = ({
...
+
+        for (const row of data) {
+            if (row.op !== undefined && !seenOps.has(row.op)) {
+                seenOps.add(row.op);
+                firstRows.add(row);
+            }
+        }
+
+        return firstRows;
+    }, [data]);
```
aidemsined: **S1 — "first-of-op" keyed off source order produces an awkward UX under arbitrary sorts**

The intent (stability across sorts) is fine, but the consequence is awkward. Under any sort other than execution order, the device rows for op N can scatter throughout the table, and the L1 value will appear on whichever of those rows happens to be the source-order-first one — which may not even be near the top of its op group in the current view. A user sorting by, say, Device Time can easily look at the top device-op row for op N and conclude "no L1 data here", then find the value rows below on what feels like an arbitrary row.

Three cleaner alternatives, in order of effort:

1. Compute `firstRowOfOpRun` against the *current display order* (i.e. after sorting in `tableFields`). Breaks the stated stability goal but matches what the user sees, which is usually what they want.
2. Render the value on **every** device-op row in the same TTNN op. The numbers are identical by construction, so the duplication is visually noisy but never wrong; mute the repeats with a class so the eye still finds the canonical row.
3. Keep the current behaviour but make op-group structure visible — e.g. a left-border rule on the first source-order row of each op, so users learn where to look.

I'd lean (1) for the smallest behavioural surprise.

src/hooks/useAPI.tsx
```diff
@@ -929,6 +930,30 @@ export const useBuffers = (bufferType: BufferType | null, useRange?: boolean) =>
...
+
+        const pressureByOperation = new Map<number, L1PressureMetrics>();
+
+        for (const operation of buffersByOperation) {
+            pressureByOperation.set(operation.id, computeL1PressureForOperation(operation.buffers, l1Start, l1End));
+        }
+
+        return pressureByOperation;
+    }, [buffersByOperation, isError, isLoading, l1End, l1Start]);
+};
```
aidemsined: **S2 — the hook collapses three states into `null` and lets stale marker defaults leak through**

`useL1PressureByOperation` returns `null` for `isLoading`, `isError`, missing data, *and* invalid markers. Downstream, `hasL1PressureData = l1PressureMap !== null` toggles the L1 column on/off, so the column pops in mid-render once the buffers finish loading and shifts the entire table horizontally.

There's a related, subtler issue with the markers: `useGetL1SmallMarker` (unchanged, lines 275–291) falls back to `l1Size ?? L1_DEFAULT_MEMORY_SIZE` when L1_SMALL buffers haven't loaded yet, and `useGetL1StartMarker` returns `0` while `useDevices()` is in flight. The `if (l1End <= l1Start)` guard catches the degenerate case (`0 <= 0` when both default to 0 is a real risk early in the load), but only by accident — once `l1Size` resolves with markers still missing, you can briefly compute pressure against the **default** L1 window before the real one arrives. The user sees one set of percentages flicker into another.

Two suggestions, smallest first:

- Make the hook return a discriminated result like `{ status: 'loading' | 'unavailable' | 'ready'; data: Map | null }`. `PerfTable` can then reserve the column (showing `—` or a spinner) when `status === 'loading'`, hide it only on `'unavailable'`, and stop the layout jump. Also drives a clearer mental model for B1's eventual fix.
- Gate the actual computation on "all inputs resolved": require `useDevices` data + (`useGetAllBuffers(BufferType.L1_SMALL).data !== undefined`) before returning anything other than `loading`. That stops the brief flicker against default-window pressure values.

src/routes/Performance.tsx
```diff
@@ -327,11 +335,17 @@ const getRowAttributes = (row: PerfTableRow): RowAttributes => {
...
+    rows: PerfTableRow[],
+    opIdsMap: { perfId?: string; opId: number }[],
+    l1PressureMap: Map<number, L1PressureMetrics> | null,
+): TypedPerfTableRow[] => {
     const typedRows = rows.map((row) => {
         const val = parseInt(row.op_to_op_gap, 10);
         const opStr = opIdsMap.find((opMap) => opMap.perfId === row.id)?.opId;
         const op = opStr !== undefined ? Number(opStr) : undefined;
+        // TTNN-op snapshot is shared by all device ops that map to the same row.op.
+        const l1Pressure = op !== undefined ? l1PressureMap?.get(op) : undefined;
```
aidemsined, line 340: **S4 — `opIdsMap.find(…)` makes `enrichRowData` O(N·M)**

Pre-existing for the `op` column (line 345 just above), so the L1 lookup doesn't make it worse — the L1 part is fine, it's a `Map.get`. But this `.find()` runs once per row, and `opIdsMap` grows with the active report's op count. For multi-thousand-row reports this is the dominant cost in the enrichment loop, and the cost now compounds across `enrichedData` plus one map call per comparison dataset.

Drop in a `Map<perfId, opId>` once at the top of `enrichRowData` and the rest becomes O(N):

```ts
const opIdByPerfId = new Map<string, number>();
for (const { perfId, opId } of opIdsMap) {
    if (perfId !== undefined) opIdByPerfId.set(perfId, opId);
}
// then inside the map:
const op = opIdByPerfId.get(row.id);
const l1Pressure = op !== undefined ? l1PressureMap?.get(op) : undefined;
```

Not blocking; flagging because the screenshot in the PR description has both an active report and a comparison stack, which is exactly when this matters.

aidemsined (commented):
Solid feature — the metric, the stacked bar, and the tooltip are well factored and well tested. Posting five inline comments inline:

- **B1, B2** (blocking): comparison-row L1 cells display the active report's metrics and bypass the per-op-group suppression. Easiest landing is dropping `ColumnKeys.L1Fullness` from `comparisonKeys` until per-report L1 data exists; that knocks out both B1 and B2 in one line.
- **S1**: the "first row of op group" pivot is keyed off source order, so under arbitrary sorts the L1 value appears on what feels like a random row. Probably the most user-visible thing to revisit.
- **S2**: `useL1PressureByOperation` collapses three states into `null`, and the markers can briefly resolve to defaults before devices/L1_SMALL load, producing a flicker and a column-pop. Worth a discriminated `{ status, data }` shape.
- **S4**: per-row `opIdsMap.find()` (pre-existing) makes `enrichRowData` O(N·M); the L1 feature compounds the cost across comparison datasets. One-line fix.

Low-priority nits (not posted inline): `mergeOverlappingChunks` actually merges *contiguous* chunks too — worth renaming or commenting; duplicate `OPERATION_COLOURS` now lives in both `perfFunctions.tsx` and `stackedPerfFunctions.tsx`, follow-up hoist; `.l1-fullness-attributes` font-size is a hard-coded `12px` rather than a token; consider `role="presentation"` instead of `aria-hidden` on the legend so the per-segment percentages aren't fully invisible to screen readers (the bar's `aria-label` does cover the summary).

Nice touches worth keeping: `firstRowOfOpRun` keying off object identity rather than `id` (avoids the `id: null` placeholder collision), the "bar collapses to 1 segment, legend still shows 3 rows" test in `L1FullnessBar.spec.tsx`, and the `Excludes circular buffers` line in the tooltip paired with the `// Perf-table v1 uses DB buffers from useBuffers(BufferType.L1) only` comment at the top of `l1Pressure.ts` — rationale captured for both the engineer and the user.

src/routes/Performance.tsx
dcblundell: Removed `ColumnKeys.L1Fullness` from `comparisonKeys` so comparison sub-rows now render an empty L1 cell instead of misattributing the active report's metrics.

src/components/performance/PerfTable.tsx
dcblundell: Resolved by the previous fix.

src/components/performance/PerfTable.tsx
dcblundell: Went with option (1).

src/hooks/useAPI.tsx
dcblundell: Addressed both points. `useL1PressureByOperation` now returns a discriminated `{ status: 'loading' | 'unavailable' | 'ready'; data }` (`L1PressureResult` in `l1Pressure.ts`). `Performance` sets `hasL1PressureData = status !== Unavailable`, so the column is reserved during loading (cells render empty until ready) and hidden only when genuinely unavailable — no more mid-render horizontal shift. The computation is also gated on all inputs being resolved: it reads `useDevices().data` and `useGetAllBuffers(BufferType.L1_SMALL).data` directly and stays in `loading` until both are defined, so we never compute pressure against the default L1 window before the real markers arrive.

src/routes/Performance.tsx
dcblundell: Done.

src/functions/l1Pressure.ts
```diff
@@ -0,0 +1,136 @@
...
+            fullnessPercent: 0,
+            freeSegments: 0,
+            largestFreeBytes: 0,
+            largestFreePercent: 0,
+        };
+    }
+
+    const clippedChunks = dedupeAndClipBuffers(buffers, l1Start, l1End);
+    const mergedChunks = mergeOverlappingChunks(clippedChunks);
+    const gapSizes = computeGapSizes(mergedChunks, l1Start, l1End);
```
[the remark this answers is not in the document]
dcblundell: Fixed. `computeL1PressureForOperation` now groups buffers by `device_id` before deduping/merging, so same-offset allocations on different devices no longer collide into one address space. Since each device has its own physical L1 (same layout), metrics are computed per device and we surface the **worst-case (most-full) device** — the binding L1 constraint for the op — keeping its fragmentation figures (free segments / largest-free) internally consistent. Single-device behaviour is unchanged. Added two regression tests in `tests/l1Pressure.spec.ts`: one asserting same-offset buffers across devices aren't merged, and one asserting the most-full device's figures are reported. Per-device L1 markers remain a separate follow-up (the marker hooks still assume identical device config).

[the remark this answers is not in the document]
dcblundell: Opened follow-up for the remaining limitation (shared L1 window from `devices[0]` / identical-device assumption in the marker hooks): #1617

src/hooks/useAPI.tsx
```diff
@@ -929,6 +935,42 @@ export const useBuffers = (bufferType: BufferType | null, useRange?: boolean) =>
...
+        if (isError) {
+            return { status: L1PressureStatus.Unavailable, data: null };
+        }
+
+        const inputsResolved =
+            !isLoading && buffersByOperation !== undefined && devices !== undefined && l1SmallBuffers !== undefined;
+
+        if (!inputsResolved) {
+            return { status: L1PressureStatus.Loading, data: null };
+        }
```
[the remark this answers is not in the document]
dcblundell: Updated.

src/routes/Performance.tsx
```diff
@@ -113,10 +120,13 @@ export default function Performance() {
...
+    const enrichedData = useMemo(
+        () => enrichRowData(rangedData, opIdsMap, l1PressureMap),
+        [rangedData, opIdsMap, l1PressureMap],
+    );
     const enrichedComparisonData = useMemo(
-        () => comparisonPerfData?.map((dataset) => enrichRowData(dataset, opIdsMap)) || [],
-        [comparisonPerfData, opIdsMap],
+        () => comparisonPerfData?.map((dataset) => enrichRowData(dataset, opIdsMap, l1PressureMap)) || [],
+        [comparisonPerfData, opIdsMap, l1PressureMap],
     );
```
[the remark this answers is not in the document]
dcblundell: Updated.

PR #1609, opened 2026-06-03
Performance overlay on the operation graph

### LINKED ISSUE #1515: Operation graph: Perf overlay (color nodes by duration or utilization)
## Context

When both a profiler report and a perf report are loaded, the perf signal lives in a separate table view. The operation graph today communicates structure only — there's no visual cue for *where the time is going*.

## Ask

When a perf report is loaded alongside the profiler report, overlay perf signal directly on operation graph nodes:

- Color nodes by duration (or utilization)
- Optionally size nodes by duration
- Legend showing the scale

Gives users an instant "where is the time?" view without scrolling the perf table.

## Possible signals to drive the overlay

- Duration (default)
- Utilization %
- DRAM / L1 bytes
- Op kind (categorical — already shown via existing styling)

## Open questions

- Default metric: duration or utilization?
- Allow user to swap the metric via a small selector?
- Color scale: linear vs log (long-tail durations dominate linear scales)
- Behaviour when only profiler report is loaded (graceful no-op vs hide control)
- Interaction with existing op-type coloring — do we replace or toggle?

## Related

- #844 Perf table prefilter from chart interactions
- #707 Add op-type filtering via highlighting

aidemsined:
This pull request adds a performance overlay feature to the operation graph, allowing users to visualize per-operation kernel durations directly on the graph. The overlay uses color and node size to indicate performance metrics, and includes UI controls and legends for clarity. Supporting utilities and documentation have been updated to ensure consistency and usability.

image

closes #1515


src/routes/GraphView.tsx
```diff
@@ -28,6 +31,24 @@ const GraphView = () => {
...
+        if (!raw || raw.length === 0) {
+            return undefined;
+        }
+        return raw.map((row) => {
+            const parsedId = parseInt(row.id, 10);
+            const parsedDeviceTime = parseFloat(row.device_time);
+            return {
+                id: Number.isFinite(parsedId) ? parsedId : null,
+                device_time: Number.isFinite(parsedDeviceTime) ? parsedDeviceTime : null,
+            };
```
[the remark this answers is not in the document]
aidemsined: Good catch — you're right. `PerfTable.ts:176` declares the column's `unit: 'µs'` and `PerfDeviceTimeChart.tsx:28` explicitly does `* 1000 /* Convert microseconds to nanoseconds */`. Every duration on the overlay was off by 1000× (the earlier "90 ns" displayed value was actually 90 µs).

Fixed by normalising at the aggregation boundary in `perfOverlay.ts` rather than at the projection site. Rationale: `PerfOverlaySource.device_time` now honestly matches the wire format (µs), and the conversion to ns happens once inside `aggregatePerfByOp` next to a `US_TO_NS = 1_000` constant. Added a docstring on `PerfOverlaySource` flagging the unit so future modifications can't silently mix units. Test fixtures + assertions updated accordingly.

src/routes/Styleguide.tsx
```diff
@@ -759,6 +761,47 @@ export default function Styleguide() {
                     latestAppVersion={LATEST_APP_VERSION}
                 />
             </div>
+
+            <div className='container'>
```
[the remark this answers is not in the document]
aidemsined: Correct — all `.perf-overlay-*` rules are scoped under `.operation-graph-component` in the component SCSS, so the styleguide preview rendered largely unstyled. Wrapped the section in `<div className='container operation-graph-component'>` so the component styles apply. Paired with the SCSS specificity bump in the other thread.

src/scss/routes/Styleguide.scss
```diff
@@ -9,6 +9,21 @@
     margin-bottom: 20px;
 }
 
+// Inline-render `.perf-overlay-legend` inside the styleguide rather than
+// pinning it to the viewport. Component default is `position: fixed`.
+.styleguide-perf-overlay-host {
+    position: relative;
```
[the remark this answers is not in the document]
aidemsined: Bumped the selector from `.styleguide-perf-overlay-host .perf-overlay-legend` to `.operation-graph-component .styleguide-perf-overlay-host .perf-overlay-legend` so the styleguide override now has specificity 0,3,0 versus the component default's 0,2,0 — wins regardless of CSS injection order. Comment in the SCSS file explains the dance.

src/components/OperationGraphComponent.tsx
```diff
@@ -131,17 +203,26 @@ const OperationGraph = ({ operationList, operationId }: OperationGraphProps) =>
...
-                        label: `${op.id} ${op.name}${
-                            op.operationFileIdentifier ? `\n${op.operationFileIdentifier}` : ''
-                        }`,
-                        shape: 'box',
-                        filterString: `${op.name}`,
-                        deviceOpFilter: op.deviceOperationNameList.join(' '),
-                    })),
+                    .map((op) => {
+                        const score = perfOverlayActive ? scoreByOpId.get(op.id) : undefined;
+                        return {
```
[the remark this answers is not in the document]
aidemsined: Valid concern, deferring to a follow-up issue: #1614.

Quick triage: this is actually a **pre-existing** pattern — `filterOutDeallocate` is already in the same `useMemo` deps list and has triggered a full network reinit on flip since before this PR. The fix (stabilise the `DataSet` via `useRef` + targeted `nodes.add/remove/update` in `useEffect`s) is best done once for **all** toolbar toggles, including the upcoming metric selector (Phase 3 of #1515 followups), aggregation selector (#1611), and any future overlay controls — each will inherit the fix for free.

Cost is also bounded: only on Switch flip, not per node interaction. `colorHighlightIO` already calls `nodes.update(...)` in-place on the existing DataSet, so clicks/hovers/selections don't pay this cost.

#1614 has the full proposed approach and acceptance criteria.

src/functions/formatting.ts
```diff
@@ -53,3 +53,26 @@ const toShortTypeLabel = (input: string) => {
...
 };
+
+/**
+ * Human-readable nanosecond formatter — picks the largest unit that still
+ * yields a reasonable mantissa (ns → µs → ms → s).
+ *
+ * Zero, negative, and non-finite input collapse to `'0 ns'` so callers can
+ * use this in legends/labels without guarding upstream.
+ */
+export const formatDuration = (ns: number): string => {
```
dcblundell, line 72: I wonder if we can leveral `formatSize` or `formatUnit` here instead of returning `toFixed(x)`. It should support all these units.

src/functions/perfOverlay.ts
```diff
@@ -0,0 +1,162 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
```
dcblundell: It's 2026.

src/scss/components/OperationGraphComponent.scss
```diff
@@ -50,6 +50,72 @@
...
+        line-height: 1.4;
+        z-index: variables.$z-index-2-interactive;
+        display: flex;
+        flex-direction: column;
+        gap: 4px;
+        pointer-events: none;
+    }
+
+    .perf-overlay-legend-title {
+        font-weight: 600;
```
dcblundell: I would say use `700`. There's probably not a visual difference but 700 is the `bold` css keyword.

```
font-weight: 100;
font-weight: 200;
font-weight: 300;
font-weight: 400; /* normal */
font-weight: 500;
font-weight: 600;
font-weight: 700; /* bold */
font-weight: 800;
font-weight: 900;
```

src/definitions/GraphColors.ts
```diff
@@ -12,3 +12,21 @@ export const GRAPH_COLORS = {
...
+// visible when the graph is zoomed out.
+export interface PerfBin {
+    color: string;
+    size: number;
+}
+
+export const PERF_BIN_DEFAULT_SIZE = 20;
+
+export const PERF_BINS: readonly PerfBin[] = [
+    { color: '#3b4a6b', size: PERF_BIN_DEFAULT_SIZE },
```
dcblundell: Size doesn't appear to do anything.

src/components/OperationGraphComponent.tsx
```diff
@@ -672,6 +753,17 @@ const OperationGraph = ({ operationList, operationId }: OperationGraphProps) =>
...
                         disabled={isLoading}
                     />
+                    <Tooltip
+                        content={PERF_OVERLAY_TOOLTIP[perfOverlayStatus]}
+                        placement={PopoverPosition.BOTTOM}
+                    >
+                        <Switch
+                            checked={perfOverlayActive}
+                            onChange={() => setPerfOverlayEnabled(!perfOverlayEnabled)}
+                            label='Perf overlay'
```
dcblundell, line 821: Not sure about this name. We don't tend to say "perf" in the UI and you need the tooltip to explain anything about this feature. Maybe "Data overlay"?

dcblundell:
Control is still available with an unlinked perf report, and when you click away from the first node, the colouring of the first node is a little weird.

Screenshot 2026-06-03 at 10 59 55

src/store/app.ts
```diff
@@ -53,6 +53,9 @@ export const activeNpeOpTraceAtom = atom<string | null>(null);
 export const activeMlirJsonAtom = atom<string | null>(null);
 export const hasClusterDescriptionAtom = atom(false);
 
+// Operation graph route
+export const perfOverlayEnabledAtom = atomWithStorage('perfOverlayEnabled', false);
```
dcblundell: We should really align how we treat our controls, i.e. do we use local storage to keep the setting or not.

dcblundell (changes requested):

src/components/OperationGraphComponent.tsx
aidemsined: we will keep perf for now, we can revise thsi once we load more data in

src/store/app.ts
aidemsined: this might even benefit from being a local state

dcblundell (approved):

PR #1616, opened 2026-06-03
Add navigation bar for charts

### LINKED ISSUE #1608: Performance: Charts should have an index so you can jump to a chart easily

dcblundell:
Adds a sticky navigation bar for charts. The bar code itself isn't complicated but it required some code reorganisation.

Screenshot 2026-06-03 at 12 46 11

Screenshot 2026-06-03 at 12 46 01

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1608.

### DIFF
--- a/src/definitions/PerformanceCharts.ts
+++ b/src/definitions/PerformanceCharts.ts
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+export enum PerfChartId {
+    OpCountVsRuntime = 'perf-chart-op-count-vs-runtime',
+    CoreCountKernelRuntime = 'perf-chart-core-count-kernel-runtime',
+    KernelDurationVsCoreCount = 'perf-chart-kernel-duration-vs-core-count',
+    MatmulCoreCountUtilization = 'perf-chart-matmul-core-count-utilization',
+    MatmulDeviceTime = 'perf-chart-matmul-device-time',
+    MatmulKernelDurationUtilization = 'perf-chart-matmul-kernel-duration-utilization',
+    MatmulUtilizationVsKernelDuration = 'perf-chart-matmul-utilization-vs-kernel-duration',
+    ConvCoreCountUtilization = 'perf-chart-conv-core-count-utilization',
+    ConvDeviceTime = 'perf-chart-conv-device-time',
+    ConvKernelDurationUtilization = 'perf-chart-conv-kernel-duration-utilization',
+    ConvUtilizationVsKernelDuration = 'perf-chart-conv-utilization-vs-kernel-duration',
+    OperationTypes = 'perf-chart-operation-types',
+}
+
+export const PERF_CHART_LABELS: Record<PerfChartId, string> = {
+    [PerfChartId.OpCountVsRuntime]: 'Operation Count vs Device Time',
+    [PerfChartId.CoreCountKernelRuntime]: 'Core Count + Device Kernel Runtime',
+    [PerfChartId.KernelDurationVsCoreCount]: 'Device Kernel Duration vs Core Count',
+    [PerfChartId.MatmulCoreCountUtilization]: 'Matmul · Core Count + Utilization',
+    [PerfChartId.MatmulDeviceTime]: 'Matmul · Device Time + Ideal Time',
+    [PerfChartId.MatmulKernelDurationUtilization]: 'Matmul · Device Kernel Duration + Utilization',
+    [PerfChartId.MatmulUtilizationVsKernelDuration]: 'Matmul · Utilization vs Device Kernel Duration',
+    [PerfChartId.ConvCoreCountUtilization]: 'Conv · Core Count + Utilization',
+    [PerfChartId.ConvDeviceTime]: 'Conv · Device Time + Ideal Time',
+    [PerfChartId.ConvKernelDurationUtilization]: 'Conv · Device Kernel Duration + Utilization',
+    [PerfChartId.ConvUtilizationVsKernelDuration]: 'Conv · Utilization vs Device Kernel Duration',
+    [PerfChartId.OperationTypes]: 'Operation Types',
+};
+
+export function getOperationTypesChartId(key: 'active' | `comparison-${number}`): string {
+    return `${PerfChartId.OperationTypes}-${key}`;
+}
+
+export function getOperationTypesChartLabel(reportTitle: string): string {
+    return reportTitle
+        ? `${PERF_CHART_LABELS[PerfChartId.OperationTypes]} — ${reportTitle}`
+        : PERF_CHART_LABELS[PerfChartId.OperationTypes];
+}
+
+export interface PerfChartIndexEntry {
+    id: string;
+    label: string;
+}
+
+export const FILTERABLE_CHART_ENTRIES: PerfChartIndexEntry[] = [
+    {
+        id: PerfChartId.OpCountVsRuntime,
+        label: PERF_CHART_LABELS[PerfChartId.OpCountVsRuntime],
+    },
+    {
+        id: PerfChartId.CoreCountKernelRuntime,
+        label: PERF_CHART_LABELS[PerfChartId.CoreCountKernelRuntime],
+    },
+    {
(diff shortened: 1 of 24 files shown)


src/hooks/useActiveSection.ts
```diff
@@ -0,0 +1,49 @@
...
+
+                    if (scrollY > sectionTop && scrollY <= sectionTop + sectionHeight) {
+                        current = id;
+                    }
+                }
+            });
+
+            if (current !== null) {
+                setActiveId(current);
+            }
```
aidemsined, line 44: **B1 (bug): stale `activeId` survives a change to `ids`.**

`activeId` is initialised from `ids[0]` exactly once at first render, and the scroll handler only ever calls `setActiveId(current)` when *some* section matched (`if (current !== null)`). When `idsKey` changes mid-mount (e.g. matmul data lands and a new chart is added, or the user toggles comparison reports and the Operation Types entries change), the effect re-runs but `activeId` is never re-seeded. Two visible consequences:

1. After the entries change, scrolling above every section keeps the *previous* set's id highlighted (the dropdown trigger keeps showing the stale label).
2. If the user lands on the Charts tab and never scrolls past `offsetTop - 250px` of any chart, the trigger sits at whatever was first computed; with the parent recomputing entries as data arrives, the button can show `'Jump to chart'` indefinitely.

Suggested fix: re-seed inside the effect when `idsKey` changes, and accept that `null` is a legitimate "no sections visible" state:

```ts
useEffect(() => {
    const sectionIds = idsKey ? (idsKey.split(ID_SEPARATOR) as TId[]) : [];
    setActiveId(sectionIds[0] ?? null);
    // ...rest unchanged
}, [idsKey, offsetPx]);
```

(Or, if you want to keep the "do not clear when scrolling past the last section" behaviour from the old BufferSummary code, gate it on `activeId === null || !sectionIds.includes(activeId)`.)

src/hooks/useActiveSection.ts
```diff
@@ -0,0 +1,49 @@
...
+        const sectionIds = idsKey ? (idsKey.split(ID_SEPARATOR) as TId[]) : [];
+
+        function navHighlighter() {
+            const { scrollY } = window;
+            let current: TId | null = null;
+
+            sectionIds.forEach((id) => {
+                const element = document.getElementById(id);
+
+                if (element?.offsetHeight && element.offsetTop) {
```
aidemsined: **B2 (bug): `element.offsetTop` truthiness skips any section flush to the top.**

`if (element?.offsetHeight && element.offsetTop)` discards a section whose `offsetTop === 0` even though that is a legitimate position. Today this matters less because of the page headings above the charts, but as soon as the first anchored element renders at the top of a scroll container (no `h1`/comparison selector above it, narrow viewport, future refactor, etc.) the very first chart can never become active.

The old BufferSummary code had the same defect, so it's not a regression — but this is the moment to fix it now that it's the shared hook:

```diff
-if (element?.offsetHeight && element.offsetTop) {
+if (element && element.offsetHeight > 0) {
```

src/routes/Performance.tsx
```diff
@@ -277,33 +275,15 @@ export default function Performance() {
...
-                                        <div>
-                                            <NonFilterablePerfCharts
-                                                chartData={enrichedData}
-                                                secondaryData={enrichedComparisonData || []}
-                                                opCodeOptions={opCodeOptions}
-                                            />
-                                        </div>
-                                    </div>
-                                </>
+                                <PerformanceChartsTab
```
aidemsined, line 278: **S1 (suggestion): hash-anchor navigation breaks across tabs.**

The popover menu items in `PerfChartsIndex` link to `${ROUTES.PERFORMANCE}#${chartId}`. The Charts tab is wrapped in `renderActiveTabPanelOnly`, so any shared/back-button URL like `/performance#perf-chart-matmul-device-time` lands on the Table tab and the browser silently fails to find the anchor (the chart isn't in the DOM). The button itself is only reachable from the Charts tab, so the in-tab case works, but cross-tab sharing of a hash link is broken.

Low-effort fix: when the route mounts (or `location.hash` changes), if the hash matches any `PerfChartId`, set `selectedTabId = PerfTabIds.CHARTS` before the browser tries to scroll. That keeps the popover URLs as plain anchors and makes them work when pasted into a new tab.

src/scss/components/PerfCharts.scss
```diff
@@ -2,16 +2,30 @@
 //
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
+// Clears the sticky "Jump to chart" nav bar when scrolling to an anchored chart
+$perf-charts-index-scroll-offset: 60px;
```
aidemsined: **S2 (suggestion): coordinate the two scroll-offset constants.**

`$perf-charts-index-scroll-offset: 60px` here exists to clear the sticky nav's visual height when a hash anchor scrolls into view, while `useActiveSection`'s default `offsetPx = 250` controls when each section becomes "active". They represent different things, but both are tied to the sticky-bar height in spirit, and right now they can drift independently — changing the bar's padding/box-shadow won't update either number, and `250` is conspicuously larger than the bar itself (which means a freshly-clicked anchor scrolls the chart into view but the dropdown trigger doesn't update its label until you scroll a further ~190px).

Suggestion: lift the bar height into a single shared constant (e.g. a new `$z-index`/`$size` entry in `_variables.scss`) and pass the matching JS value to `useActiveSection(ids, NAV_BAR_HEIGHT_PX)`, so a single change keeps the visual offset and the active-section threshold aligned.

src/definitions/PerformanceCharts.ts
```diff
@@ -0,0 +1,147 @@
...
+        entries.push({
+            id: getOperationTypesChartId('active'),
+            label: getOperationTypesChartLabel(comparisonReportNames ? activeReportName : ''),
+        });
+    }
+
+    comparisonReportNames?.forEach((report, index) => {
+        entries.push({
+            id: getOperationTypesChartId(`comparison-${index}`),
+            label: getOperationTypesChartLabel(activeReportName !== null ? report : ''),
```
aidemsined: **Nit:** the active/comparison label logic is a little tangled — `getOperationTypesChartLabel(comparisonReportNames ? activeReportName : '')` flips its argument *only* when the other branch is present. A reader has to cross-reference both lines to convince themselves the labels stay symmetric with `NonFilterablePerfCharts`' `reportTitle={comparisonReportList ? performanceReport.reportName : ''}`. Worth a one-line code comment, or a helper like `getOperationTypesEntryLabel(reportName, hasComparison)` that lives next to the chart component, so the symmetry is enforced by a single function rather than two mirroring ternaries in different files.

aidemsined (commented):
Overall: the reorg (split into `PerformanceChartsTab`, central `PerformanceCharts.ts` for ids/labels, shared `useActiveSection`) reads well and is well-tested (typecheck ✓, eslint ✓, vitest ✓ on changed files locally).

Main concerns:
- **B1** — `useActiveSection` keeps the previous `activeId` after `ids` change, so the dropdown label can sit stale across tab/data updates.
- **B2** — `element.offsetTop` truthiness skips sections at `offsetTop === 0` (inherited from the old BufferSummary code, worth fixing while it's getting moved to a shared hook).

Suggestions:
- **S1** — hash navigation to a chart anchor doesn't auto-switch to the Charts tab (`renderActiveTabPanelOnly` hides the anchor); breaks share-a-URL.
- **S2** — the SCSS `$perf-charts-index-scroll-offset` (60px) and the hook's default `offsetPx` (250px) drift independently; worth a single shared source of truth.
- **S3** — hook listens only on `window`; document the assumption or consider `IntersectionObserver`.

Plus one tiny readability nit on the Operation Types label ternary.

src/scss/components/PerfChartsIndex.scss
```diff
@@ -0,0 +1,45 @@
...
+
+.perf-charts-index {
+    position: sticky;
+    top: 0;
+    z-index: variables.$z-index-2-interactive; // Above base content and allows internal dropdown to be fully visible
+    margin-bottom: 10px;
+    padding: 5px;
+    background-color: $tt-purple-shade;
+    border-radius: 2px;
+    box-shadow: 6px 6px 8px 0 rgb(0 0 0 / 25%);
```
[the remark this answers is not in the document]
dcblundell: Already fixed.

PR #1621, opened 2026-06-04
Deduplicate displayed edges by endpoint

aidemsined:
Prevent duplicate consumer/producer rows in the Outputs and Inputs sections by de-duplicating edges on the user-visible endpoint tuple (target/producer node + source output port + target input port). Adds a seenKeys set and composite key checks when building outgoing and incoming edge views so distinct worker edge ids that share the same endpoint tuple do not produce duplicated rows or inflated counts. Change is in MlGraphInner (MLIRViewReactFlow.tsx).

### DIFF
--- a/src/components/mlir/MLIRViewReactFlow.tsx
+++ b/src/components/mlir/MLIRViewReactFlow.tsx
@@ -659,14 +659,30 @@ const MlGraphInner = ({ data }: ViewProps) => {
                 buckets.push(partnerBucket);
             }
         }
+        // De-dupe on the tuple the user actually reads off the row
+        // (target node + source output port + target input port). The
+        // self and partner buckets can both surface the same logical
+        // wire when the worker's `addEdgeSafe` keeps two edge ids that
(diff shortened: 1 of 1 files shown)


src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -659,14 +659,30 @@ const MlGraphInner = ({ data }: ViewProps) => {
...
+        // wire when the worker's `addEdgeSafe` keeps two edge ids that
+        // share an endpoint tuple (its id-dedup only blocks exact id
+        // collisions, and pair-dedup is bypassed for non-top-level
+        // endpoints). Without this guard the Outputs section renders
+        // duplicated consumer rows for region-pair selections.
         const result: OutgoingEdge[] = [];
+        const seenKeys = new Set<string>();
         for (const bucket of buckets) {
             for (const edge of bucket) {
+                const sourceOutputId = edge.sourceHandle ?? '0';
```
dcblundell, line 674: String IDs 😞

dcblundell (approved):

PR #1624, opened 2026-06-05
Allow users to toggle perf table columns

### LINKED ISSUE #1607: Performance: Allow toggling of table columns
The table is getting bigger and some columns might only be relevant in some contexts.

dcblundell:
Adds a control for toggling columns in the performance table.

Screenshot 2026-06-08 at 14 29 06

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1607.


src/components/performance/PerfColumnPicker.tsx
```diff
@@ -0,0 +1,101 @@
...
+import 'styles/components/PerfTableToolbar.scss';
+import { ColumnDefinition, ColumnKeys, LOCKED_PERF_COLUMN_KEYS } from '../../definitions/PerfTable';
+import { TEST_IDS } from '../../definitions/TestIds';
+import { userPerfColumnsAtom } from '../../store/app';
+
+interface PerfTableToolbarProps {
+    eligibleColumns: ColumnDefinition[];
+}
+
+function PerfTableToolbar({ eligibleColumns }: PerfTableToolbarProps) {
```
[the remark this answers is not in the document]
dcblundell: Deleted.

src/store/app.ts
```diff
@@ -96,6 +96,7 @@ export const stackedGroupByAtom = atom<StackedGroupBy>(StackedGroupBy.OP);
 // so a selection can't leak across reports. Cleared on active-report change (Performance.tsx)
 // and on drawer close / row removal / unsynced reports (PerfTable.tsx).
 export const selectedPerfRowIdAtom = atom<number | null>(null);
+export const userPerfColumnsAtom = atomWithStorage<ColumnKeys[]>('userPerfColumns', []);
```
aidemsined: Is the cross-report persistence intentional here? `atomWithStorage` means a user who hides "L1 Fullness" on report A still has it hidden when they open report B.

That's the opposite of the choice made for `selectedPerfRowIdAtom` two lines above, where the comment explicitly says the atom is per-report so selections "can't leak across reports." Both behaviours are defensible — global "I always want my preferred layout" vs per-report "this column doesn't apply to this run" — but which one are we going for, and should it be called out in the comment block above for the next person reading this file?

src/components/performance/PerfTable.tsx
```diff
@@ -428,19 +432,17 @@ const PerformanceTable = ({
...
-                                    ))}
+                                footerColumns.map((header) => (
+                                    <td
+                                        key={header.key}
+                                        className={classNames({
+                                            'pre-wrap': header.key === ColumnKeys.OpCode,
+                                        })}
+                                        colSpan={header.footerSpan ?? undefined}
+                                    >
+                                        {getTotalsForFooter(header, data, hideHostOps)}
```
aidemsined, line 443: Slightly orthogonal, but the column-toggle UX makes this more visible: `getTotalsForFooter` walks the full `data` array (`reduce` / `filter`) per footer column on every render, even though its inputs (`data`, `hideHostOps`) don't depend on user column selection — so every toggle re-runs ~6 × O(N) of identical work.

A single memo collapses it into one pass:

```suggestion
                                        {footerTotals[header.key] ?? ''}
```

…paired with something like this near the other memos:

```tsx
const footerTotals = useMemo(
    () => computeFooterTotals(data, hideHostOps),
    [data, hideHostOps],
);
```

where `computeFooterTotals` walks `data` once and returns a `Record<ColumnKeys, string>`. Same pattern exists in `StackedPerfTable.tsx:209` if you want to land both.

Happy to send a follow-up if you'd rather keep this PR scoped to the toggle feature.

src/components/performance/PerfTable.tsx
```diff
@@ -67,6 +64,7 @@ const PerformanceTable = ({
     activeReportComparisonIndex = null,
 }: PerformanceTableProps) => {
     const hideHostOps = useAtomValue(hideHostOpsAtom);
+    const userColumns = useAtomValue(userPerfColumnsAtom);
```
aidemsined: Not a blocker for this PR — the column-toggle code itself is well-memoized (`eligibleColumns` / `visibleColumns` / `footerColumns` are correctly scoped, `tableFields` doesn't re-sort on toggle, no refetches). Flagging a related perf concern for tracking only.

Subscribing to `userPerfColumnsAtom` here means every toggle re-renders the entire `<tbody>`: N rows × C columns × (1 + M comparison datasets) `formatCell` calls, even when most cells are referentially unchanged. For typical reports that's fine; for large reports (thousands of rows × tens of columns × comparisons) the toggle starts to feel sluggish.

The cleanest fix is extracting a `React.memo`-wrapped `<PerfTableRow>` keyed on the row's own props so toggling a column only triggers header/footer plus cells that actually changed. That's a structural refactor with its own test surface, so suggest splitting it into a separate perf-only PR rather than expanding scope here — and only doing it if profiling on a representative report shows it's worth the complexity. Filing here so it doesn't get lost.

aidemsined (commented):
Two perf observations from auditing the column-toggle re-render path. Neither is a blocker for this PR.

tests/PerfColumnPicker.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+                <HiddenColumnsProbe />
+            </TestProviders>,
+        );
+
+        fireEvent.click(screen.getByTestId(TEST_IDS.PERF_COLUMN_PICKER_TRIGGER));
+
+        const picker = screen.getByTestId(TEST_IDS.PERF_COLUMN_PICKER);
+        const opCodeCheckbox = within(picker).getByTestId(
+            `${TEST_IDS.PERF_COLUMN_PICKER_OPTION}-${ColumnKeys.OpCode}`,
+        ) as HTMLInputElement;
```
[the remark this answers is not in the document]
dcblundell: Blueprint's `Checkbox` forwards `data-testid` (and other html props) to the underlying `<input>`, not the `<label>` wrapper, so `getByTestId(...)` resolves to the input here. The spec passes — `toBeChecked()` would throw if the target weren't a real checkable input — so no change needed.

src/components/performance/PerfTable.tsx
dcblundell: Agreed, and thanks for flagging. This is out of scope for the column-toggle feature and needs its own test surface, so I'll handle the `React.memo` `<PerfTableRow>` extraction in a separate perf-only PR, gated on profiling a representative report. Resolving here to keep this PR scoped.

src/components/performance/PerfTable.tsx
dcblundell: Good catch — `getTotalsForFooter` does re-walk `data` per footer column on every render. I'd like to keep this PR scoped to the toggle feature, so yes please to the follow-up; the same pattern in `StackedPerfTable.tsx:209` can land with it. Leaving this thread open to track that.

src/store/app.ts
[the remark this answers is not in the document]
dcblundell: Renamed.

dcblundell: Yes, added a comment to explain intent.

src/components/performance/PerfTable.tsx
dcblundell: Decided to fix it in this PR rather than defer. `getTotalsForFooter` is replaced by `computeFooterTotals(data, hideHostOps)`, which walks `data` once into a `Record<ColumnKeys, string>` and is memoized on `[data, hideHostOps]`; the footer cell just does `footerTotals[header.key] ?? ''`. Applied the same single-pass treatment to `StackedPerfTable.tsx`. All tests pass.

dcblundell: Created https://github.com/tenstorrent/ttnn-visualizer/issues/1635.

PR #1625, opened 2026-06-06
dev: relax deps version

mert-kurttutan:
When running on my nix based environment, the dependency versions were too strict. So, this introduces relaxes some of these too strict versions (most of them pinned to minor versions)

I tested their running based on [here](https://docs.tenstorrent.com/ttnn-visualizer/src/running-from-source.html)

The error log example for versions: 

```sh

       > ERROR Missing dependencies:
       >        setuptools==78.1.1
       For full logs, run:
         nix log /nix/store/xgkwic9vaz63xsp9shkpsslc7f8wgqk6-python3.13-ttnn-visualizer-0.88.0.drv
error: Cannot build '/nix/store/f8z7q779vyn432k56phsdk5pbd8nr8hs-python3-3.13.13-env.drv'.

```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
     "gevent==25.9.1",
     "gunicorn~=23.0.0",
     "orjson>=3.9.0",
-    "pandas>=2.0.0,<3.0",
+    "pandas>=2.2.0,<3.0",
     "pydantic_core==2.41.5",
     "pydantic==2.12.5",
     "python-dotenv==1.2.2",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,21 +8,21 @@ requires-python = ">=3.10"
 dependencies = [
     "alembic~=1.18.0",
     "build",
-    "Flask-Cors==6.0.0",
-    "flask-socketio==5.6.1",
+    "Flask-Cors>=6.0.0,<7.0",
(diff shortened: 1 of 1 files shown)


mert-kurttutan:
As a side note, if we want to be careful about versions, we should move forward with uv.lock type of solution instead of pinning it down to minor versions.

mert-kurttutan:
Applied the suggestiof github copilot except for setuptools one since python packges does not promise for backwards compatibility. Copilot seems extra eager to make changes :)

pyproject.toml
```diff
@@ -8,21 +8,21 @@ requires-python = ">=3.10"
...
+    "Flask-Cors>=6.0.0,<7.0",
+    "flask-socketio>=5.6.0,<6.0",
     "flask-sqlalchemy==3.1.1",
     "Flask-Static-Digest==0.4.1",
-    "Flask==3.1.3",
+    "Flask>=3.1.0,<4.0",
     "gevent==25.9.1",
     "gunicorn~=23.0.0",
     "orjson>=3.9.0",
-    "pandas==2.2.3",
```
dcblundell, line 19: Preferably use `2.2.0` as the minimum.

mert-kurttutan: done

smountenay-tt (approved):

smountenay-tt (approved):

PR #1626, opened 2026-06-08
Gunicorn as module

mert-kurttutan:
This changes the way that gunicorn server is initiated. Instead of using gunicorn binary, we  WSGIApplication api and use only python modules from gunicorn and start app.

This should also solve the problems and extra logic required around finding gunicorn binaries, e.g. https://github.com/tenstorrent/ttnn-visualizer/issues/1012
No need for `find_gunicorn_path` function

This also handles environments like nix (where you have tricky logics around sym links and path manipulation)


mert-kurttutan:
just applied to formatting rules

mert-kurttutan:
I would also appreciate if @smountenay-tt could take a look at this since you are the original person on this problem.

backend/ttnn_visualizer/app.py
```diff
@@ -23,6 +23,7 @@
 from dotenv import load_dotenv
 from flask import Flask, abort, jsonify
 from flask_cors import CORS
+from gunicorn.app.wsgiapp import WSGIApplication
```
aidemsined, line 25: **Wrong base class for the intent.** Subclassing `WSGIApplication` pulls in CLI machinery (`init(parser, opts, args)`, paste plumbing, the implicit assumption that `super().load_config()` parses CLI args). The current override skips all of that, so it works today — but the failure mode is one line away (anyone adding `super().load_config()` to honor `gunicorn.conf.py` would suddenly start parsing `ttnn-visualizer`'s own CLI flags as gunicorn args).

`BaseApplication` is the documented embedding base class. Same behavior, no footgun:

```python
from gunicorn.app.base import BaseApplication
from gunicorn import util

class StandaloneGunicornApplication(BaseApplication):
    def __init__(self, app_uri: str, options=None):
        self.options = options or {}
        self.app_uri = app_uri
        super().__init__()

    def load_config(self):
        for key, value in self.options.items():
            if key in self.cfg.settings and value is not None:
                self.cfg.set(key.lower(), value)

    def load(self):
        return util.import_app(self.app_uri)
```

backend/ttnn_visualizer/app.py
```diff
@@ -465,29 +481,18 @@ def main():
             "Found gunicorn.conf.py in current directory - this may override environment settings"
```
aidemsined, line 493: **`gunicorn.conf.py` is silently no longer loaded.** Pre-PR `subprocess.run([gunicorn, ...])` invoked the gunicorn CLI, which loads `./gunicorn.conf.py` by default via `get_default_config_file()`. Post-PR the override never calls `super().load_config()`, so the file is never loaded. This warning is no longer accurate — the file is silently ignored.

Either delete the warning, rewrite it to say "will be ignored by embedded Gunicorn", or preserve the old behavior inside the override:

```python
def load_config(self):
    cfg_file = Path("gunicorn.conf.py")
    if cfg_file.exists():
        self.load_config_from_file(str(cfg_file))
    for key, value in self.options.items():
        if key in self.cfg.settings and value is not None:
            self.cfg.set(key.lower(), value)
```

backend/ttnn_visualizer/app.py
```diff
@@ -45,6 +45,22 @@
 logger = logging.getLogger(__name__)
 
 
+class StandaloneGunicornApplication(WSGIApplication):
+    def __init__(self, app_uri: str, options=None):
+        self.options = options or {}
+        self.app_uri = app_uri
+        super().__init__()
+
+    def load_config(self):
```
aidemsined, line 53: **`GUNICORN_CMD_ARGS` env var is no longer honored.** The parent `Application.load_config()` reads it via `parser.parse_args(self.cfg.get_cmd_args_from_env())`. The override replaces that entire method, so the env var is silently ignored. Rarely used but documented gunicorn behavior. If intentional, please call it out in the PR description so anyone debugging "my `GUNICORN_CMD_ARGS` isn't applying anymore" can find it.

backend/ttnn_visualizer/app.py
```diff
@@ -45,6 +45,22 @@
...
+        config = {
+            key: value
+            for key, value in self.options.items()
+            if key in self.cfg.settings and value is not None
```
aidemsined, line 61: **`sys.path` no longer gets `'.'` injected.** `Application.load_config()` calls `self.chdir()` which does `sys.path.insert(0, self.cfg.chdir)` (default `.`). The override skips this. Wheel installs don't care (package already on sys.path), but editable / development installs from arbitrary cwds that load auxiliary modules relative to cwd would. Probably fine here, just flagging as a behavioral diff.

backend/ttnn_visualizer/app.py
```diff
@@ -45,6 +45,22 @@
...
+        }
+        for key, value in config.items():
```
aidemsined, line 63: **Silent option-key drop.** `if key in self.cfg.settings and value is not None` silently drops typos like `gunicorn_options["worker_clas"] = "sync"`. The pre-PR CLI version would have errored on an unknown flag. Consider an `else: logger.warning("Unknown gunicorn option: %s", key)` branch so misconfigurations surface.

backend/ttnn_visualizer/app.py
aidemsined, line 1: **Dead `subprocess` import.** The only caller (`subprocess.run(gunicorn_args)`) is gone. No other consumer of `subprocess` in `app.py`. Strip the `import subprocess` at the top of the module.

aidemsined, line 1: **Fork-after-thread on macOS — new minor exposure.** The browser thread is started before `.run()`:

```python
threading.Thread(target=open_browser, args=[host, port, instance_id]).start()
```

- Pre-PR: this thread lived in the parent ttnn-visualizer process; gunicorn was a child via `subprocess.run` (`execve`, not `fork`) → no fork-after-thread.
- Post-PR: the thread runs in what is now the gunicorn master; workers `fork()` from this same process. Fork-after-thread on macOS is famously dicey (libdispatch / atfork hooks, also a `DeprecationWarning` on Python 3.14).

The thread does `webbrowser.open()` and exits, and gunicorn doesn't fork until after Arbiter setup, so likely benign — but it's a new interaction that didn't exist before. Consider either joining the thread before `.run()` or starting it from a gunicorn `post_fork` / `when_ready` hook so it never crosses a fork boundary.

backend/ttnn_visualizer/app.py
```diff
@@ -514,7 +519,9 @@ def main():
     run_alembic_migrations(config.SQLALCHEMY_DATABASE_URI)
 
     try:
-        subprocess.run(gunicorn_args)
+        StandaloneGunicornApplication(
+            config.GUNICORN_APP_MODULE, gunicorn_options
+        ).run()
```
aidemsined, line 536: **Gevent monkey-patching is now late — highest-impact regression for wheel installs.** The default in `settings.py` is `GUNICORN_WORKER_CLASS="gevent"`. Gunicorn's gevent worker patches inside `init_process()`:

```python
def init_process(self):
    self.patch()                        # monkey.patch_all()
    super().init_process()              # eventually load_wsgi() → import flask/app
```

- **Pre-PR (subprocess):** the gunicorn subprocess starts a fresh interpreter. Workers fork inside that interpreter, call `monkey.patch_all()` first, **then** import flask/ttnn_visualizer via `util.import_app(...)`. Flask sees patched socket. ✓
- **Post-PR (embedded):** `app.py` does `from flask import Flask, ...` and `from flask_cors import CORS` at module top, importing the full flask + werkzeug + flask_cors trees in the **master** process before `.run()`. Workers fork, call `monkey.patch_all()`, but `flask`/`werkzeug`/`flask_cors` are already in `sys.modules` from the master — inherited via fork. Modules that did `from socket import socket` or `from threading import Lock` at their own import time hold unpatched objects in workers.

Symptoms when it fails are nasty: hangs, "cannot release un-acquired lock" tracebacks, gevent workers silently blocking the event loop on sync I/O. CI smoke test (`curl http://localhost:8000` once) does **not** exercise gevent under sustained or concurrent load and would not catch this.

Mitigations (any one):
- Apply `from gevent import monkey; monkey.patch_all()` at the **very top** of `app.py`, before any other imports, guarded on `GUNICORN_WORKER_CLASS == "gevent"`.
- Change the default to `sync` (loses async benefits for SSE / websockets).
- Document that `GUNICORN_WORKER_CLASS=sync` should be set when running embedded.

backend/ttnn_visualizer/app.py
aidemsined, line 1: **`create_app()` runs in master when CLI flags are passed — workers inherit SQLAlchemy state via fork.**

In `main()` (around line 455):

```python
if args.profiler_path or args.performance_path:
    app = create_app()                   # imports flask, sqlalchemy, initializes db
    with app.app_context():
        session = create_instance_from_local_paths(...)
    del app
```

`create_app()` calls `extensions(app)` which calls `db.init_app(app)` and `run_alembic_migrations(...)`. The `del app` removes the local reference but does **not** dispose SQLAlchemy connection pools, clear `sys.modules`, or close inherited file descriptors. After `.run()` workers fork inheriting:

- Possibly-open SQLite file descriptors from alembic
- SQLAlchemy engine state with pool internals (locks, weakrefs)
- Loaded flask + sqlalchemy modules (compounds the gevent issue noted above)

Pre-PR this was harmless: `subprocess.run` launched a fresh gunicorn interpreter — workers got a clean slate. Post-PR workers fork from this process.

SQLite specifically has well-known fork problems: open connection FDs inherited across fork → "database disk image is malformed" or "database is locked" depending on what races first.

Mitigation: dispose engines and clear state before fork. After `create_instance_from_local_paths(...)` finishes, explicitly call `app.extensions["sqlalchemy"].db.engine.dispose()` (or the equivalent on the extension you use) before `del app`. Better: move the profiler/performance handling out of master entirely, into a worker-side hook.

aidemsined (changes requested):
The direction is right and removing the `find_gunicorn_path` heuristic is welcome. Leaving inline comments on the concrete items I'd like resolved before merge. The two highest-impact ones are:

- **Gevent monkey-patching is now too late** (worker class defaults to `gevent`, master imports flask before workers fork → unpatched primitives inherited).
- **`create_app()` in master leaks SQLAlchemy / SQLite state to forked workers** when CLI flags trigger the early profiler/performance-path branch.

The rest are smaller behavioral diffs (`gunicorn.conf.py` silently no longer loaded, `GUNICORN_CMD_ARGS` ignored, `sys.path` not getting `'.'`, silent option-key drop, fork-after-thread on the browser thread, dead `subprocess` import) and one base-class hygiene change (`WSGIApplication` → `BaseApplication`).

CI green only proves no-args boot; it does not cover gevent under load, `--daemon`, or CLI args.

smountenay-tt:
@mert-kurttutan I also welcome a clean-up to how gunicorn is run, but this needs to be tested all three ways the app runs:

* Locally in dev mode
* Hosted environments (server mode)
* Running from the wheel/PyPI package (not server mode), with the `ttnn-visualizer` CLI command

The CLI command accepts a few CLI args, that we need to ensure are still supported. Copilot has flagged that as an issue, and I agree.

We also need to make sure the gunicorn config file is loaded, as Copilot flagged. When running in a hosted environment, the gunicorn config files may be getting loaded in a different way than how the config is read when running in local mode.

Otherwise, I do agree it is a step in the right direction.

mert-kurttutan:
Just exploring around the design and the issues you raised, I think using gunicorn as a module while maintaining the older behaviour introduces more complexity than the logic around finding path. 

I will be exploring for other solutions without such workaround/complexities on the side. You can close this PR for the time being if you want

dcblundell:
> Just exploring around the design and the issues you raised, I think using gunicorn as a module while maintaining the older behaviour introduces more complexity than the logic around finding path.
> 
> I will be exploring for other solutions without such workaround/complexities on the side. You can close this PR for the time being if you want

Will do. 

FYI we've also just introduced `uv` for package version management.

PR #1627, opened 2026-06-08
Remove uvicorn dep

mert-kurttutan:
Removed the uvicorn deps since it is not used in the project


smountenay-tt (approved):

smountenay-tt:
I confirmed this is true - not used.

PR #1628, opened 2026-06-08
Support per-core CB allocation calculation and core range parsing 

### LINKED ISSUE #1622: Peak L1 per core: account for CBs per core, not device-wide
Sub-issue of #1604.

`processMemoryAllocations` keeps a single device-wide `totalCb` running total. On the `memory_repro` workload (1 MiB CB on core (0,0), 750 KiB CB on core (1,0), each core also holding a 200 KiB tensor shard), the "Peak L1 memory load per core" header reports 1.73 MiB. `worker_l1_size` is 1.5 MiB and no individual core ever holds more than 1.20 MiB.

aidemsined:
Track circular buffer (CB) allocations per core and compute peak CB usage per-core. Added getCoresInRangeList (and getCoresInRange wrapper) to math.ts to parse and expand core_range_set strings (supports legacy and modern formats, multi-rect unions, and deduplicates cores). processMemoryAllocations.ts now accumulates CB bytes by core, computes the per-core peak (cbPeak) for total_cb/total_memory and global peakMemoryLoad, and clears per-core state on deallocation. MemoryLegendElement.tsx was updated to only show core counts for per-core buffers (not DRAM/SYSTEM_MEMORY) and to pluralize "core/cores" correctly.

image

closes #1622

### DIFF
--- a/src/components/operation-details/MemoryLegendElement.tsx
+++ b/src/components/operation-details/MemoryLegendElement.tsx
@@ -76,7 +76,9 @@ export const MemoryLegendElement = ({
     );
 
     const derivedTensor = operationDetails.getTensorForAddress(chunk.address);
-    const numCoresLabel = numCores && numCores > 1 ? ` x ${numCores} cores` : '';
+    const isPerCoreBuffer = bufferType !== StringBufferType.DRAM && bufferType !== StringBufferType.SYSTEM_MEMORY;
+    const numCoresLabel =
+        isPerCoreBuffer && numCores && numCores > 0 ? ` x ${numCores} ${numCores === 1 ? 'core' : 'cores'}` : '';
 
     const memorySquare = {
         ...(!chunk.empty &&
--- a/src/functions/math.ts
+++ b/src/functions/math.ts
@@ -115,24 +115,52 @@ export const isEqual = <T>(value: T, other: T): boolean => {
     });
 };
 
+export interface CoreCoord {
+    x: number;
+    y: number;
+}
+
+const CORE_RANGE_RECT_RE = /\[([^\]]+)\]/g;
+const CORE_COORD_RE = /\(x=(\d+),y=(\d+)\)|(\d+)-(\d+)/g;
+
 /**
- @description Count the number of cores in a range string
- @param {string} rangeString - The range string to parse {[(x=0,y=0) - (x=7,y=7)]}
- @returns {number} The number of cores
+ * Expand a core_range_set string to the deduplicated list of (x,y) cores it covers.
+ * Accepts both `{[(x=N,y=N) - (x=N,y=N)]}` (legacy) and `{[N-N - N-N]}` (modern),
+ * multi-rectangle unions, and `{}`.
  */
-export const getCoresInRange = (rangeString: string): number => {
-    const regex = /\(x=(\d+),y=(\d+)\)/g;
-    const matches = [...rangeString.matchAll(regex)];
-
-    if (matches.length !== 2) {
-        return 0;
+export const getCoresInRangeList = (rangeString: string): CoreCoord[] => {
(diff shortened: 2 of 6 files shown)


src/functions/processMemoryAllocations.ts
```diff
@@ -26,26 +27,24 @@ export function processMemoryAllocations(
     let peakMemoryLoad = 0;
     const memoryAllocationList: AllocationDetails[] = [];
     const curOpList: { name: string; id: number; deviceId?: string | number }[] = [];
-    let totalCb = 0;
+    const cbBytesByCore = new Map<string, number>();
     let totalBuffer = 0;
```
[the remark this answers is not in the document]
aidemsined: this currently doesnt need to be tracked intdependently

src/components/operation-details/MemoryLegendElement.tsx
```diff
@@ -76,7 +76,9 @@ export const MemoryLegendElement = ({
     );
 
     const derivedTensor = operationDetails.getTensorForAddress(chunk.address);
-    const numCoresLabel = numCores && numCores > 1 ? ` x ${numCores} cores` : '';
+    const isPerCoreBuffer = bufferType !== StringBufferType.DRAM && bufferType !== StringBufferType.SYSTEM_MEMORY;
+    const numCoresLabel =
+        isPerCoreBuffer && numCores && numCores > 0 ? ` x ${numCores} ${numCores === 1 ? 'core' : 'cores'}` : '';
```
dcblundell, line 81: `numCores > 0` used to be `> 1`. Is this correct?

src/functions/math.ts
```diff
@@ -115,25 +115,53 @@ export const isEqual = <T>(value: T, other: T): boolean => {
...
+        for (const m of rect[1].matchAll(CORE_COORD_RE)) {
+            const x = m[1] !== undefined ? Number(m[1]) : Number(m[3]);
+            const y = m[2] !== undefined ? Number(m[2]) : Number(m[4]);
+            corners.push({ x, y });
+            if (corners.length === 2) {
+                break;
+            }
+        }
+        if (corners.length !== 2) {
+            continue; // eslint-disable-line no-continue
```
dcblundell, line 144: Preferably don't disable eslint rule or we remove the rule.

dcblundell (changes requested):

src/functions/processMemoryAllocations.ts
```diff
@@ -86,20 +94,22 @@ export function processMemoryAllocations(
             }
         }
 
+        const cbPeak = maxCbPerCore();
```
dcblundell, line 97: `cbPeak` is the max CB across cores, but `totalBuffer` is L1 buffer amortized (size / numCores), i.e. an average. 

Adding max-per-core CB + avg-per-core buffer isn't a true per-core peak. It's a reasonable improvement over the old sum and probably acceptable, but the mismatch deserves a short comment so it's not mistaken for exact.

src/components/operation-details/MemoryLegendElement.tsx
aidemsined: yes

src/functions/math.ts
aidemsined: lets remove it then. separately

src/functions/processMemoryAllocations.ts
aidemsined: this is not an average. its a max per core. i.e. [0.5] [0.1][1][0] - max is 1.we are calculating *peak* load

src/functions/math.ts
dcblundell: It's a pretty minor change but I won't hound you about it 🤷

dcblundell (approved):

PR #1630, opened 2026-06-08
More gracefully (re)load performance table data

### LINKED ISSUE #1629: Performance Table: Whole view is rerendered when toggling data options
We can make the table reload more gracefully after initial mount because the changes only affect the perf table contents not the rest of the view.

dcblundell:
Adds lightly animated skeletal UI when the perf table is loading its data instead of reloading the entire view.

Screenshot 2026-06-08 at 11 29 25

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1629.

### DIFF
--- a/src/components/performance/PerfReport.tsx
+++ b/src/components/performance/PerfReport.tsx
@@ -124,6 +124,7 @@ const PerformanceReport = ({
     );
 
     const isSignpostsDisabled = !signposts || signposts.length === 0;
+    const isTableLoading = isLoading || isComparisonLoading;
     const comparisonIndex = (activeComparisonReportList ?? []).findIndex((value) => value === selectedTabId);
     const isGroupedByMemory = stackedGroupBy === StackedGroupBy.MEMORY;
     const filterScopeHelperText =
@@ -625,7 +626,7 @@ const PerformanceReport = ({
                                     filters={filters}
                                     stackedComparisonData={filteredComparisonStackedRowsList}
                                     reportName={activePerformanceReport?.reportName || null}
-                                    isLoading={isLoading}
+                                    isLoading={isTableLoading}
                                 />
                             ) : (
                                 <PerfTable
@@ -637,7 +638,7 @@ const PerformanceReport = ({
                                     reportName={activePerformanceReport?.reportName || null}
                                     showHashColumn={false}
                                     hasL1PressureData={hasL1PressureData}
-                                    isLoading={isLoading}
+                                    isLoading={isTableLoading}
                                 />
                             )
                         }
@@ -676,7 +677,7 @@ const PerformanceReport = ({
                                         ]}
                                         filters={filters}
                                         reportName={report}
-                                        isLoading={isComparisonLoading}
+                                        isLoading={isTableLoading}
                                     />
                                 ) : (
                                     <PerfTable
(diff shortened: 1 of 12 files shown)


src/components/performance/PerfReport.tsx
```diff
@@ -120,6 +124,7 @@ const PerformanceReport = ({
     );
 
     const isSignpostsDisabled = !signposts || signposts.length === 0;
+    const isTableLoading = isLoading || isComparisonLoading;
```
aidemsined, line 126: When `isComparisonLoading` is true, this OR replaces the primary table — which already has data — with a skeleton. The new test in `tests/PerfReport.spec.tsx` documents this as intentional ("hidden behind the skeleton rather than popping in alongside the incoming comparison sub-rows"), trading visibility of the primary data for row-jank avoidance when sub-rows arrive.

Worth a UX confirmation: do we want primary rows to disappear during a comparison fetch, or should the skeleton only cover the comparison sub-rows so users can keep reading the primary table?

src/routes/Performance.tsx
```diff
@@ -59,7 +59,7 @@ export default function Performance() {
         isLoading: isLoadingPerformance,
         error: perfDataError,
     } = usePerformanceReport(activePerformanceReport?.reportName || null);
-    const { data: comparisonData } = usePerformanceComparisonReport();
+    const { data: comparisonData, isLoading: isLoadingComparison } = usePerformanceComparisonReport();
```
aidemsined, line 62: Sanity check on React Query semantics: the linked issue (#1629) is specifically about *toggling data options* causing a re-render — i.e., refetches on an already-loaded view, not initial loads.

`useQuery` exposes `isLoading` (true only while there's no cached data yet — first load) and separately `isFetching` (true on first load **and** on every background refetch). If the intent is to show the skeleton on every option-toggle refetch, we likely want `isFetching` here (and the same swap for `usePerformanceComparisonReport`). With `isLoading` alone, refetches will silently swap in new data without any loading indicator, which is the pre-PR behaviour.

Worth confirming against the hook in `hooks/useAPI.tsx` whether `isLoading` actually flips for the case the issue describes.

aidemsined (commented):
Two questions before merge — both about UX/behavioural intent rather than the refactor itself, which looks good. Nothing blocking on the code shape.

src/routes/Performance.tsx
dcblundell: Good catch on the `isLoading` vs `isFetching` distinction in general — here the option toggles from #1629 are all part of the React Query `queryKey` in both `usePerformanceReport` and `usePerformanceComparisonReport` (`hideHostOps`, `mergeDevices`, `tracingMode`, `groupBy`, signpost filters). Changing any of those produces a new cache entry, so `isLoading` is true on every toggle, not only on first mount.

`isFetching` would mainly cover same-key background refetches; with `staleTime: Infinity` those are rare for this flow. `isLoading` is the right signal for the issue described in #1629.

src/components/performance/PerfReport.tsx
dcblundell: Intentional — `isTableLoading = isLoading || isComparisonLoading` hides the full table (including primary rows) during a comparison fetch so comparison sub-rows do not pop in alongside existing primary rows and cause row jank. That trade-off is documented in `tests/PerfReport.spec.tsx`. Keeping primary rows visible would be a reasonable alternative UX, but we are deliberately favouring stable layout over continued readability of primary data during comparison loads.

PR #1640, opened 2026-06-10
Updated performance table footer totals computation

### LINKED ISSUE #1635: Performance: Footer totals recomputed per-column on every PerfTable render
### Context

Surfaced during review of #1624 (perf-table column toggle): https://github.com/tenstorrent/ttnn-visualizer/pull/1624#discussion_r3375717977

### Problem

`PerfTable` rendered its footer by calling `getTotalsForFooter(header, data, hideHostOps)` inside `footerColumns.map(...)`. Each footer column re-walked the full `data` array (`reduce`/`filter`), so a single render did ~6 × O(N) of largely identical work — even though the totals depend only on `data` and `hideHostOps`, not on column selection. The column-toggle feature made this hot: every toggle re-ran all of it. The same pattern existed in `StackedPerfTable.tsx`.

### Resolution (done in #1624)

- Replaced `getTotalsForFooter` with `computeFooterTotals(data, hideHostOps)`, which walks `data` once into a `Record<ColumnKeys, string>` and is memoized on `[data, hideHostOps]`. Footer cells now read `footerTotals[header.key] ?? ''`.
- Applied the same single-pass treatment to `StackedPerfTable.tsx`.

### Possible follow-up

- The broader concern that toggling a column re-renders the entire `<tbody>` (N rows × C columns × comparison datasets of `formatCell`) is **not** covered here — it needs a `React.memo`'d `<PerfTableRow>` and is tracked separately.

This issue documents the footer-totals optimization for traceability; it can be closed once #1624 merges.

dcblundell:
Recomputes footer totals after work done in https://github.com/tenstorrent/ttnn-visualizer/pull/1624.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1635.

### DIFF
--- a/src/components/performance/PerfTable.tsx
+++ b/src/components/performance/PerfTable.tsx
@@ -474,13 +474,14 @@ const PerformanceTable = ({
 };
 
 const computeFooterTotals = (data: TypedPerfTableRow[], hideHostOps: boolean): Partial<Record<ColumnKeys, string>> => {
+    const rows = data ?? [];
     let deviceTimeSum = 0;
     let opToOpGapSum = 0;
     let hostOpsCount = 0;
     let nonUniqueOpsCount = 0;
     let cacheHits = 0;
 
-    for (const row of data) {
+    for (const row of rows) {
         deviceTimeSum += row.device_time || 0;
         opToOpGapSum += row.op_to_op_gap || 0;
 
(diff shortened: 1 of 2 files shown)


aidemsined (approved):
LGTM. Clean single-pass refactor, behavior is preserved on every footer key, and the memoization removes the per-column-toggle re-render cost called out in #1635.

Optional nit (no need to block on this): `computeFooterTotals` always populates the same set of keys, so the return type could be tightened from `Partial<Record<ColumnKeys, string>>` to a non-partial `Record<ColumnKeys, string>` (or just an inline object literal type). That would let the call sites drop the `?? ''` for the known keys and surface any future missing-key mistakes at the type level instead of silently rendering an empty cell. Same applies to `StackedColumnKeys` in `StackedPerfTable.tsx`.

Mergeable as-is.

PR #1642, opened 2026-06-10
Support profiler report files with and without rank suffixes

### LINKED ISSUE #1641: Add support for report files both with and without rank and world size in the filenames
An open PR in TT-Metal changes the the filenames used in reports to have `_<rank>_of_<world_size>` added to the filenames, when the world size is greater than 1. In some places the code was updated to support those filenames, but the assumption was that it would always be there.

The open PR on the TT-Metal side changes it so that `config.json` is still used for single-host reports, `config_x_of_y.yaml` would be used if world size is greater than 1.

The TTNN Visualizer code should be updated to ensure support for both filename format in the reports - both with and without rank and world size.

smountenay-tt:
## Summary

- Load cluster and mesh descriptor YAML files with or without `_<rank>_of_<world>` suffixes, matching tt-metal's single-host vs multi-host naming.
- Reuse the same ranked-file resolution pattern already used for `config.json` / `config_<n>_of_<world>.json`.
- Expose optional `?rank=` on `/api/cluster-descriptor` and `/api/mesh-descriptor` for multi-host debugging.

Closes #1641

## Test plan

- [x] `pytest backend/ttnn_visualizer/tests/test_utils.py -k "mesh_descriptor or cluster_descriptor or profiler_config"`
- [ ] Load a single-host report with `cluster_descriptor.yaml` and `physical_chip_mesh_coordinate_mapping.yaml` and confirm topology view works
- [ ] Load a multi-host report with ranked YAML files and confirm topology view works

### DIFF
--- a/backend/ttnn_visualizer/sftp_operations.py
+++ b/backend/ttnn_visualizer/sftp_operations.py
@@ -30,6 +30,7 @@ from ttnn_visualizer.sockets import FileProgress, FileStatus, emit_file_status
 from ttnn_visualizer.ssh_client import SSHClient, raise_for_ssh_subprocess_error
 from ttnn_visualizer.utils import (
     PROFILER_CONFIG_BASENAME,
+    pick_cluster_descriptor_path,
     ranked_profiler_config_basenames,
     update_last_synced,
 )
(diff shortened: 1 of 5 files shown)


aidemsined (approved):
looks good to me granted Copilot feedbakc is triaged and addressed where appropriate

PR #1650, opened 2026-06-15
Stop mutating cached BufferChunk objects in TensorVisualisationComponent

### LINKED ISSUE #1633: TensorVisualisationComponent: mutates React Query-cached BufferChunk objects
## Context

Follow-up from PR #1631 review thread [discussion_r3374923828](https://github.com/tenstorrent/ttnn-visualizer/pull/1631#discussion_r3374923828). Left for a separate fix because the pattern is a direct port from the legacy per-page code that PR #1631 replaces — not a regression introduced by that PR, but worth cleaning up.

## Problem

`TensorVisualisationComponent` mutates the `BufferChunk` objects returned by React Query in-place:

```tsx
// src/components/tensor-sharding-visualization/TensorVisualisationComponent.tsx, ~L86–105
data.forEach((chunk: BufferChunk) => {
    if (!buffersByBankId[chunk.bank_id]) buffersByBankId[chunk.bank_id] = [];

    if (tensorByAddress) {
        const tensor = tensorByAddress?.get(chunk.address);
        chunk.tensor_id = tensor?.id;      // mutates cached query data
        chunk.color = getTensorColor(tensor?.id);
    } else if (tensorId) {
        chunk.tensor_id = tensorId;        // mutates cached query data
        chunk.color = getTensorColor(tensorId);
    }
    if (chunk.tensor_id === undefined) {
        chunk.color = getBufferColor(chunk.address);   // mutates cached query data
    }

    buffersByBankId[chunk.bank_id].push(chunk);
    coordsByBankId[chunk.bank_id] = { x: chunk.core_x, y: chunk.core_y };
});
```

`data` is the array returned by `useBufferChunks(...)` — i.e. the React Query cache entry. Assigning `chunk.tensor_id` and `chunk.color` mutates that cached object. Two real failure modes:

1. **UI-only fields leak into other consumers.** Any other component that later reads the same cache key sees `tensor_id` / `color` populated by *this* invocation's `tensorByAddress` map, which may be wrong for the other consumer's context.
2. **Stale identity on re-render.** When this component re-renders against the same cached array (e.g. props change), the second pass sees previously-mutated chunks and may produce different `color` values for chunks that fell through different branches the first time.

The pattern was inherited from the legacy per-page code (`page.tensor_id = ...`, `page.color = ...`) and ports across line-for-line in PR #1631. Today it happens to work because there's only one consumer of `useBufferChunks`, but it's a tripwire for any future caller.

## Proposed fix

Project each chunk into a derived shape before pushing into the bank bucket, e.g.:

```tsx
type DecoratedChunk = BufferChunk & { tensor_id?: number; color?: string };

data.forEach((chunk: BufferChunk) => {
    if (!buffersByBankId[chunk.bank_id]) buffersByBankId[chunk.bank_id] = [];

    const tensor = tensorByAddress?.get(chunk.address) ?? null;
    const resolvedTensorId = tensor?.id ?? tensorId;
    const color = resolvedTensorId !== undefined
        ? getTensorColor(resolvedTensorId)
        : getBufferColor(chunk.address);

    const decorated: DecoratedChunk = { ...chunk, tensor_id: resolvedTensorId, color };
    buffersByBankId[chunk.bank_id].push(decorated);
    coordsByBankId[chunk.bank_id] = { x: chunk.core_x, y: chunk.core_y };
});
```

Moving the projection into a `useMemo` keyed on `[data, tensorByAddress, tensorId]` also drops the per-render cost.

## Acceptance criteria

- [ ] No assignment to fields of objects that came directly from `useBufferChunks` (or any React Query hook).
- [ ] `BufferChunk` interface either retains the optional `tensor_id` / `color` for the derived type, or those fields move to a dedicated `DecoratedBufferChunk` type local to the component.
- [ ] Existing rendering / colour assignment behaviour preserved; SVGBufferRenderer still receives one decorated chunk per row with the correct `color`.
- [ ] Add a small focused test (or extend an existing one) that mounts the component twice against the same cached `data` and asserts no field on the input objects changes between renders.

## Out of scope

- Wider audit of React Query mutation patterns elsewhere in the codebase.
- Memoising `tensorByAddress` upstream (separate question if it shows up as a hot path).

## Refs

- PR #1631 (introduced — well, ported — the chunk version of this pattern)
- Original review comment: https://github.com/tenstorrent/ttnn-visualizer/pull/1631#discussion_r3374923828

aidemsined:
## Summary

`TensorVisualisationComponent` was assigning `tensor_id` and `color` in place on the `BufferChunk` objects returned by `useBufferChunks`, i.e. directly on the React Query cache entry. There's only one consumer of that cache key today, so this is currently dormant — but any future caller would see fields populated by another render's `tensorByAddress` map. Same pattern called out in the original PR #1631 review thread.

- Drop `tensor_id?` / `color?` from `BufferChunk` and add a sibling `DecoratedBufferChunk extends BufferChunk` (with `color: string` required, `tensor_id?: number`). Cache type stays pure API data; the render-side projection is the only place those fields live.
- Replace the in-place `data.forEach` mutation with a `useMemo` keyed on `[data, tensorByAddress, tensorId]` that emits `DecoratedBufferChunk[][]` and the existing `coordsByBankId`. Branching matches the original line-for-line (including the `tensorId` falsy-0 skip and the `'red'` final fallback that used to live in `SVGBufferRenderer`).
- Thread `DecoratedBufferChunk` through `SVGBufferRenderer` and `bufferChunksToColoredChunks` — both are only consumed by this component, so no ripple.

Closes #1633.

## Perf

Same O(N) outer pass. One extra shallow spread per chunk (~14-field copy). On re-renders where the memo deps don't change (tensix selection, `showHex` toggle, parent re-renders), the new version skips the per-chunk loop entirely — the original recomputed it every render. Net: equal or faster for any realistic render pattern.

`details.tensorListByAddress` is a class field on `OperationDetails` (set once per `.update()`), so the `tensorByAddress` prop is reference-stable across renders and the memo actually hits.

### DIFF
--- a/src/components/tensor-sharding-visualization/TensorVisualisationComponent.tsx
+++ b/src/components/tensor-sharding-visualization/TensorVisualisationComponent.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 import { Button, ButtonVariant, Card, Overlay2, Size } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import { PlotData } from 'plotly.js';
@@ -11,7 +11,7 @@ import { useAtomValue } from 'jotai';
 import { BufferType } from '../../model/BufferType';
 import { useBufferChunks, useDevices } from '../../hooks/useAPI';
 import 'styles/components/TensorVisualizationComponent.scss';
-import { BufferChunk, Tensor } from '../../model/APIData';
+import { DecoratedBufferChunk, Tensor } from '../../model/APIData';
 import SVGBufferRenderer from './SVGBufferRenderer';
 import { getBufferColor, getTensorColor } from '../../functions/colorGenerator';
 import getChartData, { bufferChunksToColoredChunks } from '../../functions/getChartData';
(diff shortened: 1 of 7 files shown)


src/components/tensor-sharding-visualization/TensorVisualisationComponent.tsx
```diff
@@ -50,6 +50,44 @@ const TensorVisualisationComponent = ({
...
     const [chartData, setChartData] = useState<Partial<PlotData>[]>([]);
 
+    // Project each cached BufferChunk into a DecoratedBufferChunk that
+    // carries the resolved tensor association and the palette colour for
+    // this render. Done off the React Query cache so the cached objects
+    // stay untouched and a future second consumer of useBufferChunks can't
+    // see fields populated by our `tensorByAddress` map. Computed in a
+    // useMemo because both downstream readers (the per-bank grid and the
+    // tensix-detail click handler) need the same shape, and the
+    // address-keyed colour lookups are stable across re-renders.
```
dcblundell, line 60: So long

dcblundell (approved):

PR #1667, opened 2026-06-19
Better Mlir upload UX

dcblundell:
Uses the same file loader UI as local uploads/remote sync, improves error messaging and current status.

Screenshot 2026-06-19 at 15 18 00

### DIFF
--- a/backend/ttnn_visualizer/mlir.py
+++ b/backend/ttnn_visualizer/mlir.py
@@ -167,7 +167,7 @@ def test_mlir_server_connection(
             _endpoint_unreachable_message(remote, http_port),
         )
     else:
-        add_status(ConnectionTestStates.OK, f"MLIR server reachable")
+        add_status(ConnectionTestStates.OK, "MLIR server reachable")
 
     return statuses
 
--- a/backend/ttnn_visualizer/mlir.py
+++ b/backend/ttnn_visualizer/mlir.py
@@ -105,7 +105,7 @@ def _remote_send_command_url(port: int) -> str:
 def _endpoint_unreachable_message(connection: RemoteConnection, port: int) -> str:
     return (
         f"MLIR server not reachable at "
-        f"http://localhost:{port}{MLIR_UPLOAD_PATH} on {connection.host}"
+        f"http://localhost:{port}{MLIR_UPLOAD_PATH} on {connection.host} - check MLIR is running on host"
     )
 
 
@@ -167,7 +167,7 @@ def test_mlir_server_connection(
             _endpoint_unreachable_message(remote, http_port),
         )
     else:
-        add_status(ConnectionTestStates.OK, f"MLIR server reachable (HTTP {http_code})")
+        add_status(ConnectionTestStates.OK, f"MLIR server reachable")
 
     return statuses
 
--- a/src/components/FileStatusOverlay.tsx
+++ b/src/components/FileStatusOverlay.tsx
@@ -34,8 +34,12 @@ const SYNC_META = {
     heading: 'Syncing remote report',
     icon: IconNames.CLOUD_DOWNLOAD,
 } as const;
+const PROCESSING_META = {
+    heading: 'Processing report',
+    icon: IconNames.COG,
+} as const;
 
-type DisplayMeta = typeof UPLOAD_META | typeof SYNC_META;
+type DisplayMeta = typeof UPLOAD_META | typeof SYNC_META | typeof PROCESSING_META;
 
 const FileStatusOverlay = () => {
     const [progress] = useAtom(fileTransferProgressAtom);
@@ -51,10 +55,20 @@ const FileStatusOverlay = () => {
     const overallPercent = getOverallFileTransferPercent(progress);
     const isActive = isActiveTransferStatus(status);
     const isUpload = status === FileStatus.UPLOADING;
-    const displayMeta: DisplayMeta = isUpload ? UPLOAD_META : SYNC_META;
+    // Bytes are sent; the backend is now working (e.g. remote MLIR conversion)
+    // with no further progress to report — show an indeterminate stage.
+    const isProcessing = status === FileStatus.PROCESSING;
+    let displayMeta: DisplayMeta = SYNC_META;
+    if (isUpload) {
+        displayMeta = UPLOAD_META;
+    } else if (isProcessing) {
+        displayMeta = PROCESSING_META;
+    }
     const showByteTotals = bytesTotal !== undefined && bytesTotal > 0;
     const showCurrentFileSize = currentFileSize !== undefined && currentFileSize > 0;
-    const showFileCount = numberOfFiles > 0;
+    // During processing there's no transfer to quantify, so suppress the
+    // file-count / byte-totals line and the percent readout.
+    const showFileCount = numberOfFiles > 0 && !isProcessing;
     // Uploads stream as a single multipart request, so the backend never
     // reports per-file completion (finishedFiles stays 0). Hide the `0/N`
     // prefix in that case to avoid showing misleading progress.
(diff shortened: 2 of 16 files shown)


src/hooks/useMlirUpload.tsx
```diff
@@ -0,0 +1,108 @@
...
+                        currentFileName: fileName,
+                        percentOfCurrent: Math.round((event.loaded * 100) / event.total),
+                        bytesTransferred: event.loaded,
+                        bytesTotal: event.total,
+                        // Once all bytes are sent, the remote MLIR server is
+                        // converting the file — switch to the indeterminate
+                        // processing stage until the request resolves.
+                        status: uploadComplete ? FileStatus.PROCESSING : FileStatus.UPLOADING,
+                    });
+                },
```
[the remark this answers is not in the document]
dcblundell: Fixing.

src/components/FileStatusOverlay.tsx
```diff
@@ -51,10 +55,20 @@ const FileStatusOverlay = () => {
     const overallPercent = getOverallFileTransferPercent(progress);
     const isActive = isActiveTransferStatus(status);
     const isUpload = status === FileStatus.UPLOADING;
-    const displayMeta: DisplayMeta = isUpload ? UPLOAD_META : SYNC_META;
+    // Bytes are sent; the backend is now working (e.g. remote MLIR conversion)
+    // with no further progress to report — show an indeterminate stage.
+    const isProcessing = status === FileStatus.PROCESSING;
+    let displayMeta: DisplayMeta = SYNC_META;
+    if (isUpload) {
```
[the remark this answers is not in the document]
dcblundell: Fixing.

backend/ttnn_visualizer/mlir.py
```diff
@@ -167,7 +167,7 @@ def add_status(status, message, detail=None):
             _endpoint_unreachable_message(remote, http_port),
         )
     else:
-        add_status(ConnectionTestStates.OK, f"MLIR server reachable (HTTP {http_code})")
+        add_status(ConnectionTestStates.OK, f"MLIR server reachable")
```
aidemsined: Backend lint (`ruff`) fails on this line with `F541 [*] f-string without any placeholders` — the `(HTTP {http_code})` substitution was dropped here, so the `f` prefix is now extraneous. Easy fix:

```suggestion
        add_status(ConnectionTestStates.OK, "MLIR server reachable")
```

src/hooks/useMlirUpload.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { AxiosError } from 'axios';
+import axiosInstance from '../libs/axiosInstance';
+import { fileTransferProgressAtom, getInactiveFileTransferProgress } from '../store/app';
+import { FileStatus } from '../model/APIData';
+import Endpoints from '../definitions/Endpoints';
+import { ConnectionStatus, ConnectionTestStates } from '../definitions/ConnectionStatus';
+import { MlirServerConnection } from '../definitions/MlirServer';
+import getResponseError from '../functions/getResponseError';
+
+const useMlirUpload = () => {
```
aidemsined: Nit/observation: `useMlirUpload` is named like a hook but doesn't use any React hook APIs (`useState` / `useEffect` / `useCallback` / `useMemo` / `useAtom`). It's effectively a plain factory returning two side-effecting functions that talk to the Jotai store via `getDefaultStore()`.

This matches the existing `useLocal.tsx` pattern, so it's consistent with house style — flagging in case you want to switch both to plain modules (e.g. `mlirUploadClient.ts`) at some point. Not a blocker.

src/hooks/useMlirRemote.tsx
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            return [{ status: ConnectionTestStates.FAILED, message: getResponseError(err, 'Connection test failed') }];
+        }
+    };
+
+    return {
+        uploadMlirFileToServer,
+        testMlirServerConnection,
+    };
```
aidemsined, line 117: Minor: the name `useMlirUpload` is narrower than what's actually exported — `testMlirServerConnection` is a reachability probe, not an upload. Something like `useMlirServer` / `useMlirRemote` would match the scope better.

Safe to ignore if you prefer to keep file names symmetric with the dialog (`MlirServerDialog`) that consumes both halves.

src/hooks/useMlirRemote.tsx
```diff
@@ -0,0 +1,108 @@
...
+                        // Once all bytes are sent, the remote MLIR server is
+                        // converting the file — switch to the indeterminate
+                        // processing stage until the request resolves.
+                        status: uploadComplete ? FileStatus.PROCESSING : FileStatus.UPLOADING,
+                    });
+                },
+            });
+        } finally {
+            resetTransferProgress();
+        }
```
aidemsined, line 94: Worth a follow-up issue (not for this PR): the `fileTransferProgressAtom` is a single global slot, so two MLIR uploads kicked off in parallel will share it — the first one's `finally` will reset progress to INACTIVE and close the overlay even while the second is still in flight.

This is a **pre-existing** concern that also affects `useLocal`'s local-folder / remote-sync flows, so it isn't introduced here. But now that there are three independent upload paths writing to the same atom, it's a bit easier to trip into. Worth filing a separate issue to either (a) make the atom a stack/per-source map, or (b) explicitly guard against starting a second transfer while one is active. Happy to open the issue if you want me to.

src/components/mlir/MlirJsonFileLoader.tsx
aidemsined, line 1: Nice-to-have on the response handling around lines 80–88: the upload response is structurally destructured via casts — `response?.data?.status`, `(response.data as { graph?: GraphBundle }).graph`, `(response.data as { name?: string }).name`. If you give `uploadMlirFileToServer` a proper return type in `useMlirUpload.tsx` — e.g.

```ts
interface MlirUploadResponse {
    status: ConnectionTestStates;
    message?: string;
    graph?: GraphBundle;
    name?: string;
}
const uploadMlirFileToServer = async (
    files: FileList,
    server: MlirServerConnection,
): Promise<AxiosResponse<MlirUploadResponse>> => { /* ... */ }
```

— these casts disappear and the `status !== ConnectionTestStates.OK` guard becomes type-safe. Optional; current code works.

aidemsined (commented):
Reviewed locally against `dev`: `tsc`, `eslint`, and the full vitest suite (53 files / 421 tests including the two new specs) are all green. Backend `ruff` flags one extraneous f-string — inline suggestion on `mlir.py`.

The `PROCESSING` stage + shared `FileStatusOverlay` is a real UX improvement; comments are mostly nice-to-haves plus a follow-up suggestion for the multi-source `fileTransferProgressAtom` re-entrancy (pre-existing concern, not introduced here).

backend/ttnn_visualizer/mlir.py
dcblundell: Good catch — dropped the extraneous `f` prefix. `ruff check` now passes.

src/hooks/useMlirUpload.tsx
dcblundell: Keeping the `use*` hook convention for consistency with `useLocal`, but renamed the file/function to `useMlirRemote` to better reflect its scope (it owns both the upload and the reachability probe).

src/hooks/useMlirRemote.tsx
dcblundell: Renamed `useMlirUpload` → `useMlirRemote` (file, function, importers and test), which covers both the upload and the `testMlirServerConnection` probe.

src/hooks/useMlirRemote.tsx
dcblundell: Agreed it's worth tracking separately. Filed #1679 for the shared `fileTransferProgressAtom` slot (covering the pre-existing `useLocal` flows too). Leaving out of scope for this PR.

src/components/mlir/MlirJsonFileLoader.tsx
dcblundell: Done — `uploadMlirFileToServer` now returns `Promise<AxiosResponse<MlirUploadResponse>>` and the casts in `MlirJsonFileLoader` are gone, so the `status !== OK` guard is type-safe.

PR #1669, opened 2026-06-22
Keeps uv.lock file in sync on release

dcblundell:
Running the release script will require generating a new uv lock due to the version bump. This PR modifies the release script to regenerate the lockfile and commit it. It also adds a guard to prevent the script from running if the user does not have `uv` installed.

### DIFF
--- a/scripts/release.mjs
+++ b/scripts/release.mjs
@@ -48,6 +48,20 @@ function getCurrentBranch() {
     }
 }
 
+function assertUvAvailable() {
+    try {
+        execSync('uv --version', { stdio: 'ignore' });
+    } catch {
+        console.error(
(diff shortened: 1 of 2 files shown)


aidemsined (approved):

smountenay-tt:
In most cases it's probably going to be fine, but I'm slightly uncomfortable with running `uv lock` in a release script. I believe that the lock file in the release should be a lockfile that has been used and tested by the CI pipeline. In theory, running `uv lock` during release could change not just the ttnn-visualizer version in the lockfile, but also other dependency versions.

It's ok because it solves the immediate problem, but I think a check to ensure it doesn't actually change anything else in the lockfile would be appropriate. Maybe as an issue for the backlog.

smountenay-tt (approved):

PR #1664, opened 2026-06-18
Avoid overlap of device operation data

dcblundell:
**Before**
Screenshot 2026-06-18 at 13 02 28

**After**
Screenshot 2026-06-18 at 13 03 42

Fixes https://github.com/tenstorrent/ttnn-visualizer/issues/1638.

Issue #1651, opened 2026-06-15
Treat globally_allocated circular buffers as tensor views, not as separate allocations


aidemsined:
## Context

Triaged from a report at `resnet50_main_jun10_2110/` (ResNet-50, ops 5 `ttnn.fold` and 8 `ttnn.conv2d`). The CB row for `HaloDeviceOperation` looked like it had CBs sitting on top of the op's L1 sharded tensors — same addresses, same sizes-per-core. Initial read was "CBs overlap tensors and start really late".

After pulling the raw `captured_graph` blob for op 8 from the report DB:

**HaloDeviceOperation CBs vs L1 buffer allocations**

| CB # | Addr | Size/core | `globally_allocated` | Matches buffer |
|--|--|--|--|--|
| 65 | 1,137,312 | 105,824 | **1** | aliased to upstream input tensor (not allocated in this block) |
| 66 | 1,360,480 | 114,080 | **1** | BUF 37 @ 1,360,480 (HEIGHT_SHARDED, `max_size_per_bank=114,080`) |
| 67 | 103,840 | 32 | 0 | anonymous CB |
| 68 | 103,872 | 32 | 0 | anonymous CB |
| 69 | 1,499,104 | 4 | **1** | BUF 43 @ 1,499,104 |
| 70 | 1,499,072 | 4 | **1** | BUF 49 @ 1,499,072 |
| 71 | 1,499,040 | 28 | **1** | BUF 55 @ 1,499,040 |
| 72 | 1,499,008 | 16 | **1** | BUF 61 @ 1,499,008 |

**Conv2dDeviceOperation** shows the same pattern: every CB with `globally_allocated=1` sits at the exact address of an L1 sharded tensor (CB 108 + CB 113 both alias BUF 97 @ 1,147,232; CB 114 aliases BUF 103 @ 1,498,848). Anonymous CBs (`globally_allocated=0`) sit in their own low-address region 103K–791K as expected.

So the "overlap" is correct semantics: a `globally_allocated=1` CB is a kernel-side view bound to an existing L1 sharded buffer (the tensor itself), not a new allocation. Common in sharded Conv/Halo paths. The data is right; we're just drawing it as if it were a fresh allocation.

## Storage

The flag round-trips intact today:

- tt-metal C++ emits `globally_allocated` on every `circular_buffer_allocate` node.
- Importer stores the captured graph as a TEXT blob in `captured_graph.captured_graph` — flag survives end-to-end.
[...]

Issue #1638, opened 2026-06-10
Operation details: Device operations can overlap


dcblundell:
CBs in hidden under Arguments button.

<img width="1314" height="652" alt="Image" src="https://github.com/user-attachments/assets/35b3a687-6c67-4372-8874-fe5632560b48" />

PR #1676, opened 2026-06-23
Align buffer and CB legend columns under a single legend row

aidemsined:
## Summary

Follow-up polish to #1664 / #1638. The CB rows now sit in a `.cb-legend-row` flex (legend width = `parent - 330px`) with a custom 5-col grid (`15px 0.5fr 2fr 10fr 5fr`, sum 17.5fr), while buffer-allocate / buffer-deallocate rows still render their `<MemoryLegendElement>` at full parent width with the default grids. Different available widths *and* different fr totals meant the swatch / address / size / shape columns never lined up between buffer and CB rows.

This PR makes them share one grid:

1. **Buffer rows reuse `.cb-legend-row`.** Each buffer-allocate / buffer-deallocate `<MemoryLegendElement>` is now wrapped in the same flex container, paired with an invisible `.memory-info-placeholder` (the running totals stay in the h4 above; the spacer just reserves the matching 320px right gutter so the legend grid is computed against the same width as CB rows).
2. **Drop the CB-only grid override.** Both row types fall back to the default `.legend-item` grids:
   - CB rows (no `extra-info`): `15px 0.5fr 2fr 13fr 3fr`
   - Buffer rows (`extra-info`):  `15px 0.5fr 2fr 10fr 3fr 3fr`

   Both sum to **18.5fr** after the 15px swatch, so swatch / address / size / shape-info columns line up column-for-column at every viewport width. CB col 4 (tensor id, 13fr) still occupies the combined width of buffer cols 4+5 (tensor id + bufferType+layout), so the inner boundaries differ semantically but the eye reads them as one shared grid.

Scoped under `.device-operations-full-render` so it doesn't bleed into the L1 / DRAM plot legends at the top of the page.

## What this fixes vs. what it doesn't

This is purely the visual-alignment polish that #1664 didn't cover. The original overlap regression that #1664 / #1638 targeted stays fixed.

## Test plan

- [ ] `pnpm exec tsc -p tsconfig.json --noEmit` — clean.
- [ ] `pnpm exec eslint src/components/operation-details/DeviceOperationsFullRender.tsx` — clean.
- [ ] `pnpm exec vitest run tests/MemoryLegendElement.spec.tsx tests/CircularBufferPressureModal.spec.tsx` — 37 / 37 passing.
- [ ] Visual: open the operation details view on a CB-heavy op (the resnet50 report from #1651 reproduces it well). Swatch / address / size columns line up between buffer-allocate, buffer-deallocate, and CB rows. Shape / dtype column lines up too. Running totals on the right stay anchored to the same 320px gutter as before. No new horizontal scroll.

### DIFF
--- a/src/components/operation-details/DeviceOperationsFullRender.tsx
+++ b/src/components/operation-details/DeviceOperationsFullRender.tsx
@@ -378,20 +378,31 @@ function useDeviceOperationsFullRenderModel(args: {
                             key={index}
                             title='Buffer allocate'
                         >
-                            <MemoryLegendElement
-                                chunk={{
-                                    address: parseInt(buffer.address, 10),
-                                    size: bufferSize,
-                                }}
-                                numCores={cores}
-                                key={buffer.address}
-                                memSize={details.l1_sizes[0] || L1_DEFAULT_MEMORY_SIZE}
-                                selectedTensorAddress={selectedAddress}
-                                operationDetails={details}
-                                onLegendClick={onLegendClick}
-                                bufferType={buffer.type}
-                                layout={buffer.layout}
-                            />
+                            {/* Same flex wrapper as CB rows below: pairs the legend
+                                with a 320px placeholder so its grid columns are
+                                computed against the same width as the CB legend.
+                                The running totals live in the h4 above, so we
+                                use an invisible spacer instead of `memoryInfo`. */}
+                            <div className='cb-legend-row'>
+                                <MemoryLegendElement
+                                    chunk={{
+                                        address: parseInt(buffer.address, 10),
+                                        size: bufferSize,
+                                    }}
+                                    numCores={cores}
+                                    key={buffer.address}
+                                    memSize={details.l1_sizes[0] || L1_DEFAULT_MEMORY_SIZE}
+                                    selectedTensorAddress={selectedAddress}
+                                    operationDetails={details}
+                                    onLegendClick={onLegendClick}
+                                    bufferType={buffer.type}
+                                    layout={buffer.layout}
+                                />
+                                <div
+                                    className='memory-info-placeholder'
+                                    aria-hidden='true'
+                                />
+                            </div>
                         </DeviceOperationNodeComponent>
                     );
                 } else if (nodeType === NodeType.buffer_deallocate) {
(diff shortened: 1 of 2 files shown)


src/components/operation-details/DeviceOperationsFullRender.tsx
```diff
@@ -378,20 +378,31 @@ function useDeviceOperationsFullRenderModel(args: {
...
-                                bufferType={buffer.type}
-                                layout={buffer.layout}
-                            />
+                            {/* Same flex wrapper as CB rows below: pairs the legend
+                                with a 320px placeholder so its grid columns are
+                                computed against the same width as the CB legend.
+                                The running totals live in the h4 above, so we
+                                use an invisible spacer instead of `memoryInfo`. */}
+                            <div className='cb-legend-row'>
+                                <MemoryLegendElement
```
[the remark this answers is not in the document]
aidemsined: Agreed — renamed to `memory-legend-row` in e95d2115 (matches the surrounding `MemoryLegendElement` / `memory-info` / `memory-info-placeholder` naming).

src/scss/components/DeviceOperationFullRender.scss
```diff
@@ -155,19 +155,29 @@
...
+    // 320px right gutter (either the in-flow `.memory-info` for CB rows, or
+    // the invisible `.memory-info-placeholder` for buffer rows where totals
+    // live in the h4 above) keeps every legend grid computed against the
+    // same available width. Combined with the default `.legend-item` grid
+    // templates (5-col `15px 0.5fr 2fr 13fr 3fr` and 6-col extra-info
+    // `15px 0.5fr 2fr 10fr 3fr 3fr` — both sum to 18.5fr after the swatch),
+    // this makes swatch / address / size / shape columns line up
+    // column-for-column across buffer and CB rows. Scoped under
+    // `.device-operations-full-render` so it doesn't bleed into the L1 /
+    // DRAM plot legends.
```
dcblundell, line 168: Loooooong

src/components/operation-details/DeviceOperationsFullRender.tsx
```diff
@@ -378,20 +378,31 @@ function useDeviceOperationsFullRenderModel(args: {
...
-                                operationDetails={details}
-                                onLegendClick={onLegendClick}
-                                bufferType={buffer.type}
-                                layout={buffer.layout}
-                            />
+                            {/* Same flex wrapper as CB rows below: pairs the legend
+                                with a 320px placeholder so its grid columns are
+                                computed against the same width as the CB legend.
+                                The running totals live in the h4 above, so we
+                                use an invisible spacer instead of `memoryInfo`. */}
```
dcblundell, line 385: Not sure this is that useful.

dcblundell (approved):

PR #1689, opened 2026-06-26
Add per-op kernel timing to Performance Table

### LINKED ISSUE #1518: Perf table: Expandable per-op kernel timing breakdown
## Context

The perf table shows op-level duration but not the per-kernel breakdown that Tracy already captures. The data is in the DB — it's just not surfaced.

## Ask

Allow expanding a perf-table row to reveal the device-kernel timings for that op:

- Per-kernel name and duration
- Optional: % of op total
- Optional: grouped by core / device when multi-device

## Verification before scoping

Confirm what's actually in the DB today (`profile_log_device.csv` ingest path — see #958) so we know the exact granularity available without a backend change.

## Open questions

- Row-expand vs side panel vs separate route
- Sorting inside the breakdown (by duration desc by default?)
- Behaviour when only host-side timing is available (no device kernel rows)
- Should this also appear in Operation Details, or keep it scoped to the perf table for now?

## Related

- #958 TT Metal: tests for Tracy CSV data integrity

dcblundell:
Adds new columns with colourised cells.

Screenshot 2026-06-26 at 10 27 33

The two-report comparison view compares on `DeviceKernelDuration` (kernel-only, the more precise measure) rather than `DeviceTime`, which is intentionally excluded.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1518.


src/definitions/PerfTable.ts
```diff
@@ -54,6 +54,13 @@ export interface PerfTableRow {
     output_subblock_w: string;
     high_dispatch?: boolean;
     pm_ideal_ns: string;
+    device_kernel_duration: string;
+    brisc_kernel_duration: string;
+    ncrisc_kernel_duration: string;
+    trisc0_kernel_duration: string;
+    trisc1_kernel_duration: string;
+    trisc2_kernel_duration: string;
+    erisc_kernel_duration: string;
```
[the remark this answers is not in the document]
dcblundell: Fixed — these are now typed as `string | null` (with a comment noting the backend `dict.get()` nullability), matching the `hash` field.

src/routes/Performance.tsx
```diff
@@ -317,6 +317,10 @@ const getRowAttributes = (row: PerfTableRow): RowAttributes => {
     };
 };
 
+const NS_PER_US = 1000;
+
+const nsToUs = (value: string | null | undefined): number | null => (value ? parseFloat(value) / NS_PER_US : null);
```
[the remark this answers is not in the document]
dcblundell: Good catch. This line moved to `nsToUs` in `src/functions/math.ts`; I've hardened it to return `null` for non-finite parses via a `Number.isFinite` guard. (Sorting and `formatCell` already treated `NaN` as empty/last, but this keeps the `number | null` field honest.) Added regression tests for whitespace/non-numeric input.

backend/ttnn_visualizer/tests/test_perf_report.py
```diff
@@ -0,0 +1,75 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
```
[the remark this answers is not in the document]
dcblundell: Fixed — updated the SPDX header to © 2026.

src/functions/enrichPerfRowData.ts
```diff
@@ -67,6 +67,13 @@ export const enrichRowData = (
             flops: row.flops ? parseFloat(row.flops) : null,
             flops_percent: row.flops_percent ? parseFloat(row.flops_percent) : null,
             pm_ideal_ns: row.pm_ideal_ns ? parseFloat(row.pm_ideal_ns) : null,
+            device_kernel_duration: row.device_kernel_duration ? parseFloat(row.device_kernel_duration) : null,
+            brisc_kernel_duration: row.brisc_kernel_duration ? parseFloat(row.brisc_kernel_duration) : null,
+            ncrisc_kernel_duration: row.ncrisc_kernel_duration ? parseFloat(row.ncrisc_kernel_duration) : null,
+            trisc0_kernel_duration: row.trisc0_kernel_duration ? parseFloat(row.trisc0_kernel_duration) : null,
+            trisc1_kernel_duration: row.trisc1_kernel_duration ? parseFloat(row.trisc1_kernel_duration) : null,
+            trisc2_kernel_duration: row.trisc2_kernel_duration ? parseFloat(row.trisc2_kernel_duration) : null,
+            erisc_kernel_duration: row.erisc_kernel_duration ? parseFloat(row.erisc_kernel_duration) : null,
```
aidemsined: These keep the raw nanosecond strings as numeric `ns`, but the column declarations in `PerfTable.ts` say `unit: 'µs', decimals: 2` and `formatCell` (`perfFunctions.tsx:239-244`) just appends the unit string — there's no implicit conversion. A 1.5 ms kernel ends up rendering as `1,500,000.00 µs` (1000× too large) on every real report.

The intent looks correct — the column comment in `PerfTable.ts` says `// Stored in µs (converted from the raw ns CSV values in enrichRowData)`, and `nsToUs` was added + tested in `math.ts` / `tests/math.spec.ts` — but `nsToUs` is never actually called. Likely lost in the `Merge branch 'dev'` resolution after #1686 pulled `enrichRowData` out of `Performance.tsx`.

Fix: swap each `row.X_kernel_duration ? parseFloat(row.X_kernel_duration) : null` for `nsToUs(row.X_kernel_duration)`. The `getKernelRiscColour` ratio is unit-independent so the colour logic and its tests stay correct.

src/functions/perfFunctions.tsx
```diff
@@ -384,10 +404,29 @@ export const getCellColour = (row: TypedPerfTableRow, key: ColumnKeys): CellColo
...
+// A neutral colour is used deliberately: being the critical path identifies where the time goes,
+// it does not imply the op is unhealthy.
+export const getKernelRiscColour = (riscUs: number, deviceKernelUs: number | null): CellColour => {
+    if (!deviceKernelUs || deviceKernelUs <= 0) {
+        return DEFAULT_COLOUR;
+    }
+
+    const share = riscUs / deviceKernelUs;
+
+    return share >= KERNEL_CRITICAL_PATH_SHARE ? CellColour.Blue : FALLBACK_COLOUR;
```
aidemsined: Returning `FALLBACK_COLOUR` (Grey, muted) for every cell below the 90% share means in practice 5 of 6 RISC cells in every row are dimmed. The Math Fidelity analogy the comment above leans on only paints one tint, it doesn't mute anything else — much lighter touch.

Suggest leaving non-critical cells at `DEFAULT_COLOUR` (White, readable) and only accenting the critical-path cell with Blue. Keeps the table legible across the full row while still drawing the eye to the bottleneck.

src/functions/perfFunctions.tsx
```diff
@@ -46,6 +46,21 @@ const WARNING_COLOUR = CellColour.Yellow;
...
+const KERNEL_RISC_KEYS: ColumnKeys[] = [
+    ColumnKeys.BriscKernelDuration,
+    ColumnKeys.NcriscKernelDuration,
+    ColumnKeys.Trisc0KernelDuration,
+    ColumnKeys.Trisc1KernelDuration,
+    ColumnKeys.Trisc2KernelDuration,
+    ColumnKeys.EriscKernelDuration,
+];
+// A RISC this close to the device kernel duration is effectively gating the op.
+const KERNEL_CRITICAL_PATH_SHARE = 0.9;
```
aidemsined, line 64: Worth either justifying `0.9` in the comment with some empirical analysis, or lowering it. Two RISCs at 88% / 87% with the kernel at 95 µs are effectively co-gating but neither lights up — the critical-path signal is harder to read once those near-miss cases start showing up on real reports.

src/definitions/PerfTable.ts
```diff
@@ -273,6 +312,13 @@ export const comparisonKeys: ColumnKeys[] = [
     ColumnKeys.OpCode,
     ColumnKeys.OpToOpGap,
     ColumnKeys.TotalPercent,
+    ColumnKeys.DeviceKernelDuration,
```
aidemsined, line 315: Adds `DeviceKernelDuration` to the comparison list but `DeviceTime` is still absent — likely intentional (kernel-only is the more precise measure), but worth a sentence in the PR body so reviewers know what's being compared across two reports. Either way, no objection.

src/functions/perfFunctions.tsx
dcblundell: With the muting removed (see the sibling thread), near-miss RISCs now stay readable rather than dimmed, so the legibility concern is addressed even when they aren't accented. I've kept the `0.9` threshold but reworded the comment to make clear it's a display heuristic (which RISC to accent), not a correctness boundary.

aidemsined (approved):

PR #1691, opened 2026-06-29
docs: add Figma nav/search/modal layout template and shared assets

ctr-fmamedov-crypto:
## Summary

- `docs/shared/_templates/layout.html`: full Jinja2 template with top navigation bar, OpenSearch search modal, Kapa.ai Ask AI, PostHog analytics, feedback widget
- `docs/shared/_static/docs-toc.js`: sticky TOC behaviour
- `docs/shared/_static/logotype.png`: Tenstorrent logotype for sidebar
- `docs/shared/_templates/feedback_widget.html`: thumbs up/down feedback widget

Scripts loaded via `pathto()` (local `shared/_static/` is on `html_static_path`).

## Merge order

Requires `docs/sphinx-config` in this repo to be merged first. Also requires `tenstorrent.github.io` [`docs/shared-static-assets`](https://github.com/tenstorrent/tenstorrent.github.io/pull/168) and [`docs/shared-theme-css`](https://github.com/tenstorrent/tenstorrent.github.io/pull/169) to be live.

## Test plan

- [ ] Navigation dropdowns render correctly
- [ ] Search modal returns results
- [ ] Ask AI button loads Kapa widget
- [ ] Feedback widget visible on each page


docs/shared/_templates/feedback_widget.html
```diff
@@ -0,0 +1,210 @@
...
+        window.posthog.capture('docs_feedback', payload);
+      } else {
+        console.log('Feedback submitted:', payload);
+      }
+      return true;
+    } catch (e) {
+      console.error('Feedback submit error', e);
+      return false;
+    }
+  }
```
aidemsined, line 145: Non-blocking: the submit path returns `true` even when PostHog isn't loaded, so the user sees "Thank you for your feedback!" while the payload only lands in `console.log`. If shared-static-assets ever ships without PostHog, or an ad-blocker is in the way, feedback evaporates with a positive UX confirmation. Suggest either treating the no-posthog branch as `return false` (which already shows the error state) or routing through a real transport before the success message. As-is it's silent data loss disguised as success.

docs/shared/_static/docs-toc.js
```diff
@@ -0,0 +1,99 @@
...
+
+    var ul = document.createElement('ul');
+
+    headings.forEach(function (h) {
+      if (!h.id) {
+        h.id = h.textContent.trim()
+          .toLowerCase()
+          .replace(/[^a-z0-9]+/g, '-')
+          .replace(/^-|-$/g, '');
+      }
```
aidemsined, line 31: Non-blocking: the auto-slug doesn't deduplicate, so two `<h2>` (or `<h3>`) elements with the same text on the same page collapse to the same `id`. The TOC's anchor for the second one will jump to the first. Sphinx normally pre-assigns unique ids so this is the fallback path, but it's worth a `Set` check to suffix duplicates with `-2`, `-3`, etc.

docs/shared/_templates/layout.html
```diff
@@ -1,34 +1,201 @@
...
 {%- set _root_doc = root_doc|default(master_doc) %}
+{%- set _home = "https://docs.tenstorrent.com/" %}
+<nav class="tt-top-nav" role="navigation" aria-label="Top navigation">
+  <div class="tt-top-nav-left">
+    <a href="{{ _home }}" class="tt-nav-logo">
+      <img src="{{ pathto('_static/logotype.png', 1) }}" alt="Tenstorrent" />
+      <span>DOCUMENTATION</span>
+    </a>
+    <ul class="tt-nav-links">
+      <li class="has-dropdown">
```
aidemsined, line 15: Non-blocking a11y: the dropdown parents are plain `<a href="...#anchor">` with no `aria-haspopup`, no `aria-expanded`, and no click handler to open the submenu. Submenus presumably reveal via `:hover` / `:focus-within` in the shared CSS, but there's no keyboard-only way to open a menu, see its items, and pick one without leaving the page (the parent `<a>` navigates away on `Enter`). Worth wiring up a small toggle + arrow-key handling in the shared JS, or at minimum making the parents `<button type="button">` so they don't navigate when expanded.

docs/shared/_templates/layout.html
```diff
@@ -1,34 +1,201 @@
...
+    <div class="tt-nav-actions">
+      <a href="https://docs.tenstorrent.com/tt-awesome/" class="tt-nav-btn-cta" target="_blank" rel="noopener">Explore</a>
+      <a href="https://github.com/tenstorrent" class="tt-nav-btn-github" target="_blank" rel="noopener">
+        <svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false">
+          <path d="M8 0C3.58 0 0 3.58 0 8c0 3.54 2.29 6.53 5.47 7.59.4.07.55-.17.55-.38 0-.19-.01-.82-.01-1.49-2.01.37-2.53-.49-2.69-.94-.09-.23-.48-.94-.82-1.13-.28-.15-.68-.52-.01-.53.63-.01 1.08.58 1.23.82.72 1.21 1.87.87 2.33.66.07-.52.28-.87.51-1.07-1.78-.2-3.64-.89-3.64-3.95 0-.87.31-1.59.82-2.15-.08-.2-.36-1.02.08-2.12 0 0 .67-.21 2.2.82.64-.18 1.32-.27 2-.27.68 0 1.36.09 2 .27 1.53-1.04 2.2-.82 2.2-.82.44 1.1.16 1.92.08 2.12.51.56.82 1.27.82 2.15 0 3.07-1.87 3.75-3.65 3.95.29.25.54.73.54 1.48 0 1.07-.01 1.93-.01 2.2 0 .21.15.46.55.38A8.013 8.013 0 0016 8c0-4.42-3.58-8-8-8z"/>
+        </svg>
+        Github
+      </a>
+    </div>
+    <i data-toggle="wy-nav-top" class="fa fa-bars tt-nav-hamburger" aria-hidden="true"></i>
```
aidemsined, line 91: Non-blocking a11y: the mobile-nav hamburger is an `<i aria-hidden="true">` element acting as an interactive control. The upstream RTD JS still picks it up via `data-toggle`, but assistive tech can't see or operate it. Pre-existing pattern inherited from sphinx_rtd_theme, but since this PR is a theme refresh anyway, swapping to `<button type="button" data-toggle="wy-nav-top" aria-label="Open navigation">` would be a cheap win.

aidemsined (commented):
Reviewed locally. Four non-blocking observations inline (one feedback-widget UX bug, one TOC edge case, two a11y nits on the nav). PR status unchanged.

PR #1693, opened 2026-06-29
Support multiple MLIR file uploads and processing

dcblundell:
Screenshot 2026-06-30 at 12 18 47

Screenshot 2026-06-30 at 12 22 27

Screenshot 2026-06-30 at 12 22 30

Closes


backend/ttnn_visualizer/views.py
```diff
@@ -1730,6 +1731,23 @@ def test_mlir_server():
     )
 
 
+def _unique_mlir_name(base: str, used: set) -> str:
```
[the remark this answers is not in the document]
dcblundell: Done — changed to `set[str]` for consistency with the rest of the backend. Note: this repo's mypy config doesn't enable `disallow_any_generics`/`strict` and these annotations sit in untyped function bodies, so the bare `set` wasn't actually producing a *Missing type parameters* error — but matching the `set[str]` convention is worth doing regardless.

backend/ttnn_visualizer/views.py
```diff
@@ -1754,45 +1772,83 @@ def upload_mlir_server():
...
-    mlir_name = Path(Path(file.filename or "model").name).stem
-    mlir_path = target_directory / f"{mlir_name}.json"
-    mlir_path.write_text(result.graph_json, encoding="utf-8")
+    # Convert every uploaded file independently. One file failing to convert
+    # must not abort the others, so per-file outcomes are collected and
+    # returned as a list; the caller surfaces them in the results overlay and
+    # picks which converted graph to make active. The active MLIR is set
+    # separately via `/mlir/active` so nothing is activated until the user
+    # chooses.
+    used_names: set = set()
```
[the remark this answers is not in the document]
dcblundell: Done — changed to `set[str]` for consistency with the rest of the backend. Note: this repo's mypy config doesn't enable `disallow_any_generics`/`strict` and these annotations sit in untyped function bodies, so the bare `set` wasn't actually producing a *Missing type parameters* error — but matching the `set[str]` convention is worth doing regardless.

backend/ttnn_visualizer/views.py
```diff
@@ -1769,45 +1787,83 @@ def upload_mlir_server():
...
-    if not instance_id:
-        return response_bad_request("Missing required query parameter: instanceId")
+        if (
+            result.status.status == ConnectionTestStates.OK.value
+            and result.graph_json is not None
+        ):
+            mlir_name = _unique_mlir_name(Path(Path(filename).name).stem, used_names)
+            used_names.add(mlir_name)
+            mlir_path = target_directory / f"{mlir_name}.json"
+            mlir_path.write_text(result.graph_json, encoding="utf-8")
```
aidemsined: Cross-batch name collisions can silently clobber prior uploads.

`_unique_mlir_name` only consults `used_names` within the current request. Concrete scenario:

- Batch A: upload `model.mlir` + `model.pb` → written as `model.json` and `model (2).json`.
- Batch B: any subsequent upload that disambig-walks into `model (2)` (e.g. another `model.mlir` + `model.pb`, or a single file already named `model (2)`) → `model (2).json` from batch A is silently overwritten.

The docstring on `_unique_mlir_name` defends the *intentional* overwrite of `model.json` when re-uploading `model.mlir`. The `model (2)` clobber is not intentional — the user never asked to refresh it, the disambig counter just happened to land there.

One-line fix at the top of `upload_mlir_server`:

```python
used_names: set[str] = {p.stem for p in target_directory.glob("*.json")}
```

`test_upload_endpoint_converts_every_file` only covers within-batch collisions; a cross-batch regression test would be worth adding alongside the fix.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -0,0 +1,127 @@
...
+        // restores them.
+        if (result.persisted) {
+            try {
+                await setActiveMlir(result.name);
+            } catch (err: unknown) {
+                createToastNotification('MLIR', getResponseError(err, 'Unable to set active MLIR'), ToastType.ERROR);
+            }
+        }
+
+        createToastNotification('MLIR', result.filename, ToastType.SUCCESS);
```
aidemsined, line 66: Double toast on persistence failure.

If `setActiveMlir` rejects, the ERROR toast fires inside the catch and execution falls through to the unconditional SUCCESS toast on line 65. The user sees "Unable to set active MLIR" immediately followed by a success toast carrying the filename — contradictory, and the success toast misrepresents the state (the in-memory active graph is set, but the persisted instance pointer isn't).

Either `return` after the error toast, or only fire the SUCCESS toast on the happy path:

```tsx
if (result.persisted) {
    try {
        await setActiveMlir(result.name);
    } catch (err: unknown) {
        createToastNotification('MLIR', getResponseError(err, 'Unable to set active MLIR'), ToastType.ERROR);
        return; // or skip the SUCCESS toast below
    }
}

createToastNotification('MLIR', result.filename, ToastType.SUCCESS);
```

src/components/mlir/MlirJsonFileLoader.tsx
```diff
@@ -41,87 +38,130 @@ interface MlirJsonFileLoaderProps {
...
+                inputProps={{ accept: acceptedExtensions, multiple: true }}
+            />
+
+            <Button
+                variant={ButtonVariant.OUTLINED}
+                intent={Intent.PRIMARY}
+                icon={IconNames.LIST}
+                text='View MLIR uploads'
+                onClick={() => setMlirFileResultsOpen(true)}
+                disabled={!mlirFileResults?.length}
```
aidemsined: `View MLIR uploads` button is enabled with stale PROGRESS rows during an in-flight upload.

`handleFileChange` writes PROGRESS spinner rows to `mlirFileResultsAtom` *before* `uploadMlirFileToServer` is awaited (lines 103–111). The button's disabled check here is `!mlirFileResults?.length` — length is > 0 while those PROGRESS rows are present, so the button is clickable mid-upload.

Clicking it opens `MlirFileResultsOverlay`, which renders "No files could be processed." (because `successCount` counts only `status === OK`) next to a list of spinning rows. The intended in-progress UX is `FileStatusOverlay`'s processing list, not this overlay.

Fix: gate on settled rows, e.g.

```tsx
const hasSettledResults = !!mlirFileResults?.some(
    (r) => r.status !== ConnectionTestStates.PROGRESS,
);
// ...
disabled={!hasSettledResults}
```

src/components/mlir/MlirJsonFileLoader.tsx
```diff
@@ -41,87 +38,130 @@ interface MlirJsonFileLoaderProps {
...
+                // FileStatusOverlay; the results overlay opens once converted.
+                const response = await uploadMlirFileToServer(files, server);
+                const results: MlirFileResult[] = (response?.data?.results ?? []).map((result) => ({
+                    filename: result.filename,
+                    name: result.name,
+                    status: result.status,
+                    message: result.message,
+                    graph: result.graph ?? null,
+                    persisted: true,
+                }));
```
aidemsined, line 123: Nit / structural: `MlirFileResult` (`MLIRJsonModel.ts`) and `MlirFileUploadResult` (`useMlirRemote.tsx`) are near-duplicates — same five core fields, only `persisted` (UI concern) and `detail` (server-side error context) differ. This hand-mapping silently drops `result.detail`, which the backend may populate via `StatusMessage.detail` on failure paths.

Two options:

1. Fold them into one type with both `persisted: boolean` and `detail?: string`, and just spread the response.
2. Keep them separate but at least surface `detail` here so it reaches the user, e.g. `message: result.message ?? result.detail`.

Not blocking, but the duplication is the kind that drifts as the API grows.

backend/ttnn_visualizer/views.py
```diff
@@ -1769,45 +1787,83 @@ def upload_mlir_server():
...
+
+            entry["name"] = mlir_name
+            # `graph_json` is already a JSON string, so embed it verbatim rather
+            # than re-serialising — the caller renders it without a follow-up
+            # `/mlir` fetch.
+            entry["graph"] = orjson.Fragment(result.graph_json.encode("utf-8"))
+
+        results.append(entry)
```
aidemsined, line 1836: Follow-up scope: uploaded MLIR files are effectively write-only.

This endpoint writes one `<name>.json` per uploaded file into the MLIR directory, and `set_active_mlir` records one of them as the instance pointer. The others stay on disk indefinitely with no list endpoint to surface them and no delete endpoint to clean them up. Combined with `mlirFileResultsAtom` being in-memory only (not `atomWithStorage`), once the user reloads, every non-active file from the batch becomes invisible — discoverable only by re-uploading (which feeds back into the cross-batch collision concern above).

Not for this PR, but worth a follow-up issue to track:

1. `GET /mlir/list` to enumerate stored reports, so the picker can rehydrate after reload.
2. `DELETE /mlir/<name>` to let users prune unwanted uploads.
3. Or — alternatively — write uploads to a per-session scratch dir and only promote the chosen file to the persisted MLIR directory on `set_active_mlir`.

dcblundell:
Addressed reviewer feedback in this PR and resolved the six in-scope threads.

Implemented fixes:
- Backend: validate and normalise active MLIR name input (type/empty checks + .json suffix normalised to stem).
- Backend: align decorator order for set_active_mlir with repository convention.
- Backend: prevent cross-batch disambiguation collisions from clobbering previously created sibling names while preserving intentional base-name refresh.
- Frontend: prevent contradictory success toast when persisting active MLIR fails.
- Frontend: keep View MLIR uploads disabled until at least one settled (non-PROGRESS) result exists.
- Frontend: preserve backend detail when message is missing in upload results.

Added regression coverage for the above backend and frontend cases, including cross-batch collision behaviour and active-MLIR validation/normalisation.

The remaining lifecycle concern is tracked separately here: https://github.com/tenstorrent/ttnn-visualizer/issues/1705

PR #1703, opened 2026-06-30
Extend CI smoke tests: upload demo report and exercise main tabs

### LINKED ISSUE #1644: Extend CI smoke tests: upload demo report and exercise main tabs
## Summary

The existing [Smoke Tests](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/.github/workflows/smoke-tests.yml) workflow already installs Playwright, starts a real `ttnn-visualizer` server, and runs [`scripts/smoke_test.py`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/scripts/smoke_test.py). That test only verifies the app loads at `http://localhost:8000` and contains expected text — it does not exercise the core memory-profiler views.

We should extend the smoke suite with **one or two additional Playwright tests** that:

1. **Upload a demo memory report through the browser** (local folder upload flow).
2. **Navigate the main memory-profiler tabs** — **Operations**, **Tensors**, and **Buffers**.
3. **Assert the app stays healthy** — no 5xx API responses, no unhandled error UI, and each tab renders without crashing.

This would catch regressions in the upload → report activation → data-fetch → render path that the current “page loads” check misses.

## Motivation

Several recent classes of bugs only show up once a report is loaded and the user visits a tab:

- Backend API errors (500s) when querying operations, tensors, or buffers for a real report shape.
- Frontend crashes when rendering tab content after upload.
- Broken navigation / disabled tabs after a successful upload.

A lightweight end-to-end smoke test on every PR to `dev` would give early signal without duplicating the full Vitest suite.

## Current state (already in place)

| Piece | Location |
|-------|----------|
| CI workflow | [`.github/workflows/smoke-tests.yml`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/.github/workflows/smoke-tests.yml) |
| Playwright install in CI | `pip install playwright` + `playwright install --with-deps chromium` |
| Existing smoke test | [`scripts/smoke_test.py`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/scripts/smoke_test.py) |
| Main nav tabs | [`src/components/MainNavigation.tsx`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/components/MainNavigation.tsx) — Operations (`/operations`), Tensors (`/tensors`), Buffers (`/buffer-summary`) |
| Upload UI test IDs | [`TEST_IDS.LOCAL_PROFILER_UPLOAD`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/definitions/TestIds.ts) |
| Key API routes | `/api/operations`, `/api/tensors`, `/api/buffers` |

**No new tooling is required** — this is mostly extending the existing Python + Playwright script (or splitting it into a small module with 2–3 test functions).

### 1. Add a small bundled demo report fixture

Commit a minimal but valid profiler report directory under something like `tests/fixtures/smoke/` (e.g. a `db.sqlite` built from the existing [`SCHEMA_V2`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/backend/ttnn_visualizer/tests/report_schemas.py) test schema with a handful of rows). Keep it small so CI stays fast.

> Alternatively, vendor a trimmed version of one of the [sample reports documented in the README](https://github.com/tenstorrent/ttnn-visualizer#sample-reports) — but a purpose-built minimal fixture is preferable for speed and stability.

### 2. Playwright test: upload + tab walkthrough

Rough flow:

```python
# Pseudocode — extend scripts/smoke_test.py or add scripts/smoke_test_report_tabs.py

async def test_upload_and_main_tabs():
    page.on("response", lambda r: assert r.status < 500 for /api/ requests)

    await page.goto("http://localhost:8000")
    await page.get_by_test_id("local-profiler-upload").set_input_files(fixture_dir)

    # Wait for upload success (toast, status indicator, or nav tabs becoming enabled)
    await page.get_by_role("button", name="Operations").click()
    await expect_no_error_state()
    await page.get_by_role("button", name="Tensors").click()
    await expect_no_error_state()
    await page.get_by_role("button", name="Buffers").click()
    await expect_no_error_state()
```

**Assertions (pick a pragmatic subset):**

- No `/api/*` response with status ≥ 500 during the flow.
- Operations / Tensors / Buffers nav buttons are enabled after upload.
- Each tab loads without showing the error overlay (`TEST_IDS.ERROR_MESSAGE`) or an empty crash state.
- Optionally: wait for a known element on each route (e.g. page title / table skeleton disappearing).

### 3. Wire into CI

Add a step (or extend the existing one) in `smoke-tests.yml`:

```yaml
- name: Run Playwright smoke test (report tabs)
  run: python scripts/smoke_test_report_tabs.py
```

Run on the same matrix as today (Python 3.10–3.14) so we know the wheel + server + browser flow works across supported versions.

## Out of scope (for this issue)

- Performance, NPE, Graph, Topology, or MLIR tabs — focus on the three core memory-profiler tabs first.
- Deep UI assertions (row counts, chart pixels, etc.) — smoke test only.
- Replacing Vitest / pytest coverage — this complements, not replaces, unit and API tests.

## Acceptance criteria

- [ ] CI smoke workflow uploads a bundled demo report via the browser.
- [ ] Playwright clicks **Operations**, **Tensors**, and **Buffers** after upload.
- [ ] Test fails on any 5xx `/api/` response or visible error state during the flow.
- [ ] Existing “app loads” smoke test continues to pass unchanged.
- [ ] Total added CI time stays reasonable (target: &lt; 30s extra per matrix job).

## Notes

- The smoke test runs against the **installed wheel** (production build), which is the right surface for catching packaging / static-asset regressions.
- Local-only upload endpoints are available because CI starts the server without `SERVER_MODE`.
- Playwright’s `aria-label` selectors (`Operations`, `Tensors`, `Buffers`) or existing CSS classes (`.operations-button`, etc.) are both viable; prefer `aria-label` / test IDs for stability.

smountenay-tt:
## Summary
- Extend `scripts/smoke_test.py` to upload each `demo-reports/` zip and click through **Operations**, **Tensors**, and **Buffers**
- Add `scripts/smoke_test_helpers.py` for fixture extraction, API 5xx tracking, and a preflight check that the production SPA is being served
- Fail fast locally when Flask is running in development mode (API-only on port 8000)

Closes #1644

## Test plan
- [x] `pnpm build && FLASK_ENV=production uv run ttnn-visualizer` then `uv run python scripts/smoke_test.py`
- [x] CI smoke workflow passes on this PR

### DIFF
--- a/scripts/smoke_test.py
+++ b/scripts/smoke_test.py
@@ -4,51 +4,176 @@
 # SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
 """
-Simple smoke test using Playwright to verify the ttnn-visualizer web app loads correctly.
+Playwright smoke tests for the ttnn-visualizer web app.
+
+Verifies the app loads, uploads demo memory reports, and exercises the core
+memory-profiler tabs without API or UI errors.
 """
+
+from __future__ import annotations
 
 import asyncio
 import sys
+import tempfile
+from pathlib import Path
 
-from playwright.async_api import async_playwright
+from playwright.async_api import Browser, Page, TimeoutError, async_playwright, expect
+
+_SCRIPTS_DIR = Path(__file__).resolve().parent
+if str(_SCRIPTS_DIR) not in sys.path:
+    sys.path.insert(0, str(_SCRIPTS_DIR))
+
+from smoke_test_helpers import (
+    BASE_URL,
+    DEMO_REPORT_ZIPS,
+    DEMO_REPORTS_DIR,
+    MAIN_TAB_NAMES,
+    ApiErrorTracker,
+    assert_no_error_ui,
+    extract_profiler_report_dir,
+    verify_server_serving_spa,
+)
+
+UPLOAD_TIMEOUT_MS = 180_000
+TAB_TIMEOUT_MS = 120_000
 
 
-async def smoke_test():
-    """Run basic smoke test to verify the web app loads."""
-    async with async_playwright() as p:
-        browser = await p.chromium.launch()
+async def smoke_test_app_loads(page: Page) -> None:
+    """Verify the web app loads and contains expected content."""
+    await page.goto(BASE_URL, timeout=10_000)
+    await page.wait_for_load_state("networkidle", timeout=10_000)
+
+    title = await page.title()
+    print(f"✅ Page loaded successfully. Title: {title}")
+
+    body_text = (await page.text_content("body")) or ""
+    if (
+        "visualizer" in body_text.lower()
+        or "ttnn" in body_text.lower()
+        or "tt-nn" in body_text.lower()
+    ):
+        print("✅ Page contains expected content")
+        return
+
+    raise RuntimeError(
+        "Page did not contain expected TT-NN Visualizer content.\n"
+        f"Body text preview: {body_text[:200]}..."
+    )
+
+
+async def upload_profiler_report(page: Page, report_dir: Path) -> None:
+    """Upload a local memory report directory via the Reports page."""
+    await page.get_by_role("button", name="Reports").click()
+    await page.wait_for_url(f"{BASE_URL}/**")
+
+    upload_input = page.get_by_test_id("local-profiler-upload")
+    await upload_input.set_input_files(str(report_dir))
+
+    operations_button = page.get_by_role("button", name="Operations")
+    try:
+        await expect(operations_button).to_be_enabled(timeout=UPLOAD_TIMEOUT_MS)
+    except AssertionError as exc:
+        status = page.get_by_test_id("local-profiler-status")
+        status_text = (
+            await status.text_content() if await status.count() > 0 else "unknown"
+        )
+        raise TimeoutError(
+            f"Operations tab did not become enabled after upload ({status_text})"
+        ) from exc
+
+    await assert_no_error_ui(page)
+    print(f"✅ Uploaded report from {report_dir.name}")
+
+
+async def exercise_main_tabs(page: Page) -> None:
+    """Click Operations, Tensors, and Buffers and wait for each view to settle."""
+    tab_urls = {
+        "Operations": "**/operations**",
+        "Tensors": "**/tensors**",
+        "Buffers": "**/buffer-summary**",
+    }
+
+    for tab_name in MAIN_TAB_NAMES:
+        tab_button = page.get_by_role("button", name=tab_name)
+        if not await tab_button.is_enabled():
+            raise AssertionError(f"{tab_name} tab is disabled after upload")
+
+        await tab_button.click()
+        await page.wait_for_url(tab_urls[tab_name], timeout=TAB_TIMEOUT_MS)
+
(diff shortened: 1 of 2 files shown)


scripts/smoke_test.py
```diff
@@ -4,51 +4,177 @@
...
+        tab_button = page.get_by_role("button", name=tab_name)
+        if not await tab_button.is_enabled():
+            raise AssertionError(f"{tab_name} tab is disabled after upload")
+
+        await tab_button.click()
+        await page.wait_for_url(tab_urls[tab_name], timeout=TAB_TIMEOUT_MS)
+
+        spinner = page.locator(".loading-spinner").first
+        try:
+            if await spinner.is_visible():
```
aidemsined: `is_visible()` returns immediately from the current DOM, so a spinner that renders even a few ms after `tab_button.click()` will be missed and the test falls through to `wait_for_load_state("networkidle")` while data is still being fetched. Consider `spinner.wait_for(state="visible", timeout=<short>)` inside a try/except before the hidden-wait, or drop the spinner check entirely and gate on a stronger per-tab signal (e.g. a known root element in each view).

scripts/smoke_test.py
```diff
@@ -4,51 +4,177 @@
...
+    """Click Operations, Tensors, and Buffers and wait for each view to settle."""
+    tab_urls = {
+        "Operations": "**/operations**",
+        "Tensors": "**/tensors**",
+        "Buffers": "**/buffer-summary**",
+    }
+
+    for tab_name in MAIN_TAB_NAMES:
+        tab_button = page.get_by_role("button", name=tab_name)
+        if not await tab_button.is_enabled():
```
aidemsined: This is a synchronous read against the current DOM — if the tab is briefly disabled during navigation between iterations, it asserts prematurely. The upload path already uses `expect(operations_button).to_be_enabled(timeout=...)` for the same reason; mirroring that here (with a short timeout) would remove the race.

scripts/smoke_test_helpers.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def extract_profiler_report_dir(zip_path: Path, work_dir: Path) -> Path:
+    """Extract the memory-profiler report folder from a demo zip archive."""
+    with zipfile.ZipFile(zip_path) as archive:
+        db_paths = [name for name in archive.namelist() if name.endswith("/db.sqlite")]
+        if not db_paths:
+            raise ValueError(f"No db.sqlite found in {zip_path}")
+
+        report_prefix = f"{db_paths[0].rsplit('/', 1)[0]}/"
```
aidemsined, line 107: `db_paths[0]` silently drops every other report if a fixture ever contains more than one `db.sqlite` — fine for today's demo zips, but not obvious from the outside. Worth either asserting `len(db_paths) == 1` with a clear error, or iterating over every found prefix and returning a list, so future multi-report fixtures don't quietly test only the first one.

aidemsined (approved):

Issue #1707, opened 2026-07-01
MLIR op filter: extract `computeFilterMatches` as a pure function with unit tests


aidemsined:
Follow-up to the op-name filter work on `feature/1545-mlir-op-filter`. The filter's core resolution logic — walking `sourceNodes`, resolving each label match to its visible representative via `containingNamespacesByNodeId`, tallying buried counts — currently lives inline as a `useMemo` in `MLIRViewReactFlow.tsx`.

## Motivation

- The logic isn't unit-testable without mounting React Flow. As the filter grows subtle behaviour (case-sensitive, whole-word, regex, negation), the memo will accumulate branches with no test coverage.
- The semantic cases the memo already handles — top-level match, buried under one collapse, buried under nested collapses, match under a synthetic `section_N_of_M` topology wrapper — all deserve fixtures.
- Extracting also isolates the `visibleNodeIds.has(anchorByNamespace[ns])` collapsibility test from the React state graph, making the "why we don't read `expandedNamespaces` here" argument easier to defend on future refactors.

## Proposed shape

```ts
// src/components/mlir/mlirFilter.ts
export interface FilterMatchInfo {
    visibleRepIds: Set<string>;
    buriedCountByRepId: Map<string, number>;
    hiddenMatchCount: number;
}

export function computeFilterMatches(
    sourceNodes: readonly MlirSourceNode[],
    visibleNodeIds: ReadonlySet<string>,
    interactionIndex: WorkerInteractionIndex | null,
    query: string,
): FilterMatchInfo | null;
```

The `MLIRViewReactFlow` memo shrinks to a `visibleNodeIds` build plus a call into `computeFilterMatches`.

## Test fixtures (minimum)

- Empty query → returns `null`.
- Query matching nothing → empty sets, `hiddenMatchCount = 0`.
- Top-level match with no enclosing namespace → source is its own rep.
- Match under a single collapsed namespace → rep is the anchor, `buriedCountByRepId[anchor] = 1`.
- Match under nested collapsed namespaces → rep is the outermost collapsed anchor.
- Match under a synthetic `section_N_of_M` topology wrapper → rep is the section anchor (proves `containingNamespacesByNodeId` covers synthetic sections).
- Match inside an expanded namespace where the source is on canvas → source is its own rep.
- Multiple matches under the same collapsed anchor → `buriedCountByRepId[anchor] > 1` and `hiddenMatchCount` reflects the total.

## Not blocking

[...]

Issue #1545, opened 2026-05-20
MLIR viewer: Live filter by op name (substring, dim non-matches)


aidemsined:
Part of `mlir-integration-2` — Basic End-to-End Graph Rendering.

Op names are not unique in practice (a typical MLIR graph has hundreds of `stablehlo.add` / `ttir.matmul` / etc.), so a single-hit search is the wrong primitive. Instead, give the user a live filter that highlights the matching subset by dimming everything else, with no layout change.

## Behaviour

- Persistent floating input in the top-left of the React Flow canvas (sits above the controls panel so `#1471` can redesign controls independently).
- Case-insensitive **substring** match on the node `label` (op name).
- Live filtering as the user types (~200 ms debounce).
- Non-matching op nodes dim via CSS opacity (~0.2). No add/remove, no re-layout.
- Edges stay bright **only if both endpoints match**; otherwise they dim too. Keeps the matched subset readable.
- Group / namespace wrappers stay neutral so structure is still legible.
- Empty query restores full brightness (the canvas is the default state).
- 0 matches: don't dim everything to invisibility — show the badge and leave the canvas untouched so the user can fix the query.
- Match-count badge in the input: `42 matches` / `no matches` / blank when query is empty.
- Prev / Next arrows cycle through matches, panning to each (no auto-pan on the bulk filter — too jarring with many hits).
- Cmd+F (Ctrl+F on Linux/Windows) focuses the input; Esc clears it.

## Out of scope (follow-ups)

- Regex mode — drop-in once the index exists (`/^stablehlo\..*/i`); follow-up issue.
- Namespace / attribute predicates (`ns:foo`, `attr:dtype=f32`) — same index, same dim mechanism.
- Paged results sidebar for >N matches — tracked separately in `#1546`.
- Global cross-view search bar — `#1524`.

## Implementation notes

- Build a flat node index once per `GraphBundle` (label, op type, namespace, source location). Reuse for the regex / typed-predicate follow-ups and `#1546`.
- Dim via a `data-dimmed` attribute (or a `dimmed` flag on `MLNodeData`) consumed by CSS, *not* by mutating positions or node IDs. Reuses the existing producer/consumer highlight pattern in `MLIRViewReactFlow.tsx`.
- Query state lives in a Jotai atom; URL-hash persistence comes with `#1556`.

PR #1725, opened 2026-07-03
Add regex mode to MLIR op-name filter

### LINKED ISSUE #1716: MLIR filter: regex mode toggle
Extend the op-name filter (#1545) with a regex mode. Substring matching is fine for most cases but power users want anchored / alternation / capture patterns.

- Small toggle inside the filter input
- Invalid regex renders zero matches without crashing
- Persist the mode across the session

aidemsined:
Extends the op-name filter shipped in #1545 with a regex mode toggle.

image

## Behaviour
- Mode toggle sits in the filter input's right slot — `.*` glyph, high-contrast active/inactive states (transparent grey vs. solid yellow).
- Patterns compile with the `i` flag, mirroring the substring path's `toLowerCase()` semantics.
- Invalid regex short-circuits to a zero-match matcher, paints the input with `Intent.DANGER`, and reads `invalid regex` in the counter — so the user can distinguish a broken pattern from a query that just doesn't hit.
- Mode persists in `sessionStorage` under `mlirFilterMode` and is restored on mount.
- Flipping mode resets the prev/next cursor since the match set can change.

## Structure
- Pure `buildFilterMatcher(mode, query)` helper extracted to `src/components/mlir/mlirFilter.ts` — makes the substring / regex / invalid paths unit-testable without wiring up the whole graph view. Also lays groundwork for #1707 (extract `computeFilterMatches` as a pure function).
- `MlirOpFilter` gains `mode` / `onModeChange` / `isRegexInvalid` props; existing `MlirFilterMode` type re-exported so callers don't shift imports.

## Tests
29 tests, split across `tests/mlirFilter.spec.ts` (matcher unit coverage, including the `co*` regression that prompted the review) and `tests/MlirOpFilter.spec.tsx` (mode toggle, placeholder swap, danger-intent styling, counter text for all states, keyboard shortcuts, imperative focus handle).

Closes #1716.

### DIFF
--- a/src/components/mlir/mlirFilter.ts
+++ b/src/components/mlir/mlirFilter.ts
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+export type MlirFilterMode = 'substring' | 'regex';
+
+export interface MlirFilterMatcher {
+    /** True if a source label should be treated as a filter match. */
+    testLabel: (label: string) => boolean;
+    /** True only in regex mode when the query fails to compile. */
+    isRegexInvalid: boolean;
+}
+
+// Empty queries and invalid regexes both surface a matcher that rejects
+// everything so the caller doesn't need branch-per-case handling; the
+// `isRegexInvalid` flag lets the UI distinguish "no query" from "bad regex".
+export function buildFilterMatcher(mode: MlirFilterMode, query: string): MlirFilterMatcher {
+    if (query.length === 0) {
+        return { testLabel: () => false, isRegexInvalid: false };
+    }
+    if (mode === 'regex') {
+        try {
+            const re = new RegExp(query, 'i');
+            return { testLabel: (label) => re.test(label), isRegexInvalid: false };
+        } catch {
+            return { testLabel: () => false, isRegexInvalid: true };
+        }
+    }
+    const needle = query.toLowerCase();
+    return {
+        testLabel: (label) => label.toLowerCase().includes(needle),
+        isRegexInvalid: false,
+    };
+}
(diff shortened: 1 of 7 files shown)


src/scss/components/MlirOpFilter.scss
```diff
@@ -36,8 +36,37 @@
...
     }
 
+    .mlir-op-filter-right-slot {
+        display: inline-flex;
+        align-items: center;
+        gap: 2px;
+    }
+
+    .mlir-op-filter-mode {
+        // Monospace so `.*` reads as a symbol, not a variable-width sentence.
```
dcblundell: Do we need this comment?

src/components/mlir/mlirFilter.ts
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+export type MlirFilterMode = 'substring' | 'regex';
```
dcblundell: Should this be an enum? Would allow us to avoid magic looking strings for this feature.

dcblundell:
Could tighten up the space between the field and "Re-layout".

dcblundell (approved):

src/scss/components/MlirOpFilter.scss
aidemsined: Good call, dropped in `b8dcdee3` — the `text='.*'` in the JSX makes the monospace choice self-evident.

src/components/mlir/mlirFilter.ts
aidemsined: Considered enum first but leaned union to match `ClusterLayoutMode = 'mesh' | 'condensed'` (`src/components/cluster/ClusterRenderer.tsx:87`) and to avoid the reverse-mapping runtime object TS enums emit. Fair point on the magic-string concern though — converting `MlirFilterMode` to a proper enum in the next commit. Backing values stay `'substring'` / `'regex'` so existing `sessionStorage` state migrates without a version bump. Leaving `ClusterLayoutMode` for now to keep this PR's scope tight; happy to open a follow-up if you'd like feature-mode enums as a broader convention.

PR #1726, opened 2026-07-03
Retry MLIR file processing

dcblundell:
Adds a retry per-mlir upload.

Screenshot 2026-07-03 at 11 14 24

Screenshot 2026-07-03 at 11 14 20


src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,95 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
         setSelectedIndex(null);
+        setRetryingIndex(null);
         setIsOpen(false);
     };
 
+    const handleRetry = async (index: number) => {
+        const result = results?.[index];
+        const file = retryFiles?.[index];
+        if (!result || result.status !== ConnectionTestStates.FAILED || !result.persisted || !file || !retryServer) {
+            return;
```
aidemsined, line 69: This early-return silently no-ops when the retry atoms are missing, but the Retry button in `MlirFileList` only gates on `isFailedServerFile && !!onRetry` — it doesn't know whether we actually have a `File` blob + `server` to retry with. So on any persisted-failed row where the retry state has been reset (Copilot's #4 will introduce more of these paths — empty results / catch — and page-reload/atom-cleared scenarios generally), the button renders and clicking it does nothing.

Either thread a `canRetry` flag through to `MlirFileList` and gate visibility on it, or surface a toast here on the miss so the user gets feedback instead of a dead click.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,95 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
+                              message: undefined,
+                              name: null,
+                              graph: null,
+                          }
+                        : entry,
+                ) ?? current,
+        );
+
+        try {
+            const response = await uploadMlirFileToServer([file], retryServer);
```
aidemsined, line 89: Worth confirming with the backend that a retry POST for a file that previously failed is treated as an idempotent replacement (server-side entry keyed by filename or a stable id, overwriting the failed slot) rather than creating a second persisted record. If the original failed upload left any on-disk artifacts under a distinct id, repeated retries could accumulate orphaned entries the UI has no way to reach.

A quick way to check: retry a failed file a few times in a row and inspect what shows up in `GET /uploads` (or wherever the persisted list lives) on the server.

src/components/mlir/MlirJsonFileLoader.tsx
```diff
@@ -126,19 +138,25 @@ const MlirJsonFileLoader = ({ server = null }: MlirJsonFileLoaderProps) => {
                     // Drop the pending spinner rows so they don't linger as
                     // permanently-converting entries behind the View button.
                     setMlirFileResults(null);
+                    setMlirRetryFiles(null);
```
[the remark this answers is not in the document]
dcblundell: Implemented, thanks. We now clear `mlirRetryFilesAtom` and `mlirRetryServerAtom` after a server upload when there are no failed conversion rows, so all-success batches do not retain uploaded File blobs in session memory. I also added a regression assertion in `tests/useMlirRemote.spec.tsx` to verify retry atoms are null after the all-success upload path.

src/components/mlir/MlirFileList.tsx
```diff
@@ -13,49 +13,102 @@ interface MlirFileListProps {
...
             const isPending = result.status === ConnectionTestStates.PROGRESS;
             const isSuccess = result.status === ConnectionTestStates.OK && !!result.graph;
+            const isFailedServerFile = result.status === ConnectionTestStates.FAILED && result.persisted;
+            const shouldShowEyeIcon = !isPending && result.status !== ConnectionTestStates.FAILED;
             const selectable = !!onSelect && isSuccess;
+            const retryAvailable = canRetry ? canRetry(index) : true;
+            const hasRetryAction = isFailedServerFile && !!onRetry && retryAvailable;
             const isSelected = !!onSelect && index === selectedIndex;
+            const isRetrying = retryingIndex === index;
+            const retryInFlight = retryingIndex !== null;
```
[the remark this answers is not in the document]
dcblundell: Inaccurate.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,104 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
+                                  filename: retried.filename,
+                                  name: retried.name,
+                                  status: retried.status,
+                                  message: retried.message ?? retried.detail,
+                                  graph: retried.graph ?? null,
+                                  persisted: true,
+                              }
+                            : entry,
+                    ) ?? current,
+            );
```
aidemsined, line 109: **Cross-batch retry writeback race — positional `setResults` with no batch identity check.**

`handleRetry` writes to `mlirFileResultsAtom` using only `entryIndex === index`. If the user closes the overlay and starts a new upload while a retry is still in flight, the loader replaces the atom with new PROGRESS rows. When the old retry resolves, this `setResults` writes the retried filename/graph into batch B's row at the same index — silent data corruption (a row bearing a filename that never belonged to the new batch). The same race is reachable without close if a new upload happens in another mount.

**Suggested action:** Capture a batch identifier at click time (a counter incremented on each `setMlirFileResults` populate, or the results-array reference itself) and no-op the writeback if the atom's current batch no longer matches. Alternatively pass an `AbortSignal` into `uploadMlirFileToServer` and abort in `handleClose`.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,104 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
+                              name: null,
+                              graph: null,
+                          }
+                        : entry,
+                ) ?? current,
+        );
+
+        try {
+            const response = await uploadMlirFileToServer([file], retryServer);
+            const retried = response.data.results?.[0];
```
aidemsined, line 90: **Retry reuses the full upload pipeline with global progress side effects.**

`uploadMlirFileToServer` unconditionally drives `fileTransferProgressAtom` (see `useMlirRemote.tsx` L68), which opens `FileStatusOverlay` on top of the already-visible results overlay. During a retry the user sees two overlays stacked: the results overlay with an in-row PROGRESS spinner, and the global processing overlay on top rendering the full results list.

**Suggested action:** Add a `{ suppressProgressOverlay?: boolean }` option (or a thin "convert-only" wrapper) so retry can skip the global progress side effect and let the row-level spinner be the only UI signal.

tests/MlirFileResultsOverlay.spec.tsx
aidemsined, line 1: **Retry state-machine tests are missing several high-signal branches.**

Current tests cover the happy path and cross-row lockout. Missing, in order of regression severity:

- **Retry catch path** — `uploadMlirFileToServer.mockRejectedValueOnce(new Error('server 502'))`; assert row reverts to FAILED with the error message, Retry button re-renders, one toast fires.
- **Retry-vs-new-upload interleave** (for the cross-batch race noted on `MlirFileResultsOverlay.tsx` L109) — render overlay with a failed batch, click Retry with a deferred upload, simulate a new-batch publish, then resolve the deferred old retry. Assert batch B's row is *not* mutated by batch A's response.
- **Retry-during-close** — click Retry, click Close, resolve the deferred upload. Assert no corruption of the retained (or replaced) results.
- **Retry empty-results path** — `mockResolvedValueOnce({ data: { results: [] } })`; assert the `throw new Error('Upload failed')` branch settles the row to FAILED.
- **Retry-success clears retry atoms when it was the last failure** — 1 failed file, retry succeeds → assert both atoms are `null` (pairs with the `MlirFileResultsOverlay.tsx` L110 cleanup gap).

The retry catch path and retry-atom cleanup assertion are the highest-value additions for regression coverage.

tests/useMlirRemote.spec.tsx
aidemsined, line 1: **Retry atom assertions are all negative (asserting null); positive assertions are missing.**

Current specs assert `mlirRetryFilesAtom` and `mlirRetryServerAtom` are `null` on all-success, network error, and empty-results paths. Missing:

- **After a mixed-outcome batch** (at least one FAILED result), both atoms should be **populated** with `selectedFiles` and `server`. Without this positive assertion a future reorder that inadvertently clears the atoms on every path would still pass all current tests — silently disabling the whole retry feature.
- The **`uses detail as fallback`** case (FAILED conversion) should assert retry atoms are **not** null on settle.

**Suggested action:** Add positive-populate assertions on the mixed-outcome path; extend `uses detail as fallback` to include the atom check.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,104 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
+                                  name: retried.name,
+                                  status: retried.status,
+                                  message: retried.message ?? retried.detail,
+                                  graph: retried.graph ?? null,
+                                  persisted: true,
+                              }
+                            : entry,
+                    ) ?? current,
+            );
+        } catch (err: unknown) {
```
aidemsined, line 110: **Retry atoms not cleared when this success drains the last failure.**

`MlirJsonFileLoader.tsx` L149-154 clears `mlirRetryFilesAtom` / `mlirRetryServerAtom` when the initial batch has no failures (via `hasRetryableFailures`). The retry success path here never re-evaluates that condition, so if a batch had one failed row and this retry resolves it, the atoms still hold the original `File[]` (potentially tens–hundreds of MB per model file) for the rest of the session. Copilot already flagged this class for the all-success upload path in an earlier round; the same fix is missing on the retry path.

**Suggested action:** After the success `setResults`, compute `results.some(r => r.status === FAILED)` on the fresh array and, if none remain, `setMlirRetryFiles(null); setMlirRetryServer(null)`.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,104 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
 
     // Reset the pending selection on close so it can't carry over to a reopen
     // or the next upload. The results themselves are retained so the overlay
     // can be reopened. Every close path — Close, View, escape/outside click —
     // routes through here.
     const handleClose = () => {
         setSelectedIndex(null);
+        setRetryingIndex(null);
         setIsOpen(false);
     };
```
aidemsined, line 55: **`handleClose` clears local state but doesn't invalidate the in-flight retry.**

Closing the overlay resets `retryingIndex` and hides the UI, but the pending `uploadMlirFileToServer` promise keeps running and its `setResults` writebacks still land on `mlirFileResultsAtom`. This is the enabling condition for the cross-batch race noted at L109, and even in isolation it means the retry side effects mutate state the user believed was "closed away."

**Suggested action:** Capture a `cancelled` flag or an abort signal at click time and skip the writebacks if closed. Ideally, propagate an `AbortSignal` into `uploadMlirFileToServer` (the hook doesn't currently accept one) so the request itself is dropped.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,104 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
+                            ? {
+                                  filename: retried.filename,
+                                  name: retried.name,
+                                  status: retried.status,
+                                  message: retried.message ?? retried.detail,
+                                  graph: retried.graph ?? null,
+                                  persisted: true,
+                              }
+                            : entry,
+                    ) ?? current,
```
aidemsined, line 108: **Duplicate upload-response → `MlirFileResult` mapping.**

This 6-field mapping (`{ filename, name, status, message: X ?? Y, graph, persisted: true }`) also lives in `MlirJsonFileLoader.tsx` L128-135 for the initial batch upload. That mapping is the API-boundary contract between `useMlirRemote.tsx` and the model; duplicating it across upload and retry paths will drift when fields change.

**Suggested action:** Extract `mapUploadResultToMlirFileResult(result: MlirFileUploadResult): MlirFileResult` — colocate with `MlirFileUploadResult` in `useMlirRemote.tsx` or as a small helper in `src/functions/`.

src/store/app.ts
```diff
@@ -61,6 +61,8 @@ export const activeMlirJsonAtom = atom<string | null>(null);
 export const activeMlirDataAtom = atom<GraphBundle | null>(null);
 export const mlirFileResultsAtom = atom<MlirFileResult[] | null>(null);
 export const mlirFileResultsOpenAtom = atom(false);
+export const mlirRetryFilesAtom = atom<File[] | null>(null);
+export const mlirRetryServerAtom = atom<MlirServerConnection | null>(null);
```
aidemsined, line 65: **Consider merging the two retry atoms into one `mlirRetryContextAtom`.**

`mlirRetryFilesAtom` and `mlirRetryServerAtom` are always written and cleared together — there are ~8 paired `set(...)` sites in `MlirJsonFileLoader.tsx` plus the retry-success cleanup path. A single atom `mlirRetryContextAtom: { files: File[]; server: MlirServerConnection } | null` gives atomic writes/clears and removes the desync class where one gets cleared but the other doesn't. Matches the domain model ("retry is one context") better than two independent nullable atoms.

**Suggested action:** Optional refactor — low priority now, higher value if the surface expands.

CONVENTIONS.md
```diff
@@ -230,6 +230,16 @@ If a pixel value, threshold, or duration is used in more than one place, promote
...
 
+### Prefer app-owned classnames over direct Blueprint selector overrides
+
+In component stylesheets, prefer styling classes we own (for example wrapper or element classes) rather than targeting Blueprint internals like `.bp6-menu-item`, `.bp6-button`, etc.
+
+- Preferred: add a project classname in TSX and style that classname.
+- Avoid by default: direct `.bp6-*` overrides in component-local SCSS.
+- Exception: global Blueprint theming/normalisation in `src/scss/_blueprintjs.scss` when there is no stable app-owned hook.
+
+This keeps styles resilient to Blueprint markup changes and avoids cross-component side effects.
```
aidemsined, line 241: **Convention is well-motivated but currently aspirational vs the codebase state.**

The new rule ("prefer app-owned classnames over `.bp6-*` overrides") is applied cleanly in this PR (`MlirFileList.scss` drops `.bp6-menu-item` for `.mlir-file-list-item`). But the current codebase has **~81 `.bp6-*` selectors across ~26 component/partial SCSS files** outside `_blueprintjs.scss`. Without an explicit "apply on touch / don't refactor unrelated files" clause, reviewers won't know whether legacy overrides are violations or grandfathered.

**Suggested action:** Add a one-line "apply on touch" scope clause. Optionally document a small allow-list of cases where `.bp6-*` remains necessary (Blueprint components without stable app-owned hooks, e.g. Popover target markup).

aidemsined (commented):
## PR review — machine-assisted

Automated review generated from a 3-agent hybrid pass (comprehensive linear reviewer + Risk-Perf sidecar + Architecture sidecar). Line-anchored comments below cover definite bugs and important findings. Nit-picks are consolidated in **General observations** to keep the noise down.

## General observations

_Nit-picks and minor cleanup — not merge-blocking._

- **`MlirFileList.canRetry` default** — `retryAvailable = canRetry ? canRetry(index) : true` (`MlirFileList.tsx` L44) means a future caller that supplies `onRetry` without `canRetry` will render Retry on every failed persisted row. Safer to default to `false`, or express the coupling in the type.
- **[perf] Callback churn + `canRetry(index)` per-row** — `onRetry={handleRetry}` and inline `canRetry={(index) => ...}` allocate on every overlay render, and `canRetry(index)` runs once per row inside `MlirFileList`'s `.map`. Negligible today (MlirFileList not memoized). Becomes a footgun if MlirFileList is ever wrapped in `React.memo`.
- **[perf] Retry produces up to 5 successive state writes for a single row status flip** — each fanout re-maps `results`. Practically negligible at current batch sizes; would matter for a future "Retry all failed" bulk action.
- **Duplicated `'Upload failed'` magic string** — `MlirJsonFileLoader.tsx` L138 and `MlirFileResultsOverlay.tsx` L92 both hard-code it. AGENTS.md's naming section calls out promoting user-visible copy to module-scope constants at two-plus occurrences.
- **Stale comment in `MlirFileList.tsx` L102-104** — still says the leading eye appears "in every state"; failed rows now render an empty `.mlir-file-icon` placeholder.
- **Factor the three identical `setResults(current?.map(...))` branches in `handleRetry`** — the `entryIndex === index ? {…} : entry` pattern appears three times. A local `patchResultAtIndex(current, index, patch)` helper would shrink the function.
- **Index-aligned parallel arrays coupling** — `retryFiles?.[index]` assumes `mlirRetryFilesAtom[i]` matches `mlirFileResultsAtom[i]`. Safe today (both seeded from `selectedFiles`), fragile if results are reordered or filtered downstream. Filename-keyed lookup would decouple.
- **`MlirFileList` retry props widen shared-list responsibility** — optional `retryingIndex`/`onRetry`/`canRetry` with defaults keep `FileStatusOverlay` unchanged. Fine for two consumers; consider a `renderTrailingActions` slot or a results-only wrapper if row actions grow.
- **`uploadMlirFileToServer(files: File[])` narrowing is low-risk** — no external consumers, `File[]` is strictly more flexible for retry. Confirmation, no change needed.
- **Document retry atom lifetime in `app.ts`** — a one-line comment that they hold in-memory `File` blobs only while retryable server failures exist would match the loader's `hasRetryableFailures` comment.
- **Run `pnpm lint` / `pnpm test` locally before merge** — CI reports green (13 checks) on head `6e5fae9`; local re-run not executed in this review.

---

_Generated with the `pr-review` skill (composes `ultimate-review`). Confidence varies by finding; the human reviewer decides what to act on._

PR #1728, opened 2026-07-06
Standardise MLIR report storage

dcblundell:
Stores JSON under `/reports/remote/name-of-server/mlir-reports` in the same way we do for other report types.


backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,43 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
...
+
+    if remote_connection is not None and getattr(remote_connection, "host", None):
+        host_scoped_path = (
+            remote_dir
+            / remote_connection.host
+            / current_app.config["MLIR_DIRECTORY_NAME"]
+            / f"{mlir_name}.json"
+        )
+        return str(host_scoped_path)
+
```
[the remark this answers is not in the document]
dcblundell: Implemented. `get_mlir_path` now only returns the host-scoped path immediately when that file exists. If it is missing, it falls through to host-agnostic candidates and then the flat legacy fallback when present. Added regression coverage in `test_get_mlir_path_falls_back_when_host_scoped_file_missing`.

backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,43 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
...
+    host_scoped_candidates = list(
+        remote_dir.glob(
+            f"*/{current_app.config['MLIR_DIRECTORY_NAME']}/{mlir_name}.json"
+        )
+    )
+    if host_scoped_candidates:
+        return str(host_scoped_candidates[0])
+
```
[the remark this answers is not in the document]
dcblundell: Implemented. Host-scoped candidates are now sorted before selection, so fallback resolution is deterministic when multiple hosts contain the same MLIR name. Added regression coverage in `test_get_mlir_path_uses_deterministic_host_candidate_order`.

backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,51 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
...
+            / f"{mlir_name}.json"
+        )
+        if host_scoped_path.exists():
+            return str(host_scoped_path)
+
+    host_scoped_candidates = sorted(
+        remote_dir.glob(
+            f"*/{current_app.config['MLIR_DIRECTORY_NAME']}/{mlir_name}.json"
+        )
     )
```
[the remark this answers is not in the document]
dcblundell: Implemented. `get_mlir_path` no longer interpolates `mlir_name` into a glob pattern. It now iterates host directories and checks literal candidate paths (`<host>/mlir-reports/<mlir_name>.json`), so wildcard metacharacters in `mlir_name` are treated as plain filename characters. Added regression coverage in `test_get_mlir_path_treats_glob_metacharacters_as_literal_name`.

backend/ttnn_visualizer/views.py
```diff
@@ -1787,8 +1787,12 @@ def upload_mlir_server():
...
         )
 
-    data_directory = current_app.config["LOCAL_DATA_DIRECTORY"]
-    target_directory = data_directory / current_app.config["MLIR_DIRECTORY_NAME"]
+    data_directory = current_app.config["REMOTE_DATA_DIRECTORY"]
+    target_directory = (
+        data_directory
+        / mlir_connection.host
+        / current_app.config["MLIR_DIRECTORY_NAME"]
+    )
```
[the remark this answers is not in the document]
dcblundell: Implemented. MLIR upload now sanitizes the host before using it as a storage path segment (`_safe_remote_host_segment`) by normalizing separators and taking a single path component. Persisted results also return the sanitized host. Added regression coverage in `test_upload_endpoint_host_traversal_stays_in_remote_dir`.

backend/ttnn_visualizer/views.py
```diff
@@ -1860,14 +1864,18 @@ def set_active_mlir(instance: Instance):
     # directory and the name is only ever a file stem. Accepting `.json`
     # input from callers is fine: normalise to the stem before lookup.
     safe_name = Path(name.strip()).stem
-    mlir_path = get_mlir_path(safe_name, current_app)
+    mlir_path = get_mlir_path(
+        safe_name,
+        current_app,
+        remote_connection=instance.remote_connection,
+    )
```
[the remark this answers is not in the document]
dcblundell: Implemented. `POST /mlir/active` now accepts optional `{ host }` and resolves directly to `REMOTE_DATA_DIRECTORY/<host>/mlir-reports/<name>.json` when provided, avoiding host-agnostic ambiguity for duplicate names. The frontend now passes host when persisting active MLIR selection, and host metadata is carried in upload results. Added regression coverage in `test_set_active_mlir_uses_host_to_disambiguate_duplicate_names`.

backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,51 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
...
-    return str(mlir_path)
+    if host_scoped_candidates:
+        return str(host_scoped_candidates[0])
+
+    fallback_path = (
+        remote_dir / current_app.config["MLIR_DIRECTORY_NAME"] / f"{mlir_name}.json"
+    )
+    if fallback_path.exists():
```
aidemsined, line 927: **Do — local-layout fallback / migration is missing.**

Pre-#1728 MLIR uploads landed under `LOCAL_DATA_DIRECTORY/mlir-reports/` (see `dev` `utils.py` and the previous `upload_mlir_server`). Post-#1728, `get_mlir_path` only searches under `REMOTE_DATA_DIRECTORY` — this `fallback_path` probes `REMOTE_DATA_DIRECTORY / mlir-reports / <name>.json`, not `LOCAL_DATA_DIRECTORY / mlir-reports /`.

On an upgrade, existing MLIR files silently become unreachable. The bite happens the moment anything triggers path recomputation via `_resolve_report_path` (e.g. the first `PUT /api/instance`) — the correct stored `mlir_path` gets overwritten with a non-existent one because none of the four resolution branches match.

Worth flagging: the regression test `test_get_mlir_path_falls_back_when_host_scoped_file_missing` seeds the "flat legacy" file under `REMOTE_DATA_DIRECTORY/mlir-reports/` too, so CI is happy while the migration path is broken.

Fix options — either:
1. Add a `LOCAL_DATA_DIRECTORY/mlir-reports/<name>.json` lookup branch here, with a companion test seeding under the local root.
2. Ship a one-time startup migration that moves legacy JSONs into the host-scoped layout (there are existing app-data migration helpers to model on).

Without one of these, upgraded deployments lose their MLIR history.

backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,51 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
     return str(npe_path)
 
 
-def get_mlir_path(mlir_name, current_app, **_kwargs):
+def get_mlir_path(mlir_name, current_app, remote_connection=None, **_kwargs):
```
aidemsined, line 886: **Discuss — function shape: reuse `PathResolver`, and consider splitting read from write.**

Two related design points on `get_mlir_path`:

**1. Duplicate host-scoping.** `PathResolver.get_base_report_path` already encodes `remote_dir / host / <directory-name>` for profiler and performance. The PR's own claim — _"in the same way we do for other report types"_ — reads more naturally when the mechanism is actually shared. Extending `PathResolver` with `report_type="mlir"` (or extracting a `_remote_host_report_path(host, dir_name, leaf)` helper) would keep MLIR / profiler / performance layout rules in one place; today the SFTP-sync layout, the profiler getters, and MLIR each spell out the same rule.

**2. Read vs write are mixed.** Branches 1–3 (host-scoped-exists, glob, flat-legacy-exists) resolve an existing file. Branch 4 (`if host_scoped_path is not None: return str(host_scoped_path)` at L927–928) returns an *intended write target* — a non-existent path. Profiler getters are construct-only; callers supply explicit paths on write. Cleaner:
- `resolve_mlir_read_path(name, host_hint?) → str | None` used by `/mlir/active`.
- Upload handler constructs its own write path inline (it already does at L1791–1794) and passes `mlir_path=` to `update_instance` explicitly.

This would remove the "host_scoped_path is not None → return non-existent path" footgun that Copilot's original resolved thread caught once already.

backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,51 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
...
+            / remote_connection.host
+            / current_app.config["MLIR_DIRECTORY_NAME"]
+            / f"{mlir_name}.json"
+        )
+        if host_scoped_path.exists():
+            return str(host_scoped_path)
+
+    host_scoped_candidates = sorted(
+        remote_dir.glob(
+            f"*/{current_app.config['MLIR_DIRECTORY_NAME']}/{mlir_name}.json"
```
aidemsined: **Discuss — [perf] this glob fires on every `PUT /api/instance`, not just on `set_active_mlir`.**

Because `update_existing_instance` now backfills `effective_remote_connection` from the stored connection (good change on its own) and calls `_resolve_report_path` for each `_sentinel` slot, every path recomputation on any PUT walks into `get_mlir_path`. When the host-scoped file is absent, we `sorted(remote_dir.glob("*/mlir-reports/<name>.json"))` across every host subdirectory.

_When it matters:_ shared / long-lived deployments with many synced remote hosts and frequent instance updates. On a dev laptop with one host it's free; on a multi-host machine each PUT is O(hosts) filesystem stat + a sort of the results. The sort itself is cheap; the fs I/O is what scales.

If `/mlir/active` ends up passing an explicit host (see the resolution-dependency thread below), the glob branch becomes essentially dead for the common case and this concern goes away.

backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,51 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
...
     if not mlir_name:
         return None
-    local_dir = Path(current_app.config["LOCAL_DATA_DIRECTORY"])
-    mlir_path = (
-        local_dir / current_app.config["MLIR_DIRECTORY_NAME"] / f"{mlir_name}.json"
+    remote_dir = Path(current_app.config["REMOTE_DATA_DIRECTORY"])
+    host_scoped_path: Path | None = None
+
+    if remote_connection is not None and getattr(remote_connection, "host", None):
+        host_scoped_path = (
```
aidemsined, line 904: **Discuss — read side also uses `remote_connection.host` as a raw path segment.**

Copilot's open thread on `views.py:1795` correctly flags the *write* side (user-supplied `mlir_connection.host` joined directly). The *read* side has the same shape here: `remote_connection.host` composed into `host_scoped_path` with no validation, and `RemoteConnection` itself has no host validator.

The attack surface is smaller than the upload side because `remote_connection` in this call typically comes from the DB — but it also means a poisoned host stored once (via any codepath that persists a `RemoteConnection`) is replayed on every path recomputation forever.

Cleanest fix is to sanitise at the `RemoteConnection` model boundary (a `Path(host).name` normaliser or a validator) so both this call site and the upload write site get it for free.

backend/ttnn_visualizer/utils.py
aidemsined, line 927: **Blocker: local-layout / pre-#1728 migration path is missing.**

The legacy fallback here only checks `REMOTE_DATA_DIRECTORY / mlir-reports / <name>.json`, but every pre-#1728 MLIR upload was written to `LOCAL_DATA_DIRECTORY / mlir-reports / <name>.json` (the previous `get_mlir_path` + `upload_mlir_server` both used the local root). After this PR merges, resolution never looks there, so existing files become unreachable on the first `PUT /api/instance` — `_resolve_report_path` will clobber the correct stored `mlir_path` with a non-existent one.

Heads-up: `test_get_mlir_path_falls_back_when_host_scoped_file_missing` seeds the flat fallback under `REMOTE_DATA_DIRECTORY` too, so CI is green on the broken behaviour.

Suggested: add a `LOCAL_DATA_DIRECTORY / MLIR_DIRECTORY_NAME / <name>.json` branch to the chain, or run a one-time startup migration to move legacy JSONs into the host-scoped layout. Companion test with the local seed.

backend/ttnn_visualizer/utils.py
aidemsined, line 904: Two design points about the shape of this function (linked but separable):

1. **`PathResolver` reuse.** Profiler / performance getters delegate host scoping to `PathResolver.get_base_report_path` (`utils.py` ~L598–659), which encodes the exact `remote_dir / host / <report-type-dir>` rule that this function now reimplements. Wiring MLIR into `PathResolver` (or extracting a `_remote_host_report_path(host, dir_name, leaf)` helper) would keep the PR's "same way we do for other report types" promise honest at the code level, not just the storage layout.
2. **Split read vs write responsibilities.** Branches 1–3 answer "find the existing file"; the tail (L927–930) answers "return the intended write target when it doesn't exist yet". Copilot open thread #1 already caught the same footgun once. Cleaner: `resolve_mlir_read_path(name, host_hint?) -> str | None` for `/mlir/active`, and let `upload_mlir_server` construct the write path inline (it already does today) and pass `mlir_path=` explicitly to `update_instance`.

backend/ttnn_visualizer/utils.py
aidemsined: **[perf]** This glob runs on every `PUT /api/instance` (not only on `set_active_mlir`), because `_resolve_report_path` calls `get_mlir_path` whenever `mlir_name` is set and no explicit `mlir_path` is passed — which is the common case for `update_current_instance`.

**When it matters:** deployments with a large number of synced remote host directories under `REMOTE_DATA_DIRECTORY`. Every host subdir is `stat`'d, then results are collected and sorted. Low priority on single-host dev, worth flagging for shared / long-lived deployments.

A cheap mitigation without redesigning the API: short-circuit to the host-agnostic scan only when `remote_connection` is absent (the host-scoped exact-path check already covers the with-connection case), and cache the list per-request if there are multiple report-path recomputations in the same handler.

backend/ttnn_visualizer/utils.py
```diff
@@ -883,25 +883,51 @@ def get_npe_path(npe_name, current_app, remote_connection=None):
...
+            remote_dir
+            / remote_connection.host
+            / current_app.config["MLIR_DIRECTORY_NAME"]
```
aidemsined, line 907: Copilot's open thread on `views.py:1795` correctly flags path traversal via `mlir_connection.host` on the **write** side. The **read** side has the same shape: `remote_connection.host` from the stored `RemoteConnection` (loaded from the DB) is used raw as a path segment here. `RemoteConnection` has no host validator today, so if a poisoned value ever makes it into the DB (host renamed by a script, imported from another instance table, etc.), it replays on every path recomputation.

Best fix is at the boundary: validate/sanitise `host` on `RemoteConnection` construction (a field validator that rejects path separators / `..`), which fixes both the write and read sides in one place — and covers future report types that use the same field.

backend/ttnn_visualizer/instances.py
```diff
@@ -170,6 +170,16 @@ def update_existing_instance(
     if clear_remote:
         clear_remote_data(instance_data)
 
+    effective_remote_connection = remote_connection
+    if (
+        effective_remote_connection is None
+        and instance_data.remote_connection is not None
+    ):
+        effective_remote_connection = RemoteConnection.model_validate(
```
aidemsined, line 178: Two things about this fallback:

1. **This is a cross-cutting bugfix, not MLIR-only plumbing.** `effective_remote_connection` is threaded through *all four* report types below (profiler / performance / npe / mlir). On `dev`, `PUT /api/instance` always passes `remote_connection=None`, so path recomputation for REMOTE profiler / performance rows likely also resolved against `LOCAL_DATA_DIRECTORY` in some flows — masked when callers passed explicit `*_path`. Worth calling out in the PR body / release notes so the wider impact isn't hidden inside "MLIR storage standardisation". The two new regression tests in `test_mlir.py` cover MLIR only.
2. **`create_new_instance` (`instances.py:252`) does not apply the same fallback.** If a create path can ever be called with `remote_connection=None` while a stored instance already has one, it will exhibit the pre-fix behaviour. Worth a follow-up.
3. **Unguarded `model_validate(..., strict=False)`.** Any partial / legacy JSON blob under `instance_data.remote_connection` will raise `ValidationError` from here, propagating as a 500 on `PUT /api/instance`. Wrap in `try / except ValidationError`, log-and-skip (keep `effective_remote_connection = None`), so a schema drift doesn't wedge instance updates.

backend/ttnn_visualizer/views.py
aidemsined: After `c665ca0d` (correctly) removed the upload-time `remote_connection` mutation, `/mlir/active` resolution now hinges on `instance.remote_connection.host` matching the MLIR upload host, or a host-agnostic glob winning the sort. That's a subtle contract that's easy to un-learn in six months.

A one-paragraph docstring or inline comment on `set_active_mlir` explaining the resolution chain would carry the design decision. Also worth pairing with the frontend fix (see the `MlirFileResultsOverlay.tsx` comment) so the payload carries `host` alongside `name` — the client already retains it via `mlirRetryServerAtom`. This is essentially the same axis as Copilot open thread #5 but framed as a documentation + wire-up follow-up rather than a security concern.

CONVENTIONS.md
```diff
@@ -230,6 +230,16 @@ If a pixel value, threshold, or duration is used in more than one place, promote
 
 Don't mix the two: a new component stylesheet doesn't need an underscore, and a new shared partial does.
 
+### Prefer app-owned classnames over direct Blueprint selector overrides
```
aidemsined, line 233: Nit on convention scope: as written this reads repo-wide and immediate. A quick grep turns up ~93 `.bp6-*` selector occurrences across ~27 files under `src/scss/components/`, so at merge time the codebase is broadly out of compliance.

A one-line qualifier ("apply to new files and on-touch", or "existing violations are grandfathered pending migration") makes the rule enforceable in review without either being ignored on day one or requiring a large upfront cleanup PR. The current MLIR SCSS migration in this diff is a good example of what on-touch looks like.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -24,24 +31,104 @@ import 'styles/components/MlirFileResultsOverlay.scss';
...
     const handleClose = () => {
         setSelectedIndex(null);
+        setRetryingIndex(null);
         setIsOpen(false);
     };
 
+    const handleRetry = async (index: number) => {
+        const result = results?.[index];
+        const file = retryFiles?.[index];
+        if (!result || result.status !== ConnectionTestStates.FAILED || !result.persisted) {
```
aidemsined, line 59: Companion for the host-scoped storage: `setActiveMlir` currently sends only `{ name }`, so activation relies on server-side glob heuristics once the same name exists under multiple hosts. `mlirRetryServerAtom` (and the initial upload path) already know which server produced this result — passing that host through would make activation deterministic and align with Copilot open thread #5 without changing the retry-atom design (which is #1726 carry-over).

Minimal wire-up: extend `setActiveMlir(name, host?)` in `useMlirRemote.tsx` to include an optional host, and forward it from `handleView` here.

backend/ttnn_visualizer/views.py
```diff
@@ -2048,8 +2056,8 @@ def update_current_instance(instance: Instance):
             # NPE is always local right now
             "npe_location": ReportLocation.LOCAL.value,
             "mlir_name": active_report.get("mlir_name"),
-            # MLIR is always local right now
-            "mlir_location": ReportLocation.LOCAL.value,
+            # MLIR is always remote right now
+            "mlir_location": ReportLocation.REMOTE.value,
```
aidemsined, line 2081: `mlir_location=ReportLocation.REMOTE.value` is set unconditionally on every `PUT /api/instance` that reaches this branch — even when the payload is a profiler-only update and `mlir_name` came along for the ride.

Combined with the missing local fallback in `get_mlir_path` (see the utils.py comment), this means any legacy instance with `mlir_name` set but `mlir_location: LOCAL` gets silently flipped to REMOTE on the next unrelated PUT, then fails to resolve.

Suggested: only set `mlir_location` when the payload actually carries a new MLIR name / activation, and leave stored values alone otherwise. `set_active_mlir` remains the explicit REMOTE-setting path.

aidemsined (commented):
## General observations

Nits and follow-ups that aren't merge-blocking, consolidated here to keep noise off the diff:

- **Dead-ish parameter:** `get_mlir_path(..., **_kwargs)` — worth confirming whether any caller still exercises the swallow, otherwise drop.
- **Symlink safety:** `Path.exists()` follows symlinks, so a poisoned symlink under `REMOTE_DATA_DIRECTORY/<host>/mlir-reports/` could point outside the tree. `Path.resolve().is_relative_to(remote_dir)` guards are cheap insurance in the same family as the open host / mlir_name sanitisation threads.
- **Upload response could pin the path.** The upload endpoint already knows `target_directory / f"{mlir_name}.json"`. Returning `{ name, path, host }` per result would let `/mlir/active` accept an optional pinned path (mirroring the profiler upload's explicit `profiler_path=`) and shrink `get_mlir_path` to instance-recompute only.
- **Convention cross-ref in `CONVENTIONS.md`.** Document the canonical remote layout — `REMOTE_DATA_DIRECTORY/<host>/{profiler|performance|mlir}-reports/…` — next to the path getters so a future report type doesn't reintroduce a third bespoke resolver.
- **MLIR host source is currently borrowed from the profiler `RemoteConnection`.** MLIR's own connection model is `MlirServerConnection`; a dedicated stored field (or reusing `selectedMlirServerAtom` server-side) would decouple MLIR from profiler remote state and remove the need for the host-agnostic glob fallback.
- **Legacy fallback timeline.** The comment on `get_mlir_path` says "Keep a flat-directory fallback for older instances" with no timeline — worth deciding whether the branches are permanent compatibility surface or a one-time move-and-remove.
- **Housekeeping:** run `pnpm flask:lint` / `pnpm flask:mypy` / `pnpm test` pre-merge (my review is readonly). Also, `.cursor/rules/ttnn-architecture.mdc` still describes MLIR as local-only under `LOCAL_DATA_DIRECTORY` — stale once this merges.
- **Sanity check on the fetched HEAD (`6d971402`):** the `mlir_name` glob-escape fix from Copilot open thread #3 doesn't appear to be present in `get_mlir_path` even though the reply on that thread says "Implemented." Possibly an un-pushed local commit — worth double-checking before the next push.

dcblundell:
Thread triage update for unresolved review comments:

- Addressed now in code and should be resolved:
  - Host-aware activation payload from frontend (`setActiveMlir(name, host)` in `MlirFileResultsOverlay` + hook/view plumbing).
  - Outdated docs/wire-up note on `/mlir/active`.
  - Outdated perf duplicate threads on `get_mlir_path` host scan.
  - Duplicate design-shape and duplicate blocker threads where the same point is already tracked by one canonical unresolved thread.

- Remaining substantive open items we still need to implement (kept unresolved):
  - Legacy local MLIR fallback/migration path (`LOCAL_DATA_DIRECTORY/mlir-reports/...`) for pre-#1728 data.
  - Centralised `RemoteConnection.host` sanitisation/validation at model boundary.

- Kept open for follow-up discussion/scope decisions:
  - Broader `instances.py` cross-cutting notes (create-path parity, validation guard, release-note scope).
  - `CONVENTIONS.md` wording nit.

I’ll resolve only duplicate/outdated/already-addressed threads and leave actionable implementation threads open.

backend/ttnn_visualizer/utils.py
dcblundell: Intentional product decision: we are not preserving pre-#1728 local MLIR artefacts in this PR. Data migration/local fallback is out of scope, and we accept potential loss of old local MLIR data as part of this storage standardisation.

backend/ttnn_visualizer/utils.py
dcblundell: Addressed in this PR by centralising host sanitisation at the model boundary. `RemoteConnection.host` and `MlirServerConnection.host` now both use a shared `sanitise_remote_host_segment` validator, so both read and write paths consume the same normalised host value. Added regression tests in `test_mlir.py` for traversal-like host input and invalid dot-segment rejection.

backend/ttnn_visualizer/utils.py
dcblundell: Addressed with the same boundary fix: host validation/sanitisation now lives centrally in the model layer (`RemoteConnection.host` + `MlirServerConnection.host` via shared `sanitise_remote_host_segment`), so this no longer relies on ad-hoc call-site handling.

backend/ttnn_visualizer/utils.py
dcblundell: Good suggestion. We’re keeping this PR scoped, so we are not doing the `PathResolver`/read-vs-write refactor here. We’ll handle that as a follow-up cleanup to keep this change set focused on storage/path correctness.

CONVENTIONS.md
dcblundell: Thanks — agreed on the wording scope. Updated `CONVENTIONS.md` to make this explicitly forward-looking: “Avoid by default in new or touched component-local SCSS: direct `.bp6-*` overrides.”

backend/ttnn_visualizer/views.py
dcblundell: Clarification: this is not unconditional in practice. `active_report` fields are only updated when the corresponding name is provided, so `mlir_location` does not mutate persisted MLIR state by itself when `mlir_name` is absent.

backend/ttnn_visualizer/instances.py
dcblundell: Acknowledged. We are skipping the optional `ValidationError` hardening in this PR to keep scope tight, so I’m leaving this thread open for follow-up (create-path parity / defensive validation handling).

PR #1729, opened 2026-07-06
Add abort processing to MLIR retry

dcblundell:
Lets the user abort processing requests.

### DIFF
--- a/src/components/mlir/MlirFileResultsOverlay.tsx
+++ b/src/components/mlir/MlirFileResultsOverlay.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 import { useAtom, useAtomValue, useSetAtom } from 'jotai';
 import { useLocation, useNavigate } from 'react-router-dom';
 import { Button, ButtonVariant, Classes, Icon, Intent, Tooltip } from '@blueprintjs/core';
@@ -41,22 +41,54 @@ const MlirFileResultsOverlay = () => {
     const navigate = useNavigate();
     const location = useLocation();
     const [selectedIndex, setSelectedIndex] = useState<number | null>(null);
-    const [retryingIndex, setRetryingIndex] = useState<number | null>(null);
+    const [retryingIndices, setRetryingIndices] = useState<Set<number>>(new Set<number>());
+    const retrySessionRef = useRef(0);
+    const retryAbortControllersRef = useRef<Map<number, AbortController>>(new Map<number, AbortController>());
 
     // Reset the pending selection on close so it can't carry over to a reopen
     // or the next upload. The results themselves are retained so the overlay
     // can be reopened. Every close path — Close, View, escape/outside click —
     // routes through here.
     const handleClose = () => {
+        retrySessionRef.current += 1;
+        retryAbortControllersRef.current.forEach((controller) => controller.abort());
+        retryAbortControllersRef.current.clear();
+        if (retryingIndices.size > 0) {
+            const cancelledRetryCount = retryingIndices.size;
+            setResults(
+                (current) =>
+                    current?.map((entry, entryIndex) =>
+                        retryingIndices.has(entryIndex) && entry.status === ConnectionTestStates.PROGRESS
+                            ? {
+                                  ...entry,
+                                  status: ConnectionTestStates.FAILED,
+                                  message: entry.message ?? 'Retry cancelled',
+                                  name: null,
+                                  graph: null,
+                              }
+                            : entry,
+                    ) ?? current,
+            );
+            createToastNotification(
+                'MLIR',
+                `Aborted ${cancelledRetryCount}x MLIR conversion${cancelledRetryCount === 1 ? '' : 's'}`,
+                ToastType.WARNING,
+            );
+        }
         setSelectedIndex(null);
-        setRetryingIndex(null);
+        setRetryingIndices(new Set<number>());
         setIsOpen(false);
     };
 
     const handleRetry = async (index: number) => {
         const result = results?.[index];
         const file = retryFiles?.[index];
-        if (!result || result.status !== ConnectionTestStates.FAILED || !result.persisted) {
+        if (
+            !result ||
+            result.status !== ConnectionTestStates.FAILED ||
+            !result.persisted ||
+            retryingIndices.has(index)
+        ) {
             return;
         }
 
@@ -69,7 +101,14 @@ const MlirFileResultsOverlay = () => {
             return;
         }
 
-        setRetryingIndex(index);
+        const retrySession = retrySessionRef.current;
+        const abortController = new AbortController();
+        retryAbortControllersRef.current.set(index, abortController);
+        setRetryingIndices((current) => {
+            const next = new Set(current);
+            next.add(index);
+            return next;
+        });
         setResults(
             (current) =>
                 current?.map((entry, entryIndex) =>
@@ -77,7 +116,6 @@ const MlirFileResultsOverlay = () => {
                         ? {
                               ...entry,
                               status: ConnectionTestStates.PROGRESS,
-                              message: undefined,
                               name: null,
                               graph: null,
                           }
@@ -86,10 +124,16 @@ const MlirFileResultsOverlay = () => {
         );
 
         try {
-            const response = await uploadMlirFileToServer([file], retryServer);
+            const response = await uploadMlirFileToServer([file], retryServer, {
+                signal: abortController.signal,
+            });
             const retried = response.data.results?.[0];
             if (!retried) {
                 throw new Error('Upload failed');
+            }
+
+            if (retrySessionRef.current !== retrySession) {
+                return;
             }
 
             setResults(
@@ -109,6 +153,10 @@ const MlirFileResultsOverlay = () => {
                     ) ?? current,
             );
         } catch (err: unknown) {
+            if (retrySessionRef.current !== retrySession) {
+                return;
+            }
+
             const message = getResponseError(err, 'Unable to retry MLIR conversion');
             setResults(
                 (current) =>
@@ -126,7 +174,15 @@ const MlirFileResultsOverlay = () => {
             );
             createToastNotification('MLIR', message, ToastType.ERROR);
         } finally {
-            setRetryingIndex(null);
+            retryAbortControllersRef.current.delete(index);
+
+            if (retrySessionRef.current === retrySession) {
+                setRetryingIndices((current) => {
+                    const next = new Set(current);
+                    next.delete(index);
+                    return next;
+                });
+            }
         }
     };
 
@@ -189,7 +245,7 @@ const MlirFileResultsOverlay = () => {
                 <MlirFileList
                     results={results ?? []}
                     selectedIndex={selectedIndex}
-                    retryingIndex={retryingIndex}
+                    retryingIndices={retryingIndices}
                     // Clicking the already-selected file deselects it.
                     onSelect={(index) => setSelectedIndex((current) => (current === index ? null : index))}
                     onRetry={handleRetry}
--- a/src/components/mlir/MlirFileResultsOverlay.tsx
+++ b/src/components/mlir/MlirFileResultsOverlay.tsx
@@ -2,9 +2,10 @@
 //
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
-import { useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useAtom, useAtomValue, useSetAtom } from 'jotai';
 import { useLocation, useNavigate } from 'react-router-dom';
+import axios from 'axios';
 import { Button, ButtonVariant, Classes, Icon, Intent, Tooltip } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import Overlay from '../Overlay';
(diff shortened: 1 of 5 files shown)


aidemsined (commented):
## General observations

Three Copilot review threads on this PR are marked resolved+outdated, but that state was triggered by commit `71f7c42` (which only removes a stale eye-icon comment in `MlirFileList.tsx`) — none of the flagged concerns are actually addressed in code. Independent 3-agent review reached the same conclusions, so surfacing them here as +1s:

- **Concur with Copilot #2 (`useMlirRemote.tsx:90`, `fileTransferProgressAtom` race across parallel retries).** The `finally` in `uploadMlirFileToServer` unconditionally resets `fileTransferProgressAtom`, so the first retry to finish clears the global `FileStatusOverlay` while other retries are still in flight. Suggested minimal fix that preserves this PR's parallel-retry UX: add `skipTransferProgress?: boolean` to `MlirUploadOptions`, set it from the retry path in `MlirFileResultsOverlay`, and guard the `setTransferProgress` / `resetTransferProgress` calls in the hook on it. Ref-counting would also work but is heavier.
- **Concur with Copilot #3 (`MlirFileResultsOverlay.tsx:93`, rapid double-click same-row race).** The `retryingIndices.has(index)` re-entry guard reads render-closure state, so two clicks in the same tick both pass, and the second `retryAbortControllersRef.current.set(index, ...)` overwrites the first controller — the first upload becomes uncancelable. Use `retryAbortControllersRef.current.has(index)` (synchronous, updated in the same tick) as the guard.
- **Concur with Copilot #1 (`MlirFileResultsOverlay.tsx:79`, close-immediately-after-retry).** `handleClose` decides "should we mark this row as cancelled?" from `retryingIndices` state, but the authoritative record is the ref map. On a fast Retry → Close sequence the state Set can still be empty when `handleClose` runs, so no row is marked FAILED and no toast is shown. Deriving both the count and the target indices from `retryAbortControllersRef.current` keeps close-time behaviour aligned with the abort. (Line-anchored comments below address related but distinct races in the same family.)

### Minor / nice-to-have

- `retrySessionRef.current += 1` in `handleClose` is a nice bailout for pre-response writes, but doesn't help pre-request writes — a comment noting that `retryAbortControllersRef` is the ref that owns pre-response cancellation would prevent future refactors from removing one and thinking the other suffices.
- Consider hoisting `new Set<number>()` into a module-level `EMPTY_RETRYING_INDICES` constant used both in `useState`'s initializer and in `handleClose`'s `setRetryingIndices(new Set<number>())` — trivial allocation win and makes the "empty state" grep-able.
- Retry-cancel toast copy says `Aborted Nx MLIR conversion(s)`, while `entry.message` writeback uses `'Retry cancelled'`. Minor UX inconsistency — worth aligning to one phrasing.
- Product-side follow-up (not this PR): now that retries can run in parallel, a per-row Cancel affordance becomes viable and would be more discoverable than "close the overlay to abort all".

---

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -109,6 +156,10 @@ const MlirFileResultsOverlay = () => {
                     ) ?? current,
             );
         } catch (err: unknown) {
+            if (retrySessionRef.current !== retrySession) {
```
aidemsined, line 176: The catch treats every error identically via `getResponseError`, relying on the `retrySessionRef` mismatch check to bail out on user-triggered aborts. That works for close-time cancellation (session was bumped) but not for other legitimate abort paths — e.g. a future per-row Cancel button that aborts a single controller without bumping the session, or a `useEffect` unmount cleanup (see the separate comment on line 46). In those cases axios rejects with a cancel error, `retrySessionRef` still matches, and we fall through to `setResults(... FAILED, message: 'Unable to retry MLIR conversion')` plus a red toast for what the user just cancelled.

Add an early `if (axios.isCancel(err)) return;` before the session-mismatch check to make the abort → skip-writeback contract explicit at each cancellation entry point rather than pinning the whole behaviour on `retrySessionRef`.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -126,7 +177,15 @@ const MlirFileResultsOverlay = () => {
             );
             createToastNotification('MLIR', message, ToastType.ERROR);
         } finally {
-            setRetryingIndex(null);
+            retryAbortControllersRef.current.delete(index);
```
aidemsined: `retryAbortControllersRef.current.delete(index)` runs unconditionally, even for a superseded retry session. Timeline that trips this:

1. User clicks Retry(row 0) — controller_A stored under key `0`, `retrySession = 1`.
2. User closes the overlay — `handleClose` bumps `retrySessionRef.current` to `2`, aborts controller_A, `.clear()`s the map.
3. User reopens the overlay and clicks Retry(row 0) again — controller_B stored under key `0`, `retrySession = 2` (the current one).
4. controller_A's awaited promise finally rejects and enters *this* `finally` (still holding its captured `index = 0`). The `retrySessionRef.current === retrySession` guard below only protects the `setRetryingIndices` call — the `.delete(index)` above the guard runs, evicting **controller_B** from the map.

Controller_B is now orphaned: a subsequent `handleClose` (or a future per-row Cancel) can no longer abort it via the map. Move the delete inside the same session-guard block, or (safer against index collisions) capture `abortController` locally and only `.delete(index)` if `retryAbortControllersRef.current.get(index) === abortController`.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -41,22 +41,56 @@ const MlirFileResultsOverlay = () => {
     const navigate = useNavigate();
     const location = useLocation();
     const [selectedIndex, setSelectedIndex] = useState<number | null>(null);
-    const [retryingIndex, setRetryingIndex] = useState<number | null>(null);
+    const [retryingIndices, setRetryingIndices] = useState<Set<number>>(new Set<number>());
+    const retrySessionRef = useRef(0);
+    const retryAbortControllersRef = useRef<Map<number, AbortController>>(new Map<number, AbortController>());
```
aidemsined, line 47: `handleClose` is the only place that aborts these controllers, and it's only invoked from user-driven close paths (Close/View/escape/outside-click). If the overlay unmounts through a route change or a parent-driven state flip while retries are in flight, the abort controllers here are never fired — the in-flight uploads finish, their `setResults` writebacks fire against an unmounted tree (React logs the warning), and the `retrySessionRef` bailout doesn't help because the ref lives with the unmounted component.

Suggest adding:

```tsx
useEffect(() => () => {
    retryAbortControllersRef.current.forEach((c) => c.abort());
    retryAbortControllersRef.current.clear();
}, []);
```

so unmount aborts everything the way `handleClose` does. This also complements the `axios.isCancel` guard suggestion on line 159 — unmount aborts should be no-ops on the writeback path.

src/hooks/useMlirRemote.tsx
```diff
@@ -29,6 +29,11 @@ export interface MlirUploadResponse {
     results: MlirFileUploadResult[];
 }
 
+interface MlirUploadOptions {
```
aidemsined: Two small hygiene items on this new interface:

1. Not exported, unlike the sibling `MlirUploadResponse` two lines above. Consumers (tests, future callers wanting to type an options object) can't `import { MlirUploadOptions }` from this module. Suggest `export interface MlirUploadOptions`.
2. Neither field is documented, and both encode non-trivial cross-file contracts:
   - `signal` — aborting propagates as an axios cancel; the caller is expected to skip writebacks (see `MlirFileResultsOverlay.handleRetry`).
   - `suppressProgressOverlay` — the retry path opts out of the global `FileStatusOverlay`; new call sites will need to know that this is a per-request opt-out (not concurrency-aware ref counting), so parallel non-retry uploads should not use it.

A JSDoc line on each field would make the contract legible at the definition site without needing to grep call sites.

tests/MlirFileResultsOverlay.spec.tsx
```diff
@@ -353,4 +405,126 @@ describe('MlirFileResultsOverlay', () => {
             expect(retryButtons[1]).toBeEnabled();
         });
     });
+
```
aidemsined: Good coverage on the new happy paths (parallel retries, close-time cancel, signal passing, double-click, `suppressProgressOverlay`). Two gaps worth filling while this test file is warm:

- **Stale-writeback after reopen** (guards the `finally`-delete race flagged at `MlirFileResultsOverlay.tsx:180`): start a retry for row 0 → close overlay → reopen → start a second retry for row 0 → resolve the *first* (old) upload → assert (a) the row still reflects the second retry's PROGRESS state, not the first's response, and (b) `retryAbortControllersRef` still has the second retry's controller (i.e., the resolved-old-promise's `finally` did not evict it).
- **Distinct AbortSignals per row.** The parallel-retry test asserts `signal: expect.any(AbortSignal)` for both calls. Capturing `signal` from each `uploadMlirFileToServer` call and asserting `signalA !== signalB` locks in that controllers are per-row (a shared-controller regression would still pass `expect.any(AbortSignal)`).

Optional: assert the exact toast copy `'Aborted 2x MLIR conversions'` / `'Aborted 1x MLIR conversion'` so future pluralization refactors don't regress silently.

aidemsined (commented):
## Correction to prior review

My previous review (posted moments ago) was generated against `71f7c422` and predates commit `c0f1fb01 "Code review fixes"`. That commit actually addresses **all three Copilot threads** I cited as unaddressed — the prior review body is stale. Please treat this review as superseding it.

What `c0f1fb01` fixes (removed from my open items):

- ✅ **Copilot #1** (close-cancel-derives-from-state) — `handleClose` now builds `cancelledRetryIndices` from `retryAbortControllersRef.current.keys()`.
- ✅ **Copilot #2** (`fileTransferProgressAtom` race) — new `suppressProgressOverlay?: boolean` on `MlirUploadOptions`, plumbed from the retry path, gates both the pre-upload atom set and the `finally` reset.
- ✅ **Copilot #3** (double-click same-row race) — re-entry guard now uses `retryAbortControllersRef.current.has(index)` (synchronous) instead of `retryingIndices` state, and there's a dedicated `ignores duplicate rapid retry clicks for the same row` test.
- ✅ The `entry.message ?? 'Retry cancelled'` writeback is now unconditional `'Retry cancelled'`; toast copy and row-message copy are aligned.
- ✅ New `does not mutate global progress state when suppressProgressOverlay is enabled` test covers the retry → progress-atom opt-out.

## Still outstanding

Five line-anchored comments below cover concerns `c0f1fb01` didn't touch:

- `MlirFileResultsOverlay.tsx:180` — `finally`-block `.delete(index)` still runs unconditionally, so a `retry → close → reopen → retry-same-row` sequence can have the *old* retry's tail evict the *new* retry's controller from the ref map.
- `MlirFileResultsOverlay.tsx:159` — the catch block relies solely on `retrySessionRef` for abort bailout; adding an explicit `axios.isCancel(err)` early-return future-proofs other abort entry points (per-row Cancel, unmount cleanup).
- `MlirFileResultsOverlay.tsx:46` — no `useEffect` unmount cleanup, so route-change unmount while retries are in flight leaks controllers and fires `setResults` on an unmounted tree.
- `useMlirRemote.tsx:32` — `MlirUploadOptions` isn't exported (unlike sibling `MlirUploadResponse`) and neither field is documented; both encode subtle cross-file contracts worth a one-line JSDoc.
- `tests/MlirFileResultsOverlay.spec.tsx:408` — two coverage gaps worth filling: stale-writeback-after-reopen (guards the finally-delete race) and distinct-`AbortSignal`-per-row (`expect.any(AbortSignal)` would still pass a shared-controller regression).

## Minor / nice-to-have

- Consider hoisting `new Set<number>()` into a module-level `EMPTY_RETRYING_INDICES` used both in `useState`'s initializer and in `handleClose`'s `setRetryingIndices(new Set<number>())` — trivial allocation win and makes the "empty state" grep-able.
- Product-side follow-up (not this PR): now that retries can run in parallel, a per-row Cancel affordance would be more discoverable than "close the overlay to abort all". Would compose naturally with the `axios.isCancel` guard suggestion above.

---

src/components/mlir/MlirFileResultsOverlay.tsx
dcblundell: Fixed in b1c2d706 — added an early `axios.isCancel(err)` return in the catch block before the session-mismatch check, so aborts from unmount or future per-row cancel skip the FAILED writeback and error toast.

src/components/mlir/MlirFileResultsOverlay.tsx
dcblundell: Fixed in b1c2d706 — the `finally` block now only deletes from the map when `retryAbortControllersRef.current.get(index) === abortController`, so a stale promise cannot evict a newer controller for the same row.

src/components/mlir/MlirFileResultsOverlay.tsx
dcblundell: Fixed in b1c2d706 — added a mount cleanup `useEffect` that aborts all controllers and clears the map on unmount, matching `handleClose` behaviour for route changes and parent-driven unmounts.

src/hooks/useMlirRemote.tsx
dcblundell: Fixed in b1c2d706 — `MlirUploadOptions` is now exported with JSDoc on both `signal` (axios cancel contract + caller writeback responsibility) and `suppressProgressOverlay` (per-request opt-out, not concurrency-aware).

tests/MlirFileResultsOverlay.spec.tsx
dcblundell: Fixed in b1c2d706 — added tests for stale-writeback-after-reopen, distinct AbortSignals per parallel row, and exact abort toast copy for singular/plural cases.

PR #1730, opened 2026-07-07
Output Git metadata

### LINKED ISSUE #1711: Use Git information when available
Memory reports can now contain the git repo and commit hash. The logic for finding the source file can be updated and this information can be output into the UI if needed.

dcblundell:
If we have the git repo url and commit hash we now output it in the app.

Screenshot 2026-07-07 at 11 43 37

Screenshot 2026-07-07 at 11 51 29

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1711.

### DIFF
--- a/src/components/FooterInfobar.tsx
+++ b/src/components/FooterInfobar.tsx
@@ -39,6 +39,7 @@ import { Instance } from '../model/APIData';
 import LoadingSpinner from './LoadingSpinner';
 import { LoadingSpinnerSizes } from '../definitions/LoadingSpinner';
 import AppVersionStatus from './AppVersionStatus';
+import { formatShortSha } from '../functions/formatting';
 
 const RANGE_DISALLOWED_ROUTES: string[] = [ROUTES.NPE];
 
@@ -154,7 +155,7 @@ function FooterInfobar() {
                                     {reportMetadata?.gitSha && (
                                         <>
                                             <br />
-                                            <strong>Commit:</strong> {reportMetadata.gitSha.slice(0, 7)}
+                                            <strong>Commit:</strong> {formatShortSha(reportMetadata.gitSha)}
                                         </>
                                     )}
                                 </>
--- a/src/components/FooterInfobar.tsx
+++ b/src/components/FooterInfobar.tsx
@@ -33,7 +33,7 @@ import ReportLinkStatus from './ReportLinkStatus';
 import Range from './RangeSlider';
 import ROUTES from '../definitions/Routes';
 import 'styles/components/FooterInfobar.scss';
-import { useGetLatestAppVersion, useInstance } from '../hooks/useAPI';
+import { useGetLatestAppVersion, useInstance, useReportMetadata } from '../hooks/useAPI';
 import getServerConfig from '../functions/getServerConfig';
 import { Instance } from '../model/APIData';
 import LoadingSpinner from './LoadingSpinner';
@@ -53,6 +53,7 @@ function FooterInfobar() {
     const performanceReportLocation = useAtomValue(performanceReportLocationAtom);
 
     const { data: instance } = useInstance();
+    const { data: reportMetadata } = useReportMetadata();
     const location = useLocation();
     const {
         data: latestAppVersion,
@@ -141,7 +142,23 @@ function FooterInfobar() {
                     {activeProfilerReportPath && (
                         <Tooltip
                             disabled={!activeProfilerReportPath}
-                            content={formatPath(activeProfilerReportPath)}
+                            content={
+                                <>
+                                    <strong>Report path:</strong> {formatPath(activeProfilerReportPath)}
+                                    {reportMetadata?.gitUrl && (
+                                        <>
+                                            <br />
+                                            <strong>Git repo:</strong> {reportMetadata.gitUrl}
+                                        </>
+                                    )}
+                                    {reportMetadata?.gitSha && (
+                                        <>
+                                            <br />
+                                            <strong>Commit:</strong> {reportMetadata.gitSha.slice(0, 7)}
+                                        </>
+                                    )}
+                                </>
+                            }
                             position={PopoverPosition.TOP}
                         >
                             <div className='title'>
--- a/src/components/FooterInfobar.tsx
+++ b/src/components/FooterInfobar.tsx
@@ -39,7 +39,7 @@ import { Instance } from '../model/APIData';
 import LoadingSpinner from './LoadingSpinner';
 import { LoadingSpinnerSizes } from '../definitions/LoadingSpinner';
 import AppVersionStatus from './AppVersionStatus';
-import { formatShortSha } from '../functions/formatting';
+import { ReportGitMetadataLines } from './operation-details/GitCommitInfo';
 
 const RANGE_DISALLOWED_ROUTES: string[] = [ROUTES.NPE];
 
@@ -146,18 +146,10 @@ function FooterInfobar() {
                             content={
                                 <>
                                     <strong>Report path:</strong> {formatPath(activeProfilerReportPath)}
-                                    {reportMetadata?.gitUrl && (
-                                        <>
-                                            <br />
-                                            <strong>Git repo:</strong> {reportMetadata.gitUrl}
-                                        </>
-                                    )}
-                                    {reportMetadata?.gitSha && (
-                                        <>
-                                            <br />
-                                            <strong>Commit:</strong> {formatShortSha(reportMetadata.gitSha)}
-                                        </>
-                                    )}
+                                    <ReportGitMetadataLines
+                                        gitUrl={reportMetadata?.gitUrl ?? null}
+                                        gitSha={reportMetadata?.gitSha ?? null}
+                                    />
                                 </>
                             }
                             position={PopoverPosition.TOP}
--- a/src/components/operation-details/GitCommitInfo.tsx
+++ b/src/components/operation-details/GitCommitInfo.tsx
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { Icon } from '@blueprintjs/core';
+import { IconNames } from '@blueprintjs/icons';
+import { buildGitCommitUrl, formatShortSha } from '../../functions/formatting';
+
+export interface GitCommitInfoProps {
+    gitUrl: string | null;
+    gitSha: string;
+}
+
+function GitCommitInfo({ gitUrl, gitSha }: GitCommitInfoProps) {
+    const commitUrl = gitUrl ? buildGitCommitUrl(gitUrl, gitSha) : null;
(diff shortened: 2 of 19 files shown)


src/components/FooterInfobar.tsx
```diff
@@ -141,7 +142,23 @@ function FooterInfobar() {
...
+                                    {reportMetadata?.gitUrl && (
+                                        <>
+                                            <br />
+                                            <strong>Git repo:</strong> {reportMetadata.gitUrl}
+                                        </>
+                                    )}
+                                    {reportMetadata?.gitSha && (
+                                        <>
+                                            <br />
+                                            <strong>Commit:</strong> {reportMetadata.gitSha.slice(0, 7)}
```
aidemsined: The short-SHA truncation `gitSha.slice(0, 7)` now appears in three call sites: here in the footer tooltip, and twice inside `GitCommitInfo` in `SourceFileOverlay.tsx` (the linked branch and the plain-text branch). That crosses the "duplication at ≥3 sites → extract" threshold that the rest of `src/functions/formatting.ts` uses (`formatDuration`, `formatMemorySize`, `stripEnum`).

Extracting a `formatShortSha(sha: string): string` into `src/functions/formatting.ts` keeps the truncation length as a single canonical decision, and gives you an obvious place for `stripGitRemoteSuffix(url)` alongside if you decide to fix the SSH-URL anchor issue (see the comment on `SourceFileOverlay.tsx`).

**Suggested action:** move `gitSha.slice(0, 7)` behind a `formatShortSha` helper in `src/functions/formatting.ts` and swap all three call sites over.

src/components/operation-details/SourceFileOverlay.tsx
```diff
@@ -216,6 +234,33 @@ function SourceFileOverlay({
...
+    gitUrl: string | null;
+    gitSha: string;
+}
+
+const GitCommitInfo = ({ gitUrl, gitSha }: GitCommitInfoProps) => (
+    <span className='stack-trace-git-info'>
+        {'Commit: '}
+        {gitUrl ? (
+            <a
+                href={`${gitUrl.replace(/\.git$/, '')}/commit/${gitSha}`}
```
aidemsined: `href={`${gitUrl.replace(/\.git$/, '')}/commit/${gitSha}`}` assumes `gitUrl` starts with an HTTP scheme, but nothing on the frontend or backend guarantees that. For a repo whose stored remote is the Git default over SSH (`git@github.com:foo/bar.git`), stripping `.git` yields the scheme-less string `git@github.com:foo/bar/commit/<sha>`. `target='_blank'` will then treat that as a relative URL and open `https://<app-origin>/git@github.com:foo/bar/commit/<sha>` in a new tab — a 404 under the ttnn-visualizer origin, not GitHub.

Same class of problem for Bitbucket (uses `/commits/<sha>` plural), for trailing-slash URLs, and for URLs with `?ref=…` query strings — but SSH is the common case that's most likely to hit users capturing reports from dev machines.

Dormant today because `dev`'s backend doesn't emit `git_url`, but the moment a backend writer lands (the whole point of #1711), any user whose report was captured with an SSH remote gets a broken outbound link.

**Suggested action:** either (a) frontend guard — refuse to render the anchor when `!/^https?:\/\//.test(gitUrl)` and fall through to the plain-text `gitSha.slice(0, 7)` branch, or (b) normalise SSH → HTTPS backend-side before emitting `git_url` (`git@host:owner/repo.git` → `https://host/owner/repo`). Frontend guard is the cheaper immediate fix; backend normalisation is the more principled one since it also fixes the raw display in the footer tooltip.

src/components/operation-details/SourceFileOverlay.tsx
```diff
@@ -216,6 +234,33 @@ function SourceFileOverlay({
     );
 }
 
+interface GitCommitInfoProps {
+    gitUrl: string | null;
+    gitSha: string;
+}
+
+const GitCommitInfo = ({ gitUrl, gitSha }: GitCommitInfoProps) => (
```
aidemsined: The new module-private `GitCommitInfo` produces two structurally-different DOM shapes — `<a href=… target=_blank>` wrapping the SHA + Icon when `gitUrl` is set, and a bare `<span>` when it's null — but neither is exercised by any spec. The only test that mocks `useReportMetadata` (`tests/PerfTensorDrawer.spec.tsx:21`) returns `{ data: undefined }`, hitting the outer `reportMetadata?.gitSha` guard before `GitCommitInfo` ever renders. `tests/SourceFileButton.spec.tsx` doesn't mock `useAPI` at all — it renders with an unset `activeProfilerReportAtom`, which disables the query and again bypasses this code.

Consequence: any future change to the href template — including the SSH-URL fix on the neighbouring comment — will land without CI coverage.

**Suggested action:** add a focused `GitCommitInfo.spec.tsx` (or extend `SourceFileButton.spec.tsx` by seeding a profiler-report atom + stubbing `useReportMetadata`). Two cases:
1. `{ gitUrl: 'https://github.com/foo/bar.git', gitSha: 'abcdef0123…' }` → `getByRole('link').getAttribute('href') === 'https://github.com/foo/bar/commit/abcdef0123…'`, visible text `abcdef0`.
2. `{ gitUrl: null, gitSha: 'abcdef0123…' }` → `queryByRole('link') === null`, visible text `abcdef0`.

src/components/FooterInfobar.tsx
```diff
@@ -53,6 +53,7 @@ function FooterInfobar() {
     const performanceReportLocation = useAtomValue(performanceReportLocationAtom);
 
     const { data: instance } = useInstance();
+    const { data: reportMetadata } = useReportMetadata();
```
aidemsined, line 57: `FooterInfobar` has no spec file — `rg -l FooterInfobar tests/` returns nothing. This PR is a good moment to add one because the tooltip's `content` prop just grew from a plain string into a JSX subtree with three conditional branches (`gitUrl` line, `gitSha` line, both), and it now depends on a fourth `useQuery` hook.

Every render path here is uncovered: the plain-tooltip regression (path only), the git-repo-only branch, the git-sha-only branch, and the both-present branch.

**Suggested action:** add `tests/FooterInfobar.spec.tsx` that mocks `useReportMetadata`, `useInstance`, and `useGetLatestAppVersion`, sets `activeProfilerReportAtom` to a stub, and asserts the four tooltip content shapes. `Tooltip`'s subtree isn't in the DOM until hover — either use `isOpen` override for testing or assert directly against the rendered `content` JSX. Should be characterisable in ~40 lines.

src/components/FooterInfobar.tsx
```diff
@@ -141,7 +142,23 @@ function FooterInfobar() {
                     {activeProfilerReportPath && (
                         <Tooltip
                             disabled={!activeProfilerReportPath}
-                            content={formatPath(activeProfilerReportPath)}
+                            content={
+                                <>
+                                    <strong>Report path:</strong> {formatPath(activeProfilerReportPath)}
+                                    {reportMetadata?.gitUrl && (
```
aidemsined: The footer gates `Git repo:` and `Commit:` independently — you can end up with a repo URL but no commit, or a commit but no repo URL. The source-file overlay, by contrast, only mounts `GitCommitInfo` when `gitSha` is truthy (so a repo-only report shows nothing in the stack-trace header).

Is that asymmetry deliberate? It reads odd — if the ticket's intent is "surface git info when available", surfacing repo-only in the footer but nothing in the overlay is inconsistent. If repo-only reports are considered a real state, the overlay should probably render a `Repo: <url>` line too; if they're vanishingly rare, the footer could gate both together for cleanliness.

**Suggested action:** confirm with product/UX which state is real, and either gate both surfaces on `gitSha` (cleanliness) or add a repo-only render to the overlay (parity).

aidemsined (commented):
## General observations

- **Ticket scope note.** Issue #1711 mentions both source-file finding logic AND UI display as "can be" done. Only the UI half ships here, which is defensible given the soft phrasing — but worth calling out in the PR description whether the source-file resolution half is a planned follow-up so #1711 doesn't auto-close with only half the intent addressed.
- **Orthogonal wording tweak.** The `FileStatusOverlay` copy change ("N% complete" → "N%" + " elapsed") is unrelated to git metadata. Consider landing it in a separate PR or calling it out in the description so reviewers know what to focus on.
- **`rel='noopener noreferrer'`** on the new external commit link (`SourceFileOverlay.tsx:249`). Modern browsers imply `noopener` under `noreferrer`, so this isn't a real security gap — but `rel='noopener noreferrer'` is the belt-and-braces convention.
- **`title={gitSha}` for full-SHA affordance.** The truncated SHA has no way to reveal the full 40-char value. Adding `title={gitSha}` on the anchor/span exposes it via hover + assistive tech, without any visual change.
- **URL-normalisation edge cases beyond SSH.** `gitUrl.replace(/\.git$/, '')` also mishandles trailing slashes (`bar//commit/<sha>`), URLs with `?ref=…` query strings (query smeared with `/commit/<sha>`), and Bitbucket's `/commits/<sha>` (plural). Backend-side canonicalisation would neutralise all of them.
- **Extend `backend/ttnn_visualizer/tests/views/test_report_metadata.py`** with `git_url` / `git_sha` fixture rows to document the passthrough contract before tt-metal starts writing them. The current success test asserts an exact key set, which will become brittle the moment the backend writer lands.
- **Document transitive `useReportMetadata` mocking.** No CONVENTIONS/AGENTS rule today says "mock `useReportMetadata` in specs that transitively render components using it". Consider adding one so the next transitive consumer doesn't repeat the `PerfTensorDrawer.spec.tsx` discovery.
- **[perf]** `useReportMetadata` fanout across concurrent `SourceFileOverlay` mounts is a non-issue today because `OperationList.tsx` virtualizes (only a viewport's worth of `SourceFileButton` mounts) and React Query dedupes subscribers to a single network request per key. If virtualization is ever removed, per-observer bookkeeping scales linearly. No action for this PR; note only.
- **[perf]** The new `Tooltip` `content={<>…</>}` in `FooterInfobar` allocates a fresh element tree per parent render. Popover2 is lazy-mounted so this is negligible in practice; noted only.
- **UX: sha-only, no URL.** When `gitSha` is present but `gitUrl` is null, `GitCommitInfo` renders `Commit: abc1234` as plain text (no link, no icon). The `Commit:` label may read like a link affordance; plain `abc1234` or a different label might be clearer.
- **Optional cleanup:** extract a small `<ProfilerPathTooltipContent>` from `FooterInfobar.tsx` for symmetry with `<GitCommitInfo>`. Very optional aesthetic.

---

src/components/operation-details/GitCommitInfo.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    return (
+        <span className='report-git-info'>
+            <strong>Commit:</strong>{' '}
+            {commitUrl ? (
+                <a
+                    href={commitUrl}
+                    target='_blank'
+                    rel='noreferrer'
+                >
```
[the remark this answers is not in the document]
dcblundell: No.

PR #1736, opened 2026-07-08
Tracking issue: keyboard-accessible chart filtering interaction

dcblundell:
## Summary
Chart-driven filtering in the Performance Charts tab is currently pointer-click only. Users can click chart segments/slices to prefilter the table and switch to the Table tab, but there is no equivalent keyboard path.

## Current behaviour
- Filtering is wired through Plot click callbacks in performance chart components.
- The UI hint text advertises click interaction.
- Keyboard users cannot trigger the same filter action via focused controls.

## Why this matters
This introduces an accessibility gap for a primary interaction path in the performance analysis workflow.

## Proposed follow-up
1. Add a keyboard-operable interaction path for chart filtering.
2. Ensure enabled/disabled semantics mirror click availability.
3. Add tests that exercise keyboard activation and verify table prefilter + tab switch behaviour.
4. Keep existing pointer-click behaviour unchanged.

## Acceptance criteria
- A keyboard user can trigger operation filtering from charts without a pointer device.
- Accessibility tests cover the new path.
- Existing chart click tests remain green.

Related work: refactor centralises chart click handler and filter hint rendering in this branch.

### DIFF
--- a/src/components/performance/NonFilterablePerfCharts.tsx
+++ b/src/components/performance/NonFilterablePerfCharts.tsx
@@ -24,6 +24,7 @@ interface NonFilterablePerfChartsProps {
     convData: TypedPerfTableRow[][];
     hasMatmulData: boolean;
     hasConvData: boolean;
+    onOpCodeClick?: (opCode: string) => void;
 }
 
 const NonFilterablePerfCharts = ({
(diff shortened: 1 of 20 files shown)


dcblundell:
Closing: created by mistake while intending to open a tracking issue.

Issue #1737, opened 2026-07-08
Add keyboard-accessible path for chart-driven table filtering


dcblundell:
## Summary
Chart-driven filtering in the Performance Charts tab is currently pointer-click only. Users can click chart segments/slices to prefilter the table and switch to the Table tab, but there is no equivalent keyboard path.

## Current behaviour
- Filtering is wired through Plot click callbacks in performance chart components.
- The UI hint text advertises click interaction.
- Keyboard users cannot trigger the same filter action via focused controls.

## Why this matters
This creates an accessibility gap for a primary interaction in the performance analysis workflow.

## Proposed follow-up
1. Add a keyboard-operable interaction path for chart filtering.
2. Ensure enabled/disabled semantics mirror click availability.
3. Add tests that exercise keyboard activation and verify table prefilter + tab switch behaviour.
4. Keep existing pointer-click behaviour unchanged.

## Acceptance criteria
- A keyboard user can trigger operation filtering from charts without a pointer device.
- Accessibility tests cover the new path.
- Existing chart click tests remain green.

Related branch work: refactor centralises chart click handler and filter hint rendering in `feature/filter-perf-from-charts`.

dcblundell:
### Scope addition: histogram duration-bucket controls (from #1843)

#1843 adds a second, distinct pointer-only affordance to the Charts tab that this issue should cover — it isn't just more trace clicks.

The `Op Duration Distribution` chart draws one control per column as a Plotly **annotation** (SVG with `captureevents`) to filter the table by duration bucket. These have no role, no accessible name, and no focus or keyboard activation. Critically, the chart also sets `showticklabels: false` on the x axis, because the annotations stand in for the tick labels — so the bucket **range text now exists only inside the inaccessible controls**. That makes it a net accessibility regression for this chart rather than unchanged debt, since a screen-reader user previously had the ranges as axis labels.

**Proposed design** (from the Copilot review on #1843, and it looks right): render a parallel set of real HTML controls for the buckets — buttons or a listbox — mirroring the annotations, reachable by keyboard and AT. That restores the range text to the accessibility tree as a side effect, which a keyboard handler bolted onto the SVG would not. It generalises to the trace-click filtering this issue already covers.

Worth handling alongside the shift-click work still open on #1754, since both change how bucket selection is driven.

The `TODO(#1737)` in `src/components/performance/PerfChartHint.tsx` now names these controls specifically.

PR #1738, opened 2026-07-08
Filter performance table from charts

### LINKED ISSUE #844: Performance: Interacting chart ops should prefilter the perf table by that op type

dcblundell:
Allows a user to set the operation name filter from two specific performance charts.

**MatmulDeviceOperation example**
Screenshot 2026-07-08 at 16 09 03

Screenshot 2026-07-08 at 16 09 08

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/844.


src/components/performance/PerfOperationTypesChart.tsx
```diff
@@ -37,17 +40,20 @@ function PerfOperationTypesChart({
     opCodes,
     className = '',
     id,
+    onOpCodeClick,
 }: PerfOperationTypesChartProps) {
     const filteredOpCodes = useMemo(
         () => [...new Set(data?.filter((row) => row.raw_op_code !== undefined).map((row) => row.raw_op_code))],
         [data],
     );
```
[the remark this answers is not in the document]
dcblundell: Good catch — aligned with the table pipeline. Both charts now use `getUniqueChartRawOpCodes`, which skips `null`, `undefined`, and empty strings before building trace labels/customdata. Covered in `tests/getUniqueChartRawOpCodes.spec.ts`.

src/scss/components/PerformanceOperationTypesChart.scss
```diff
@@ -2,6 +2,8 @@
 //
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
+@use './PerfCharts' as perf-charts;
+
```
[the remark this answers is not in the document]
dcblundell: Agreed. Removed the local `&.perf-chart-clickable` block — the component still adds the `perf-chart-clickable` class and picks up hint + cursor styling from the global rule in `PerfCharts.scss` (loaded via the charts tab).

src/components/performance/PerfOpCountVsRuntimeChart.tsx
```diff
@@ -36,6 +38,8 @@ function PerfOpCountVsRuntimeChart({ selectedOpCodes, datasets = [] }: PerfOpCou
         [opCodes, selectedOpCodeSet],
     );
```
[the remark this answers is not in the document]
dcblundell: Addressed via the same `getUniqueChartRawOpCodes` helper used by the pie chart — op codes are filtered to non-nullish strings before `selectedOpCodeSet.has(opCode)` runs, so null values cannot reach `customdata` or hover templates.

src/routes/Performance.tsx
```diff
@@ -178,7 +180,8 @@ export default function Performance() {
 
     useEffect(() => {
         setSelectedPerfRowId(null);
-    }, [activePerformanceReport?.path, setSelectedPerfRowId]);
+        setRawOpCodeFilterList([]);
+    }, [activePerformanceReport?.path, setRawOpCodeFilterList, setSelectedPerfRowId]);
```
aidemsined: The report-change reset raises three questions worth answering together before merge.

**1. Reset scope is asymmetric with the sibling filter atoms.** `mathFilterListAtom`, `bufferTypeFilterListAtom`, and `layoutFilterListAtom` still persist across report switches (pruned only by the `PerfReport.tsx:293-311` self-heals). Nothing in the PR explains why `rawOpCodeFilterListAtom` alone should be cleared at the route level. Pick one policy — co-clear all four (an extracted `resetPerfTableSessionState()` would fit here), or leave route-level clears out and lean on the child self-heals.

**2. User-typed op-code filters get clobbered on report change.** The atom is shared with the `MultiSelectField` chip in `PerfReport.tsx:99`; the effect can't tell "user typed `Matmul`" from "chart click set `Matmul`". Users lose their manual chip on every report switch. If that's intentional UX, worth calling out in the PR body; if not, a companion "auto-set" flag would let the reset target only chart-driven prefilters.

**3. The reset also fires on initial `/performance` mount, not only on report *change*.** `useEffect` always runs its initial cycle, so any op-code filter carried from a prior route visit is lost on route re-entry. The new `tests/Performance.spec.tsx:1063` "does not re-clear while unchanged" test happens within a single mount, so it doesn't cover this path. Gate with a `prevPath` ref if unintended.

**Suggested action:** decide the intended scope (all filter atoms or none at route level) and the intended trigger (change-only vs mount-included), then either narrow the effect or extend it symmetrically.

src/components/performance/PerfOpCountVsRuntimeChart.tsx
```diff
@@ -47,6 +49,7 @@ function PerfOpCountVsRuntimeChart({ selectedOpCodes, datasets = [] }: PerfOpCou
                             type: 'bar',
                             name: getPlotLabel(dataIndex, perfReport?.reportName, comparisonReportList),
                             hovertemplate: `${opCode}<br />%{y:.1%}`,
+                            customdata: [[opCode]],
```
aidemsined: Two producers ship different `customdata` shapes: bars use `[[opCode]]` here (and again at L76), while pie sets flat `customdata: [...filteredOpCodes]` in `PerfOperationTypesChart.tsx:54`. `getRawOpCodeFromPlotClick` papers over the divergence with `raw[0]`-unwrap and a `label` fallback (`src/functions/getRawOpCodeFromPlotClick.ts:18-19`), so today it works — but the `label` fallback in particular is fragile: any producer that sets a label distinct from `raw_op_code` (localized string, human-readable form) silently misfilters.

**Suggested action:** normalize on `customdata: opCode` (flat per-point string) at both producers. The extractor then collapses to `event.points[0]?.customdata` string-guard — the `label` fallback and the array-unwrap both go away, and the spec at `tests/getRawOpCodeFromPlotClick.spec.ts` shrinks accordingly.

src/hooks/useHandlePerfChartPlotClick.ts
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { useCallback } from 'react';
+import { PlotMouseEvent } from 'plotly.js';
+import { getRawOpCodeFromPlotClick } from '../functions/getRawOpCodeFromPlotClick';
+
+export type OnOpCodeClick = (opCode: string) => void;
```
aidemsined: This callback type is imported from five components (`PerfCharts.tsx`, `NonFilterablePerfCharts.tsx`, `PerfOpCountVsRuntimeChart.tsx`, `PerfOperationTypesChart.tsx`, `PerformanceChartsTab.tsx`). A type that widely imported reads more naturally in `src/definitions/` next to `PERF_CHART_TABLE_FILTER_HINT` — `AGENTS.md` puts cross-cutting primitives there.

**Suggested action:** move `OnOpCodeClick` into `src/definitions/PerformanceCharts.ts`, keep the hook importing it back. Not blocking, but the current placement will keep drawing imports across the hook boundary as the click surface grows.

src/components/performance/PerfClickableChartFrame.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface PerfClickableChartFrameProps {
+    id?: string;
+    className?: string;
+    title: string;
+    subtitle?: ReactNode;
+    isClickable: boolean;
+    children: ReactNode;
+}
+
+function PerfClickableChartFrame({
```
aidemsined: Six of the eight sites that render through this component pass `isClickable={false}` — every non-filterable `PerfChart` consumer (kernel duration, kernel runtime, device time, core count, kernel utilisation × 2). The name suggests it wraps only clickable charts, but it's actually the shared chrome frame (h3 + optional subtitle + hint slot) for the whole perf chart family.

**Suggested action:** rename to `PerfChartFrame`. Small mechanical change, but the intent reads more accurately for a future contributor landing here from a non-clickable chart wondering why they render through a "clickable" wrapper.

src/components/performance/PerfOperationTypesChart.tsx
```diff
@@ -63,22 +67,25 @@ function PerfOperationTypesChart({
...
             <Plot
                 className='chart'
                 data={[chartData]}
                 layout={LAYOUT}
                 config={PerfChartConfig}
+                onClick={isClickable ? handlePlotClick : undefined}
                 useResizeHandler
             />
-        </div>
+        </PerfClickableChartFrame>
```
aidemsined: Pie duplicates the "frame + Plot + onClick + hint" plumbing that `PerfChart` already provides for bar-family charts (`PerfChart.tsx:74` and below). This PR extends the duplication by wiring `onClick` into two separate `<Plot>` sites (here and inside `PerfChart`). Pre-existing divergence, but this is a natural moment to consolidate: routing the pie through `PerfChart` (or extracting a shared inner shell) means future click/hint changes touch one file, not two.

**Suggested action:** worth a follow-up. Not blocking this PR, but the seam is now clearly visible in both directions.

aidemsined (commented):
## General observations

- `setRawOpCodeFilterList([])` at `src/routes/Performance.tsx:183` allocates a new empty-array reference every mount. Jotai compares with `Object.is`, so table subscribers re-render even on `empty → empty`. Guard: `setRawOpCodeFilterList((prev) => (prev.length === 0 ? prev : []))`.
- On report change, `PerfReport.tsx:292-298` self-heal runs before the new route-level clear; on that path the child self-heal for `rawOpCodeFilterListAtom` becomes dead work. Not urgent — the self-heal still fires meaningfully on data mutation without a path change.
- Coverage seam: `tests/PerformanceChartPrefilter.integration.spec.tsx` starts with an empty filter. `tests/usePrefilterPerfTableByOpCode.spec.tsx` proves atom-replace semantics in isolation, but no integration test seeds a manual filter first and asserts end-to-end replacement.
- Coverage seam: no explicit test for "user manually clears the chart-set chip" — mechanically covered by the shared clear path, but a short assertion would close the loop on ghost state (tab, hint, filter).
- The "does not re-clear while unchanged" test at `tests/Performance.spec.tsx:1063-1078` re-selects the *same* report reference; `useAtom` short-circuits on `Object.is`, so the effect dep never changes. Assertion reads broader than what the test proves — consider a `path → null → same path` variant.
- Integration test add-on: assert `TEST_IDS.PERF_CHART_TABLE_FILTER_HINT` is visible on the Charts tab pre-click, and assert the `MultiSelectField` chip reflects the atom after tab switch. Strengthens the UX + atom-binding legs of AC #844.
- `useHandlePerfChartPlotClick(onOpCodeClick)` is called unconditionally in both chart components even when `onOpCodeClick` is omitted (`PerfOpCountVsRuntimeChart.tsx:39`, `PerfOperationTypesChart.tsx:44`). Harmless, but a `useCallback` per non-clickable render can be avoided by an early conditional.
- **[perf, pre-existing]** `flattenedData = datasets.flat()` at `PerfOpCountVsRuntimeChart.tsx:22` allocates a new array each render, defeating `useMemo(..., [flattenedData])` for `opCodes`. Not introduced by this PR — surfacing only.
- Keyboard-operable path is already tracked via `TODO(#1737)` in `PerfChartFilterHint.tsx:12` — pointer-only is fine for #844, follow-up is correctly filed.

---

src/routes/Performance.tsx
dcblundell: Will reset all filters on report change.

dcblundell: Agreed. Policy: clear **all** table chip filters (op code, math, buffer type, layout) plus the selected row when the active report path **changes**.

Implemented via `useResetPerfTableSessionState()`, gated with a previous-path ref so the initial mount / route remount does not wipe filters. Child self-heals in `PerfReport` remain as a safety net.

src/components/performance/PerfOpCountVsRuntimeChart.tsx
dcblundell: Agreed. Both bar producers now set flat `customdata: [opCode]` (per-point string on click), matching the pie. `getRawOpCodeFromPlotClick` is a string guard only — array unwrap and `label` fallback removed.

src/hooks/useHandlePerfChartPlotClick.ts
dcblundell: Agreed. Moved `OnOpCodeClick` to `src/definitions/PerformanceCharts.ts` next to `PERF_CHART_TABLE_FILTER_HINT`; call sites import from there.

src/components/performance/PerfClickableChartFrame.tsx
dcblundell: Agreed. Renamed to `PerfChartFrame` — shared chrome for clickable and non-clickable charts.

src/components/performance/PerfOperationTypesChart.tsx
dcblundell: Agreed this is a natural consolidation seam. Tracking as a follow-up rather than expanding this PR — leaving pie/PerfChart duplication for a dedicated cleanup.

dcblundell: Follow-up filed: https://github.com/tenstorrent/ttnn-visualizer/issues/1756

PR #1750, opened 2026-07-09
Expand file transfer atom storage

### LINKED ISSUE #1679: fileTransferProgressAtom is a single global slot shared across concurrent transfers
### Summary

`fileTransferProgressAtom` is a single global slot. When two file transfers run in parallel, they share it: the first one's `finally`/reset path sets progress to `INACTIVE` and closes the shared `FileStatusOverlay` while the second transfer is still in flight.

### Affected paths

This is **pre-existing** — it already affects `useLocal`'s local-folder / remote-sync flows. It became easier to trip into with PR #1667, which adds a third independent upload path (MLIR server upload in `useMlirRemote`) writing to the same atom:

- `src/hooks/useLocal.tsx` (local folder + remote sync)
- `src/hooks/useMlirRemote.tsx` (MLIR server upload)

All write to `fileTransferProgressAtom` in `src/store/app`.

### Possible fixes

- (a) Make the atom a stack / per-source map keyed by transfer source, so each transfer tracks its own progress and the overlay reflects aggregate/active state; or
- (b) Explicitly guard against starting a second transfer while one is active.

### Context

Raised by @aidemsined in review of PR #1667: https://github.com/tenstorrent/ttnn-visualizer/pull/1667#discussion_r3467932560

dcblundell:
This pull request introduces a significant refactor of the file transfer progress tracking system. The core change is the migration from a single global progress atom to a new per-source registry, enabling independent tracking and UI updates for multiple concurrent file transfers (e.g., remote sync, local upload, MLIR upload). This is achieved by introducing a new `FileTransferSource` enum, a registry module, and updating all related components and hooks to use the new architecture. Additionally, the `jotai-family` package is added to support this atom family pattern.

The most important changes are:

**File Transfer Progress Refactor:**

* Introduced a new per-source file transfer registry in `fileTransferRegistry.ts`, with utility functions and atom families for accessing and updating progress by source. This replaces the previous single global atom approach. (`src/functions/fileTransferRegistry.ts`)
* Added the `FileTransferSource` enum to define distinct transfer contexts (REMOTE_SYNC, LOCAL_UPLOAD, MLIR_UPLOAD). (`src/definitions/FileTransferSource.ts`)
* Refactored all usages of file transfer progress in hooks and components (`useLocal`, `useMlirRemote`, `FileStatusOverlay`, `RemoteSyncButton`, `RemoteSyncConfigurator`, `SocketProvider`, etc.) to use the new per-source registry and utility functions. [[1]](diffhunk://#diff-43c7bd7b98d931e14e474f9a18e98f57bb722aa2f3f2050af7fc3aaab1e9e6ccL45-R45) [[2]](diffhunk://#diff-bc0111215de66ab133230c3ee4f917e6e2344e481497d49c135d55208b2c4e02L9-R10) [[3]](diffhunk://#diff-bc0111215de66ab133230c3ee4f917e6e2344e481497d49c135d55208b2c4e02L36-R37) [[4]](diffhunk://#diff-780d8222a8c22db347c306b8b563a0d18201eef1fbacd5f67d7934220019e6b0L9-L22) [[5]](diffhunk://#diff-780d8222a8c22db347c306b8b563a0d18201eef1fbacd5f67d7934220019e6b0L35-R35) [[6]](diffhunk://#diff-4ee8a8320020eed5af3685433a85b5c6351822f7c2900f1342d1be650866bb11L99-R102) [[7]](diffhunk://#diff-4ee8a8320020eed5af3685433a85b5c6351822f7c2900f1342d1be650866bb11L111-R114) [[8]](diffhunk://#diff-b55165b971acd54ec976979c1cc187aab20abdf9a113b55a0e551d942ca20f1aL52-R68) [[9]](diffhunk://#diff-b55165b971acd54ec976979c1cc187aab20abdf9a113b55a0e551d942ca20f1aL90-R94) [[10]](diffhunk://#diff-b55165b971acd54ec976979c1cc187aab20abdf9a113b55a0e551d942ca20f1aL113-R117) [[11]](diffhunk://#diff-2e11e7b1f151fac8271b3811e4d1d83f7cd8f4ccb2a2fa15f611cbbc935d5c35L8-R13) [[12]](diffhunk://#diff-2e11e7b1f151fac8271b3811e4d1d83f7cd8f4ccb2a2fa15f611cbbc935d5c35L28-R36) [[13]](diffhunk://#diff-2e11e7b1f151fac8271b3811e4d1d83f7cd8f4ccb2a2fa15f611cbbc935d5c35L64-R62) [[14]](diffhunk://#diff-2e11e7b1f151fac8271b3811e4d1d83f7cd8f4ccb2a2fa15f611cbbc935d5c35L89-R87) [[15]](diffhunk://#diff-52a0eb62b4f502da21d786251a5ba88bf4b0f6c058f2cc6e1f88555fa44449c3L22-R29)

**Dependency Updates:**

* Added the `jotai-family` package to dependencies and lockfiles to enable atom families. (`package.json`, `pnpm-lock.yaml`, `LICENSE`) [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R69) [[2]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR47-R49) [[3]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR3694-R3699) [[4]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR9817-R9820) [[5]](diffhunk://#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7R132)

These changes make file transfer progress reporting more robust and scalable, allowing for better user experience and maintainability.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1679.


src/functions/fileTransferRegistry.ts
```diff
@@ -0,0 +1,83 @@
...
+    FileTransferSource.REMOTE_SYNC,
+    FileTransferSource.MLIR_UPLOAD,
+    FileTransferSource.LOCAL_UPLOAD,
+];
+
+export const fileTransferRegistryAtom = atom<FileTransferRegistry>({});
+
+export const fileTransferProgressBySourceAtom = atomFamily((source: FileTransferSource) =>
+    atom((get) => get(fileTransferRegistryAtom)[source] ?? INACTIVE_FILE_TRANSFER_PROGRESS),
+);
```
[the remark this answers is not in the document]
dcblundell: Fair point on AGENTS.md — these atoms are re-exported from `store/app.ts` for discoverability.

They are defined alongside the registry helpers in `fileTransferRegistry.ts` to avoid a circular import: `app.ts` imports `aggregateFileTransferProgress` to build the derived `fileTransferProgressAtom`, while the setters need `fileTransferRegistryAtom`. Defining atoms in `app.ts` and helpers in `functions/` with cross-imports recreates that cycle unless we split further (e.g. a dedicated atoms module).

Added brief comments in both files documenting the exception. Happy to extract to a small `fileTransferAtoms.ts` if we want strict placement without the cycle.

src/functions/fileTransferRegistry.ts
```diff
@@ -0,0 +1,86 @@
...
+export function aggregateFileTransferProgress(registry: FileTransferRegistry): FileProgress {
+    for (const source of FILE_TRANSFER_SOURCE_PRIORITY) {
+        const progress = registry[source];
+        if (progress && isActiveTransferStatus(progress.status)) {
+            return progress;
+        }
+    }
+
+    return INACTIVE_FILE_TRANSFER_PROGRESS;
+}
```
aidemsined, line 63: The aggregator's contract is doing product-visible work but has no in-code rationale, and its `FAILED`-swallowing behaviour deserves an explicit call.

`FILE_TRANSFER_SOURCE_PRIORITY` at L28-32 hardcodes `REMOTE_SYNC > MLIR_UPLOAD > LOCAL_UPLOAD` — fine, but future readers should be told *why* ("remote sync is the most disruptive; MLIR upload beats a local file drop") so nobody reorders it thinking it's arbitrary. More consequentially: `isActiveTransferStatus` excludes `FAILED` alongside `FINISHED` and `INACTIVE` (`src/functions/getFileStatusLabel.ts:21-30`), so the aggregator skips a failed high-priority slot and falls through to any active lower-priority source. Concretely: if REMOTE_SYNC ends up in `FAILED` while LOCAL_UPLOAD is `UPLOADING`, the overlay switches to LOCAL_UPLOAD and the failure never surfaces there. The sync flow's error toast still fires — but a socket-emitted `FAILED` without a matching axios rejection (or backend-driven terminal `FAILED` mid-download) would silently disappear from the overlay.

The existing `tests/fileTransferRegistry.spec.ts:85-96` case (`REMOTE_SYNC=FINISHED + MLIR_UPLOAD=PROCESSING`) accidentally satisfies priority already — it doesn't exercise the concerning path.

**Suggested action:** pick a semantics and lock it in — (a) surface `FAILED` explicitly in the aggregator (prefer FAILED over active lower-priority), (b) filter `FAILED` writes at the `SocketProvider` seam so it can't linger in the registry, or (c) document the "toasts own errors, overlay follows active work" contract inline. Regardless of choice, add a `REMOTE_SYNC=FAILED + LOCAL_UPLOAD=UPLOADING` test in `tests/fileTransferRegistry.spec.ts` so a future change can't drift the contract silently.

src/store/app.ts
```diff
@@ -20,25 +19,27 @@ import { SortingOptions } from '../definitions/SortingOptions';
...
+    clearFileTransferProgressForSource,
+    clearFileTransferProgressForSourceIfInactive,
+    fileTransferProgressBySourceAtom,
+    fileTransferRegistryAtom,
+    getInactiveFileTransferProgress,
+    setFileTransferProgressForSource,
+} from '../functions/fileTransferRegistry';
 
-export const fileTransferProgressAtom = atom<FileProgress>(getInactiveFileTransferProgress());
+export const fileTransferProgressAtom = atom((get) => aggregateFileTransferProgress(get(fileTransferRegistryAtom)));
```
aidemsined, line 42: This symbol used to be a writable primitive atom; it is now a derived atom. The export name is identical, so the contract change is invisible at import sites. Any missed `set(fileTransferProgressAtom, ...)` (or `useAtom` destructuring the setter) will throw at runtime with an unhelpful "read-only atom" error — confirmed no such call remains in `src/` today, but a fork, a rebase, or a future contributor reading the name won't know the write path moved.

**Suggested action:** rename to `aggregatedFileTransferProgressAtom` to force call-site attention, OR add a JSDoc above the export making the read-only contract explicit and pointing writers to `setFileTransferProgressForSource`. A rename is the safer choice since the semantic *did* change; a JSDoc is the cheaper choice if you want to preserve the import path across the repo.

tests/fileTransferRegistry.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+            ...REMOTE_SYNCING,
+            status: FileStatus.FINISHED,
+        });
+
+        clearFileTransferProgressForSourceIfInactive(FileTransferSource.REMOTE_SYNC);
+
+        const registry = getDefaultStore().get(fileTransferRegistryAtom);
+        expect(registry[FileTransferSource.REMOTE_SYNC]).toBeUndefined();
+        expect(registry[FileTransferSource.MLIR_UPLOAD]).toEqual(MLIR_UPLOADING);
+    });
```
aidemsined, line 141: The reconnect fix in `SocketProvider.tsx:32-38` is entirely load-bearing on `clearFileTransferProgressForSourceIfInactive` preserving an *active* slot, but this suite only covers `FINISHED → cleared`. Two negatives are missing:

1. **Active slot must be preserved.** Add a case where `REMOTE_SYNC` is `DOWNLOADING` (or `UPLOADING` / `PROCESSING` / `STARTED`), call the helper, and assert the entry is untouched. Without this, a future refactor to `isActiveTransferStatus` or to `ACTIVE_TRANSFER_STATUSES` in `getFileStatusLabel.ts` could silently break the reconnect-mid-upload guarantee.
2. **`FAILED` must be cleared.** Per `ACTIVE_TRANSFER_STATUSES` (`src/functions/getFileStatusLabel.ts:21-30`), `FAILED` is not active, so the guard should let it through. Currently only `FINISHED` is asserted.

**Suggested action:** two additional test cases here, both one-liners against the existing `MLIR_UPLOAD` / `REMOTE_SYNC` seeding pattern.

src/libs/SocketProvider.tsx
```diff
@@ -25,17 +26,14 @@ interface SocketProviderProps {
...
         socket.on('connect', () => {
-            // Only clear stale progress if the previous transfer is no longer
+            // Only clear stale remote-sync progress if that transfer is no longer
             // active. A reconnect mid-upload (axios still running on the same
-            // tab) must not close the overlay or drop the live progress.
-            setFileTransferProgress((previous: FileProgress) =>
-                isActiveTransferStatus(previous.status) ? previous : getInactiveFileTransferProgress(),
-            );
+            // tab) must not close the overlay or drop live client-driven progress.
+            clearFileTransferProgressForSourceIfInactive(FileTransferSource.REMOTE_SYNC);
```
aidemsined, line 36: This guard is correct for the mid-upload reconnect case: an axios upload still running client-side will keep pushing progress into the registry, so preserving `REMOTE_SYNC` on `connect` prevents the overlay from flickering closed.

The less obvious side-effect: if the *backend* dies mid-transfer without emitting a terminal `FAILED` (e.g. socket disconnect with the process gone rather than a graceful end), `REMOTE_SYNC` stays `DOWNLOADING` in the registry — the reconnect guard now preserves it. `syncRemoteFolder` has no client-side timeout configured, so the axios call can hang indefinitely, and the overlay stays frozen on the last-known progress until the user closes the tab. This isn't a regression — pre-PR the same atom-based clear had the same failure mode. But the rename from a generic clear to the explicit "if inactive" wording now makes the semantics look intentional across all cases, which they aren't.

**Suggested action:** either document the "server-death still leaks" gap in a follow-up comment right here, or add a bounded-staleness fallback (timestamp on the registry entry + age check on reconnect) so a stale `DOWNLOADING` slot older than, say, 30s gets cleared even if it's nominally "active".

src/functions/fileTransferRegistry.ts
```diff
@@ -0,0 +1,86 @@
...
+];
+
+// Atoms are co-located with mutators here rather than in store/app.ts to avoid a circular
+// import: app.ts derives fileTransferProgressAtom from aggregateFileTransferProgress, while
+// setters need fileTransferRegistryAtom. Both are re-exported from store/app.ts.
+export const fileTransferRegistryAtom = atom<FileTransferRegistry>({});
+
+export const fileTransferProgressBySourceAtom = atomFamily((source: FileTransferSource) =>
+    atom((get) => get(fileTransferRegistryAtom)[source] ?? INACTIVE_FILE_TRANSFER_PROGRESS),
+);
```
aidemsined, line 44: The `jotai-family` package is landing in `package.json` and `LICENSE` for exactly one consumer: `RemoteSyncButton.tsx:37` reading `fileTransferProgressBySourceAtom(REMOTE_SYNC)`. `MLIR_UPLOAD` and `LOCAL_UPLOAD` don't have source-specific readers.

Jotai 2.x removed `atomFamily` from `jotai/utils`, so `jotai-family` is the official path if you want the family shape — but for one consumer the family adds a dependency where three named derived atoms wouldn't, and the family memoization overhead beats what's actually needed. Alternatives:

1. Three named derived atoms (`remoteSyncFileTransferProgressAtom` and the two currently-unused others when someone needs them). Zero deps, direct import.
2. A tiny `useFileTransferProgress(source)` selector hook over the registry atom (≈ 4 lines: `const registry = useAtomValue(fileTransferRegistryAtom); return registry[source] ≠? INACTIVE_FILE_TRANSFER_PROGRESS;`). Also zero deps.

**Suggested action:** happy with either alternative here; the family really only earns its keep once the source-specific reader pattern shows up in ≥ 3 places. If you're planning that expansion in a follow-up PR, worth calling out in the PR body so reviewers can weigh the dependency.

src/components/report-selection/RemoteSyncButton.tsx
```diff
@@ -33,7 +34,7 @@ const RemoteSyncButton = ({
     isDisabled,
     handleClick,
 }: RemoteSyncButtonProps) => {
-    const fileTransferProgress = useAtomValue(fileTransferProgressAtom);
+    const fileTransferProgress = useAtomValue(fileTransferProgressBySourceAtom(FileTransferSource.REMOTE_SYNC));
```
aidemsined, line 39: This narrowing is the right call for a source-scoped control — pre-PR the button re-rendered on any progress update including MLIR/LOCAL, and post-PR Jotai's `Object.is` bailout on the frozen `INACTIVE_FILE_TRANSFER_PROGRESS` singleton skips the button entirely for unrelated writes. Strict improvement on that axis.

The UX consequence worth confirming: when REMOTE_SYNC is empty/inactive but MLIR_UPLOAD is `UPLOADING`, `FileStatusOverlay` (aggregate reader) shows "MLIR upload" while this button's tooltip falls through to whatever `getTooltipContent` says without a live progress — typically "Syncing report folder…" or a stale label depending on `isSyncingReportFolder`. That's internally consistent for a source-scoped control, but diverges from the pre-PR aggregate-driven tooltip text. `tests/RemoteSyncButton.spec.tsx` locks the source-scoped behaviour in, but doesn't call out the divergence from aggregate.

**Suggested action:** confirm this is intentional product behaviour (per-control source scoping), and add a one-line comment right here noting that the button intentionally reads REMOTE_SYNC only, so the tooltip stays scoped to sync progress even when the aggregate overlay is showing something else.

tests/useMlirRemote.spec.tsx
```diff
@@ -129,6 +136,73 @@ describe('useMlirRemote progress lifecycle', () => {
...
+        clearFileTransferProgressForSource(FileTransferSource.LOCAL_UPLOAD);
+
+        onUploadProgress?.({ loaded: 10, total: 10 } as AxiosProgressEvent);
+        expect(getDefaultStore().get(fileTransferProgressAtom).status).toBe(FileStatus.PROCESSING);
+
+        deferred.resolve({ data: { status: ConnectionTestStates.OK } });
+        await uploadPromise;
+
+        expect(getDefaultStore().get(fileTransferProgressAtom)).toEqual(getInactiveFileTransferProgress());
+    });
```
aidemsined, line 171: This block adds two happy-path concurrency tests ("keeps MLIR progress active when another source is cleared" and, below, "keeps remote sync progress active when MLIR upload completes") — both good — but the error path is still uncovered for the registry slot.

The pre-existing test at `tests/useMlirRemote.spec.tsx:303-327` ("clears the pending rows and disables View when a server upload fails") mocks `postMock.mockRejectedValue(new Error('network down'))` and asserts `mlirFileResultsAtom` / `mlirRetryFilesAtom` / `mlirRetryServerAtom` clear — but is silent on whether the `MLIR_UPLOAD` slot itself clears. The `finally` in `uploadMlirFileToServer` (`src/hooks/useMlirRemote.tsx:125-131`) does the work; without an assertion it can regress unnoticed the moment someone refactors the `try/catch` shape.

**Suggested action:** add an error-path case alongside the two new tests here, or amend the existing `303-327` failure test with `expect(getDefaultStore().get(fileTransferRegistryAtom)[FileTransferSource.MLIR_UPLOAD]).toBeUndefined()`. `useLocal.tsx` upload variants have the same gap and don't have an error-path spec at all — worth a follow-up there too.

aidemsined (commented):
## General observations

- **Compile-time exhaustiveness for `FILE_TRANSFER_SOURCE_PRIORITY`.** Hand-maintained `readonly FileTransferSource[]` at `src/functions/fileTransferRegistry.ts:28-32`. A `Record<FileTransferSource, number>` shape, or `satisfies readonly FileTransferSource[]` plus a completeness check, would catch a future missing enum member at compile time instead of silently omitting a new source from aggregation.
- **Add an MLIR_UPLOAD demo button in Styleguide.** `src/routes/Styleguide.tsx:772-787` demos REMOTE_SYNC (×2) and LOCAL_UPLOAD but not MLIR_UPLOAD. First-class `FileTransferSource` without an overlay preview path.
- **Amend AGENTS.md / CONVENTIONS.md with the co-location exception.** Inline comments at `src/functions/fileTransferRegistry.ts:34-36` and `src/store/app.ts:30-31` document the circular-import rationale (Copilot's atom-placement thread resolution). AGENTS.md:138-140 still reads as unconditional; a one-line exception prevents the next atom from re-opening the same debate.
- **Optional: extract `src/store/fileTransferAtoms.ts`.** Author already offered this to satisfy strict AGENTS placement without the cycle. Tradeoff: one more module vs. atoms living beside their mutators.
- **Shared registry-update helper for the four setters.** `setFileTransferProgressForSource` / `clearFileTransferProgressForSource` / `clearFileTransferProgressForSourceIfInactive` / `clearAllFileTransferProgress` each repeat `getDefaultStore().set(fileTransferRegistryAtom, …)`. A private `mutateFileTransferRegistry(updater)` would DRY this; at current scale inline is fine.
- **Cross-link `INACTIVE_FILE_TRANSFER_PROGRESS` singleton and `getInactiveFileTransferProgress()` factory in one comment.** Both are intentional (factory for resets that mutate later, frozen singleton for referential stability in `fileTransferProgressBySourceAtom`), but the roles could be signposted in a single doc block at `src/functions/fileTransferRegistry.ts:12-24`.
- **Import `fileTransferProgressBySourceAtom` via `store/app.ts` barrel in `RemoteSyncButton.tsx`.** PR imports it directly from `functions/fileTransferRegistry.ts` (`RemoteSyncButton.tsx:10`) while other consumers use `store/app.ts`. Re-export already exists; going through the barrel keeps discovery consistent.
- **Extend `SocketProvider.spec.tsx` with one real-registry test.** Current spec mocks `clearFileTransferProgressForSourceIfInactive` — verifies call-site wiring, not end-to-end preservation of active progress. One test using the real registry (seed `DOWNLOADING`, emit `connect`, assert slot survives) would tighten the coverage.
- **Reference-stability test for `aggregateFileTransferProgress`.** `tests/fileTransferRegistry.spec.ts:57-59` proves the frozen singleton is returned for empty registries. A companion `.toBe` assertion for the "same active slot across unrelated writes" case would lock in the ref-stability property that lets `RemoteSyncButton` skip re-renders against future accidental spreads.
- **Overall:** the per-source refactor is correct, complete for #1679, and the test suite covers the primary isolation guarantees. No **[must]** items — the four **[should]** items above (aggregator FAILED contract, read-only signposting, `clearIfInactive` positive-guard tests, MLIR error-path registry assertion) are the ones worth landing before merge.

---

src/store/app.ts
dcblundell: Ignoring, not important.

src/functions/fileTransferRegistry.ts
dcblundell: Agreed — locked option (c): the overlay only tracks live work (`isActiveTransferStatus`); toasts / connection UI own errors. Documented the priority order and FAILED-skipping contract on `aggregateFileTransferProgress`, and added a `REMOTE_SYNC=FAILED + LOCAL_UPLOAD=UPLOADING` test so that contract cannot drift silently.

tests/fileTransferRegistry.spec.ts
dcblundell: Good catch. Added both cases: active `DOWNLOADING` is preserved, and `FAILED` is cleared (same as `FINISHED`).

src/components/report-selection/RemoteSyncButton.tsx
dcblundell: Confirmed intentional — the button is source-scoped to sync. Added a one-line comment noting that the tooltip reads `REMOTE_SYNC` only even when the aggregate overlay is showing another source.

tests/useMlirRemote.spec.tsx
dcblundell: Done — the existing failure-path test now also asserts `fileTransferRegistryAtom[MLIR_UPLOAD]` is cleared. Agree `useLocal` deserves the same coverage; leaving that as a follow-up.

src/functions/fileTransferRegistry.ts
dcblundell: Not important to justify.

src/libs/SocketProvider.tsx
dcblundell: Creating an issue for this.

dcblundell: Agreed this is a pre-existing gap, not a regression from this PR. Tracked as follow-up: https://github.com/tenstorrent/ttnn-visualizer/issues/1757

PR #1751, opened 2026-07-10
Performance table heuristic flags 

### LINKED ISSUE #1519: Perf table: Heuristic flags per op ("why is this slow?")
## Context

Users currently have to interpret raw perf numbers to figure out *why* an op is expensive. The data we already collect (utilization, DRAM/L1 bytes, tensor shapes, op kind) is enough to produce cheap, useful heuristic flags.

## Ask

Tag each perf-table row with one or more heuristic flags surfacing likely cost drivers, e.g.:

- **DRAM-bound** — high DRAM traffic vs duration
- **Low L1 utilization** — small fraction of L1 active during op
- **Padding-heavy** — large gap between logical shape and tile-aligned shape
- **Recompute candidate** — same op shape/dtype computed multiple times
- **Underutilized cores** — op active on << total cores

Flags should be cheap to compute, easy to extend, and rendered as compact chips/icons in the row with hover explainers.

## Scope

Companion to / scoping issue for #1186 ("Recommending improvements based on the data"). This is the *flagging layer*; #1186 is the broader recommendation surface.

## Open questions

- Which heuristics ship in v1? (Suggest: DRAM-bound + Low utilization — cheap and high-signal)
- Thresholds: hardcoded defaults vs user-tunable?
- False-positive surfacing: how do we let the user dismiss / mark as expected?
- Should flags also propagate to op graph (color/badge) and op list?

## Related

- #1186 Recommending improvements based on the data
- #580 Utilization calculations only work for Matmul/Conv (limits the "Low utilization" flag for other ops)

dcblundell:
Adds a new column where known issues can be flagged.

Screenshot 2026-07-10 at 15 35 00

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1519.


src/components/performance/PerfTable.tsx
```diff
@@ -42,7 +43,7 @@ interface PerformanceTableProps {
     provideMatmulAdvice: boolean;
     hiliteHighDispatch: boolean;
     reportName: string | null;
-    showHashColumn: boolean;
+    heuristicContext: PerfHeuristicContext;
```
aidemsined: Requiring the full `PerfHeuristicContext` here (and on `PerfReport`/`StackedPerfTable`) overstates what the table needs: it only renders `row.heuristicFlags` and uses `heuristicContext.maxCores` once. Consider building + annotating once in `Performance.tsx` and passing `maxCores: number` down instead — it drops the `defaultHeuristicContext` test boilerplate and makes it clear that flags already live on the row rather than being recomputed here.

src/components/performance/PerfDeviceArchitecture.tsx
```diff
@@ -15,13 +15,14 @@ const NO_META_DATA = 'Unknown';
...
+    maxCores?: number;
 }
 
-const PerfDeviceArchitecture = ({ data, reportName }: PerfDeviceArchitectureProps) => {
+const PerfDeviceArchitecture = ({ data, reportName, maxCores: maxCoresFromParent }: PerfDeviceArchitectureProps) => {
     const { data: deviceMeta, isLoading: isLoadingDeviceLog } = usePerfMeta(reportName);
 
     const architecture = deviceMeta?.architecture ?? DeviceArchitecture.WORMHOLE;
-    const maxCores = deviceMeta?.max_cores ?? getCoreCount(architecture, data) ?? 0;
+    const maxCores = maxCoresFromParent ?? buildPerfHeuristicContext(deviceMeta, data).maxCores;
```
aidemsined: This makes a display component import the flag-compute module just for footer metadata — a display → compute dependency inversion. `buildPerfHeuristicContext` is really device-capacity resolution (`deviceMeta` → `getCoreCount` → `DEFAULT_MAX_CORES`); consider moving it next to `getCoreCount.ts` (e.g. `resolveMaxCores`) so both this footer and the route can share it without pulling in heuristics.

src/routes/Performance.tsx
```diff
@@ -118,16 +121,33 @@ export default function Performance() {
...
+    );
+
+    // Comparison rows reuse the active report's device meta for core thresholds.
+    // Assumes comparison reports target the same device architecture as the active report.
+    const enrichedComparisonData = useMemo(() => {
+        return (
+            comparisonPerfData?.map((dataset) => {
+                const comparisonTypedRows = enrichRowData(dataset, opIdsMap, null);
+
+                return annotatePerfHeuristicFlags(comparisonTypedRows, perfHeuristicContext);
```
aidemsined: Comparison rows are thresholded with the *active* report's `maxCores`. On a cross-architecture compare (e.g. Wormhole 64 vs Blackhole 130), `getCoreUtilization` under-counts → false LOW_UTILISATION, and `cores / maxCores` over-counts → missed UNDERUTILISED_CORES. The comment above acknowledges the same-arch assumption; consider building per-dataset context, or surfacing a UI caveat when the comparison device meta differs.

src/functions/computePerfHeuristicFlags.ts
```diff
@@ -0,0 +1,162 @@
...
+
+export function getPerfHeuristicFlagTooltipDetail(
+    flag: PerfHeuristicFlag,
+    row: TypedPerfTableRow,
+    context: PerfHeuristicContext,
+): string | null {
+    return evaluateRowHeuristics(row, context).details?.[flag] ?? null;
+}
+
+export function annotatePerfHeuristicFlags(
```
aidemsined: `annotatePerfHeuristicFlags` mutates each row in place and returns the same array reference, so `enrichedData` never changes identity when re-annotated. It's correct today only because the sort/filter path passes row references through and `formatCell` reads at render time — but wrapping `PerfTable` in `React.memo` on `data`, or deep-cloning rows in sort/filter, would silently leave flags stale. Prefer returning new row objects.

src/functions/computePerfHeuristicFlags.ts
```diff
@@ -0,0 +1,162 @@
...
+        flags,
+        details: flags.length > 0 ? details : undefined,
+    };
+}
+
+export function computePerfHeuristicFlags(row: TypedPerfTableRow, context: PerfHeuristicContext): PerfHeuristicFlag[] {
+    return evaluateRowHeuristics(row, context).flags;
+}
+
+export function getPerfHeuristicFlagTooltipDetail(
```
aidemsined: `getPerfHeuristicFlagTooltipDetail` (and `computePerfHeuristicFlags`) no longer have production call sites now that tooltip text is pre-annotated onto `row.heuristicFlagDetails` — this path re-runs `evaluateRowHeuristics`. Consider removing/un-exporting and keeping coverage through `annotatePerfHeuristicFlags`, so there aren't two parallel code paths that can drift.

src/definitions/PerfTable.ts
```diff
@@ -234,16 +240,16 @@ export const L1PressureColumns: ColumnDefinition[] = [
...
 const L1_PRESSURE_INSERTION_POINT = 2;
-const HIGH_DISPATCH_INSERTION_POINT = 5;
-const CACHE_HIT_INSERTION_POINT = 15;
+const HIGH_DISPATCH_INSERTION_POINT = 6;
 
 export const LOCKED_PERF_COLUMN_KEYS: ColumnKeys[] = [ColumnKeys.Id, ColumnKeys.OpCode];
 
+export const DISPLAY_COLUMNS_LABEL = 'Display columns';
+
 export interface EligiblePerfColumnsFlags {
```
aidemsined, line 250: Dropping `showHashColumn` from `EligiblePerfColumnsFlags` makes Hash (and Cache Hit) always eligible. Previously `PerfReport` always passed `showHashColumn={false}`, so Hash was never user-visible — this is a behavior change bundled into a flags PR. Please confirm it's intentional for #1519 (recompute investigation) rather than incidental.

src/functions/computePerfHeuristicFlags.ts
```diff
@@ -0,0 +1,162 @@
...
+
+export function buildPerfHeuristicContext(
+    deviceMeta: DeviceMetaLike | null | undefined,
+    rows: TypedPerfTableRow[],
+): PerfHeuristicContext {
+    const architecture = deviceMeta?.architecture ?? DeviceArchitecture.WORMHOLE;
+    const maxCores = deviceMeta?.max_cores ?? getCoreCount(architecture, rows);
+
+    return {
+        maxCores: maxCores > 0 ? maxCores : DEFAULT_MAX_CORES,
```
aidemsined: Before `usePerfMeta` resolves, this fallback runs at `DEFAULT_MAX_CORES = 64`, so on a real Blackhole report the first annotation is computed at 64 cores and then re-annotated when device meta lands — advisory chips can flicker/mis-flag on first paint. Minor for a heuristic, but worth noting alongside the in-place re-annotation.

src/components/performance/PerfDeviceArchitecture.tsx
```diff
@@ -15,13 +15,14 @@ const NO_META_DATA = 'Unknown';
 interface PerfDeviceArchitectureProps {
     data: TypedPerfTableRow[];
     reportName: string | null;
+    maxCores?: number;
```
aidemsined: `maxCores` is optional here with a `buildPerfHeuristicContext` fallback, while `PerfTable` requires `heuristicContext`. In the happy path the fallback is dead, but a standalone/test mount can annotate rows with one `maxCores` and render the footer with another. Recommend one policy (propagate a required `maxCores` from the route, or a single shared optional resolver) plus consistent naming across `perfHeuristicContext` / `heuristicContext` / `maxCores`.

src/definitions/PerfTable.ts
dcblundell: Intentional.

src/functions/computePerfHeuristicFlags.ts
dcblundell: Acknowledged — deferring for this PR. First paint can still annotate at the Wormhole/`DEFAULT_MAX_CORES` fallback before `usePerfMeta` lands; fine for an advisory heuristic here. Can follow up later (e.g. defer annotation until meta settles) if the flicker shows up in practice.

src/components/performance/PerfTable.tsx
dcblundell: Done — `PerfTable` / `PerfReport` now take `maxCores: number` (plus `comparisonMaxCores` for compare tabs). Annotation stays in `Performance.tsx`.

src/components/performance/PerfDeviceArchitecture.tsx
dcblundell: Done — extracted `resolveMaxCores` into `getCoreCount.ts`. `PerfDeviceArchitecture` no longer imports the heuristics module.

src/routes/Performance.tsx
dcblundell: Done — each comparison dataset is annotated with its own `resolveMaxCores` via `usePerfMetas` (per-report device meta when available, else row/architecture fallback).

src/functions/computePerfHeuristicFlags.ts
dcblundell: Done — `annotatePerfHeuristicFlags` now returns new row objects instead of mutating in place.

src/functions/computePerfHeuristicFlags.ts
dcblundell: Done — removed `computePerfHeuristicFlags` / `getPerfHeuristicFlagTooltipDetail`; coverage goes through `annotatePerfHeuristicFlags` only.

src/components/performance/PerfDeviceArchitecture.tsx
dcblundell: Done — single policy: required `maxCores` from the route everywhere (footer included). Dropped the optional fallback / mixed naming.

PR #1755, opened 2026-07-13
Operation duration histogram

### LINKED ISSUE #1585: Performance: Op-duration histogram with click-to-filter
## Context

The perf table answers "which ops are slow" by sort-desc. It doesn't easily answer the next question: *what's the shape of the distribution?* A handful of dominant ops vs a long tail of medium-cost ops vs many tiny ops are very different optimisation targets, but they all look similar in a sorted table.

## Ask

Add an **op-duration histogram** to the performance view:

- X-axis: log-scale op duration (so sub-microsecond and millisecond ops are both visible).
- Y-axis: op count per bucket.
- Hover a bar → see bucket range + sample ops.
- Click a bar → filter the perf table to ops in that bucket (or shift-click for range).
- Stack / colour bars by op kind so the user can see "the long tail is mostly matmuls".

Answers *"where's the long tail?"* in one glance.

## Open questions

- Auto-bucketing strategy: fixed log-decade bins vs adaptive (Freedman-Diaconis etc.)? Fixed log-decade is cheaper and probably good enough.
- Default colouring: by op kind (informative but noisy) vs single colour (cleaner but loses one signal). Tabs?
- Placement: dedicated chart panel on the perf page, or modal opened from a toolbar button?
- Multi-device: aggregate vs per-device tabs?

## Related

- #1519 Perf table heuristic flags (long-tail ops are often where the cheap heuristics fire)
- #1516 Operation list: inline perf sparkline (per-op view; this is the population view)
- #1582 Report onboarding: health-check summary (could surface "long tail is X% of total time" as one of the bullets)

dcblundell:
New chart showing the op distribution by duration. 

Screenshot 2026-07-14 at 14 01 30

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1585.


src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -0,0 +1,118 @@
...
+                x: bucketLabels,
+                y: histogramData.buckets.map((bucket) => {
+                    const segment = bucket.segmentsByOpCode.find((entry) => entry.rawOpCode === rawOpCode);
+                    return segment?.count ?? 0;
+                }),
+                type: 'bar',
+                name: rawOpCode,
+                customdata: customData as unknown as PlotData['customdata'],
+                hovertemplate:
+                    '%{x}<br />%{fullData.name}: %{y}<br />Bucket total: %{customdata.totalCount}<extra></extra>',
```
aidemsined: The `hovertemplate` only references `%{customdata.totalCount}` — the per-bucket `sampleOps` are computed (a full sort + `slice`) and threaded into `customData` but never shown, so the "per-bucket sampling for hover" the PR highlights is dead computation. Either extend the hover to render the sampled op names or drop the sampling machinery.

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -0,0 +1,118 @@
...
+        },
+        yAxis: {
+            title: {
+                text: 'Op count',
+            },
+        },
+    };
+
+    if (histogramData.buckets.length === 0) {
+        return <p className='perf-duration-histogram-empty'>No device ops available for duration histogram.</p>;
```
aidemsined: The empty state returns a bare `<p>` and never mounts `PerfChartFrame`, so the anchor node `#perf-chart-op-duration-histogram` is missing. Since this chart is first in `FILTERABLE_CHART_ENTRIES` and `useActiveSection` anchors off `document.getElementById(id)`, jump-to-chart and scroll-spy break on reports with no eligible ops. Match the sibling charts: always render the frame with `id={PerfChartId.OpDurationHistogram}` and put the empty copy inside it.

src/components/performance/PerfCharts.tsx
```diff
@@ -12,14 +13,26 @@ interface PerfChartsProps {
     filteredPerfData: TypedPerfTableRow[];
     comparisonData?: TypedPerfTableRow[][];
     selectedOpCodes: Marker[];
+    opCodeOptions: Marker[];
```
aidemsined: `Performance.tsx` already syncs `selectedOpCodes` from `opCodeOptions` on load, and `PerfOpCountVsRuntimeChart` colours its traces via `selectedOpCodes.find(...)?.colour` with no extra prop. Threading `opCodeOptions` through `PerformanceChartsTab` → `PerfCharts` widens the aggregate seam for a single consumer — consider colouring from the existing `selectedOpCodes` instead.

src/functions/buildDurationHistogram.ts
```diff
@@ -0,0 +1,106 @@
...
+                bucketIndex: 0,
+                minUs: 0,
+                maxUs: 1,
+                label: formatDurationBucketRange(0, 1),
+            },
+        ];
+    }
+
+    const minExponent = Math.floor(Math.log10(Math.min(...positiveDurations)));
+    const maxExponent = Math.ceil(Math.log10(Math.max(...positiveDurations)));
```
aidemsined: `Math.ceil(Math.log10(max))` overshoots: a value `v` belongs in decade `floor(log10(v))`, so whenever `max` isn't an exact power of 10 (the common case) this produces a top bucket `[10^maxExp, 10^(maxExp+1))` that no row can land in — the chart nearly always shows a spurious empty high bucket. Use `floor(log10(max))` for the span. (The existing test only asserts `length >= 2` / total, so it slips through.)

src/functions/buildDurationHistogram.ts
```diff
@@ -0,0 +1,106 @@
...
+    DurationBucket,
+    DurationHistogramBucket,
+    DurationHistogramData,
+    SAMPLE_OPS_PER_BUCKET,
+} from '../definitions/PerfDurationHistogram';
+import { OpType } from '../definitions/Performance';
+import { TypedPerfTableRow } from '../definitions/PerfTable';
+import { formatDurationBucketRange } from './formatDurationBucketRange';
+
+const isEligibleHistogramRow = (row: TypedPerfTableRow): boolean =>
```
aidemsined, line 20: `isEligibleHistogramRow` only rejects null/undefined, so negative (and `NaN`) `device_time` rows stay eligible. They're excluded from `positiveDurations`, but `findBucketForDuration` matches no bucket and falls through to `?? buckets[buckets.length - 1]` — i.e. a `-5µs` op is charted in the *slowest* bucket, the opposite of intent. Consider rejecting non-positive durations here (or routing them explicitly).

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -0,0 +1,118 @@
...
+function PerfDurationHistogram({ id, title, rows, opCodeOptions }: PerfDurationHistogramProps) {
+    const histogramData = useMemo(() => buildDurationHistogram(rows), [rows]);
+
+    const bucketLabels = useMemo(
+        () => histogramData.buckets.map((bucket) => bucket.bucket.label),
+        [histogramData.buckets],
+    );
+
+    const colourByOpCode = useMemo(
+        () => new Map(opCodeOptions.map((marker) => [marker.opCode, marker.colour])),
```
aidemsined: `MAX_LEGEND_OP_CODES` only toggles `showLegend`; the trace builder still emits one stacked trace per distinct `raw_op_code` with no "other"/rollup. Beyond ~12–30 op codes (realistic) the bars become unreadable, palette colours recycle, and any op code missing from this colour map falls back to Plotly auto-colour (non-deterministic). Consider capping rendered traces / grouping the tail into an "other" segment.

src/definitions/PerformanceCharts.ts
```diff
@@ -56,6 +58,10 @@ export interface PerfChartIndexEntry {
 }
 
 export const FILTERABLE_CHART_ENTRIES: PerfChartIndexEntry[] = [
+    {
+        id: PerfChartId.OpDurationHistogram,
```
aidemsined, line 62: This chart is registered as the first `FILTERABLE_CHART_ENTRIES` entry, but the component has no `onOpCodeClick` / `useHandlePerfChartPlotClick` wiring that the other filterable charts gained — clicking a segment doesn't prefilter the table. Either add click-to-filter for parity or move it out of the filterable group and document it as read-only.

src/functions/buildDurationHistogram.ts
```diff
@@ -0,0 +1,106 @@
...
+            const opCode = row.raw_op_code ?? 'unknown';
+            const opRows = countByOpCode.get(opCode) ?? [];
+            opRows.push(row);
+            countByOpCode.set(opCode, opRows);
+        });
+
+        const segmentsByOpCode = [...countByOpCode.entries()].map(([rawOpCode, opRows]) => ({
+            rawOpCode,
+            count: opRows.length,
+            sampleOps: [...opRows]
```
aidemsined: Segments are stored as an array and the chart builder does repeated `segmentsByOpCode.find(...)` per (op code × bucket), which is O(opCodes × buckets × segments). Building a `Map<rawOpCode, segment>` per bucket here would let the chart assembly do O(1) lookups instead. Immaterial below a few thousand ops, but cheap to fix.

src/components/performance/PerfCharts.tsx
```diff
@@ -12,14 +13,26 @@ interface PerfChartsProps {
...
+    opCodeOptions,
+    onOpCodeClick,
+}: PerfChartsProps) => {
     const data = [filteredPerfData, ...(comparisonData || [])].filter((set) => set.length > 0);
 
     return (
         <>
+            <PerfOpDurationHistogramChart
+                rows={filteredPerfData}
+                opCodeOptions={opCodeOptions}
```
aidemsined: Sibling charts here build `data = [filteredPerfData, ...comparisonData]`, but the histogram is passed only `filteredPerfData` — so comparison reports silently drop out of this chart. Either facet/stack the comparison distribution for consistency, or document single-report scope in the chart title/label.

src/components/performance/PerfOpDurationHistogramChart.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { PERF_CHART_LABELS, PerfChartId } from '../../definitions/PerformanceCharts';
+import PerfDurationHistogram from './PerfDurationHistogram';
+import 'styles/components/PerfDurationHistogram.scss';
+
+interface PerfOpDurationHistogramChartProps {
+    rows: TypedPerfTableRow[];
+    opCodeOptions: Marker[];
+}
+
+function PerfOpDurationHistogramChart({ rows, opCodeOptions }: PerfOpDurationHistogramChartProps) {
```
aidemsined: This wrapper only forwards constants (`id`, `title`), adds a margin wrapper, and duplicates the SCSS import. Sibling filterable charts (`PerfOpCountVsRuntimeChart`, `PerfDeviceKernelDurationChart`) are single components that bind `PerfChartId`/`PERF_CHART_LABELS` and call `PerfChart` directly. Consider collapsing this into the chart component to match the convention (keep `buildDurationHistogram`/`formatDurationBucketRange` as the pure testable layer).

src/functions/buildDurationHistogram.ts
```diff
@@ -0,0 +1,106 @@
...
+            {
+                bucketIndex: 0,
+                minUs: 0,
+                maxUs: 1,
+                label: formatDurationBucketRange(0, 1),
+            },
+        ];
+    }
+
+    const minExponent = Math.floor(Math.log10(Math.min(...positiveDurations)));
```
aidemsined: `Math.min(...positiveDurations)` / `Math.max(...positiveDurations)` spread the full array as function args. Fine for typical reports, but at very high op counts (~100k+) V8 throws `RangeError: Maximum call stack size exceeded`. A `reduce` avoids the spread if unbounded input is possible.

src/components/performance/PerfDurationHistogram.tsx
aidemsined, line 1: This chart inlines a `PlotConfiguration` with a custom `margin.b` and `barMode: 'stack'`, while #1758 (in review in parallel) centralizes chart layout/config. Whoever merges second should adopt the shared layout defaults and re-run `PerfChart.spec.tsx` so this chart doesn't get left on the pre-consolidation path.

tests/buildDurationHistogram.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+const row = (overrides: Partial<TypedPerfTableRow>): TypedPerfTableRow =>
+    ({
+        op_type: OpType.DEVICE_OP,
+        op_code: 'Matmul',
+        raw_op_code: 'Matmul',
+        device_time: 1,
+        ...overrides,
+    }) as TypedPerfTableRow;
+
+describe('buildDurationHistogram', () => {
```
aidemsined, line 20: Coverage gaps worth adding: empty / all-ineligible input → `buckets: []` (guards the empty state); exact decade boundaries (values `1`, `10`, `100` under half-open `[minUs, maxUs)`); `SAMPLE_OPS_PER_BUCKET` overflow (6+ rows → length 5, longest kept); and multi-op-code segmentation counts.

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -0,0 +1,118 @@
...
+import PerfChart from './PerfChart';
+import 'styles/components/PerfDurationHistogram.scss';
+
+interface DurationHistogramPointMeta {
+    totalCount: number;
+    sampleOps: string[];
+}
+
+interface PerfDurationHistogramProps {
+    id: string;
```
aidemsined: `id` is typed as `string` while every sibling passes a `PerfChartId` enum member. Typing it as `PerfChartId` keeps the coupling that `PerformanceCharts.ts` and scroll-spy rely on.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Fixed — samples are now shown in the hover via `%{customdata[2]}`.

src/functions/buildDurationHistogram.ts
```diff
@@ -0,0 +1,106 @@
...
+        const minUs = exponent === minExponent && hasNonPositive ? 0 : 10 ** exponent;
+        const maxUs = 10 ** (exponent + 1);
+
+        return {
+            bucketIndex: offset,
+            minUs,
+            maxUs,
+            label: formatDurationBucketRange(minUs, maxUs),
+        };
+    });
```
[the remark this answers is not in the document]
dcblundell: Agreed — switched min/max to a loop (no spread), and non-positive / non-finite durations are now excluded from eligibility instead of widening the first decade bucket (which also stops negatives falling through to the slowest bucket).

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { PlotConfiguration } from '../../definitions/PlotConfigurations';
+import { TEST_IDS } from '../../definitions/TestIds';
+import buildDurationHistogram from '../../functions/buildDurationHistogram';
+import PerfChart from './PerfChart';
+import 'styles/components/PerfDurationHistogram.scss';
+
+interface DurationHistogramPointMeta {
+    totalCount: number;
+    sampleOps: string[];
+}
```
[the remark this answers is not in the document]
dcblundell: Done — hover now uses a `[rawOpCode, totalCount, sampleOpsSummary]` tuple; the old object meta type is gone.

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return (
+        <div
+            className='perf-duration-histogram'
+            data-testid={TEST_IDS.PERF_DURATION_HISTOGRAM}
+        >
+            <PerfChart
+                id={id}
+                title={title}
+                chartData={chartData}
+                configuration={configuration}
```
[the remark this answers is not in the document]
dcblundell: Agreed. Wired click-to-filter via `onOpCodeClick` / `useHandlePerfChartPlotClick` so a stacked segment filters the table by that raw op code (parity with the other filterable charts). Full bucket-range / shift-click filtering from #1585 still needs a duration-range table filter and is left as follow-up — happy to switch the PR body from Closes to Refs if we want to keep #1585 open for that.

src/components/performance/PerfOpDurationHistogramChart.tsx
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { Marker, TypedPerfTableRow } from '../../definitions/PerfTable';
+import { PERF_CHART_LABELS, PerfChartId } from '../../definitions/PerformanceCharts';
+import PerfDurationHistogram from './PerfDurationHistogram';
+import 'styles/components/PerfDurationHistogram.scss';
```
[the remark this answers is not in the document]
dcblundell: Removed — collapsed the wrapper into `PerfDurationHistogram`, so the stylesheet is imported once.

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -0,0 +1,118 @@
...
+                    return segment?.count ?? 0;
+                }),
+                type: 'bar',
+                name: rawOpCode,
+                customdata: customData as unknown as PlotData['customdata'],
+                hovertemplate:
+                    '%{x}<br />%{fullData.name}: %{y}<br />Bucket total: %{customdata.totalCount}<extra></extra>',
+                marker: {
+                    color: colourByOpCode.get(rawOpCode),
+                },
```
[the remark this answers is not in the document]
dcblundell: Fixed — hovertemplate now uses `%{customdata[1]}` / `%{customdata[2]}` and surfaces the sampled op names.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Fixed — empty state always mounts `PerfChartFrame` with `id={PerfChartId.OpDurationHistogram}` so jump-to-chart / scroll-spy keep working.

src/components/performance/PerfCharts.tsx
dcblundell: Agreed — colouring now comes from `selectedOpCodes` like `PerfOpCountVsRuntimeChart`; dropped the extra `opCodeOptions` seam through `PerfCharts`.

src/functions/buildDurationHistogram.ts
dcblundell: Fixed — top decade now uses `floor(log10(max))`, with a regression test that asserts no empty high bucket for a typical max.

src/functions/buildDurationHistogram.ts
dcblundell: Fixed — eligibility now requires a finite `device_time > 0`, so negatives / NaN / zero are excluded rather than mis-bucketed.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Agreed this is still a real gap. Left as follow-up for this PR — legend hide remains the current guard; capping traces / rolling into “Other” can land next.

src/definitions/PerformanceCharts.ts
dcblundell: Fixed — wired `onOpCodeClick` so segment clicks prefilter the table (same path as the sibling filterable charts). Bucket-range click from #1585 remains follow-up.

src/functions/buildDurationHistogram.ts
dcblundell: Fixed in the chart builder — segments are mapped once per bucket (`Map<rawOpCode, segment>`) so assembly is O(1) lookups.

src/components/performance/PerfCharts.tsx
dcblundell: Agreed — left deliberately single-report for now (comment in `PerfCharts`). Comparison faceting deferred rather than silently implying parity with the other charts.

src/components/performance/PerfOpDurationHistogramChart.tsx
dcblundell: Done — collapsed `PerfOpDurationHistogramChart` into `PerfDurationHistogram`, which now binds `PerfChartId` / label itself.

src/functions/buildDurationHistogram.ts
dcblundell: Fixed — min/max no longer use a spread; they iterate with a loop.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Noted — will adopt the shared layout from #1758 when that lands / on merge conflict cleanup. Leaving the local `PlotConfiguration` for now.

tests/buildDurationHistogram.spec.ts
dcblundell: Added coverage for empty/all-ineligible input, exact decade boundaries, sample-cap overflow, and multi-op-code segmentation.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Done — chart now owns `PerfChartId.OpDurationHistogram` internally; no string `id` prop.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Follow-up done: when there are more than `MAX_LEGEND_OP_CODES` distinct codes we keep the top N−1 by total count and roll the tail into a single **Other** trace (stable grey). Clicking Other does not prefilter (empty `customdata[0]`).

src/components/performance/PerfCharts.tsx
dcblundell: Follow-up done: still single-report for the histogram, but when comparison data is present the chart now shows an **Active report only** subtitle so scope is not silent.

Issue #1747, opened 2026-07-08
MLIR viewer: worker protocol hygiene (trim WorkerInteractionIndex + dead type fields + versioning)


aidemsined:
MLIR audit follow-up (#1734) — **Group I**. Protocol-surface hygiene.

## Scope

1. **Trim `WorkerInteractionIndex`** to fields the main thread actually reads.
    - `outerNamespaceByNodeId` is shipped from worker to view (`src/components/mlir/mlirLayoutWorker.ts:118-125`) but **never read on the main thread** — only mentioned in a code comment at `src/components/mlir/MLIRViewReactFlow.tsx:922`. Drop from wire payload (keep in builder-local `GraphIndex` if still used there).
    - `anchorNamespaceByNodeId` — verify usage on main thread; drop if similarly unread.
2. **Add explicit `subgraphNamespaces: string[]`** to `WorkerInteractionIndex` rather than deriving from `Object.keys(anchorByNamespace)` at `MLIRViewReactFlow.tsx:781-784`. Document invariant: `keys(anchorByNamespace) === set(subgraphNamespaces)`.
3. **Remove dead type fields**:
    - `WorkerNode.dragHandle` at `src/components/mlir/mlirGraphTypes.ts:95` — declared, never assigned repo-wide.
    - `WorkerEdge.labelShowBg` at `:124` — declared, never assigned.
    - `WorkerEdge.labelBgStyle` at `:125` — declared, never assigned.
    - `WorkerBuiltMessage.cacheKey` echoed on reply but never read by the hook.
4. **Add `protocolVersion: 1` field** to `SetGraphMessage`. Worker rejects unknown versions. Forward-compat currently relies on implicit deploy lockstep between the app bundle and the worker chunk.

## Acceptance criteria

- Type file loses the four dead fields.
- `WorkerInteractionIndex` shrinks by unused entries; wire payload verified via structured-clone size measurement.
- `subgraphNamespaces` is an explicit contract.
- `protocolVersion` field added; worker logs a warning on mismatch.
- All existing tests + a new type-level test that the removed fields are not accessible anywhere.

## Effort

~0.5–1 engineer-day.

## Prerequisite

Bundle with Group L (worker spec) if landing that concurrently.

Issue #1741, opened 2026-07-08
MLIR viewer: extract MlirOpNode + MlirGroupNode + region-partner maps


aidemsined:
MLIR audit follow-up (#1734) — **Group C**. Composer decomposition (round one).

## Scope

1. **Extract `MlirOpNode.tsx` and `MlirGroupNode.tsx`** from `src/components/mlir/MLIRViewReactFlow.tsx:109-339` + their contexts (`MlirNodeBodyContext`, `MlirGroupContext`). Cuts the composer by ~230 LOC and isolates per-node re-render analysis. Sibling files, same folder.
2. **Extract `buildRegionPartnerMaps(interactionIndex, expandedNamespaces)` as a pure function** with a spec. Currently in-view at `MLIRViewReactFlow.tsx:928-961` — re-derived on every `expandedNamespaces` change, untested. Region-partnership facts already exist in the worker's `WorkerInteractionIndex`; this extraction just makes the view-side inversion testable in isolation.

## Acceptance criteria

- `MlirOpNode.tsx` + `MlirGroupNode.tsx` sibling files with their contexts and props types.
- `buildRegionPartnerMaps` with a `tests/mlirRegionPartnerMaps.spec.ts` covering the empty / expanded / nested / synthetic-section cases.
- No behavioural change.

## Effort

~1.5 engineer-days.

## Prerequisite

Group B item 3 (`resolveVisibleRepId`) is a soft prereq — trivial to unblock if needed.

Issue #1740, opened 2026-07-08
MLIR viewer: extract pure-function modules + unify tensor formatting


aidemsined:
MLIR audit follow-up (#1734) — **Group B**. Independent, low-risk, high test-value.

Small pure-function extractions from `MLIRViewReactFlow.tsx` (each closes a specific untested regression risk) plus consolidation of three parallel tensor formatters.

## Scope

1. **Extract `computeDisplayedEdges` and `computeFilterMatchInfo` as pure modules with characterization tests.** Currently at `src/components/mlir/MLIRViewReactFlow.tsx:1241-1351` (132 LOC) and `:644-694`. Each is trapped inside a `useMemo` closure with no direct spec. Extract to `src/components/mlir/mlirDisplayedEdges.ts` and `src/components/mlir/mlirFilterMatch.ts`; add characterization specs alongside the existing `tests/mlirGraphBuilder.spec.ts`. Highest untested regression surface on the subsystem.
2. **Extract `useMlirViewportBaton` hook.** Currently: refs at `MLIRViewReactFlow.tsx:374-387`, consumer at `:475-522`, arming at `:738-801, 1187-1217`. Bundle `viewportAnchorRef` / `pendingFocusNodeIdRef` / `pendingFitAllRef` / `hasFitInitiallyRef` + `applyBuiltGraph` consumer into a hook returning `{ armAnchor, armFocus, armFitAll, consumeAfterBuild }`. Unlocks direct testability of the mutual-clear discipline that landed in #1733.
3. **Extract shared `resolveVisibleRepId(nodeId, expandedNamespaces, index)`.** Currently duplicated: private helper in `src/components/mlir/mlirGraphBuilder.ts:238-244`, re-implementation at `MLIRViewReactFlow.tsx:669-678` (including the `?? source.id` fallback branch). Export from `mlirGraphHelpers.ts` (or a new `mlirVisibility.ts`) so filter badges + buried-match counts + builder edge emission all use one function.
4. **Unify tensor formatting.** Three parallel formatters render subtly different strings for the same port:
    - `getTensorInfoFromAttrs` in `src/components/mlir/mlirGraphBuilder.ts:154` (edge labels).
    - `compactPortShape` in `src/components/mlir/mlirNodeBodySummary.ts:43` (node-body overlay).
    - `formatCompactTensor` in `src/components/mlir/MlirPortAttrs.tsx:34` (details panel).
[...]

Issue #1734, opened 2026-07-08
MLIR viewer: code + architecture audit (post milestone-2)


aidemsined:
Landing the milestone-2 MLIR viewer arc (up to and including #1733) accumulated design decisions and deferred cleanups that deserve one dedicated pass rather than piecemeal follow-ups. This ticket collects the audit scope so the review is deliberate.

## Purpose

Re-examine the MLIR viewer end-to-end — code, architecture, and product decisions — now that the surface has stabilised. Goal is to surface pre-existing debt, cross-cutting inconsistencies, and next-milestone refactors before the milestone-3 work extends the surface further.

## Seeded scope (from PR #1733 review — @dcblundell)

Items triaged as out-of-scope for #1733 but worth revisiting in this pass:

- **`MLIRViewReactFlow.tsx` has no test file at all.** Every feature in the viewer has been landed with leaf-component unit tests only; the parent's state-machine (worker protocol, viewport batons, `applyBuiltGraph`, `filterMatchInfo`, `regionOutputPartnerByNodeId`, focus/select/expand/collapse handlers) is entirely uncovered. Design a testable seam (mock worker + `fitView` + React Flow) and characterise the current behaviour.
- **Post-rebuild stale-window between state flip and worker rebuild.** Callers set `expandedNamespaces` synchronously; memos keyed on it (`filterMatchInfo`, buried-match counts, `regionOutputPartnerByNodeId`, `inputArgIdxByArgIdByAnchor`) recompute against the previous build's node set until the worker reply lands. Consider deferring state flips until `onBuilt`, gating memos, or making the transient state explicit.
- **In-flight build state is not exposed.** Bulk toggles trigger a full `buildVisibleGraph` + `setNodes/setEdges` + animated `fitView` on the worst-case path with no user feedback. Consider a busy indicator, disabling the controls during in-flight builds, and/or `duration: 0` on bulk fits to skip the animation cost.
- **SCSS "floating panel" chrome is duplicated in three widget files** (`MlirOpFilter.scss`, `MlirNodeBodyToggles.scss`, `MlirExpandCollapseControls.scss`) — identical `background` / `border` / `border-radius` / `box-shadow`. Extract a `%mlir-graph-widget-pill` placeholder / mixin (see `CONVENTIONS.md` § "Same rule for magic layout numbers"). Applies repo-wide.
[...]

PR #1759, opened 2026-07-14
MLIR viewer: perf hot-path fixes (memo deps + context stability) (#1739)

### LINKED ISSUE #1739: MLIR viewer: perf hot-path fixes (memo deps + context stability)
MLIR audit follow-up (#1734) — **Group A / highest leverage**. Ship first.

All seven fixes are pure memo/context/dep-array refactor with no behavioural change. Together they close the top-of-list perf defects that make drag, filter, and toggle interactions feel laggy on realistic graphs (≥2K visible nodes).

## Scope

1. **`displayedEdges` LCA lift re-runs on every RF interaction frame.** Deps `[edges, nodes, selectedNodeId]` at `src/components/mlir/MLIRViewReactFlow.tsx:1351`. `nodes` is React Flow's mutable `useNodesState` array — mutates on every drag/select/hover. Cache the parent-chain map on structural node topology (mirror the `visibleOpNodeIds` cached-set pattern at `:576-590`); new deps `[edges, nodeParentChainById, selectedNodeId]`. Biggest single main-thread win.
2. **`moveGroup` mousemove is not RAF-batched.** `MLIRViewReactFlow.tsx:880-889` fired from manual `MlirGroupNode` mousemove at `:223-242`. Every event → full O(N) map allocation + downstream memo cascade. Coalesce mousemoves within a frame via `requestAnimationFrame`; use `useReactFlow().updateNode()` in place of full `setNodes(current.map(...))`.
3. **`cacheByGraphId` grows monotonically within a session on one graph.** `src/components/mlir/mlirLayoutWorker.ts:11-13`. Cross-graph eviction is handled by `key={graph.id}` remount, but every distinct expansion state visited within one graph stays. Add LRU cap (e.g. N=32) on `cacheByGraphId.get(graphId)`.
4. **`filterMatchInfo` re-runs on every RF interaction while a filter is active.** `MLIRViewReactFlow.tsx:644-694`. Deps include `nodes` at `:694`. Drop `nodes` from deps; use the already-stabilised `visibleOpNodeIds` set to check anchor visibility. Lift `buildFilterMatcher` (`src/components/mlir/mlirFilter.ts:22`) into `useMemo([filterMode, appliedFilterQuery])` so the RegExp is compiled once per query, not once per frame.
5. **Overlay toggle-flip fans out via `MlirNodeBodyContext`.** `MLIRViewReactFlow.tsx:1499-1502`. On toggle flip, `overlayLinesByNodeId` legitimately recomputes → new context value reference → every mounted `MlirOpNode` re-renders. Push per-node overlay strings through the RF node `data` payload (`data.shapeLines` / `data.locationLines`) so RF's per-node diff drives which nodes re-render, and drop the context.
6. **`toggleNamespaceFromGroup` closure over `nodes` churns `MlirGroupContext`.** `MLIRViewReactFlow.tsx:869-898`. Callback identity changes on every drag frame; `groupContextValue` identity churns; every mounted `MlirGroupNode` re-renders whenever any node mutates position. Swap `nodes.find(...)` for `useReactFlow().getNodes()` (stable getter) or a `latestNodesRef` synced in a passive effect; empty deps.
7. **Wrap widgets in `React.memo`.** `MlirNodeDetailsPanel`, `MlirOpFilter`, `MlirNodeBodyToggles`, `MlirExpandCollapseControls`. Parent already builds props via `useMemo`, so shallow-eq is honest. Cheap defensive win — parent commits stop cascading reconciliation into these widgets. Also memoise `tryParseAttrValue` inside `MlirAttrValue` on `[value]`.

## Acceptance criteria

- All 7 fixes land in one PR.
- Existing tests remain green.
- No user-visible behaviour change on graphs ≤ 500 visible nodes.
- Observable improvement on ≥ 2K visible-node graphs during drag / filter / toggle-flip.
- Add characterization tests where trivially applicable (drop-`nodes`-from-deps changes should be spec-verifiable).

## Effort

~3–4 engineer-days. Highest leverage-per-hour on the entire subsystem.

## Non-goals

- Structural refactor of `MLIRViewReactFlow.tsx` — that's Group C.
- Extraction of pure modules — that's Group B.
- Worker protocol changes — that's Group I.

See also: `~/Documents/repos/mlir-viewer-audit-run2.md` (audit report).

aidemsined:
Closes #1739. Group A of the MLIR audit follow-up (#1734) — highest leverage-per-hour perf pass. All changes are memo/context/dep-array refactor with no behavioural change; they close the top-of-list defects that make drag, filter, and toggle interactions lag on graphs with 2K+ visible nodes.

## What's in the PR

All seven items from the ticket, plus a follow-up commit from a local ultimate-review pass:

### `displayedEdges` (item 1) + topology derivation moved to the build boundary

Structural topology (`nodeTopologyById`) and the op-only visibility set (`visibleOpNodeIds`) are now derived inside `applyBuiltGraph` at the same time as `setNodes`, and stored in `useState`. The previous approach kept them as `useMemo([nodes])` sources with a computed-cached-state bail-out — that stabilised downstream references, but still re-walked every parent chain on every drag / select / hover frame (~3N + N·depth ops).

Structural fields (`parentId`, `type`, `groupKind`, `collapsedSubgraphNamespace`, `subgraphToggleState`) are worker-owned and never mutate outside a rebuild; RF's other write paths (selection reflection effect, `onNodesChange`, `updateNode`) all preserve them. `applyBuiltGraph` is the sole ingress for structural change, so the maps only need to update there.

Result: `displayedEdges`, `filterMatchInfo`, and `overlayLinesByNodeId` do zero work on drag / select / hover.

### `moveGroup` RAF-batched (item 2)

Mousemove deltas coalesce into a per-frame `Map<groupId, {dx, dy}>` and flush via one `updateNode` call per group per frame, replacing per-mousemove `setNodes(current.map(...))` cascades. xyflow's `updateNode` is `setNodes(prev.map(...))` internally, so per-flush cost against N nodes is unchanged — the win is write rate (mousemove cadence → 60 Hz RAF cadence).

`applyBuiltGraph` cancels any pending RAF and clears the pending Map at the top of the callback, so a mid-drag worker rebuild can't apply the accumulated delta against the freshly-committed canonical position ~16 ms later.

### Worker cache LRU (item 3)

`cacheByGraphId.get(graphId)` capped at 32 entries per graph, move-to-end LRU. The helper (`touchLruCache`) is extracted into `src/components/mlir/mlirLayoutWorkerCache.ts` and unit-tested in isolation (10 cases: hit-refresh-no-evict, insertion-order-under-eviction, limit=1 degenerate, monotonic cap, non-string keys, ...). A `touchGraphCache` wrapper inside the worker DRYs the two production callsites without narrowing the public helper.

### `filterMatchInfo` (item 4) + `resolveFilterMatches` extraction

`buildFilterMatcher` lifted into its own `useMemo([filterMode, appliedFilterQuery])` so the RegExp compiles once per query (not once per interaction frame). `filterMatchInfo` now depends on `visibleOpNodeIds` instead of `nodes`, and the resolution body is extracted into a pure `resolveFilterMatches` in `mlirFilter.ts`. Component-side memo just delegates; behaviour is byte-identical.

The extraction unlocks 8 spec cases in `tests/mlirFilter.spec.ts` — direct visibility, single-namespace bury, outer-preferred-over-inner, off-canvas anchor drops the match (locks the op-only visibility narrowing that made this whole item safe), inner-anchor after outer expand, source visible after full expand, coalesced counts on shared anchor, and missing-anchor fallback to source id.

### Overlay via `data.shapeLines` / `data.locationLines` (item 5)

`MlirNodeBodyContext` removed. Overlay strings ride on the RF node `data` payload and are read directly by `MlirOpNode`, so React Flow's per-node diff drives which ops re-render on toggle flip — no more full-canvas provider-value fanout.

### `toggleNamespaceFromGroup` (item 6)

Reads `nodes` via a `latestNodesRef` (passive-mirror pattern, synced by a `useEffect`) so the callback identity survives drag frames. `groupContextValue` stops churning; every mounted `MlirGroupNode` no longer re-renders whenever any node's position changes.

### `React.memo` on widgets (item 7)

`MlirNodeDetailsPanel`, `MlirOpFilter`, `MlirNodeBodyToggles`, `MlirExpandCollapseControls` wrapped in `memo`. `MlirOpFilter.displayName` on the memo wrapper for parity with the other three (React DevTools no longer shows a `Memo(...)` fallback). `MlirAttrValue` memoises `tryParseAttrValue(value)`. `onPrev` / `onNext` in the parent stabilised via `useCallback` so shallow-eq is honest.

## Testing

- **New unit tests** — 10 cases for `touchLruCache` (`tests/mlirLayoutWorkerCache.spec.ts`), 8 cases for `resolveFilterMatches` (added to `tests/mlirFilter.spec.ts`).
- **Regression coverage** — full MLIR suite green: **175 / 175 across 14 test files**.
- **Type + lint** — `tsc --noEmit` clean on touched files (the one pre-existing error in `tests/PerfOperationTypesChart.spec.tsx` is unrelated; tracked separately in #1731). `eslint` clean on all touched files.
- **Manual QA** — drag / filter / toggle flip on a 2K-visible-node fixture no longer stutters. No benchmark harness added; the audit called that out as a separate follow-up (out of Group A scope).

## Deferred (from local self-review)

- Export `MLNodeData` / `NodeTopologyEntry` to a shared view-types module — belongs in #1741 (`MlirOpNode` / `MlirGroupNode` extraction).
- `useLatestRef` helper — 2 callsites of the passive-mirror pattern in `MLIRViewReactFlow.tsx`, but the file is on the chopping block for #1741 so extraction is better handled during that split.
- Fixed the stale `nodeBodyContextValue` comment reference (removed as part of the item-1 refactor).

## Non-goals

Per #1739 — no structural refactor of `MLIRViewReactFlow.tsx` (Group C / #1741), no pure-module extractions beyond the LRU helper and `resolveFilterMatches` (Group B / #1740), no worker protocol changes (Group I / #1747).

### DIFF
--- a/src/components/mlir/MLIRViewReactFlow.tsx
+++ b/src/components/mlir/MLIRViewReactFlow.tsx
@@ -75,9 +75,14 @@ const NODE_BODY_OVERLAY_LINE_PX = 11;
 //    bleeds behind the dashed body and hides children.
 //  - `buriedMatchCount`: hidden filter matches under a collapsed anchor;
 //    drives the "+N" badge.
+//  - `shapeLines` / `locationLines`: overlay strings for the node-body
+//    toggles. Routed through per-node `data` (not a context provider) so a
+//    toggle flip only re-renders ops that actually have overlay content.
 type MLNodeData = WorkerNode['data'] & {
     highlight?: 'input' | 'output';
     buriedMatchCount?: number;
+    shapeLines?: readonly string[];
+    locationLines?: readonly string[];
 };
 
 interface ViewProps {
@@ -91,26 +96,35 @@ type MLNode = Node<MLNodeData>;
 // has no overlay entry.
 const EMPTY_OVERLAY_LINES: readonly string[] = Object.freeze([]);
 
-type MlirNodeBodyOverlayLines = { shapes: string[]; location: string[] };
+const EMPTY_CHAIN: readonly string[] = Object.freeze([]);
 
-// Shared empty-map instance returned when both toggles are off, so
-// `nodeBodyContextValue`'s identity stays stable and every MlirOpNode
-// avoids a context-driven re-render on unrelated graph rebuilds.
-const EMPTY_OVERLAY_MAP: Map<string, MlirNodeBodyOverlayLines> = new Map();
-
-type MlirNodeBodyContextValue = {
-    overlayLinesByNodeId: Map<string, MlirNodeBodyOverlayLines>;
+const parentChainsEqual = (a: readonly string[], b: readonly string[]): boolean => {
+    for (let i = 0; i < a.length; i++) {
+        if (a[i] !== b[i]) {
+            return false;
+        }
+    }
+    return true;
 };
 
-const MlirNodeBodyContext = createContext<MlirNodeBodyContextValue>({
-    overlayLinesByNodeId: new Map(),
-});
+// `parentChain` is root-first, terminating at the immediate parent.
+type NodeTopologyEntry = {
+    parentChain: readonly string[];
+    parentId: string | undefined;
+    type: string | undefined;
+    groupKind: MLNodeData['groupKind'];
+    collapsedSubgraphNamespace: MLNodeData['collapsedSubgraphNamespace'];
+    subgraphToggleState: MLNodeData['subgraphToggleState'];
+};
 
-const MlirOpNode = memo<NodeProps<MLNode>>(({ id, data }) => {
-    const { overlayLinesByNodeId } = useContext(MlirNodeBodyContext);
-    const overlay = overlayLinesByNodeId.get(id);
-    const shapeLines = overlay?.shapes ?? EMPTY_OVERLAY_LINES;
-    const locationLines = overlay?.location ?? EMPTY_OVERLAY_LINES;
+type MlirNodeBodyOverlayLines = { shapes: string[]; location: string[] };
+
+// Shared empty-map used when both toggles are off; keeps map identity stable.
+const EMPTY_OVERLAY_MAP: Map<string, MlirNodeBodyOverlayLines> = new Map();
+
+const MlirOpNode = memo<NodeProps<MLNode>>(({ data }) => {
+    const shapeLines = data.shapeLines ?? EMPTY_OVERLAY_LINES;
+    const locationLines = data.locationLines ?? EMPTY_OVERLAY_LINES;
 
     return (
         <>
@@ -352,7 +366,7 @@ function builtGraphToReactFlow(built: BuiltGraph): { nodes: MLNode[]; edges: Edg
 }
 
 const MlGraphInner = ({ data }: ViewProps) => {
-    const { fitView, getViewport, setViewport } = useReactFlow();
+    const { fitView, getViewport, setViewport, updateNode } = useReactFlow<MLNode, Edge>();
     const graph = data.graphs[0];
     const [nodes, setNodes, onNodesChange] = useNodesState<MLNode>([]);
     const [edges, setEdges, onEdgesChange] = useEdgesState<Edge>([]);
@@ -371,6 +385,9 @@ const MlGraphInner = ({ data }: ViewProps) => {
     const [nodeBodyToggles, setNodeBodyToggles] = useAtom(mlirNodeBodyTogglesAtom);
     const filterRef = useRef<MlirOpFilterHandle>(null);
     const selectedNodeIdRef = useRef<string | null>(null);
+    // Passive mirror of `nodes` so callbacks stored in context (see
+    // `toggleNamespaceFromGroup`) don't churn identity per drag frame.
+    const latestNodesRef = useRef<MLNode[]>([]);
     const viewportAnchorRef = useRef<{
         toNodeId: string;
         fromPosition: { x: number; y: number };
@@ -395,6 +412,10 @@ const MlGraphInner = ({ data }: ViewProps) => {
     useEffect(() => {
         selectedNodeIdRef.current = selectedNodeId;
     }, [selectedNodeId]);
+
+    useEffect(() => {
+        latestNodesRef.current = nodes;
+    }, [nodes]);
 
     // `pendingFocusNodeIdRef` is a one-shot baton armed by `navigateToNode`
     // when the locate target lives in a collapsed namespace and consumed by
@@ -570,9 +591,9 @@ const MlGraphInner = ({ data }: ViewProps) => {
         return next;
     }, [nodes]);
     // "Computed cached state" pattern (React docs): only replace the
-    // tracked set when contents actually change. Prevents a drag from
-    // fanning out a re-render to every mounted `MlirOpNode` via the
-    // `MlirNodeBodyContext` provider.
+    // tracked set when contents actually change. Downstream memos
+    // (`filterMatchInfo`, `overlayLinesByNodeId`) must not invalidate
+    // on drag frames.
     const [visibleOpNodeIds, setVisibleOpNodeIds] = useState<Set<string>>(visibleOpNodeIdsSource);
     if (visibleOpNodeIds !== visibleOpNodeIdsSource) {
         let unchanged = visibleOpNodeIds.size === visibleOpNodeIdsSource.size;
@@ -589,17 +610,64 @@ const MlGraphInner = ({ data }: ViewProps) => {
         }
     }
 
-    // Per-node overlay lines for the node-body toggles. Scoped to op
-    // nodes currently on the canvas (`visibleOpNodeIds`) — sources hidden
-    // inside collapsed subgraphs never render overlay lines and don't
-    // contribute to `extraLinesByNodeId` height. Computed exactly once
-    // per (toggle, visible-set) change and shared with both the
-    // height-growth path (`extraLinesByNodeId` below → `styledNodes`) and
-    // the render path (via `MlirNodeBodyContext` → `MlirOpNode`), so
-    // `collectShapeLines` / `collectLocationLines` never run twice for
-    // the same node on a single toggle flip. When both toggles are off
-    // we return a shared empty Map so the context Provider value stays
-    // referentially stable and doesn't fan out a re-render to every op.
+    // Structural topology per node — set once by the worker, invariant
+    // under drag / selection / hover. Stabilised via computed-cached-state
+    // so `displayedEdges` bails out on unrelated frames.
+    const nodeTopologyByIdSource = useMemo<Map<string, NodeTopologyEntry>>(() => {
+        const parentById = new Map<string, string | undefined>();
+        for (const n of nodes) {
+            parentById.set(n.id, n.parentId);
+        }
+        const result = new Map<string, NodeTopologyEntry>();
+        for (const n of nodes) {
+            const chain: string[] = [];
+            let pid = n.parentId;
+            while (pid) {
+                chain.push(pid);
+                pid = parentById.get(pid);
+            }
+            chain.reverse();
+            result.set(n.id, {
+                parentChain: chain,
+                parentId: n.parentId,
+                type: n.type,
+                groupKind: n.data?.groupKind,
+                collapsedSubgraphNamespace: n.data?.collapsedSubgraphNamespace,
+                subgraphToggleState: n.data?.subgraphToggleState,
+            });
+        }
+        return result;
+    }, [nodes]);
+    const [nodeTopologyById, setNodeTopologyById] = useState<Map<string, NodeTopologyEntry>>(nodeTopologyByIdSource);
+    if (nodeTopologyById !== nodeTopologyByIdSource) {
+        let unchanged = nodeTopologyById.size === nodeTopologyByIdSource.size;
+        if (unchanged) {
+            for (const [id, next] of nodeTopologyByIdSource) {
+                const prev = nodeTopologyById.get(id);
+                if (
+                    !prev ||
+                    prev.parentId !== next.parentId ||
+                    prev.type !== next.type ||
+                    prev.groupKind !== next.groupKind ||
+                    prev.collapsedSubgraphNamespace !== next.collapsedSubgraphNamespace ||
+                    prev.subgraphToggleState !== next.subgraphToggleState ||
+                    prev.parentChain.length !== next.parentChain.length ||
+                    !parentChainsEqual(prev.parentChain, next.parentChain)
+                ) {
+                    unchanged = false;
+                    break;
+                }
+            }
+        }
+        if (!unchanged) {
+            setNodeTopologyById(nodeTopologyByIdSource);
+        }
+    }
+
+    // Per-node overlay lines for the node-body toggles. Only op-nodes
+    // with non-empty overlay content appear in this map; `styledNodes`
+    // applies the arrays onto `data.shapeLines` / `data.locationLines`
+    // so React Flow's per-node diff drives re-renders on toggle flip.
     const overlayLinesByNodeId = useMemo<Map<string, MlirNodeBodyOverlayLines>>(() => {
         if (!nodeBodyToggles.location && !nodeBodyToggles.shapes) {
             return EMPTY_OVERLAY_MAP;
@@ -619,18 +687,6 @@ const MlGraphInner = ({ data }: ViewProps) => {
         return result;
     }, [nodeBodyToggles.location, nodeBodyToggles.shapes, visibleOpNodeIds, sourceNodeById]);
 
-    // Derived count map consumed by `styledNodes` to grow each node's DOM
-    // height inline. The layout worker is intentionally not told about
-    // these — trading occasional vertical overlap with the row below for
-    // zero layout shift when the user flicks a toggle.
-    const extraLinesByNodeId = useMemo<Map<string, number>>(() => {
-        const result = new Map<string, number>();
-        for (const [id, { shapes, location }] of overlayLinesByNodeId) {
-            result.set(id, shapes.length + location.length);
-        }
-        return result;
-    }, [overlayLinesByNodeId]);
-
     // Maps each label-matching source to its visible representative: itself
     // if on canvas, otherwise the anchor of its outermost collapsed
     // ancestor. Reads `containingNamespacesByNodeId` (not `source.namespace`)
@@ -641,22 +697,30 @@ const MlGraphInner = ({ data }: ViewProps) => {
     // expanded parent or as the collapsed anchor of their own namespace), so
     // `visibleNodeIds.has(anchorId)` can't distinguish collapsed from
     // expanded. Mirrors the graph builder's `resolveRenderedNodeId`.
+    // Lifted from `filterMatchInfo` so an active filter doesn't recompile
+    // the RegExp on every interaction frame.
+    const filterMatcher = useMemo(() => {
+        if (appliedFilterQuery.length === 0) {
+            return null;
+        }
+        return buildFilterMatcher(filterMode, appliedFilterQuery);
+    }, [filterMode, appliedFilterQuery]);
+
     const filterMatchInfo = useMemo<{
         visibleRepIds: Set<string>;
         buriedCountByRepId: Map<string, number>;
         hiddenMatchCount: number;
         isRegexInvalid: boolean;
     } | null>(() => {
-        if (appliedFilterQuery.length === 0) {
+        if (!filterMatcher) {
             return null;
         }
-        const { testLabel, isRegexInvalid } = buildFilterMatcher(filterMode, appliedFilterQuery);
+        const { testLabel, isRegexInvalid } = filterMatcher;
         const anchorByNamespace = interactionIndex?.anchorByNamespace ?? {};
         const containingNamespacesByNodeId = interactionIndex?.containingNamespacesByNodeId ?? {};
-        const visibleNodeIds = new Set<string>();
-        for (const node of nodes) {
-            visibleNodeIds.add(node.id);
-        }
+        // Anchor and source ids are always op-node ids, so the stabilised
+        // op set is the correct visibility check and keeps the memo out
+        // of the per-frame `nodes` invalidation path.
         const visibleRepIds = new Set<string>();
         const buriedCountByRepId = new Map<string, number>();
         let hiddenMatchCount = 0;
@@ -673,13 +737,13 @@ const MlGraphInner = ({ data }: ViewProps) => {
                 for (const ns of containing) {
                     if (!expandedNamespaces.has(ns)) {
                         const anchorId = anchorByNamespace[ns] ?? source.id;
-                        repId = visibleNodeIds.has(anchorId) ? anchorId : null;
+                        repId = visibleOpNodeIds.has(anchorId) ? anchorId : null;
                         break;
                     }
                 }
             }
             if (repId === null) {
-                repId = visibleNodeIds.has(source.id) ? source.id : null;
+                repId = visibleOpNodeIds.has(source.id) ? source.id : null;
             }
             if (repId === null) {
                 continue;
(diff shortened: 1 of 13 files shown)


src/components/mlir/mlirLayoutWorkerCache.ts
```diff
@@ -0,0 +1,22 @@
...
+// LRU order; the head is then always the least-recently-used entry.
+export const touchLruCache = <K, V>(cache: Map<K, V>, key: K, value: V, limit: number): void => {
+    if (cache.has(key)) {
+        cache.delete(key);
+    } else if (cache.size >= limit) {
+        const oldestKey = cache.keys().next().value;
+        if (oldestKey !== undefined) {
+            cache.delete(oldestKey);
+        }
+    }
```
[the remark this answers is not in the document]
aidemsined: Valid catch — fixed in 5a8e44fc. Not exploitable in the current usage (`cacheByGraphId: Map<string, BuiltGraph>` — the string cache keys are namespace-join strings and never `undefined` at runtime), but the helper is generic and exported, and one of the existing spec cases already claims non-string-key coverage, so silently violating the `limit` contract for the `undefined` LRU-position case was worth locking down. Now checks the iterator's `done` sentinel and evicts the head regardless of value; new spec case exercises `undefined` as the LRU-position key.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -641,57 +670,34 @@ const MlGraphInner = ({ data }: ViewProps) => {
     // expanded parent or as the collapsed anchor of their own namespace), so
     // `visibleNodeIds.has(anchorId)` can't distinguish collapsed from
     // expanded. Mirrors the graph builder's `resolveRenderedNodeId`.
+    // Lifted from `filterMatchInfo` so an active filter doesn't recompile
+    // the RegExp on every interaction frame.
```
dcblundell: **Must fix — stale / misplaced `filterMatcher` comment**

This block still describes filter-match *resolution* (`containingNamespacesByNodeId`, collapse-to-anchor) and cites the dead name `` `visibleNodeIds.has(anchorId)` ``, but the symbol below only builds/returns a matcher. Resolution now lives in `resolveFilterMatches` (`mlirFilter.ts`), which already has the accurate comment.

Per CONVENTIONS: stale comments are bugs. Trim this to the why for the matcher lift (avoid RegExp recompile on interaction frames) and drop or retarget the `visibleNodeIds` / resolution prose.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -641,57 +670,34 @@ const MlGraphInner = ({ data }: ViewProps) => {
...
+            expandedNamespaces,
+            anchorByNamespace: interactionIndex?.anchorByNamespace ?? {},
+            containingNamespacesByNodeId: interactionIndex?.containingNamespacesByNodeId ?? {},
+            visibleOpNodeIds,
+        });
+        return { ...resolution, isRegexInvalid: filterMatcher.isRegexInvalid };
+    }, [filterMatcher, sourceNodes, expandedNamespaces, interactionIndex, visibleOpNodeIds]);
 
     // Visible reps in canvas order so prev/next walks top-to-bottom.
     const matchedNodesInOrder = useMemo<string[]>(() => {
```
dcblundell, line 711: **Must fix — `matchedNodesInOrder` still keys off `nodes` and returns a fresh `[]`**

Every drag frame (position-only `nodes` updates) recomputes this memo, churns `goToMatch` / `goToPrevMatch` / `goToNextMatch`, and defeats `memo(MlirOpFilter)`.

Canvas order does not change on drag — drop `nodes` from the dep list (order from a rebuild-time id list / topology), return a module-level empty sentinel when idle, and keep the previous array when contents are identical.

Verify: React Profiler on group-drag with filter chrome mounted — `MlirOpFilter` should not re-render.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -1419,20 +1447,26 @@ const MlGraphInner = ({ data }: ViewProps) => {
...
                         ...next,
                         height: grownHeight,
                         style: { ...(next.style ?? {}), height: grownHeight },
+                        data: {
+                            ...next.data,
+                            shapeLines: overlay.shapes,
+                            locationLines: overlay.location,
+                        },
                     };
                 }
```
dcblundell, line 1441: **Should fix — overlay toggles undo `data`-stability on drag**

With shapes/location toggles on, `styledNodes` always allocates new `data` (`shapeLines` / `locationLines`) for every overlay op whenever `nodes` changes. That re-renders all overlay-bearing `MlirOpNode`s each drag frame — worse than the old stable body context for this path.

Prefer baking overlay fields into store nodes once when `overlayLinesByNodeId` changes, or reuse prior styled-node / `data` identity when overlay content and grown height are unchanged.

Also worth a thin test (after extract if needed) that assert toggles land on `data` and grow `height`, and that badge/`buriedMatchCount` survives the same rewrite.

src/components/mlir/mlirFilter.ts
```diff
@@ -37,3 +39,68 @@ export function buildFilterMatcher(mode: MlirFilterMode, query: string): MlirFil
...
+        const containing = containingNamespacesByNodeId[source.id];
+        let repId: string | null = null;
+        if (containing && containing.length > 0) {
+            for (const ns of containing) {
+                if (!expandedNamespaces.has(ns)) {
+                    const anchorId = anchorByNamespace[ns] ?? source.id;
+                    repId = visibleOpNodeIds.has(anchorId) ? anchorId : null;
+                    break;
+                }
+            }
```
dcblundell, line 90: **Should fix — duplicated collapsed-rep walk vs graph builder**

This outer→inner “first non-expanded namespace → `anchorByNamespace[ns] ?? nodeId`” loop mirrors `resolveRenderedNodeId` in `mlirGraphBuilder.ts`. Comments/tests claim they stay in sync; that will drift.

Extract a pure `resolveCollapsedRepId(...)` (shared module both can import) and keep filter-only visibility / burial counting on top.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -91,26 +96,71 @@ type MLNode = Node<MLNodeData>;
...
+// are invariant under drag / selection / hover — RF only mutates `position`
+// and `selected` on those paths — so recomputing here means downstream memos
+// keying on the topology map bail out for free on non-structural frames,
+// without a per-render source rebuild + equality walk.
+const buildNodeTopologyById = (nodes: readonly MLNode[]): Map<string, NodeTopologyEntry> => {
+    const parentById = new Map<string, string | undefined>();
+    for (const n of nodes) {
+        parentById.set(n.id, n.parentId);
+    }
+    const result = new Map<string, NodeTopologyEntry>();
```
dcblundell, line 129: **Should fix — Map accumulator named `result`**

AGENTS.md / CONVENTIONS: name `Map`/`Record` accumulators after the relation they encode. Prefer `topologyById` / `entryByNodeId` here instead of bare `result`.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -91,26 +96,71 @@ type MLNode = Node<MLNodeData>;
...
-});
+const buildVisibleOpNodeIds = (nodes: readonly MLNode[]): Set<string> => {
+    const next = new Set<string>();
+    for (const n of nodes) {
+        if (n.type === 'mlirOp') {
+            next.add(n.id);
+        }
+    }
+    return next;
+};
```
dcblundell, line 158: **Should fix — untested structural boundary (`buildNodeTopologyById` / `buildVisibleOpNodeIds`)**

These are the new rebuild-time inputs for filter + LCA edges, but stay private and untested. Sibling logic was extracted to `mlirFilter.ts` so it could be locked.

Prefer extracting to a small module (same pattern as `resolveFilterMatches`) and asserting: nested `parentChain` order (root-first), missing-parent short-circuit, groups excluded from the op set, section vs real-group field copy.

tests/mlirFilter.spec.ts
```diff
@@ -79,3 +79,153 @@ describe('buildFilterMatcher', () => {
...
+            expandedNamespaces: new Set(),
+            anchorByNamespace: { 'region.a': 'anchor-op' },
+            containingNamespacesByNodeId: { 'buried-op': ['region.a'] },
+            visibleOpNodeIds: new Set(),
+        });
+
+        expect(result.visibleRepIds.size).toBe(0);
+        expect(result.buriedCountByRepId.size).toBe(0);
+        expect(result.hiddenMatchCount).toBe(0);
+    });
```
dcblundell: **Should fix — weak “op-only visibility” regression**

Comment claims to lock “op-only vs full RF node array / group id must not satisfy visibility”, but only passes `visibleOpNodeIds: new Set()` — equivalent to “anchor off-canvas → drop”. It never asserts that a present group id fails while a matching op id would succeed.

Tighten with something like `visibleOpNodeIds: new Set(['group:region.a'])` + anchor `anchor-op` → empty reps, and a twin with `anchor-op` → match.

dcblundell (commented):
Combined code-review conductor findings (Must / Should only). One thread per item.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -472,13 +534,33 @@ const MlGraphInner = ({ data }: ViewProps) => {
...
     const applyBuiltGraph = useCallback(
         (built: BuiltGraph) => {
+            // Drop any in-flight group-drag delta: the pending RAF would
+            // otherwise apply it against the just-committed canonical
+            // position (shift by delta on top of the worker's placement).
+            if (moveGroupRafRef.current !== null) {
+                window.cancelAnimationFrame(moveGroupRafRef.current);
+                moveGroupRafRef.current = null;
+            }
+            moveGroupPendingRef.current.clear();
```
dcblundell: **Should fix — no coverage for RAF cancel-on-rebuild**

RAF-coalesced `moveGroup` + clear/cancel in `applyBuiltGraph` (and unmount) is new async intent with a real bug class: pending delta applied on top of worker placement.

Prefer extracting delta-accumulate + flush / cancel into a testable helper, then a small fake-timer / `raf` mock test for “pending RAF + rebuild ⇒ no post-rebuild position shift” (and optionally “multiple mousemove deltas coalesce to one `updateNode`”). Full `MlGraphInner` mount is heavier than this repo’s current MLIR test style.

dcblundell:
**Consider / nit** (from combined conductor review — non-blocking):

- Extract `clearPendingMoveGroup()` for the duplicated RAF cancel + `moveGroupPendingRef.clear()` in `applyBuiltGraph` and the unmount effect (`MLIRViewReactFlow.tsx`).
- Single-pass `buildStructuralIndices(nodes) → { topology, visibleOpIds }` instead of walking the node list twice in `buildNodeTopologyById` / `buildVisibleOpNodeIds`.
- Type `filterMatchInfo` as `FilterMatchResolution & { isRegexInvalid: boolean }` so the view type can’t drift from the helper.
- Add a one-line why on the `mlirFilter.ts` file-level `eslint-disable no-continue` (or restructure the loop to avoid `continue`).
- Batch multi-id `updateNode` into one `setNodes` if multi-group pending ever shows up in traces; for the usual single-group drag, RAF already caps rate.
- Heap-check the worker after cycling ~32 distinct expansion keys on a large report — cap of 32 `BuiltGraph`s per graph is a win vs unbounded growth, but memory can still be heavy.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Fixed in `aab4a3d5`. Trimmed the block to just the matcher-lift rationale (avoid recompiling the RegExp on interaction frames) and dropped the stale `visibleNodeIds.has(anchorId)` / resolution prose — that contract now lives solely on `resolveFilterMatches` in `mlirFilter.ts`.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Fixed in `aab4a3d5`. `matchedNodesInOrder` now iterates `visibleOpNodeIds` (build-time state — its `Set` insertion order is canvas order because `buildVisibleOpNodeIds` walks the same node array) instead of `nodes`, and returns a shared module-level `EMPTY_MATCHED_ORDER` sentinel when idle. Deps are now `[visibleOpNodeIds, filterMatchInfo]`, both stable across position-only drag frames, so the memo — and `MlirOpFilter` — no longer churn on drag. `visibleRepIds ⊆ visibleOpNodeIds`, so the emitted order is identical to the old `nodes`-filtered walk.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Agreed the overlay branch re-allocates `data` for every overlay op on each `nodes` change. I tried both suggested fixes and both collide with this repo's react-hooks lint policy (`--max-warnings 0`):

- Reusing prior styled-node / `data` identity needs a cross-render cache — a ref read+written during render → `react-hooks/refs` ("Cannot access/update ref during render").
- Baking overlay onto store nodes when `overlayLinesByNodeId` changes needs a `setNodes` in an effect → `react-hooks/set-state-in-effect`, which this file deliberately avoids (see the `MlGraphInner` remount comment).

The lint-clean version is to bake overlay onto the store nodes from the toggle event-handler + `applyBuiltGraph` (mirroring how `nodeTopologyById` / `visibleOpNodeIds` are already build-time state). That's a real change to the overlay data flow and it overlaps the `MlirOpNode` / overlay extraction planned in #1741, so I'd rather land it (with the suggested `data`/`height`/`buriedMatchCount` test) there than hot-patch it here. Tracking against #1741.

src/components/mlir/mlirFilter.ts
aidemsined: This is already scoped: #1740 (Group B, item 3) calls for extracting a shared `resolveVisibleRepId(nodeId, expandedNamespaces, index)` and pointing both the builder (`resolveRenderedNodeId` in `mlirGraphBuilder.ts`) and the filter-side walk at it. The ticket cites the pre-move location (`MLIRViewReactFlow.tsx:669-678`); this PR relocated that walk into `resolveFilterMatches`, so it's the same duplication, now at `mlirFilter.ts:90`. Leaving as-is here and folding it into that extraction rather than widening this PR.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Fair — the accumulator should be named for the relation it encodes. Rather than rename it in place in the composer, I'll give it the convention-correct name (`entryByNodeId`) when `buildNodeTopologyById` is extracted into its own tested module — which is exactly the extraction flagged in the next thread. Keeping this PR's diff scoped to the hot-path changes.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Agreed these are the right things to lock. The module + view-types extraction is already earmarked: `MLNodeData` / `NodeTopologyEntry` moving to a shared view-types module is called out for #1741 in the PR description, and the "extract a pure function + characterization spec" pattern (same as `resolveFilterMatches`) is Group B's remit (#1740). I'd rather land `buildNodeTopologyById` / `buildVisibleOpNodeIds` as tested modules there — asserting root-first `parentChain` order, missing-parent short-circuit, group exclusion, and section-vs-real-group field copy — than grow this hot-path PR. Tracking against #1740/#1741.

tests/mlirFilter.spec.ts
aidemsined: Agreed — the current case only exercises "anchor off-canvas → drop" and never proves a present group id fails while a matching op id succeeds. The tightening you describe (`visibleOpNodeIds: new Set(['group:region.a'])` + `anchor-op` → empty reps, plus a twin with `anchor-op` present → match) is the right assertion. I'll add it alongside the `resolveFilterMatches` / topology extraction specs in the follow-up rather than here.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Fixed in `aab4a3d5`. Extracted the accumulate/flush/cancel logic into `createMoveGroupBatcher` (`mlirMoveGroupBatch.ts`); the composer now just holds the batcher and calls `.accumulate()` / `.cancel()` (the two loose refs + `flushMoveGroup` are gone). New `tests/mlirMoveGroupBatch.spec.ts` uses an injected frame scheduler to cover: multi-delta coalescing into one `applyDelta` per group per frame, per-group independence, a fresh frame after flush, and the cancel-on-rebuild path (`cancel()` drops the pending delta + cancels the frame, so a stale frame applies nothing) plus resume-after-cancel. The "pending delta applied on top of worker placement" bug class is now asserted directly, without mounting `MlGraphInner`.

src/components/mlir/MLIRViewReactFlow.tsx
dcblundell: Verified — the residual is real: with shape/location toggles on, `styledNodes` still allocates fresh `data` for every overlay op on each `nodes` change, so drag re-renders those `MlirOpNode`s.

Lint objections to a render-time ref cache / `setNodes`-in-effect are fair for this file.

Please don't track this against #1741 as currently written — that ticket is Op/Group extract + `buildRegionPartnerMaps`, and it doesn't mention baking overlay fields into store nodes. Either:
1. land the toggle-handler + `applyBuiltGraph` bake (plus the `data` / `height` / `buriedMatchCount` characterization) here, or
2. add an explicit acceptance criterion on a follow-up (#1741 AC update or a small dedicated issue) and link it from this thread.

Leaving open until one of those is chosen.

tests/mlirFilter.spec.ts
dcblundell: The case still only proves "empty visible set → drop". The comment claims a group id must not satisfy visibility, but nothing asserts that.

Prefer landing the twin cases in this PR — they're small and lock the narrowing this PR introduced:

```ts
// group id present, matching op anchor absent → no reps
visibleOpNodeIds: new Set(['group:region.a'])
anchorByNamespace: { 'region.a': 'anchor-op' }
// → empty

// same fixture with the op anchor present → match
visibleOpNodeIds: new Set(['group:region.a', 'anchor-op'])
// → ['anchor-op']
```

Happy to accept a follow-up only if there's a concrete linked issue with that assertion called out; otherwise please patch here. Leaving open.

aidemsined: Patched here in `40aafc34`. Replaced the weak `visibleOpNodeIds: new Set()` case with your twin:

- **group id present, op anchor absent** → `visibleOpNodeIds: new Set(['group:region.a'])`, `anchorByNamespace: { 'region.a': 'anchor-op' }` → empty reps (proves a group id can't stand in for the op anchor).
- **same fixture + anchor present** → `new Set(['group:region.a', 'anchor-op'])` → `['anchor-op']`, `buriedCountByRepId('anchor-op') === 1`, `hiddenMatchCount === 1`.

Both green in `tests/mlirFilter.spec.ts` (18 passing). Agreed this locks the op-only narrowing this PR introduced — kept it here rather than a follow-up.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Fair correction — #1741's AC is Op/Group extract + `buildRegionPartnerMaps` and doesn't mention overlay baking, so that was the wrong home. Went with your option 2: filed #1762 with the explicit AC.

#1762 spells out: bake `shapeLines` / `locationLines` + grown `height` onto store nodes from the toggle event-handler + `applyBuiltGraph` (mirroring `nodeTopologyById` / `visibleOpNodeIds` build-time state), drop the overlay branch from `styledNodes`, handle toggle-off, plus the `data` / `height` / `buriedMatchCount` characterization test and a React Profiler check that overlay `MlirOpNode`s don't re-render on group-drag. Soft-sequenced after #1741 but independent. Linking here so this thread has the concrete AC.

PR #1760, opened 2026-07-14
Clear abandoned remote file transfers

### LINKED ISSUE #1757: Clear stale REMOTE_SYNC progress after backend death mid-transfer
## Context

From review on #1750 ([discussion](https://github.com/tenstorrent/ttnn-visualizer/pull/1750#discussion_r3573619893)).

`SocketProvider` reconnect uses `clearFileTransferProgressForSourceIfInactive(REMOTE_SYNC)` so a mid-upload socket drop does not wipe live client-driven progress while axios is still running. That is correct for the reconnect-mid-upload case.

## Gap

If the **backend** dies mid-transfer without emitting a terminal `FAILED` (socket disconnect, process gone), `REMOTE_SYNC` can stay `DOWNLOADING` in the registry. The reconnect guard preserves that active slot, `syncRemoteFolder` has no client-side timeout, and the overlay can freeze on last-known progress until the tab is closed.

This is **not a regression** from #1750 — the pre-PR single atom had the same failure mode. Capturing it so we do not treat the "if inactive" wording as covering server-death.

## Possible directions

- Document the limitation next to the reconnect clear.
- Bounded staleness: timestamp registry entries and clear a nominally "active" slot older than N seconds on reconnect.
- Client-side timeout / abort on the sync axios call paired with clearing the slot.

dcblundell:
This pull request introduces robust handling for stale or orphaned remote file transfer progress, particularly for remote sync operations. It adds time-based logic to automatically clear outdated progress entries after reconnects, preventing overlays from persisting indefinitely if the backend dies mid-transfer. The changes also enforce request timeouts for remote sync operations and provide comprehensive tests for these new behaviors.

**File transfer staleness handling and registry improvements:**

- Introduced `REMOTE_SYNC_PROGRESS_STALE_MS` and `REMOTE_SYNC_REQUEST_TIMEOUT_MS` constants in `FileTransfer.ts` to define staleness and request timeout windows for remote sync operations.
- Updated the file transfer registry:
  - Added an `updatedAtMs` timestamp to `FileProgress` and ensured it is set whenever progress is updated. [[1]](diffhunk://#diff-b2d02bc0f094787f8016fcd55e5472330b972c2ba36ff2496e1710c4805b7fdcR135-R136) [[2]](diffhunk://#diff-c2f3910ef58d210aa29d90623e8873c2c9d595bb42f0337203851a12795d5b11L68-R83)
  - Enhanced `clearFileTransferProgressForSourceIfInactive` to accept a `staleAfterMs` option, enabling the clearing of active but stale or orphaned slots after reconnects. [[1]](diffhunk://#diff-c2f3910ef58d210aa29d90623e8873c2c9d595bb42f0337203851a12795d5b11R12-R23) [[2]](diffhunk://#diff-c2f3910ef58d210aa29d90623e8873c2c9d595bb42f0337203851a12795d5b11L79-R119)
- On socket reconnect, the app now clears terminal or stale remote sync slots while preserving fresh, active transfers. [[1]](diffhunk://#diff-2e11e7b1f151fac8271b3811e4d1d83f7cd8f4ccb2a2fa15f611cbbc935d5c35R9-L13) [[2]](diffhunk://#diff-2e11e7b1f151fac8271b3811e4d1d83f7cd8f4ccb2a2fa15f611cbbc935d5c35L33-R38)

**Remote sync request timeout enforcement:**

- Updated `useRemoteConnection` to use the new `REMOTE_SYNC_REQUEST_TIMEOUT_MS` for the `/api/remote/sync` call, ensuring the UI doesn't hang indefinitely if the backend is unresponsive. [[1]](diffhunk://#diff-f7cfb535298cc5b5a94f68f56d793e7cd060556c507047571948dbf426d7d0ccR8-L13) [[2]](diffhunk://#diff-f7cfb535298cc5b5a94f68f56d793e7cd060556c507047571948dbf426d7d0ccL109-R118)

**Testing improvements:**

- Added and expanded tests to verify:
  - Stale progress slots are cleared and fresh slots are preserved on reconnect. [[1]](diffhunk://#diff-b8c3d554e989d77ac1b54326d4336521c8d76b2456c4904d4108d438f9b0c76fR7-R9) [[2]](diffhunk://#diff-b8c3d554e989d77ac1b54326d4336521c8d76b2456c4904d4108d438f9b0c76fR69) [[3]](diffhunk://#diff-b8c3d554e989d77ac1b54326d4336521c8d76b2456c4904d4108d438f9b0c76fR96-R183)
  - The registry correctly stamps and checks `updatedAtMs` for staleness. [[1]](diffhunk://#diff-7228d5ce4a5b4ca70da463ec8c9ff08810e32c8c35a23e960f9b519d7061e73dL7-R8) [[2]](diffhunk://#diff-7228d5ce4a5b4ca70da463ec8c9ff08810e32c8c35a23e960f9b519d7061e73dR112-R130) [[3]](diffhunk://#diff-7228d5ce4a5b4ca70da463ec8c9ff08810e32c8c35a23e960f9b519d7061e73dR141-R149) [[4]](diffhunk://#diff-7228d5ce4a5b4ca70da463ec8c9ff08810e32c8c35a23e960f9b519d7061e73dL140-R221) [[5]](diffhunk://#diff-7228d5ce4a5b4ca70da463ec8c9ff08810e32c8c35a23e960f9b519d7061e73dL162-R235)
  - Remote sync requests are issued with the correct timeout. [[1]](diffhunk://#diff-84188e640666692040240bd2e650ae66ba8e5faa617d1440273e5258a57d0e8eR8) [[2]](diffhunk://#diff-84188e640666692040240bd2e650ae66ba8e5faa617d1440273e5258a57d0e8eR187-R216)

Fixes https://github.com/tenstorrent/ttnn-visualizer/issues/1757.


src/hooks/useRemote.tsx
```diff
@@ -106,11 +107,15 @@ const useRemoteConnection = () => {
...
-            performance: performanceRemoteFolder,
-        });
+        return axiosInstance.post<RemoteFolder>(
+            `${Endpoints.REMOTE}/sync`,
+            {
+                connection,
+                profiler: profilerRemoteFolder,
+                performance: performanceRemoteFolder,
+            },
+            { timeout: REMOTE_SYNC_REQUEST_TIMEOUT_MS },
```
aidemsined: **Definite** — Ticket #1757 direction (c) asks for the timeout *paired with* clearing the slot, but here the two halves live in different layers. The `timeout` bounds the axios hang, yet the REMOTE_SYNC registry slot is only cleared by `RemoteSyncConfigurator`'s pre-existing `finally { resetFileTransferProgress() }`, and the component-local `isSyncing` lock only releases when the promise settles (there's no `AbortController`). So on a timeout, a reconnect in the 6–30 min window clears the overlay but the sync form stays locked for up to ~24 min, and any future `syncRemoteFolder` caller gets the timeout without guaranteed cleanup. Consider co-locating clear-on-timeout/failure at the hook layer (here in `syncRemoteFolder`) and adding a test that a rejected/timed-out sync clears the slot — that linkage is currently assumed, not proven (the new test only asserts the `{ timeout }` option reaches axios).

src/model/APIData.ts
```diff
@@ -132,6 +132,8 @@ export interface FileProgress {
     bytesTotal?: number;
     currentFileSize?: number;
     timestamp?: string; // Optional, with default handled elsewhere if necessary
+    /** Client-only: ms since epoch of the last registry write. Used for reconnect staleness. */
+    updatedAtMs?: number;
```
aidemsined, line 136: **Definite** — `FileProgress` already carries `timestamp?: string` from the backend socket emit, but `SocketProvider` currently drops `data.timestamp`. Adding a parallel client-only `updatedAtMs` duplicates freshness semantics, blurs server- vs client-provenance on a shared model type, and forces `setFileTransferProgressForSource` to stamp *every* source (MLIR_UPLOAD, LOCAL_UPLOAD) when only REMOTE_SYNC reconnect cleanup consumes it. Worth considering mapping `data.timestamp` through and basing staleness on parsed ISO age instead of a new field.

src/definitions/FileTransfer.ts
```diff
@@ -0,0 +1,18 @@
...
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+/**
+ * Reconnect orphan window for REMOTE_SYNC progress.
+ * Backend emits are file-granular and a single download can run up to the 5-minute
+ * SFTP/SCP per-file timeout with no further events — keep this above that bound
+ * (+ margin) so a mid-file socket reconnect does not wipe a live overlay.
+ */
+export const REMOTE_SYNC_PROGRESS_STALE_MS = 6 * 60 * 1000;
```
aidemsined: **Possible** — The 6-min window sits only ~1 min above the backend's hardcoded 300s per-file cap (`sftp_operations.py:729`). A single large file can run that full 5 min with no intermediate socket emits, so a reconnect at ~6 min after the pre-download emit could clear a still-live overlay. Options: bump to 7–8 min, refresh `updatedAtMs` on `disconnect`, and/or tie this constant's comment to the backend literal so the two don't silently drift.

src/libs/SocketProvider.tsx
```diff
@@ -30,10 +31,11 @@ export const SocketProvider = ({ children }: SocketProviderProps) => {
...
     useEffect(() => {
         socket.on('connect', () => {
-            // Only clear stale remote-sync progress if that transfer is no longer
-            // active. A reconnect mid-upload (axios still running on the same
-            // tab) must not close the overlay or drop live client-driven progress.
-            clearFileTransferProgressForSourceIfInactive(FileTransferSource.REMOTE_SYNC);
+            // REMOTE_SYNC only: clear terminal or orphaned slots (#1757). Fresh
+            // actives survive mid-transfer reconnects while axios is still running.
+            clearFileTransferProgressForSourceIfInactive(FileTransferSource.REMOTE_SYNC, {
+                staleAfterMs: REMOTE_SYNC_PROGRESS_STALE_MS,
```
aidemsined: **Possible** — This stale-clear is reconnect-triggered only, not timer-driven. If the backend dies and socket.io can't reconnect (server fully down), `connect` never re-fires, so the overlay persists until the 30-min axios timeout + `RemoteSyncConfigurator` `finally`. Not infinite (good improvement over the pre-PR freeze-until-tab-close), but worth a comment here or a follow-up issue so the reconnect-only trigger is explicit.

src/functions/fileTransferRegistry.ts
```diff
@@ -76,11 +91,30 @@ export function clearFileTransferProgressForSource(source: FileTransferSource):
...
 
-export function clearFileTransferProgressForSourceIfInactive(source: FileTransferSource): void {
+/**
+ * Drops a terminal source slot. When `staleAfterMs` is provided, also drops
+ * active slots that are missing a freshness stamp or are at least as old as
+ * that window — so a reconnect can clear orphaned DOWNLOADING state after
+ * backend death without a FAILED event (#1757), while fresh mid-transfer
+ * progress (axios still running) is kept.
+ */
+export function clearFileTransferProgressForSourceIfInactive(
```
aidemsined, line 105: **Possible** — `staleAfterMs`/`nowMs` widen this shared helper and export `ClearFileTransferIfInactiveOptions` as a semi-public contract, yet `SocketProvider` is the only consumer that passes options. A named wrapper that owns the policy (e.g. `clearStaleRemoteSyncOnReconnect()`) would keep the call site declarative and shrink the public surface. Reasonable to keep the generic form if you expect more sources to need stale-active cleanup soon — otherwise it reads as over-factored for one caller.

src/definitions/RemoteSync.ts
```diff
@@ -0,0 +1,18 @@
...
+ * (+ margin) so a mid-file socket reconnect does not wipe a live overlay.
+ */
+export const REMOTE_SYNC_PROGRESS_STALE_MS = 6 * 60 * 1000;
+
+/**
+ * Upper bound for the remote sync HTTP call. Per-file server downloads already
+ * cap at 5 minutes; this bounds the whole folder sync so a dead backend cannot
+ * leave axios (and the overlay) hanging indefinitely.
+ */
+export const REMOTE_SYNC_REQUEST_TIMEOUT_MS = 30 * 60 * 1000;
```
aidemsined, line 19: **Possible** — `FileTransfer.ts` reads as cross-source transfer policy, but it only holds two REMOTE_SYNC-specific constants. Sibling conventions colocate: `RemoteConnection.ts` already hosts sync UX constants and `UiConfig.ts` holds timing primitives. A `RemoteSync.ts` module (or private constants beside `fileTransferRegistry.ts`) would name the scope more honestly.

src/hooks/useRemote.tsx
dcblundell: Good catch — addressed.

`syncRemoteFolder` now owns an `AbortController` + clears the `REMOTE_SYNC` slot in `finally` (timeout/reject/success), with tests for reject/resolve cleanup. On reconnect, `clearStaleRemoteSyncOnReconnect()` aborts the in-flight HTTP when it drops an *active* orphan so the component `isSyncing` lock can release via that rejection; `RemoteSyncConfigurator` skips the error toast for `axios.isCancel`.

src/model/APIData.ts
dcblundell: Keeping `updatedAtMs` deliberately.

Backend `timestamp` is still unused on the wire mapping, but staleness for orphan cleanup wants “when the client last wrote the registry slot” (receive-time `Date.now()`), not server ISO age — avoids clock skew / parse fragility. Stamping MLIR/LOCAL is unused by the stale path today; harmless. Happy to map `data.timestamp` through later for debugging if useful, but not as the freshness clock.

src/definitions/FileTransfer.ts
dcblundell: Agreed the margin was thin. Bumped `REMOTE_SYNC_PROGRESS_STALE_MS` to 8 minutes and tied the comment explicitly to the hardcoded 300s per-file timeout in `sftp_operations.py`.

src/libs/SocketProvider.tsx
dcblundell: Documented on the `connect` handler: reconnect-triggered only; if the socket never comes back, the axios timeout in `syncRemoteFolder` is the backstop.

src/functions/fileTransferRegistry.ts
dcblundell: Fair. Kept the generic clear (and `nowMs` for tests) but added a declarative `clearStaleRemoteSyncOnReconnect()` wrapper that owns the REMOTE_SYNC policy + abort-on-orphan-clear; `SocketProvider` now calls that.

src/definitions/RemoteSync.ts
dcblundell: Moved the two constants into `src/definitions/RemoteSync.ts` and removed `FileTransfer.ts`.

PR #1775, opened 2026-07-16
Load synced remote reports when not connected to host

### LINKED ISSUE #1767: Synced remote reports should be selectable even without SSH access
If you don't currently have access to the remote machine but have previously synced a report, you should be able to select it. Currently it fails because it will always try to connect to the remote machine first.

### LINKED ISSUE #1774: Expand linked report status reporting
Currently only showing this in a limited capacity for local uploads.

dcblundell:
Allows a user to load any synced reports from remote hosts even when not connected to the host.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1767.

### DIFF
--- a/backend/ttnn_visualizer/local_remote_reports.py
+++ b/backend/ttnn_visualizer/local_remote_reports.py
@@ -1,0 +1,133 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+"""Discover already-synced remote reports on local disk (no SSH)."""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import Callable, List, Optional
+
+from ttnn_visualizer.models import RemoteConnection, RemoteReportFolder
+from ttnn_visualizer.utils import read_last_synced_file
+
+logger = logging.getLogger(__name__)
+
+PROFILER_REQUIRED_FILES = frozenset({"db.sqlite"})
+PERFORMANCE_REQUIRED_FILES = frozenset(
+    {"profile_log_device.csv", "tracy_profile_log_host.tracy"}
+)
+PERFORMANCE_OPS_PERF_PREFIX = "ops_perf_results"
+
+
+def is_valid_local_profiler_report_dir(directory: Path) -> bool:
+    """True when the folder has the files needed to load a memory/profiler report."""
+    return all((directory / name).is_file() for name in PROFILER_REQUIRED_FILES)
+
+
+def is_valid_local_performance_report_dir(directory: Path) -> bool:
+    """True when the folder has the files needed to load a performance report.
+
+    Matches local performance folder listing / upload validation:
+    profile_log_device.csv, tracy_profile_log_host.tracy, and ops_perf_results*.
+    """
+    if not all((directory / name).is_file() for name in PERFORMANCE_REQUIRED_FILES):
+        return False
+    try:
+        entries = directory.iterdir()
+    except OSError:
+        return False
+    return any(
+        entry.is_file() and entry.name.startswith(PERFORMANCE_OPS_PERF_PREFIX)
+        for entry in entries
+    )
+
+
+def _synthetic_remote_path(configured_path: Optional[str], report_name: str) -> str:
+    """Build a remotePath whose final segment is the local folder name (used by /use)."""
+    if configured_path:
+        return f"{configured_path.rstrip('/')}/{report_name}"
+    return report_name
+
+
+def list_local_synced_report_folders(
+    *,
+    remote_data_directory: Path,
+    host: str,
+    report_directory_name: str,
+    configured_remote_path: Optional[str],
+    is_valid_report_dir: Callable[[Path], bool],
+) -> List[RemoteReportFolder]:
+    """
+    List report folders under REMOTE_DATA_DIRECTORY/<host>/<report_directory_name>/.
+
+    Matches the layout written by sync_remote_*_folders / PathResolver.
+    Only directories that pass ``is_valid_report_dir`` are included.
+    """
+    host_reports_dir = remote_data_directory / host / report_directory_name
+    if not host_reports_dir.is_dir():
+        logger.info("No local synced reports directory at %s", host_reports_dir)
+        return []
+
+    folders: List[RemoteReportFolder] = []
+    for entry in sorted(host_reports_dir.iterdir(), key=lambda p: p.name.lower()):
+        if not entry.is_dir() or entry.name.startswith("."):
+            continue
+        if not is_valid_report_dir(entry):
+            logger.debug("Skipping incomplete local report folder: %s", entry)
+            continue
+
+        last_synced = read_last_synced_file(str(entry))
+        # Without a remote lastModified, treat synced copies as current so select
+        # does not force an SSH refresh while offline.
+        try:
+            mtime = int(entry.stat().st_mtime)
+        except OSError:
+            mtime = 0
+        last_modified = last_synced if last_synced is not None else mtime
+
+        folders.append(
+            RemoteReportFolder(
+                reportName=entry.name,
+                remotePath=_synthetic_remote_path(configured_remote_path, entry.name),
+                lastModified=last_modified,
+                lastSynced=last_synced,
+            )
+        )
+
+    folders.sort(key=lambda folder: folder.lastModified, reverse=True)
+    return folders
+
+
+def list_local_synced_profiler_folders(
+    connection: RemoteConnection,
+    remote_data_directory: Path,
+    profiler_directory_name: str,
+) -> List[RemoteReportFolder]:
+    if not connection.profilerPath:
+        return []
+    return list_local_synced_report_folders(
+        remote_data_directory=remote_data_directory,
+        host=connection.host,
+        report_directory_name=profiler_directory_name,
+        configured_remote_path=connection.profilerPath,
+        is_valid_report_dir=is_valid_local_profiler_report_dir,
+    )
+
+
+def list_local_synced_performance_folders(
+    connection: RemoteConnection,
+    remote_data_directory: Path,
+    performance_directory_name: str,
+) -> List[RemoteReportFolder]:
+    if not connection.performancePath:
+        return []
+    return list_local_synced_report_folders(
+        remote_data_directory=remote_data_directory,
+        host=connection.host,
+        report_directory_name=performance_directory_name,
+        configured_remote_path=connection.performancePath,
+        is_valid_report_dir=is_valid_local_performance_report_dir,
+    )
--- a/backend/ttnn_visualizer/local_remote_reports.py
+++ b/backend/ttnn_visualizer/local_remote_reports.py
@@ -50,7 +50,10 @@ def _is_valid_local_performance_report_dir(directory: Path) -> bool:
     if not all((directory / name).is_file() for name in PERFORMANCE_REQUIRED_FILES):
         return False
     try:
-        return any(directory.glob(f"{PERFORMANCE_OPS_PERF_PREFIX}*"))
+        # Require a matching file — a directory named ops_perf_results* is not enough.
+        return any(
+            path.is_file() for path in directory.glob(f"{PERFORMANCE_OPS_PERF_PREFIX}*")
+        )
     except OSError:
         return False
 
--- a/backend/ttnn_visualizer/local_remote_reports.py
+++ b/backend/ttnn_visualizer/local_remote_reports.py
@@ -22,12 +22,26 @@ PERFORMANCE_REQUIRED_FILES = frozenset(
 PERFORMANCE_OPS_PERF_PREFIX = "ops_perf_results"
 
 
-def is_valid_local_profiler_report_dir(directory: Path) -> bool:
+def local_synced_report_path(
+    remote_data_directory: Path,
+    host: str,
+    report_directory_name: str,
+    report_name: Optional[str] = None,
+) -> Path:
+    """REMOTE_DATA_DIRECTORY/<host>/<report_directory_name>[/<report_name>].
+
+    Must match sync_remote_*_folders / PathResolver remote layout.
+    """
+    base = remote_data_directory / host / report_directory_name
+    return base / report_name if report_name is not None else base
+
+
+def _is_valid_local_profiler_report_dir(directory: Path) -> bool:
     """True when the folder has the files needed to load a memory/profiler report."""
     return all((directory / name).is_file() for name in PROFILER_REQUIRED_FILES)
 
 
-def is_valid_local_performance_report_dir(directory: Path) -> bool:
+def _is_valid_local_performance_report_dir(directory: Path) -> bool:
     """True when the folder has the files needed to load a performance report.
 
     Matches local performance folder listing / upload validation:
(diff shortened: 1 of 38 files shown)


src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -274,6 +478,25 @@ const RemoteSyncConfigurator = () => {
...
+
+        if (connection) {
+            loadLocalSyncedFolders(connection).catch(() => {
+                // Local scan is best-effort; cached folders remain if it fails.
+            });
+        }
+        // Intentionally once on mount for the persisted selection; connection changes go
+        // through updateSelectedConnection which also calls loadLocalSyncedFolders.
+        // eslint-disable-next-line react-hooks/exhaustive-deps -- mount seed only
+    }, []);
```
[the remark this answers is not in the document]
dcblundell: Fixed: mount-seed effect now returns a cleanup that aborts `localSyncedFoldersAbortRef`, so unmount cancels in-flight local scans before any `updateSaved*` / setState.

backend/ttnn_visualizer/local_remote_reports.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Matches local performance folder listing / upload validation:
+    profile_log_device.csv, tracy_profile_log_host.tracy, and ops_perf_results*.
+    """
+    if not all((directory / name).is_file() for name in PERFORMANCE_REQUIRED_FILES):
+        return False
+    try:
+        return any(directory.glob(f"{PERFORMANCE_OPS_PERF_PREFIX}*"))
+    except OSError:
+        return False
```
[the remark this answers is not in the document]
dcblundell: Fixed: `_is_valid_local_performance_report_dir` now requires `path.is_file()` for `ops_perf_results*` matches. Added a regression test where only a directory named `ops_perf_results_0` is present.

backend/ttnn_visualizer/local_remote_reports.py
```diff
@@ -0,0 +1,156 @@
...
+    host: str,
+    report_directory_name: str,
+    report_name: Optional[str] = None,
+) -> Path:
+    """REMOTE_DATA_DIRECTORY/<host>/<report_directory_name>[/<report_name>].
+
+    Must match sync_remote_*_folders / PathResolver remote layout.
+    """
+    base = remote_data_directory / host / report_directory_name
+    return base / report_name if report_name is not None else base
```
aidemsined: > [!CAUTION]
> `local_synced_report_path` re-implements the disk layout that `PathResolver.get_base_report_path` already owns. The segment-order bug this PR fixes in the existing remote-list handlers is exactly the failure mode a second source of truth produces. Have this delegate to `PathResolver` (or route both through it) so the sync destination, mount resolution, and local listing can't drift apart again.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -427,22 +610,21 @@ const RemoteSyncConfigurator = () => {
...
-                                    }
+                            setSelectedReportFolder(folder);
+
+                            const response = await remote.mountRemoteFolder(
+                                remote.persistentState.selectedConnection,
+                                folder,
+                            );
+
+                            if (response.status === HttpStatusCode.Ok) {
+                                updateReportSelection(folder);
```
aidemsined: > [!WARNING]
> Select now always mounts the local copy via `/api/remote/use` and never syncs, so choosing a folder the SSH list surfaced but that was never actually synced activates a report whose local files don't exist — `/use` returns 200 and the user gets no feedback. Wrap `mountRemoteFolder` (here and the performance analogue at ~659) in try/catch and surface the failure. ★ flagged independently by the linear pass and the risk/perf review.

backend/ttnn_visualizer/local_remote_reports.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    if not all((directory / name).is_file() for name in PERFORMANCE_REQUIRED_FILES):
+        return False
+    try:
+        # Require a matching file — a directory named ops_perf_results* is not enough.
+        return any(
+            path.is_file() for path in directory.glob(f"{PERFORMANCE_OPS_PERF_PREFIX}*")
+        )
+    except OSError:
+        return False
```
aidemsined: > [!WARNING]
> These profiler/performance completeness rules now exist in ≥3 places with slight divergence — here (`ops_perf_results*` file), `utils.py` (~872-878, `ops_perf_results*.csv`), and inline in `views.py` (~1075-1080). Pull the required-file sets plus the ops-perf check into one shared helper so the offline listing and the upload/discovery paths can't accept different things. ★ flagged by the linear pass and the architecture review.

backend/ttnn_visualizer/local_remote_reports.py
```diff
@@ -0,0 +1,156 @@
...
+        last_synced = read_last_synced_file(str(entry))
+        # Without a remote lastModified, treat synced copies as current so select
+        # does not force an SSH refresh while offline.
+        if last_synced is not None:
+            last_modified = last_synced
+        else:
+            try:
+                last_modified = int(entry.stat().st_mtime)
+            except OSError:
+                last_modified = 0
```
aidemsined: > [!WARNING]
> `read_last_synced_file` does an unguarded `int(...)` on the marker contents, so a corrupt/empty/unreadable `.last-synced` throws `ValueError`/`OSError`. That propagates to the local-list endpoint, which only catches `pydantic.ValidationError`, so a single bad marker 500s the entire host listing and defeats the offline fallback. The mtime branch here only covers the missing (`None`) case — guard the parse (or catch broadly) so one bad marker degrades to mtime instead of failing the whole scan.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -101,22 +124,136 @@ const RemoteSyncConfigurator = () => {
...
+
+            // Always replace cached lists on a successful scan — including []. Otherwise
+            // never-synced / empty local dirs from an older remote fetch stay visible.
+            if (localProfilerFolders.status === 'fulfilled') {
+                updateSavedReportFolders(connection, localProfilerFolders.value);
+            }
+
+            if (localPerformanceFolders.status === 'fulfilled') {
+                updateSavedPerformanceFolders(connection, localPerformanceFolders.value);
+            }
```
aidemsined, line 159: > [!WARNING]
> Between the `signal.aborted` check above and these `updateSaved*` calls nothing re-checks that this scan is still the current one. A scan that resolves just as a newer connection's scan supersedes it can write folders into the wrong connection's UI. Guard with `localSyncedFoldersAbortRef.current === abortController` (the same identity check the `finally` already uses) before applying results.

backend/ttnn_visualizer/local_remote_reports.py
```diff
@@ -0,0 +1,156 @@
...
+
+        folders.append(
+            RemoteReportFolder(
+                reportName=entry.name,
+                remotePath=_synthetic_remote_path(configured_remote_path, entry.name),
+                lastModified=last_modified,
+                lastSynced=last_synced,
+            )
+        )
```
aidemsined, line 93: > [!WARNING]
> Local-only rows can carry `lastSynced = None` here, and `isRemoteFolderOutdated` treats a missing `lastSynced` as always-outdated — so the Sync badge reads “needs refresh” for a copy you just loaded offline, even though select is now mount-only. Either default `lastSynced` from mtime for local-only rows, or scope the outdated predicate to remote-fetched metadata so the list contract and the badge agree.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -101,22 +124,136 @@ const RemoteSyncConfigurator = () => {
...
         }
+
+        // Populate report dropdowns from on-disk synced copies for this host (no SSH).
+        // Fire-and-forget so connection switch stays cache-fast; abort cancels prior scans.
+        loadLocalSyncedFolders(connection).catch(() => {
+            // Local scan is best-effort; cached folders remain if it fails.
+        });
+    };
+
+    const loadLocalSyncedFolders = async (connection: RemoteConnection) => {
```
aidemsined, line 133: > [!WARNING]
> `loadLocalSyncedFolders`, `applyRemoteOrLocalFolderList`, and `fetchRemoteFolderLists` together form a fetch/fallback/abort state machine bolted onto an already ~700-line UI component, while `useRemote.tsx` already owns the HTTP layer. Consider lifting this into a `useRemoteFolderLists` / `refreshFolderListsWithLocalFallback` hook so the offline policy lives next to the transport and the component just wires buttons/effects.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -30,6 +33,24 @@ import { updateInstance, useReportMetadata } from '../../hooks/useAPI';
 import { ActiveReport } from '../../model/APIData';
 import { DBVersionValidation, evaluateDbVersion } from '../../functions/compareDbVersion';
 
+export const LOCAL_SYNCED_REPORTS_TOAST_TITLE = 'Loaded local synced reports';
+export const LOCAL_SYNCED_REPORTS_TOAST_DETAIL =
+    'Remote host unreachable; showing reports already synced on this machine.';
+export const FOLDER_LIST_SYNC_ERROR_TOAST_TITLE = 'Folder list sync error';
```
aidemsined: > [!WARNING]
> These list-level fallback toast strings live inline in the component, while the per-folder fallback copy lives in `notifyFolderSyncLocalFallback.ts` (mirroring `notifyFolderSyncError.ts`). Same UX family, two homes — co-locate these with the notify helpers so the toast copy is discoverable in one place.

backend/ttnn_visualizer/local_remote_reports.py
dcblundell: Fixed: introduced `remote_host_report_path` as the canonical layout helper in `utils.py`. `PathResolver.get_base_report_path` (remote mode) and the offline listing path both go through it, so sync destination / mount resolution / local listing share one source of truth.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Fixed: select handlers now try/catch `mountRemoteFolder` and toast via `notifyRemoteFolderMountError`. Also tightened `/api/remote/use` to return 404 when the local synced report is missing/incomplete (try/catch alone would not catch the previous 200-with-missing-files case).

backend/ttnn_visualizer/local_remote_reports.py
dcblundell: Fixed: shared `is_valid_profiler_report_dir` / `is_valid_performance_report_dir` (plus required-file constants) in `utils.py`. Offline listing, `/profiler`, `/performance`, and `get_available_reports` discovery all use them.

backend/ttnn_visualizer/local_remote_reports.py
dcblundell: Fixed: `read_last_synced_file` now catches `ValueError`/`OSError` and returns `None`, so a corrupt marker falls back to mtime instead of 500ing the whole local listing.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Fixed: before applying `updateSaved*`, we now require `localSyncedFoldersAbortRef.current === abortController` (in addition to `signal.aborted`), matching the `finally` identity check.

backend/ttnn_visualizer/local_remote_reports.py
dcblundell: Fixed: local-only rows without a `.last-synced` marker now set both `lastSynced` and `lastModified` from mtime, so `isRemoteFolderOutdated` no longer treats them as always-outdated.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Agreed this belongs in a dedicated hook next to the transport layer, but pulling the fetch/fallback/abort state machine out of `RemoteSyncConfigurator` is a larger refactor than this PR’s scope. Deferring to a follow-up.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Fixed: list-level fallback/error toast copy and helpers now live with the notify modules (`notifyLocalSyncedReportsListFallback` / `notifyFolderListSyncError`), alongside the per-folder fallback helpers.

src/functions/reportLinks.ts
```diff
@@ -4,71 +4,173 @@
...
+    }
+
+    return null;
+};
+
+const pathBasename = (value: string): string => {
+    const normalised = value.replace(/\\/g, '/');
+    const segments = normalised.split('/').filter(Boolean);
+    return segments.length > 0 ? segments[segments.length - 1]! : normalised.trim();
+};
```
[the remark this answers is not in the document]
dcblundell: Fixed: pathBasename now returns empty string when there are no non-empty path segments (e.g. '/', '\'), so getReportId falls through instead of treating '/' as an id. Added a unit test.

src/functions/reportLinks.ts
```diff
@@ -4,71 +4,173 @@
...
+
+export const capReportLinks = (links: ReportLink[]): ReportLink[] =>
+    links.length <= MAX_REPORT_LINKS ? links : links.slice(links.length - MAX_REPORT_LINKS);
 
 /**
- * Performance report paths recorded as linked with the given (active) memory report.
+ * Insert or update a pair (identity ignores status). Moves the pair to the end so
+ * recency-based capping keeps actively compared pairs. Returns the same reference
+ * when the pair already exists with the same status (avoids jotai update loops).
  */
```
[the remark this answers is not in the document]
dcblundell: Fixed: updated the upsertReportLink doc comment to match behaviour — same-status hits return the existing array reference and do not bump recency/recordedAt (intentional to avoid jotai update loops).

tests/remoteFolderSelector.spec.tsx
```diff
@@ -239,39 +249,197 @@ it('handles API errors gracefully', () => {
...
+        <TestProviders>
+            <RemoteSyncConfigurator />
+        </TestProviders>,
+    );
+
+    await selectPerformanceFolder(selectedReport.remotePath);
+
+    const { remotePath } = selectedReport;
+    const formattedPath = remotePath.split('/').slice(-1);
+
```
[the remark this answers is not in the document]
dcblundell: Fixed: extract the final path segment as a string (.filter(Boolean).at(-1)) instead of passing a string[] from .slice(-1) into chai .contain.

backend/ttnn_visualizer/views.py
```diff
@@ -1488,78 +1485,103 @@ def create_npe_files():
...
+    dir_name = current_app.config[directory_config_key]
+    for rf in folders:
+        directory_name = Path(rf.remotePath).name
+        local_path = local_synced_report_path(
+            remote_data, host, dir_name, directory_name
+        )
+        logger.info("Checking last synced for %s", directory_name)
+        rf.lastSynced = read_last_synced_file(str(local_path))
+        if not rf.lastSynced:
+            logger.info("%s not yet synced", directory_name)
```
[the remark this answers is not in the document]
dcblundell: Fixed: lowered the per-folder last-synced checks in _annotate_last_synced from INFO to DEBUG.

Issue #1743, opened 2026-07-08
MLIR viewer: Expand-all progress affordance + tight-tree default + fitView constants


aidemsined:
MLIR audit follow-up (#1734) — **Group E**. User-visible UX; own review boundary.

## Scope

1. **Progress affordance for Expand-all.** Currently the button click arms `pendingFitAllRef` and sets `expandedNamespaces` to the full set; the worker rebuild takes 1–5 seconds on realistic MLIR graphs with no user feedback — the button feels broken. Show a spinner (or disable the button with an inline `laying out N nodes…` label) while the rebuild is in flight. The state exists — `runBuild` posts, `onBuilt` returns — no new plumbing needed.
2. **Default `dagreLayout` to `tight-tree` without the try/catch fallback**, or gate the fallback on `nodes.length > 5000`. `src/components/mlir/mlirGraphBuilder.ts:52-59` currently tries `tight-tree` first and re-runs with default network-simplex on catch — for extreme graphs the retry path silently doubles the layout cost.
3. **Replace `fitView({ padding: 0.X, duration: 200 })` with `duration: 0`** for the initial fit (`MLIRViewReactFlow.tsx:517`) and Expand-all. Keep 200ms only for local jumps (recenter-on-selected, navigateToNode, goToMatch). Also extract the fitView constants (`padding: 0.2 | 0.3`, `duration: 200 | 0`) as `MLIR_FIT_VIEW_OPTIONS` in `src/definitions/` — already flagged in #1734 seed items.

## Acceptance criteria

- Expand-all shows visible progress while the worker builds; button is disabled during in-flight builds.
- Initial fit and bulk fits are instant (`duration: 0`); local jumps still animate.
- All fitView calls read from `MLIR_FIT_VIEW_OPTIONS` — no more inlined numeric literals.
- No hang, no double-fitView.

## Effort

~1 engineer-day.

## Prerequisite

None. Independent.

PR #1778, opened 2026-07-17
MLIR viewer: expand-all progress, fitView constants, dagre size-gate

aidemsined:
## Summary

MLIR audit follow-up (#1743). Three independent, low-risk improvements to the MLIR graph viewer:

- **Expand-all progress affordance** — the worker rebuild can take 1–5s with no feedback today. `useMlirLayoutWorker` now exposes `isBuilding`; while a rebuild is in flight the expand/collapse controls disable and show a "Laying out N nodes…" spinner. A `worker.onerror` clears the flag so an uncaught worker failure can't strand the controls disabled with no recovery.
- **Centralized fitView tuning** — `MLIR_FIT_VIEW_OPTIONS` (`bulk` = instant, `localJump` = animated). Initial load and expand/collapse-all fit instantly; recenter / navigate / match-cycling keep the short animation. No more inline padding/duration literals.
- **dagre size-gate** — graphs above `DAGRE_NODE_LIMIT` take a single `network-simplex` pass; smaller graphs keep `tight-tree` with a fallback, avoiding a double layout at the scale that hurts most.

## Test plan

- [x] `tsc --noEmit`, `eslint`, production build all clean
- [x] `useMlirLayoutWorker`: `isBuilding` flips true on dispatch and clears on built / error / worker crash; stale replies ignored
- [x] Controls: buttons lock + status region shows the node count while building; callbacks don't fire mid-build
- [x] `MLIR_FIT_VIEW_OPTIONS`: bulk instant, localJump animated
- [ ] Manual: expand-all on a large graph shows the spinner; initial/bulk fits are instant, local jumps animate

Refs #1743

### DIFF
--- a/src/components/mlir/MLIRViewReactFlow.tsx
+++ b/src/components/mlir/MLIRViewReactFlow.tsx
@@ -48,6 +48,7 @@ import type {
     WorkerNode,
 } from './mlirGraphTypes';
 import { GRAPH_COLORS } from '../../definitions/GraphColors';
+import { MLIR_FIT_VIEW_OPTIONS } from '../../definitions/MlirFitView';
 import { useMlirLayoutWorker } from './useMlirLayoutWorker';
 import MlirNodeDetailsPanel from './MlirNodeDetailsPanel';
 import MlirOpFilter, { MlirOpFilterHandle } from './MlirOpFilter';
@@ -603,12 +604,12 @@ const MlGraphInner = ({ data }: ViewProps) => {
                 // isn't in the build (e.g. synthetic id that never reaches the
                 // canvas) — a missing fitView is preferable to a noisy error.
                 requestAnimationFrame(() => {
-                    void fitView({ nodes: [{ id: pendingFocusId }], padding: 0.3, duration: 200 });
+                    void fitView({ nodes: [{ id: pendingFocusId }], ...MLIR_FIT_VIEW_OPTIONS.localJump });
                 });
             } else if (shouldFitAll || !hasFitInitiallyRef.current) {
                 hasFitInitiallyRef.current = true;
                 requestAnimationFrame(() => {
-                    void fitView({ padding: 0.2, duration: 200 });
+                    void fitView({ ...MLIR_FIT_VIEW_OPTIONS.bulk });
                 });
             }
         },
@@ -643,7 +644,7 @@ const MlGraphInner = ({ data }: ViewProps) => {
         }
         return result;
     }, [sourceNodes]);
-    const { interactionIndex, runBuild } = useMlirLayoutWorker(graph.id, sourceNodes, applyBuiltGraph);
+    const { interactionIndex, runBuild, isBuilding } = useMlirLayoutWorker(graph.id, sourceNodes, applyBuiltGraph);
 
     useEffect(() => {
         runBuild(expandedNamespaces);
@@ -738,7 +739,7 @@ const MlGraphInner = ({ data }: ViewProps) => {
             }
             const targetId = matchedNodesInOrder[nextIdx];
             if (targetId) {
-                void fitView({ nodes: [{ id: targetId }], padding: 0.3, duration: 200 });
+                void fitView({ nodes: [{ id: targetId }], ...MLIR_FIT_VIEW_OPTIONS.localJump });
             }
             setCurrentMatchIndex(nextIdx);
         },
@@ -1187,7 +1188,7 @@ const MlGraphInner = ({ data }: ViewProps) => {
         if (!selectedNodeId) {
             return;
         }
-        void fitView({ nodes: [{ id: selectedNodeId }], padding: 0.3, duration: 200 });
+        void fitView({ nodes: [{ id: selectedNodeId }], ...MLIR_FIT_VIEW_OPTIONS.localJump });
     }, [fitView, selectedNodeId]);
 
     // Click handler for the "locate" affordance next to each producer /
@@ -1220,7 +1221,7 @@ const MlGraphInner = ({ data }: ViewProps) => {
                 }
             }
             if (prefixesToAdd.length === 0) {
-                void fitView({ nodes: [{ id: targetNodeId }], padding: 0.3, duration: 200 });
+                void fitView({ nodes: [{ id: targetNodeId }], ...MLIR_FIT_VIEW_OPTIONS.localJump });
                 return;
             }
             viewportAnchorRef.current = null;
@@ -1552,6 +1553,8 @@ const MlGraphInner = ({ data }: ViewProps) => {
                 <MlirExpandCollapseControls
                     namespaceCount={allExpandableNamespaces.length}
                     expandedCount={expandedNamespaces.size}
+                    isBuilding={isBuilding}
+                    nodeCount={sourceNodes.length}
                     onExpandAll={expandAllNamespaces}
                     onCollapseAll={collapseAllNamespaces}
                 />
--- a/src/components/mlir/MlirExpandCollapseControls.tsx
+++ b/src/components/mlir/MlirExpandCollapseControls.tsx
@@ -3,7 +3,7 @@
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
 import { memo } from 'react';
-import { Button, ButtonVariant, Size, Tooltip } from '@blueprintjs/core';
+import { Button, ButtonVariant, Size, Spinner, Tooltip } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import 'styles/components/MlirExpandCollapseControls.scss';
 
@@ -12,6 +12,9 @@ export interface MlirExpandCollapseControlsProps {
     expandedCount: number;
     onExpandAll: () => void;
     onCollapseAll: () => void;
+    // Worker rebuild in flight: lock the buttons and show progress.
+    isBuilding?: boolean;
+    nodeCount?: number;
 }
 
 const MlirExpandCollapseControlsInner = ({
@@ -19,9 +22,11 @@ const MlirExpandCollapseControlsInner = ({
     expandedCount,
     onExpandAll,
     onCollapseAll,
+    isBuilding = false,
+    nodeCount,
 }: MlirExpandCollapseControlsProps) => {
-    const canExpand = namespaceCount > 0 && expandedCount < namespaceCount;
-    const canCollapse = expandedCount > 0;
+    const canExpand = !isBuilding && namespaceCount > 0 && expandedCount < namespaceCount;
+    const canCollapse = !isBuilding && expandedCount > 0;
     return (
         <div className='mlir-expand-collapse-controls'>
             <Tooltip
@@ -52,6 +57,15 @@ const MlirExpandCollapseControlsInner = ({
                     onClick={onCollapseAll}
                 />
             </Tooltip>
+            {isBuilding && (
+                <span
+                    className='mlir-layout-status'
+                    role='status'
+                >
+                    <Spinner size={14} />
+                    <span>Laying out{typeof nodeCount === 'number' ? ` ${nodeCount} nodes` : ''}…</span>
+                </span>
+            )}
         </div>
     );
 };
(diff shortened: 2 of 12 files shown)


src/components/mlir/useMlirLayoutWorker.ts
```diff
@@ -122,9 +133,10 @@ export function useMlirLayoutWorker(
                 expandedNamespaces: expandedSorted,
```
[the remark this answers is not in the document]
aidemsined: Fixed in 441c62bf — `runBuild`'s `postMessage` is now wrapped in try/catch; a synchronous send failure (InvalidStateError / DataCloneError on a crashed or terminated worker) clears `isBuilding` and logs instead of propagating into the React tree. Added a `useMlirLayoutWorker` regression test asserting a throwing `postMessage` doesn't throw out of `runBuild` and leaves `isBuilding` false.

src/components/mlir/useMlirLayoutWorker.ts
```diff
@@ -34,17 +34,26 @@ export function useMlirLayoutWorker(
...
 
     useEffect(() => {
         const worker = new Worker(new URL('./mlirLayoutWorker.ts', import.meta.url), { type: 'module' });
         workerRef.current = worker;
+        // A worker-level crash (uncaught throw, structured-clone failure on
+        // post) emits no terminal reply, which would strand `isBuilding` true
+        // and lock the expand/collapse controls with no recovery.
+        worker.onerror = () => {
+            setIsBuilding(false);
+        };
```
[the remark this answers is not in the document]
aidemsined: Addressed in 441c62bf. The concrete crash vector you flag — a subsequent `runBuild()` throwing on `postMessage` after a crash — is now caught: the dispatch is wrapped in try/catch and clears `isBuilding` on failure, so a dead worker degrades to a no-op rebuild rather than crashing the React tree. I deliberately did *not* terminate + null `workerRef` in `onerror`: an `error` event doesn't guarantee the worker is dead, and the guarded dispatch already makes future posts safe. A genuinely dead worker is replaced on the next graph switch (the inner view remounts on `graphId`). Happy to add explicit teardown + respawn as a follow-up if you'd prefer it.

src/components/mlir/MlirExpandCollapseControls.tsx
```diff
@@ -12,16 +12,21 @@ export interface MlirExpandCollapseControlsProps {
     expandedCount: number;
     onExpandAll: () => void;
     onCollapseAll: () => void;
+    // Worker rebuild in flight: lock the buttons and show progress.
```
dcblundell: Is this comment useful?

dcblundell (changes requested):

src/components/mlir/mlirLayoutWorker.ts
```diff
@@ -64,13 +64,17 @@ function processLatestBuild(graphId: string): void {
...
-                    });
+                    try {
+                        postMessage({
+                            type: 'built',
+                            requestId: request.requestId,
+                            graphId,
+                            cacheKey: request.cacheKey,
+                            graph: cached,
+                        });
+                    } catch (error) {
```
dcblundell: **Should fix:** Extract shared `postBuilt` helper

Two near-identical `type: 'built'` posts (cache-hit here vs fresh-build ~88–94). Only the cache-hit path wraps `postMessage` in `try/catch`; the fresh-build path relies on the outer `buildVisibleGraph` `try` and diverges on failure gating. Extract one `postBuilt` / `safePostBuilt` used by both arms so post-failure handling cannot drift again.

src/components/mlir/useMlirLayoutWorker.ts
```diff
@@ -34,17 +34,26 @@ export function useMlirLayoutWorker(
 ): {
     interactionIndex: WorkerInteractionIndex | null;
     runBuild: (expanded: Set<string>) => void;
+    isBuilding: boolean;
```
dcblundell, line 41: **Should fix:** Update stale module JSDoc for `isBuilding`

True location: module JSDoc at `useMlirLayoutWorker.ts:~8–28`. It still lists responsibilities only through `runBuild` / `interactionIndex`, but the hook now returns `isBuilding` and owns crash/`postMessage` recovery for that flag. Stale comments are bugs per CONVENTIONS — add a responsibility covering the `isBuilding` lifecycle (set on dispatch; clear on built/error/onerror/sync post failure).

src/components/mlir/MlirExpandCollapseControls.tsx
```diff
@@ -52,6 +57,15 @@ const MlirExpandCollapseControlsInner = ({
                     onClick={onCollapseAll}
                 />
             </Tooltip>
+            {isBuilding && (
+                <span
+                    className='mlir-layout-status'
+                    role='status'
+                >
+                    <Spinner size={14} />
+                    <span>Laying out{typeof nodeCount === 'number' ? ` ${nodeCount} nodes` : ''}…</span>
```
dcblundell: **Should fix:** Promote layout-status copy to a named constant

User-visible `Laying out…` / `Laying out N nodes…` is inlined here and re-asserted as string literals in `tests/MlirExpandCollapseControls.spec.tsx`. Prefer a `SCREAMING_SNAKE_CASE` constant or small label helper shared by component and specs so the status text cannot drift.

src/components/mlir/mlirGraphBuilder.ts
```diff
@@ -48,12 +48,14 @@ function dagreLayout(nodes: WorkerNode[], edges: WorkerEdge[]): WorkerNode[] {
     if (nodes.length === 0) {
         return nodes;
     }
-    const opts: DagreOptions =
-        nodes.length > DAGRE_NODE_LIMIT
-            ? { nodesep: 10, ranksep: 50, edgesep: 5, ranker: 'tight-tree' }
-            : { ranker: 'tight-tree' };
+    // Large graphs: single network-simplex pass — tight-tree can throw, and a
+    // retry would double the layout cost at exactly the size that hurts most.
+    if (nodes.length > DAGRE_NODE_LIMIT) {
```
dcblundell, line 53: **Should fix:** Assert the dagre size-gate in tests

`tests/mlirGraphBuilder.spec.ts` only checks that graphs above `DAGRE_NODE_LIMIT` don't crash + invariants. A regression back to `tight-tree` above the limit would still pass. Mock dagre `setGraph` and assert `ranker` is default/`network-simplex` (single layout call) above the limit, and `tight-tree` below it.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -1552,6 +1553,8 @@ const MlGraphInner = ({ data }: ViewProps) => {
                 <MlirExpandCollapseControls
                     namespaceCount={allExpandableNamespaces.length}
                     expandedCount={expandedNamespaces.size}
+                    isBuilding={isBuilding}
```
dcblundell, line 1556: **Should fix:** Cover parent wiring of `isBuilding` / `nodeCount`

`isBuilding` and `nodeCount={sourceNodes.length}` are wired only here; leaf + hook tests pass even if these props are dropped. Add a thin mount that mocks `useMlirLayoutWorker` / React Flow and asserts controls receive `isBuilding` and show `Laying out N nodes…` for `sourceNodes.length`.

src/components/mlir/MLIRViewReactFlow.tsx
```diff
@@ -603,12 +604,12 @@ const MlGraphInner = ({ data }: ViewProps) => {
...
                 // canvas) — a missing fitView is preferable to a noisy error.
                 requestAnimationFrame(() => {
-                    void fitView({ nodes: [{ id: pendingFocusId }], padding: 0.3, duration: 200 });
+                    void fitView({ nodes: [{ id: pendingFocusId }], ...MLIR_FIT_VIEW_OPTIONS.localJump });
                 });
             } else if (shouldFitAll || !hasFitInitiallyRef.current) {
                 hasFitInitiallyRef.current = true;
                 requestAnimationFrame(() => {
-                    void fitView({ padding: 0.2, duration: 200 });
+                    void fitView({ ...MLIR_FIT_VIEW_OPTIONS.bulk });
```
dcblundell, line 612: **Should fix:** Assert bulk vs `localJump` fitView options at call sites

`tests/MlirFitView.spec.ts` only snapshots constants. Wrong spread (`localJump` on expand-all / `bulk` on match cycle) would not fail CI. Prefer a focused test that mocks `fitView` and drives `pendingFitAllRef` vs focus id, asserting `duration: 0` (bulk, here) vs `duration: 200` (`localJump`).

dcblundell (commented):
Conductor review: Should-fix items from the combined DRY/TEST/PERF/STANDARDS/GUARD pass (Must-fix: none).

src/components/mlir/MlirExpandCollapseControls.tsx
aidemsined: Fair — removed in 3749fc60. `isBuilding` / `nodeCount` are self-describing and the render below shows the behaviour, so the comment was just narration.

src/components/mlir/mlirLayoutWorker.ts
aidemsined: Done in 3749fc60. Both arms now go through a single `safePostBuilt(requestId, graphId, cacheKey, graph)` that wraps the `postMessage` in try/catch and routes a structured-clone failure to `postWorkerError`. The build-gating (`!latestBuildByGraphId.has(...)` / `!newerRequestExists && !graphVersionChanged`) stays at the call sites; only the post + failure handling is unified, so it can't drift again.

src/components/mlir/useMlirLayoutWorker.ts
aidemsined: Fixed in 3749fc60 — added responsibility #6 to the module JSDoc covering the `isBuilding` lifecycle (set on every `build` dispatch; cleared on the matching `built`/`error` reply, on `worker.onerror`, and on a synchronous `postMessage` failure).

src/components/mlir/MlirExpandCollapseControls.tsx
aidemsined: Done in 3749fc60. Extracted an exported `layoutStatusLabel(nodeCount?)` helper (it has to interpolate the count, so a helper reads better than a bare SCREAMING_SNAKE constant). The component renders it and `MlirExpandCollapseControls.spec.tsx` now asserts against `layoutStatusLabel(42)` / `layoutStatusLabel()` instead of string literals, so the copy can't drift.

src/components/mlir/mlirGraphBuilder.ts
aidemsined: Added in 3749fc60. `dagreLayout` is called both per-group (interiors) and top-level, so a large graph doesn't necessarily produce a single >limit pass through `buildVisibleGraph` — a mount-level assertion would be brittle. Instead I exported `dagreLayout` / `DAGRE_NODE_LIMIT` and added `tests/mlirDagreSizeGate.spec.ts`, which mocks dagre to record the `ranker` per graph: `tight-tree` (single pass) at/below the limit, default `network-simplex` (single pass) above it, plus a regression guard that it never falls back to `tight-tree` above the limit.

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: Agree it's a real gap, but I'd rather not land a second full `MLIRViewReactFlow` mount harness here: this branch predates the `MLIRViewMiniMap.spec.tsx` harness (#1773, now on `dev`) and the whole view is about to be decomposed in #1740 (MLIR viewer: extract pure-function modules). The natural home for the composer-wiring characterization specs — including this `isBuilding` / `nodeCount={sourceNodes.length}` seam — is that decomposition, reusing the existing mount harness rather than duplicating one on a behind-`dev` branch. Happy to fold it in there (or add it here if you'd prefer it not wait).

src/components/mlir/MLIRViewReactFlow.tsx
aidemsined: This one is squarely covered by planned work: the bulk-vs-`localJump` selection at the call sites (`pendingFitAllRef` vs focus id) is exactly the `armFitAll` / `consumeAfterBuild` logic being extracted into `useMlirViewportBaton` in #1740. Once that's a standalone module the "which fitView options at which call site" assertion becomes a clean unit test instead of a `fitView`-mock against the monolith. Kept the constant snapshot here and I'll add the call-site coverage with the baton extraction.

PR #1780, opened 2026-07-20
dev: relax version bound on tt-perf-report

mert-kurttutan:
This relaxes the strict version control.
This makes it more convenient for some of the setup I had and also we have uv.lock that can take care of exact versioning.

I could not see any justification on this strict condition so I am not sure about this one hence it being a  draft PR.

### DIFF
--- a/backend/ttnn_visualizer/tests/test_perf_report_contract.py
+++ b/backend/ttnn_visualizer/tests/test_perf_report_contract.py
@@ -22,7 +22,7 @@ Chain of trust: tt-perf-report  ->  this golden  ->  the frontend.
 
 To refresh the golden after an intentional tt-perf-report change:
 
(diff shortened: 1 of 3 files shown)


dcblundell:
We use strict versioning because `tt-perf-report` outputs formatted data which the visualizer consumes. In the past updates have broken functionality in the visualizer, hence the strict version control.

What problem are you trying to solve?

mert-kurttutan:
I am usually using the latest versions available, which, for the case of tt-perf-report, is 1.2.8. This is certainly not crucial but rather minor convenience on my side. Can be closed if you want.

dcblundell:
> I am usually using the latest versions available, which, for the case of tt-perf-report, is 1.2.8. This is certainly not crucial but rather minor convenience on my side. Can be closed if you want.

Yeah for the reason I outlined above we want to keep this, will close. I didn't notice we'd gotten so many versions behind so I'll look into upgrading this week.

PR #1792, opened 2026-07-22
Load up to two MLIR files

dcblundell:
Screenshot 2026-07-22 at 12 50 30

Screenshot 2026-07-22 at 12 50 15

### DIFF
--- a/src/components/mlir/MlirJsonFileLoader.tsx
+++ b/src/components/mlir/MlirJsonFileLoader.tsx
@@ -13,9 +13,9 @@ import {
     activeMlirJsonAtom,
     mlirFileResultsAtom,
     mlirFileResultsOpenAtom,
+    mlirLoadedReportsAtom,
     mlirRetryFilesAtom,
     mlirRetryServerAtom,
-    setComparisonMlirAtom,
 } from '../../store/app';
 import { GraphBundle, MlirFileResult } from '../../model/MLIRJsonModel';
 import getResponseError from '../../functions/getResponseError';
@@ -51,13 +51,18 @@ const MlirJsonFileLoader = ({ server = null }: MlirJsonFileLoaderProps) => {
     const setMlirFileResultsOpen = useSetAtom(mlirFileResultsOpenAtom);
     const setMlirRetryFiles = useSetAtom(mlirRetryFilesAtom);
     const setMlirRetryServer = useSetAtom(mlirRetryServerAtom);
-    const setComparisonMlir = useSetAtom(setComparisonMlirAtom);
+    const setMlirLoadedReports = useSetAtom(mlirLoadedReportsAtom);
     const [uploadStatus, setUploadStatus] = useState<ConnectionTestStates>(ConnectionTestStates.IDLE);
 
-    // Publish a fresh batch of results and open the overlay. Clearing comparison
-    // here (and before pending server rows) avoids a stale cross-file right pane.
+    // Drop split peers on a new results batch so a stale second report cannot
+    // linger under a previous primary. Keep index 0 if present.
+    const clearSplitPeers = () => {
+        setMlirLoadedReports((current) => current.slice(0, 1));
+    };
+
+    // Publish a fresh batch of results and open the overlay.
     const showResults = (results: MlirFileResult[]) => {
-        setComparisonMlir(null);
+        clearSplitPeers();
         setMlirFileResults(results);
         setMlirFileResultsOpen(true);
     };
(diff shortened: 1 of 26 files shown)


src/components/mlir/MlirJsonFileLoader.tsx
```diff
@@ -67,11 +76,12 @@ const MlirJsonFileLoader = ({ server = null, disabled = false }: MlirJsonFileLoa
             Array.from(files).map(async (file): Promise<MlirFileResult> => {
                 try {
                     const graph = JSON.parse(await file.text()) as GraphBundle;
+                    const name = sanitiseFileName(file.name);
```
aidemsined: > [!WARNING]
> Local loads derive the report `name` from `sanitiseFileName(file.name)` (extension strip only) with no within-batch de-duplication. The server upload path already disambiguates via `_unique_mlir_name(base_name, unavailable_names)` in `views.py`, but the local path does not.
>
> Two same-stem files in one batch (`model.json` + `model.mlir`) therefore produce identical `MlirFileResult.name` → identical `MlirSplitReport.key` and `<option value>`, and `getReport()` in `MlirSplitView.tsx` first-matches. Net effect: **both split panes silently render the first file** — the compare feature quietly no-ops. `relabelMlirGraphIds` disambiguates graph ids, not report names, so it doesn't cover this.
>
> Suggest mirroring the backend suffix scheme client-side (`model`, `model (2)`, …) at local ingest, plus a regression test loading two same-stem files and asserting two distinct panes. (Expands on @copilot's still-open thread on this line.)

src/store/app.ts
```diff
@@ -141,4 +120,47 @@ export const selectedPerfRowIdAtom = atom<number | null>(null);
...
 // NPE
+export const activeNpeOpTraceAtom = atom<string | null>(null);
 export const altCongestionColorsAtom = atomWithStorage('altCongestionColors', false);
+
+// MLIR
+// Session-loaded MLIR reports (0–2). Index 0 is persisted/nav-active.
+export const mlirLoadedReportsAtom = atom<MlirLoadedReport[]>([]);
+// Derived for nav, restore, and FileInput label. Writing replaces the list with
+// a name-only primary (reload fetch) or clears it.
+export const activeMlirJsonAtom = atom(
```
aidemsined, line 131: > [!NOTE]
> Two optional observations on this atom, both non-blocking:
>
> 1. **The setter reads as a leaky abstraction.** `activeMlirJsonAtom` presents a `string | null` interface to nav/`useAPI`/restore, but *writing* it replaces the entire `mlirLoadedReportsAtom` list (clearing the peer slot except the narrow same-name-with-data no-op). Only `useRestoreInstance` writes it today, behind a length guard, so it's safe — but a future writer expecting "set active filename" would silently drop the peer. Consider making this a **read-only** projection and moving primary-restore to an explicit `mlirLoadedReportsAtom` mutator.
> 2. **The preserve-data branch (`current[0]?.name === name && current[0]?.data`) is untested.** It's the fix from `a9d6a9ab` and non-obvious; a small unit test (re-writing the same name keeps peer at `[1]`) would lock it in.
>
> Bigger-picture (defer-able): a capped `{ primary, peer }` shape would encode primary-vs-comparison semantics in the type and remove the `[0]`/`[1]` index-magic spread across the route, loader, and this atom.

src/hooks/useRestoreInstance.tsx
```diff
@@ -85,7 +87,11 @@ const useRestoreInstance = () => {
             setPerformanceReportLocation(activeReports.performanceLocation);
 
             setActiveNpe(activeReports.npe);
-            setActiveMlirJson(activeReports.mlir);
+            // Writing activeMlirJsonAtom replaces the whole loaded-reports list.
+            // Skip when a View (including multi-file split) already seeded memory.
+            if (store.get(mlirLoadedReportsAtom).length === 0) {
```
aidemsined, line 92: > [!NOTE]
> This skip-when-loaded guard is exactly the race the state migration introduces (restore clobbering a two-file split that's already seeded in memory), but it has no MLIR test coverage. Worth a unit test: with `mlirLoadedReportsAtom` pre-seeded (length > 0), assert restore does **not** call `setActiveMlirJson`. Cheap insurance against a regression that would be hard to spot manually.

dcblundell: Added test.

src/store/app.ts
dcblundell: Gonna ignore atm.

PR #1796, opened 2026-07-23
NPE Error Handling and Data Parsing Improvements

dcblundell:
This pull request introduces several improvements to how NPE (Neural Processing Engine) data is fetched, validated, and handled in the UI, with a focus on robust error handling, better testability, and code maintainability. The changes include centralized management of React Query keys, improved error typing and propagation, new utility functions for parsing and error handling, and enhanced test coverage through test IDs.

**Key improvements and changes:**

### Error Handling and Validation

* Introduced `getNpeValidationErrorFromFetch`, `throwNpeClientAxiosError`, and `parseNpeAxiosResponseData` utility functions to robustly map, throw, and parse errors from NPE fetches, ensuring that client-side and server-side validation errors are handled and surfaced correctly in the UI. [[1]](diffhunk://#diff-a5f9c3315507cf61b04486b81aaee33efbd303dfc6bd30f323ae8776cf1cdb93R1-R50) [[2]](diffhunk://#diff-19e8b6ed22270a4a3cc93ee70da6b96c3e88762fa3c19b1b56fbe2a5fd4c37d4R1-R33) [[3]](diffhunk://#diff-f9d8aaa57d5ad3c7d900dd7fb8db1292181aa1082ed0013566a68c1e55dff88bR1-R36)
* Updated NPE data fetching and validation in `useAPI.tsx` to use these new utilities, replacing ad-hoc error throwing with standardized, typed Axios errors. This improves error reporting for invalid JSON and shape errors, and ensures React Query caches only valid object graphs. [[1]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604R437-R490) [[2]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L592-R650) [[3]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L617-R661) [[4]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L648-R690)

### Centralized Query Key Management

* Moved NPE React Query key constants (`NPE_QUERY_KEY`, `NPE_SUMMARY_QUERY_KEY`, etc.) to `NPEData.ts` for reusability and to prevent magic strings, ensuring cache busting and query isolation work consistently across the codebase. [[1]](diffhunk://#diff-cdb5883f3778cf3517b2ef3deda4244019371bf6ab81e2eb1b41814487c568fcR8-R13) [[2]](diffhunk://#diff-40472c92ef3729fce7652a3fd19420e45d8a8d6da628574fe7d5f6ae6e881278L11-R12) [[3]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604R69-R75) [[4]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L592-R650)

### UI and Testability Enhancements

* Added `data-testid` attributes using new constants in `TestIds.ts` to key NPE UI elements, improving test coverage and enabling more robust UI testing. [[1]](diffhunk://#diff-ae82d9591262b77ba2a2ef854c2cc09fdbbd067490d29d043f17730827b8f1feR54-R57) [[2]](diffhunk://#diff-312c2d0fe5fd948a64592fd95d40e9e5356143de0932350f54d1194bd602959bR56) [[3]](diffhunk://#diff-312c2d0fe5fd948a64592fd95d40e9e5356143de0932350f54d1194bd602959bL391-R395) [[4]](diffhunk://#diff-1928b18315a6a0bf3ab38def56136345f7c57f450f47d7f12b56444656ee041eL57-R60) [[5]](diffhunk://#diff-ade291ae70f8d3c5d5c164d94c3c959343ab126a59e442bf4fb27c1ea2a93b54R136-R138)
* Refactored `NpeWindowedView` to use a single `content` variable and render all NPE error and loading states in a consistent, testable manner, including using `NPEProcessingStatus` for empty trace handling. [[1]](diffhunk://#diff-ade291ae70f8d3c5d5c164d94c3c959343ab126a59e442bf4fb27c1ea2a93b54R10-R13) [[2]](diffhunk://#diff-ade291ae70f8d3c5d5c164d94c3c959343ab126a59e442bf4fb27c1ea2a93b54R67-R117) [[3]](diffhunk://#diff-ade291ae70f8d3c5d5c164d94c3c959343ab126a59e442bf4fb27c1ea2a93b54R136-R138)

### Type and Interface Improvements

* Added `NpeClientErrorKind` and `NpeClientErrorBody` types to distinguish between parse and shape errors, ensuring UI labels and error handling logic are accurate and maintainable.

These changes collectively make NPE data handling more robust, easier to test, and simpler to extend or debug in the future.


src/hooks/useAPI.tsx
```diff
@@ -423,18 +435,72 @@ export const useGetNPEManifest = () => {
...
+};
+
+const NPE_TEXT_GET_OPTIONS = {
+    timeout: NPE_FETCH_TIMEOUT_MS,
+    responseType: 'text' as const,
+    transitional: { forcedJSONParsing: false },
+    // Fail before buffering past Chromium's approximate string ceiling.
+    maxContentLength: NPE_MAX_CONTENT_LENGTH,
+    maxBodyLength: NPE_MAX_CONTENT_LENGTH,
+};
```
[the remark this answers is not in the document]
dcblundell: Updated comment.

src/hooks/useNpeLoadRenderLifecycle.ts
```diff
@@ -0,0 +1,179 @@
...
+    useEffect(() => {
+        hasRenderedViewRef.current = false;
+        loadStartedAtRef.current = null;
+        renderStartedAtRef.current = null;
+        // eslint-disable-next-line react-hooks/set-state-in-effect -- reset gates when report identity changes
+        setHasRenderedView(false);
+        setMountView(false);
+        setLoadTimedOut(false);
+        setRenderTimedOut(false);
+    }, [npeFileName, filepath, selectedDemoKey]);
```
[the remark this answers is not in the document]
dcblundell: Fixing.

aidemsined: > [!CAUTION]
> The timeout screen is terminal on a same-report retry. `loadTimedOut`/`renderTimedOut` reset only on `[npeFileName, filepath, selectedDemoKey]`, so a refetch or retry of the *same* file (window-focus reconnect, or re-selecting the same report) leaves the flags true — `isLoading` never re-trues and `errorCode` stays pinned to `LOAD_TIMEOUT`/`RENDER_TIMEOUT` (`NPE.tsx:97-102`). The "Try loading the report again" copy then leads nowhere. This is Copilot thread #2, marked resolved but not fixed at this head. Reset the gates when `isFetchingData` transitions `false → true` (and, belt-and-suspenders, when `hasValidData` becomes true after a timeout).

src/hooks/useAPI.tsx
```diff
@@ -423,18 +435,74 @@ export const useGetNPEManifest = () => {
...
+    activeNpeRequestAbort?.abort();
+    activeNpeRequestAbort = null;
+};
+
+/** Cancel in-flight NPE downloads and drop cached payloads after a UI timeout. */
+export const discardNpeQueries = (queryClient: QueryClient): void => {
+    abortActiveNpeRequest();
+    for (const queryKey of NPE_QUERY_KEYS) {
+        queryClient.cancelQueries({ queryKey }).catch(() => undefined);
+        queryClient.removeQueries({ queryKey });
```
aidemsined: > [!CAUTION]
> `discardNpeQueries` aborts the download and then immediately re-triggers it. `removeQueries` on `['fetch-npe']`/`['get-npe-timeline']` runs while `useNpe` is still mounted and `enabled` — `isNpeQueryEnabled` (`NPE.tsx:37`) never depends on the timeout flags — so React Query v5 re-initiates the fetch on the still-observed query within a tick. The oversized file that just timed out is re-downloaded, now unbounded (the load-timer effect's `if (!isFetchingData || loadTimedOut) return` blocks re-arming once `loadTimedOut` is true). For the near-OOM large-file case this PR targets, silently re-downloading a body the browser already choked on can crash the tab. Gate the queries' `enabled` on `!loadTimedOut && !renderTimedOut` (or unmount them) so the discard actually sticks.

tests/NPERoute.spec.tsx
aidemsined, line 1: > [!CAUTION]
> Missing the one regression that matters for the two timeout bugs above. This suite covers identity-*change* clearing the timeout (`switch-npe-file`) but not same-file recovery, and it mocks `discardNpeQueries` so the real remove-then-refetch never runs. Add a case: time out on a file, then have the same file's fetch settle with data (no identity change) — assert the timeout UI clears and the view mounts. Ideally also one against a real `QueryClient` with a mounted `useNpe`, asserting the query does not re-fetch after `discardNpeQueries`.

src/routes/NPE.tsx
```diff
@@ -33,30 +52,54 @@ const NPE = () => {
...
+    const { isRendering, isLoading, isDataReady, loadTimedOut, renderTimedOut, mountView, handleViewRendered } =
+        useNpeLoadRenderLifecycle({
+            npeFileName,
+            filepath,
+            selectedDemoKey: selectedDemo?.reportFile ?? null,
+            isFetchingData,
+            hasValidData,
+        });
+
+    let errorCode = fetchErrorCode;
```
aidemsined: > [!WARNING]
> The lifecycle hook leaks raw timeout booleans that the route has to reassemble. `useNpeLoadRenderLifecycle` returns `loadTimedOut`/`renderTimedOut`, and this block re-merges them into `errorCode` here. Consider having the hook own the whole error surface — expose a single `lifecycleErrorCode: NPEValidationError | null` (or a phase enum) — so the route consumes one value instead of rebuilding the state machine.

src/scss/definitions/_variables.scss
```diff
@@ -29,10 +29,12 @@ $npe-highlight-src: #ff0;
 $npe-highlight-dst: #ff4500;
 
 // z-index scale
+// Keep $z-index-3-overlay below Blueprint overlays (`.bp6-overlay` is 20) so
+// in-page floating chrome (e.g. NPE Run summary) cannot cover FileStatusOverlay.
 $z-index-0-background: -1;
 $z-index-1-base: 0;
 $z-index-2-interactive: 1;
-$z-index-3-overlay: 1000;
+$z-index-3-overlay: 10;
```
aidemsined, line 37: > [!WARNING]
> This global `$z-index-3-overlay` change (1000 → 10) is orthogonal to the NPE loading work and touches shared overlay stacking — NPE Run summary, ClusterView, BufferSummary, ProducerConsumers all consume it. The rationale (stay below Blueprint's `.bp6-overlay` at 20) may well be right, but it widens the review/rollback blast radius of a loading-state PR. Consider splitting it into its own change, and either way please eyeball those surfaces stack correctly above page content but below modals.

src/hooks/useAPI.tsx
```diff
@@ -423,18 +435,74 @@ export const useGetNPEManifest = () => {
...
+export const NPE_OP_TRACE_QUERY_KEY = ['fetch-npe'] as const;
+export const NPE_TIMELINE_QUERY_KEY = ['get-npe-timeline'] as const;
+
+const NPE_QUERY_KEYS = [NPE_OP_TRACE_QUERY_KEY, NPE_TIMELINE_QUERY_KEY] as const;
+
+// Own AbortController (not React Query's signal): Strict Mode remount must not
+// abort a multi-hundred-MB download. Timeout / report switch call abortActiveNpeRequest.
+let activeNpeRequestAbort: AbortController | null = null;
+
+export const abortActiveNpeRequest = (): void => {
```
> The `activeNpeRequestAbort` singleton is shared by both NPE queries — `fetchNpeOpTrace` and `fetchNPETimeline` both route through `fetchNpeText`, which aborts any prior request at entry. It's safe today only because `NPE.tsx:37-38` makes op-trace and timeline mutually exclusive; a future caller mounting both enabled would have one silently abort the other's in-flight multi-hundred-MB download with no error surfaced. Worth pinning that invariant with a one-line contract comment (and maybe a test). Extracting this single-flight lifecycle into its own module (mirroring `remoteSyncRequest.ts`) would also keep it out of the 1200-line `useAPI.tsx` and make the contract discoverable.

src/hooks/useAPI.tsx
```diff
@@ -127,6 +127,18 @@ export const updateInstance = async (payload: Partial<Instance>): Promise<Instan
     return response?.data ?? null;
 };
 
+export const INSTANCE_QUERY_KEY = ['fetch-instance'] as const;
+
+/** Prefer this over bare `updateInstance` so callers cannot forget to refresh the stable instance query. */
+export const updateInstanceAndInvalidate = async (
```
> Now that `INSTANCE_QUERY_KEY` is stable, calling bare `updateInstance` (just above) without invalidating is a genuine stale-cache bug rather than a style nit. All three current call sites use `updateInstanceAndInvalidate`, so consider module-scoping / not exporting the bare `updateInstance` to make this wrapper the only public instance-write path.

src/components/npe/NPEViewComponent.tsx
```diff
@@ -77,7 +79,41 @@ const getRootZoneKey = (proc: KERNEL_PROCESS, address: NPE_COORDINATES): Rootzon
...
+const NPEView = ({ npeData, onRendered }: NPEViewProps) => {
+    // Yield once so the route spinner can paint and the render-budget timer can
+    // fire before sync chip/timeline work begins. Co-mounting the spinner with
+    // an empty shell is cheap; the heavy body mounts on the next frame.
+    const [mountBody, setMountBody] = useState(false);
+
+    useEffect(() => {
+        const frameId = window.requestAnimationFrame(() => {
+            startTransition(() => {
+                setMountBody(true);
```
aidemsined: > [!WARNING]
> The production two-stage render defer (`requestAnimationFrame` → `startTransition` → `setMountBody` → `NPEViewBody`) has no coverage — `NPERoute.spec.tsx` mocks `NPEViewComponent` with a stub that fires `onRendered` synchronously on mount. So the rAF/`startTransition` deferral, the prepaint shell, and its interaction with the route's `setTimeout(0)` mount defer are all untested. A focused test (or a hook-level `useNpeLoadRenderLifecycle` spec) would lock the render-budget contract.

src/hooks/useAPI.tsx
```diff
@@ -425,18 +434,60 @@ export const useGetNPEManifest = () => {
...
 };
 
+const NPE_TEXT_GET_OPTIONS = {
+    responseType: 'text' as const,
+    transitional: { forcedJSONParsing: false },
+};
+
+// Own AbortController (not React Query's signal): Strict Mode remount must not
+// abort a multi-hundred-MB download. Report switch / overlapping fetch aborts the prior.
+let activeNpeRequestAbort: AbortController | null = null;
```
aidemsined, line 444: > [!WARNING]
> `activeNpeRequestAbort` is a module-global single-flight controller shared by `fetchNpeOpTrace` (useNpe) and `fetchNPETimeline` (useNPETimelineFile). It's safe today only by two implicit, untested invariants: the two queries are mutually exclusive (`filepath` gating in `NPE.tsx`), and the disabled sibling's error is masked.

The abort spec drives `fetchNpeText` directly rather than the hooks, so nothing pins the mutual-exclusivity invariant. A future third `fetchNpeText` caller — or making both queries concurrently enabled — would let one multi-hundred-MB download silently abort another. The aborted request rejects with a bare `CanceledError` (no `.status`), which `getNpeValidationErrorFromFetch` maps to `null` → `validateNpeData(undefined)` → a misleading `INVALID_NPE_DATA`.

**Suggested action:** document the "only one NPE download globally" invariant at the `fetchNpeText` export, or key the controller by URL so concurrent NPE downloads can't clobber each other.

tests/getNpeValidationErrorFromFetch.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { describe, expect, it } from 'vitest';
+import { AxiosError, HttpStatusCode } from 'axios';
+import { NPEValidationError, NpeClientErrorKind } from '../src/definitions/NPEData';
+import { getNpeValidationErrorFromFetch } from '../src/functions/getNpeValidationErrorFromFetch';
+import { throwNpeClientAxiosError } from '../src/functions/throwNpeClientAxiosError';
+
+const makeStatusError = (status: number, message = 'error'): AxiosError => {
```
aidemsined, line 11: > [!WARNING]
> The real-server `.status` path isn't faithfully exercised. `makeStatusError` does `new AxiosError(message); error.status = status` — it hand-sets `.status` and never attaches a `.response`, so no test builds the error the way axios's `settle.js` does at runtime (response passed as the 5th constructor arg, which is what actually copies `response.status → error.status`).

The mapping is correct in production today (verified against axios 1.18.1), but this fixture masks the dependency: an axios upgrade that stopped copying `response.status → error.status` would turn every real server 4xx/5xx into a silent `null` fall-through to `validateNpeData`, and these tests would stay green. The synthetic PARSE/SHAPE paths in `useNpeApi.spec.tsx` are faithful (they go through the real `throwNpeClientAxiosError`); only the real-server-status path is masked.

**Suggested action:** add one case that constructs the error via `new AxiosError(msg, code, cfg, req, { status: 422, data: { kind } })` (or a mocked settle path), so the `.status`-from-`.response` contract is under test.

src/functions/throwNpeClientAxiosError.ts
```diff
@@ -0,0 +1,36 @@
...
+        });
+    }
+
+    throw new AxiosError(message, AxiosError.ERR_BAD_RESPONSE, undefined, undefined, {
+        data,
+        status: HttpStatusCode.UnprocessableEntity,
+        statusText: 'Unprocessable Entity',
+        headers: {},
+        // No real request config when the failure is shape-check only (summary/window).
+        config: undefined as unknown as AxiosResponse['config'],
```
aidemsined: > [!WARNING]
> The no-`response` branch fabricates `config: undefined as unknown as AxiosResponse['config']`, which deviates from the caveat in `CONVENTIONS.md:900` ("Pass the original `response.config` and `response.request` — omitting them breaks callers that assume they exist"). The SHAPE callers `fetchNpeSummary`/`fetchNpeWindow` destructure `{ data }` and discard the response, so there's nothing to pass through.

Safe today only because the NPE consumers read just `.status`/`.data` — but it's a latent trap for any future generic `AxiosError` consumer (e.g. a `getResponseError`-style helper reaching into `.config`), which would break specifically on these synthetic errors.

**Suggested action:** keep the full response in the two fetchers (`const response = await axiosInstance.get(...); const { data } = response;`) and pass it to `throwNpeClientAxiosError`, bringing the SHAPE path in line with the PARSE path and the convention.

aidemsined (commented):
Verified the two headline hazards against axios 1.18.1 source and they hold up — no blockers. Real 4xx/5xx populate top-level `error.status` (the `AxiosError` constructor copies `response.status`), and the synthetic 422 does the same, so keying `getNpeValidationErrorFromFetch` on `.status` is sound for both. The huge-string GC drop nulls `response.data` *before* `throwNpeClientAxiosError` spreads the response, so the body isn't recaptured. Nice work; the inline notes below are the should-level items.

## Unanchored findings

- Inconsistent return-type contract across the `validate*` family: `validateNpeData` returns the `NPEValidationError` enum, while `validateNpeSummary`/`validateNpeWindow` return `string | null`. Three shape-validators in one subsystem now speak two dialects — worth aligning on one convention (or a short doc note) so a reader touching the family doesn't re-learn the contract per file.

## General observations

_Minor / optional — not merge-blocking._

- Canceled fetches aren't mapped: an aborted `fetchNpeText` propagates a `CanceledError` (no `.status`) → `null` → `INVALID_NPE_DATA`. Harmless only because `NPE.tsx` masks the disabled query's error; a neutral/loading state would be more correct if a cancel ever reached an enabled query.
- `NPE_TEXT_GET_OPTIONS` is spread before `...config` in `fetchNpeText`, so a caller passing `responseType`/`transitional` in `config` could override the string-parse contract `parseNpeAxiosResponseData` depends on. No current caller does.
- `parseNpeAxiosResponseData`'s string path returns the `JSON.parse` result with no `typeof === 'object'` guard, so a JSON primitive (`"42"`, `"null"`) is cast to `NPEData` (caught later by `validateNpeData`). Applying the same guard the object path uses would put it in the right error bucket.
- `shape-check-then-throw` is duplicated verbatim in `fetchNpeSummary` and `fetchNpeWindow`; a `validateOrThrow(data, validateX)` helper would collapse both (below the ≥3-site bar, so optional — pairs naturally with the no-`response` fix).
- A one-line comment on `NpeClientErrorKind` noting it's a deliberately-minimal wire discriminant (UI taxonomy lives in `NPEValidationError`) would pre-empt a future "these enums overlap, merge them" churn.
- Worth a CI confirmation that lint/`tsc` are clean given the `import/first` and `react-hooks/set-state-in-effect` disables.

---

tests/getNpeValidationErrorFromFetch.spec.ts
dcblundell: I don't think the failure mode holds. The `response.status -> error.status` copy is already under test via the PARSE and SHAPE cases in this same file: `throwNpeClientAxiosError` throws with the five-argument constructor, passing a response object carrying `status: 422` as the fifth argument, and `getNpeValidationErrorFromFetch` reads `error.status`. That is the same `this.status = response.status` assignment (`axios/lib/core/AxiosError.js:136`, verified on 1.18.1) that `settle.js` relies on for real server errors. If an upgrade dropped that copy, both cases would fail rather than stay green.

The fair point is that `makeStatusError` is an unfaithful fixture for the unmarked-server-422 case, so I've taken the cheap half: both client-error cases now pass a real `AxiosResponse` (which they need anyway after making the `response` argument required, see the `throwNpeClientAxiosError` thread). Not adding a separate constructor-based server-status case, since it would duplicate coverage that already exists.

src/hooks/useAPI.tsx
dcblundell: Agreed that the invariant is implicit, and the `CanceledError -> null -> validateNpeData(undefined) -> INVALID_NPE_DATA` path checks out. Taken the documentation option: the `fetchNpeText` export now spells out that at most one NPE download may be in flight application-wide, that callers must be mutually exclusive (pointing at `isNpeQueryEnabled`/`isTimelineQueryEnabled`), and what a concurrent caller would cost.

Skipping the URL-keying alternative because it wouldn't actually separate the two callers: the timeline filename travels in `config.params`, not the URL, so both timeline fetches pass the same `/npe/timeline` string and would still share a key.

src/functions/throwNpeClientAxiosError.ts
dcblundell: Agreed, fixed as suggested. `fetchNpeSummary` and `fetchNpeWindow` now keep the response and pass it through, so the `response` argument is required and the no-response branch (with its `config: undefined` cast) is gone entirely. Both SHAPE and PARSE paths now go through the single `CONVENTIONS.md`-conformant construction. No memory concern here since these are small JSON bodies, unlike the op-trace text.

PR #1797, opened 2026-07-24
Fix remote sync

dcblundell:
Fixes remote sync being kept behind a gate which is impossible to get past.

### DIFF
--- a/backend/ttnn_visualizer/sftp_operations.py
+++ b/backend/ttnn_visualizer/sftp_operations.py
@@ -29,7 +29,7 @@ from ttnn_visualizer.models import (
     Instance,
     RemoteConnection,
     RemoteReportFolder,
-    sanitise_path_segment,
+    folder_segment_from_remote_path,
 )
 from ttnn_visualizer.sockets import FileProgress, FileStatus, emit_file_status
 from ttnn_visualizer.ssh_client import SSHClient, raise_for_ssh_subprocess_error
@@ -1163,14 +1163,14 @@ def get_remote_profiler_folders(
 
 def _safe_sync_destination_segment(remote_folder_path: str) -> str:
     """Collapse remotePath to a single local folder segment; reject ``.`` / ``..`` / empty."""
-    try:
-        return sanitise_path_segment(Path(remote_folder_path).name)
-    except (TypeError, ValueError) as err:
+    segment = folder_segment_from_remote_path(remote_folder_path)
+    if not segment:
         raise RemoteConnectionException(
             message="Invalid report path",
             status=ConnectionTestStates.FAILED,
             http_status_code=HTTPStatus.BAD_REQUEST,
-        ) from err
+        )
+    return segment
 
 
 @remote_exception_handler
--- a/backend/ttnn_visualizer/sftp_operations.py
+++ b/backend/ttnn_visualizer/sftp_operations.py
@@ -25,7 +25,12 @@ from ttnn_visualizer.exceptions import (
     RemoteConnectionException,
     SSHException,
 )
-from ttnn_visualizer.models import Instance, RemoteConnection, RemoteReportFolder
+from ttnn_visualizer.models import (
+    Instance,
+    RemoteConnection,
+    RemoteReportFolder,
+    sanitise_path_segment,
+)
 from ttnn_visualizer.sockets import FileProgress, FileStatus, emit_file_status
 from ttnn_visualizer.ssh_client import SSHClient, raise_for_ssh_subprocess_error
 from ttnn_visualizer.utils import (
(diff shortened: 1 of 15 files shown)


src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -397,6 +397,38 @@ const RemoteSyncConfigurator = () => {
...
+        if (!connection) {
+            return;
+        }
+
+        await withActivatingReport(async () => {
+            try {
+                const response = await mount(connection, folder);
+
+                if (response.status === HttpStatusCode.Ok) {
+                    activateWithToast(folder);
```
aidemsined, line 421: > [!CAUTION]
> A local report activated mid-sync gets silently clobbered when the sync completes. Now that the SSH transfer no longer holds `isActivatingReportAtom` (the fix at ~L460), and local pickers gate only on that atom, the user can activate a *local* report during a long remote sync. When the sync then resolves, this `activateWithToast(folder)` runs unconditionally — `setActive*Report(...)` back to the remote folder and `queryClient.clear()` — wiping the local report's cache and switching the active report out from under the user. `withActivatingReport` is a boolean flag, not a mutex, so nothing serialises the two activations. Guard the terminal activate on "is `folder` still the intended/selected report" (or re-hold the lock across mount+activate). Remote↔remote and sync↔sync are safe via `isDisabled`; only local-during-remote is exposed — which is exactly the interleave this PR newly enables, and it's untested.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1156,6 +1161,18 @@ def get_remote_profiler_folders(
     return sorted(remote_folder_data, key=lambda x: x.lastModified, reverse=True)
 
 
+def _safe_sync_destination_segment(remote_folder_path: str) -> str:
```
aidemsined, line 1164: > [!WARNING]
> This and the `remote_path` branch of `_safe_report_folder_name` in `views.py` both implement `sanitise_path_segment(Path(remotePath).name)` behind their own try/except. That's a single safety invariant — sync's write segment must equal mount's lookup segment — duplicated across the `sftp_operations` ↔ `views` seam, which is exactly where drift produces "sync wrote `foo`, mount looked in `bar`". Parity holds today, but there's already one divergence: `_safe_report_folder_name` falls back to `report_name` on an empty basename while this function 400s, so a `"/"`/empty `remotePath` could mount an unrelated `report_name` folder. Consider one shared `folder_segment_from_remote_path()` (e.g. in `models.py`, which both modules already import), leaving each call site only its HTTP error mapping.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -486,42 +507,33 @@ const RemoteSyncConfigurator = () => {
...
-
         setSelected(folder);
 
-        await withActivatingReport(async () => {
-            try {
-                const response = await mount(connection, folder);
-
-                if (response.status === HttpStatusCode.Ok) {
-                    activateWithToast(folder);
+        if (isRemoteFolderOutdated(folder)) {
```
aidemsined: > [!WARNING]
> This makes selecting an outdated folder block on a full SSH transfer. `isRemoteFolderOutdated` is true for merely-stale folders (`lastModified > lastSynced`), not just never-synced ones — so a report you already have locally now forces a network round-trip on a mere dropdown select, where before select always mounted the on-disk copy instantly (sync was manual). It's intended and has the offline fallback, but the "just show me the copy I have" case regresses to network-bound with no signal to the user beyond `FileStatusOverlay`. Worth confirming that's the desired trade-off (or only auto-syncing never-synced folders, and leaving stale-but-present as mount + optional sync).

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -660,23 +672,23 @@ const RemoteSyncConfigurator = () => {
                 subLabel='Select a memory report'
             >
                 <RemoteFolderSelector
-                    remoteFolder={isProfilerRemote ? selectedReportFolder : undefined}
+                    remoteFolder={selectedReportFolder}
```
aidemsined: > [!WARNING]
> `remoteFolder` is now passed unconditionally (was gated on `isProfilerRemote` / `isPerformanceRemote`), and the sync button likewise renders whenever a folder is selected. So a user whose active report location is LOCAL can still see a remote folder name here and a Sync affordance, since the selector label comes from `remoteFolder?.reportName`. Either restore the location gate on display/sync, or clear the remote selection when the location flips — worth checking against real location-switch flows. (Same on the performance selector at L709.)

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1156,6 +1161,18 @@ def get_remote_profiler_folders(
...
+    """Collapse remotePath to a single local folder segment; reject ``.`` / ``..`` / empty."""
+    try:
+        return sanitise_path_segment(Path(remote_folder_path).name)
+    except (TypeError, ValueError) as err:
+        raise RemoteConnectionException(
+            message="Invalid report path",
```
aidemsined, line 1169: > [!WARNING]
> Same validation failure surfaces as `"Invalid report path"` here but `"Invalid report name"` on the mount side (`views.py::_safe_report_folder_name`). Since both reject the same bad segment, consider unifying the copy so the API surface reads consistently.

tests/remoteFolderSelector.spec.tsx
aidemsined, line 1: > [!WARNING]
> Two coverage gaps worth closing: (1) all the new sync-on-select tests use *performance* folders — add a memory/profiler mirror, since `selectAndActivateFolder` wiring is identical but unverified for that path; (2) segment parity between sync-write (`_safe_sync_destination_segment`) and mount-lookup (`_safe_report_folder_name`) is by-construction only — a unit test asserting equal output for a sanitiser-rewriting name (backslash / trailing whitespace / unicode) would stop a future edit to either side from silently reintroducing "synced but not found."

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Guarded post-sync activate: skip mount/activate if the active report path changed mid-transfer (local pick mid-sync). Covered by a FE regression test.

PR #1799, opened 2026-07-24
Align commit link icon centrally with text

dcblundell:
**Improved alignment**
Screenshot 2026-07-24 at 12 58 57

Screenshot 2026-07-24 at 12 58 50

### DIFF
--- a/src/scss/components/GitCommitInfo.scss
+++ b/src/scss/components/GitCommitInfo.scss
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+.report-git-info {
+    display: block;
+    font-size: 12px;
+    margin-top: 5px;
+
+    .report-git-info-link {
+        display: inline-flex;
+        align-items: center;
+        gap: 0;
+    }
+
+    .report-git-info-icon {
+        margin-left: 3px;
+        margin-bottom: 2px;
+        vertical-align: middle;
+    }
+}
--- a/src/scss/components/GitCommitInfo.scss
+++ b/src/scss/components/GitCommitInfo.scss
@@ -10,12 +10,10 @@
     .report-git-info-link {
         display: inline-flex;
         align-items: center;
-        gap: 0;
     }
 
     .report-git-info-icon {
         margin-left: 3px;
-        margin-bottom: 2px;
-        vertical-align: middle;
+        margin-bottom: 1px;
     }
 }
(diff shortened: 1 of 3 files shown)


src/scss/components/GitCommitInfo.scss
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+.report-git-info {
+    display: block;
+    font-size: 12px;
+    margin-top: 5px;
```
aidemsined, line 8: > [!CAUTION]
> These base block styles now apply to the footer tooltip, not just the stack trace — an unclaimed cross-consumer change.

On `dev`, `display: block; font-size: 12px; margin-top: 5px` (and the icon margins below) lived in the `@mixin report-git-info`, which was `@include`d only under `.stack-trace-path` / `.stack-trace-error .stack-trace-path`. Moving them onto the root `.report-git-info` block in a stylesheet that's side-effect-imported from `GitCommitInfo.tsx` makes them global (Vite `.scss` imports aren't scoped).

The other consumer — `FooterInfobar.tsx` → `ReportGitMetadataLines` → `<span class="report-git-info">` — had **no** base styling in the footer tooltip before. It now gets block layout + 12px type + 5px top margin. And since `ReportGitMetadataLines` already emits a `<br/>` right before the span, `display: block` + `margin-top: 5px` stacks on that break (extra vertical gap), and 12px diverges from the sibling "Report path:" / "Git repo:" lines.

Suggest keeping only the link/icon alignment rules here and restoring the block-layout rules under the two stack-trace selectors in `StackTrace.scss` — or, if the footer change is intended, confirming it with a footer before/after screenshot (the PR screenshots only show the stack-trace context).

src/scss/components/GitCommitInfo.scss
```diff
@@ -0,0 +1,21 @@
...
+    .report-git-info-link {
+        display: inline-flex;
+        align-items: center;
+        gap: 0;
+    }
+
+    .report-git-info-icon {
+        margin-left: 3px;
+        margin-bottom: 2px;
+        vertical-align: middle;
```
aidemsined: > [!WARNING]
> A couple of these declarations are now dead or redundant given the link is `inline-flex`.

- `gap: 0` (line 13) restates the flex default (`gap: normal` resolves to `0` for flex) — it can be dropped.
- `vertical-align: middle` (line 19) has no effect on a flex item; the centering is now done by `align-items: center` on `.report-git-info-link`. It's leftover from the pre-flex approach.
- `margin-bottom: 2px` (line 18) looks like a pre-flex baseline hack and may now fight `align-items: center`. Worth checking whether the icon is still optically centered without it.

dcblundell: Removed redundant rules and yes, the margin is desired.

src/scss/components/GitCommitInfo.scss
dcblundell: It's fine.

src/scss/components/GitCommitInfo.scss
```diff
@@ -0,0 +1,21 @@
...
+    }
```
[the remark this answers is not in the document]
dcblundell: Removed.

Issue #1803, opened 2026-07-27
NPE: scrub render performance — timeline full-repaint + unmemoized SVG tiles


aidemsined:
Scrubbing large NPE reports is slow, and the cost is in the **paint layer**, not JS lookups/allocation. Profiled on `reports_with_zones-3/npe_viz/UnknownOP_ID5.npeviz.zst` (652 MB decompressed, **76,007 timesteps**, 8 chips, `link_demand` per step p50 148 / max 623, `active_transfers` p50 14 / max 63).

### Bottleneck #1 (dominant): timeline canvas fully repaints on every scrub
`NPETimelineComponent`'s draw `useEffect` lists `currentTimestep` in its deps, so every scrubber tick re-runs the full heatmap paint loop: **4 rows × 76,007 = ~304,000 `fillRect` calls** (+142 zone rects) just to reposition a 2px playhead line. The heatmap pixels don't change while scrubbing — only the playhead does. Scales with `n_timesteps`, so it's worst on exactly the large files this path targets.

```
// src/components/npe/NPETimelineComponent.tsx (draw effect, deps include currentTimestep)
for (let row = 0; row < numLines; row++) {
    pointList.forEach((point, index) => {
        ctx.fillStyle = point.color;
        ctx.fillRect(index * chunkWidth, y, chunkWidth, rowHeight);
    });
}
```

**Fix:** split into a static heatmap canvas (repaint only when `congestionMapPerTimestamp` / zones / width change) + a cheap playhead overlay canvas (clear + 1 `fillRect` per scrub). Optionally downsample the heatmap to ~`canvasWidth` columns instead of 76k sub-pixel rects.

### Bottleneck #2: unmemoized SVG link tiles
`links.link_demand.map(...)` in `NPEViewComponent` rebuilds one `<button>` + `<TensixTransferRenderer>` (SVG) per link-demand entry. `TensixTransferRenderer` is **not** `React.memo`'d, and `npeData` / `links` / `transfers` get new references every scrub, so React reconciles + repaints **~150 (p50) to 623 (max)** SVG subtrees + `8 × A` `RouteOriginsRenderer` instances per tick. Scales with `D`.

**Fix:** `React.memo` `TensixTransferRenderer` (with stable `data`/`style`), memoize per-tile render so unchanged tiles skip repaint.

### Note
[...]

Issue #1802, opened 2026-07-24
NPE windowed loading: decide hosted-safety + promote out of the DEV gate


aidemsined:
Follow-up to the #861 windowed-loading PoC. The windowed path (server-side SQLite index + `/api/npe/summary` + `/api/npe/window`) is currently gated to `import.meta.env.DEV` for uploaded reports (`src/routes/NPE.tsx`); production/hosted still serves the whole-file `/api/npe`, which is exactly the path that fails on large reports. The fork needs a tracked exit criterion before it ossifies.

### Decide
- **Hosted safety of the two routes.** `get_npe_summary` / `get_npe_window` are `@with_instance @timer` but **not** `@local_only`. Under `SERVER_MODE` an untrusted caller can trigger `ensure_index`, which does a full in-memory parse (~5 GB transient RSS on the 885 MB repro) and writes a `.npeidx.sqlite` sidecar next to the source. Either gate them `@local_only`, add a size/disk-quota guard, or consciously accept the exposure (hosted NPE is currently low-priority, graceful-degradation-OK, no cross-user data).
- **Single decision point instead of `import.meta.env.DEV`.** Replace the DEV gate with a backend-reported capability or file-size heuristic so local production builds and a future hosted rollout share one switch.
- **Retire the dual read path.** Once windowed is trusted, either drop the whole-file `/api/npe` path or keep both behind the explicit switch above, so `useNpe` vs `useNpeSummary`/`useNpeWindow` don't drift.

### Deferred hardening surfaced by the round-2 review (not blocking the PoC)
- Cross-process single-flight (`fcntl.flock`) is untested — the concurrency test only exercises the in-process `threading.Lock`.
- `_is_index_fresh` validates version + mtime only; a truncated-but-readable DB isn't caught (a `PRAGMA quick_check` / row-count sanity would fail fast).
- `read_window`'s chunked `IN(...)` doesn't preserve `active_ids` order and dedups by PK — cosmetic divergence from whole-file order.
- The `links` memo in `NPEViewComponent` mutates the React-Query-cached window `link_demand` (`FABRIC_EVENT_SCOPE`), which now persists under `staleTime: Infinity`.

Context: round-2 ultimate review of `feature/861-npe-windowed-loading`.

PR #1804, opened 2026-07-27
NPE windowed loading: render very large reports without a whole-file client parse (#861)

### LINKED ISSUE #861: NPE: files above ~50 MB do not upload correctly
Image

aidemsined:
## Summary
Large NPE reports (tens of MB compressed → hundreds of MB / ~900 MB JSON) fail to open: the browser's `JSON.parse` hits V8's string-length limit and silently returns an empty object (#861). This adds a **server-side SQLite index + windowed per-timestep API** so the browser only ever parses a small per-step payload.

Closes #861

**Scope:** the windowed path replaces the whole-file `/api/npe` for uploaded reports in **both local dev and local prod**, and is disabled only under `SERVER_MODE` (hosted) — the same trust boundary as the `@local_only` gate on `/api/npe/{summary,window}`. Hosted keeps the whole-file path (currently low-priority: graceful degradation, no cross-user data). The gate rationale + exit criterion are documented inline in `src/routes/NPE.tsx`, and the hosted-safety decision is tracked in #1802.

### Backend
- `backend/ttnn_visualizer/npe_index.py` — builds a timestep-keyed SQLite sidecar from the full parsed report (one-time, single-flighted via an in-process `threading.Lock` + cross-process `fcntl.flock`, atomic temp-file replace). Columnar summary wire format (~9 MB → ~2.6 MB at 54k steps), float truncation to 3 dp, chunked `IN(...)` for busy timesteps.
- `views.py` — `GET /api/npe/summary` (per-step aggregates + static metadata) and `GET /api/npe/window?t=` (one step's transfers + `link_demand`), both `@local_only`. Narrowed status mapping: 404 (missing file / out-of-range `t`), 422 (malformed / corrupt-`.zst` report), 400 (non-integer `t`).

### Frontend
- `useNpeSummary` / `useNpeWindow` hooks (`keepPreviousData` on the window; `instanceId` in keys under `staleTime: Infinity`), with client-side shape validation of **both** the summary and each window at the fetch boundary (`validateNpeSummary` / `validateNpeWindow`, thrown as `AxiosError`).
- `assembleWindowedNpeData` + `buildTimestepSkeleton` reconstruct a renderer-ready `NPEData`: the ~54k-step aggregate skeleton is built once per summary and only the visited step is patched per scrub.
- `NpeWindowedView` container drives the selected timestep and feeds the unchanged `NPEView` (now accepting controlled-timestep + stable `timelineData` props).
- Window fetch failures degrade in place (keep summary + scrubber) instead of replacing the whole view; re-uploads drop the NPE query cache so the rebuilt index is refetched.
- Gated on `!getServerConfig()?.SERVER_MODE` (was `import.meta.env.DEV`, which compiled to `false` in any prod build and wrongly excluded local prod installs).

## Known limitation (out of scope, tracked separately)
Scrub **render** performance on very large timelines is unchanged — the bottleneck is the paint layer (timeline canvas full-repaint per scrub + unmemoized SVG link tiles), not the JS this PR touches. Analyzed and tracked in #1803 to keep this PR's scope bounded.

## Test plan
- [x] Backend: `pytest backend/ttnn_visualizer/tests/test_npe_index.py backend/ttnn_visualizer/tests/test_npe_routes.py` — index build/read, columnar shape, empty-trace round-trip, freshness/version invalidation, `.zst` round-trip, mtime re-upload rebuild, missing-id skip, window 404/OOB, IN-list chunking (2500 ids), single-flight build lock (8-thread stampede), route 200/404/422/400, corrupt-`.zst` → 422, 403 under `SERVER_MODE`.
- [x] Frontend: `pnpm test` for `assembleWindowedNpeData` (incl. skeleton immutability + link_demand clone), `validateNpeSummary`, `validateNpeWindow`, `NpeWindowedView` (error surface, degrade-in-place, auto-jump, empty-trace, re-upload re-clamp).
- [x] `tsc`, `eslint`, `black`/`isort`/`mypy` clean on touched files.
- [ ] Manual: upload a large `.npeviz.zst`, confirm it renders + scrubs in a local (non-`SERVER_MODE`) build — dev **and** prod; re-upload a same-named file and confirm fresh data.

## Follow-ups
- #1802 — hosted-safety decision + whether to promote the windowed path into `SERVER_MODE`.
- #1803 — scrub render performance (timeline repaint + SVG tile memoization).
- #1805 — manage stored NPE reports (list / load / delete + index cleanup).

> Note: `devtools/npe_render_probe.py` is a profiling script intentionally kept on the branch — it's the harness for measuring the deferred renderer / paint-layer work in #1803, so it'll be reused there rather than dropped. It self-aborts under `SERVER_MODE` and is excluded from SAST via `.cycodeignore`.


backend/ttnn_visualizer/views.py
```diff
@@ -2216,6 +2217,60 @@ def get_npe_data(instance: Instance):
...
+        logger.exception("Malformed NPE report while building index")
+        return response_unprocessable_entity()
+    except Exception:
+        logger.exception("Unexpected error reading NPE window")
+        return response_internal_server_error()
+
+    if window is None:
+        return response_not_found()
+
+    return Response(orjson.dumps(window), mimetype="application/json")
```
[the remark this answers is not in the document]
aidemsined: Addressed in 37fb5590. This is a taint-flow false positive — `t` is the only user input and it's `int()`-coerced before use (a `ValueError` returns 400), so it can't carry markup and never reaches the response body; the payload is server-side report data serialized as `application/json`. The identical `Response(orjson.dumps(...))` pattern is used across ~6 other routes here without a flag — it's triggered only because `request.args` is in scope. Still applied the correct defence-in-depth control: both `/npe/summary` and `/npe/window` now send `X-Content-Type-Options: nosniff` so the JSON can't be MIME-sniffed as HTML, with a header assertion added to the route tests.

devtools/npe_render_probe.py
```diff
@@ -0,0 +1,62 @@
...
+about the real bottleneck.
+"""
+import statistics
+import sys
+
+import orjson
+import zstd
+
+path = sys.argv[1]
+raw = zstd.uncompress(open(path, "rb").read())
```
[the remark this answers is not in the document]
aidemsined: Addressed in 37fb5590. `_tmp/npe_render_probe.py` is a local, non-shipping profiling CLI — `sys.argv[1]` is a developer-supplied path on their own machine, not a web-facing input, so path confinement doesn't apply (the tool is meant to open any report you point it at). It's intentionally retained as the harness for the deferred render/paint-layer work in #1803. Excluded the `_tmp/` scratch dir via `.cycodeignore` so this false positive stops recurring without pulling the script from the branch.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -0,0 +1,374 @@
...
+def _truncate_link_demand(link_demand: list) -> list:
+    # Mutates in place — the parsed object is discarded after the build. Only the
+    # trailing demand (index 4) is a float; coords / noc-id are left untouched.
+    for entry in link_demand:
+        if len(entry) > _LINK_DEMAND_INDEX and type(entry[_LINK_DEMAND_INDEX]) is float:
+            entry[_LINK_DEMAND_INDEX] = (
+                int(entry[_LINK_DEMAND_INDEX] * _TRUNCATE_SCALE) / _TRUNCATE_SCALE
+            )
+    return link_demand
+
```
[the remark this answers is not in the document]
aidemsined: Fixed in 932df4be. `_truncate_link_demand` now drops a trailing `None` so windows reach the wire as 5-tuples — a serialized `null` would have read as a present scope in `NPEView` and mis-annotated FABRIC/LOCAL/BOTH, exactly as you describe. Real reports emit 5-tuples, so this only bites a report carrying an explicit null, but the guard is cheap and the contract is now unambiguous.

Root cause of the miss: the test fixtures fabricated 6-tuples with a trailing `None`, which both hid the bug and "proved" the null path safe. Corrected the fixtures to the real 5-tuple shape and added a `read_window` regression asserting a trailing null is stripped. Bumped `INDEX_VERSION` so stale on-disk caches rebuild.

src/functions/validateNpeWindow.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    if (!Array.isArray(timestep.active_transfers)) {
+        return 'NPE window timestep is missing its active_transfers array.';
+    }
+
+    if (!Array.isArray(timestep.link_demand)) {
+        return 'NPE window timestep is missing its link_demand array.';
+    }
+
+    return null;
```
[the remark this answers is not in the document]
aidemsined: Fixed in 932df4be. `validateNpeWindow` now goes past array-presence: each `link_demand` row is checked for minimum arity `[chip, y, x, noc_id, demand]`, a string `NOC_ID`, and a numeric demand; each transfer is checked for a numeric `id` and an array `route` — the fields `NPEView` indexes and walks without guards. A malformed row/transfer now surfaces as a friendly Callout instead of throwing deep in a render memo. Added spec cases for short rows, non-string NOC id, non-numeric demand, and transfers missing `id`/`route`, plus a positive case for the real 5-tuple shape.

backend/ttnn_visualizer/views.py
```diff
@@ -2216,6 +2217,60 @@ def get_npe_data(instance: Instance):
     return Response(npe_data, mimetype="application/json")
 
 
+@api.route("/npe/summary", methods=["GET"])
+@with_instance
+@timer
+def get_npe_summary(instance: Instance):
```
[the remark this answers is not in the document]
aidemsined: Known deferral, not an oversight — tracked in #1802. The windowed path is DEV-gated on the frontend (`import.meta.env.DEV` in `src/routes/NPE.tsx`), so `/npe/summary` and `/npe/window` aren't reachable in the hosted build today. Promoting the feature to hosted is a deliberate, separate step that has to decide the `@local_only` boundary plus build/RSS/disk quotas, precisely because `ensure_index()` parses the whole report and writes a large sidecar — the DoS/disk-growth surface you flag.

Since this kept getting re-raised without that context, 932df4be documents the deferral inline above both routes so future automated reviewers see the rationale at the site.

devtools/npe_render_probe.py
```diff
@@ -0,0 +1,62 @@
+# SPDX-License-Identifier: Apache-2.0
```
dcblundell, line 1: As discussed, move this file to something like `/devtools` instead of `/_tmp`.

dcblundell (changes requested):

backend/ttnn_visualizer/views.py
```diff
@@ -2216,6 +2221,76 @@ def get_npe_data(instance: Instance):
     return Response(npe_data, mimetype="application/json")
 
 
+# NOTE: /npe/summary and /npe/window are intentionally NOT @local_only yet. The
+# windowed path is DEV-gated on the frontend (src/routes/NPE.tsx), so it isn't
+# reachable in the hosted build today. Promoting it to hosted requires an explicit
+# hosted-safety decision — the @local_only boundary plus build/RSS/disk quotas,
+# since ensure_index() parses the whole report and writes a large sidecar. Tracked
+# in #1802. SAST/Copilot flag the missing gate; that's this known deferral.
```
smountenay-tt: Pushing back on this one, because the premise isn't right.

`import.meta.env.DEV` gates the *client*, not the server. These routes are registered on the blueprint unconditionally, so on the hosted deploy `curl 'https://ttnn-visualizer.tenstorrent.com/api/npe/summary?instanceId=…'` reaches `ensure_index()` regardless of what the SPA does. "Not reachable in the hosted build" is true of the UI affordance, not of the endpoint.

And the attack path is complete, not theoretical: `/local/upload/npe` (views.py:1448) is itself **not** `@local_only`, so an untrusted hosted user can upload the report *and* then trigger the whole-report parse against it. AGENTS.md is explicit that the `@local_only` boundary *is* the security boundary, precisely because there's no auth in front of it.

The reason I'd rather not defer: **the fix is free today.** The frontend only calls these under `import.meta.env.DEV`, and a DEV install never runs with `SERVER_MODE`, so adding `@local_only` now breaks exactly nothing, and #1802 removes it when the hosted story is decided. A deferral that costs nothing to close isn't really a deferral — it's a live unauthenticated endpoint that parses an arbitrary uploaded report into memory. Copilot flagged this correctly.

Worth noting the amplifier while you're in here: `_BUILD_LOCK` is a single module-global held for the *entire* build, so one cold build serializes every other NPE index build in the process, not just concurrent builds of the same report.

backend/ttnn_visualizer/views.py
```diff
@@ -2216,6 +2221,76 @@ def get_npe_data(instance: Instance):
...
+        db_path = ensure_index(instance.npe_path)
+        summary = read_summary(db_path)
+    except FileNotFoundError:
+        return response_not_found()
+    except orjson.JSONDecodeError:
+        logger.exception("Malformed NPE report while building index")
+        return response_unprocessable_entity()
+    except Exception:
+        logger.exception("Unexpected error building/reading NPE index")
+        return response_internal_server_error()
```
smountenay-tt: `zstd.Error` isn't a `ValueError`, so it doesn't match the `orjson.JSONDecodeError` arm and falls through to the generic handler. I ran the same corrupt file through both routes:

- `/api/npe/summary` → **500**
- `/api/npe` (existing whole-file route) → **422**

A truncated or corrupt upload is exactly the "malformed report" case the PR description maps to 422, and `.npeviz.zst` is the format that motivated the issue, so this is the likely-in-practice path rather than an exotic one. `except (orjson.JSONDecodeError, zstd.Error):` covers it. Same arm on the window route at 2277.

Minor related divergence: `_load_npe_object` branches on `suffix == ".zst"` only, whereas `get_npe_data` falls back to *trying* zstd for unrecognised suffixes. Low stakes, but the two paths disagree on what an unknown extension means.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -0,0 +1,388 @@
...
+        for transfer in transfers:
+            batch.append((transfer["id"], orjson.dumps(transfer)))
+            if len(batch) >= _TRANSFER_INSERT_BATCH:
+                conn.executemany(
+                    "INSERT OR REPLACE INTO transfer (id, blob) VALUES (?, ?)", batch
+                )
+                batch.clear()
+        if batch:
+            conn.executemany(
+                "INSERT OR REPLACE INTO transfer (id, blob) VALUES (?, ?)", batch
```
smountenay-tt: `transfer["id"]` is unguarded, so a transfer object without an `id` key raises `KeyError` here → generic handler → 500. And because the *build* is what fails, that report is **permanently** 500 on both endpoints rather than failing one request. Confirmed with a fixture whose transfer omits `id`.

This isn't defensive theatre: the whole-file path renders such a report fine, and `NPEViewComponent.tsx:362` explicitly guards `transfer.id === null || transfer.id === undefined` — so the frontend already treats a missing id as a live possibility. Windowed loading turns a report the current path displays into a hard 500.

The `"id": null` sibling is quieter and worse, because `INSERT OR REPLACE ... VALUES (NULL, ?)` doesn't error — SQLite auto-assigns `max(rowid)+1` for a NULL `INTEGER PRIMARY KEY`. Verified:

```
INSERT OR REPLACE INTO transfer (id, blob) VALUES (5, 'REAL-5'), (NULL, 'NULL-ID-GHOST')
→ rows:  [(5, b'REAL-5'), (6, b'NULL-ID-GHOST')]
→ SELECT blob FROM transfer WHERE id IN (6)  →  b'NULL-ID-GHOST'
```

So a null-id transfer squats a real id, and a lookup for that id silently returns the wrong blob. One guard fixes both cases — skip (and log) transfers whose id isn't an `int`, rather than crashing on one and aliasing on the other.

While you're here: `INSERT OR REPLACE` also silently makes duplicate ids last-wins, where the whole-file path is first-wins via `.find()`. Probably academic, but if `OR REPLACE` isn't deliberately defending against dupes, a plain `INSERT` would surface the assumption instead of hiding it.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -0,0 +1,388 @@
...
+            return
+        lock_path = Path(f"{db_path}.lock")
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        fd = os.open(lock_path, os.O_CREAT | os.O_RDWR, 0o644)
+        try:
+            fcntl.flock(fd, fcntl.LOCK_EX)
+            yield
+        finally:
+            fcntl.flock(fd, fcntl.LOCK_UN)
+            os.close(fd)
```
smountenay-tt, line 147: `{db_path}.lock` is created but never unlinked, so every indexed report leaves a stray lock file next to it alongside the sidecar.

Harmless in isolation, but it lengthens the list of artifacts #1805 has to clean up — worth calling out there so the cleanup covers `.npeidx.sqlite`, `.npeidx.sqlite.lock`, and any orphaned `.tmp` files left by a build that died between `mkstemp` and `replace`.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -0,0 +1,388 @@
...
+):  # Windows local installs lack fcntl; single-user, in-process lock suffices.
+    fcntl = None  # type: ignore[assignment]
+
+logger = logging.getLogger(__name__)
+
+# Serialises index builds within a single process; the file lock below extends
+# that across gunicorn worker processes.
+_BUILD_LOCK = threading.Lock()
+
+INDEX_SUFFIX = ".npeidx.sqlite"
```
smountenay-tt, line 42: Worth quantifying the sidecar for #1802/#1805, because "a large sidecar" undersells it.

The index stores every transfer and every timestep's `link_demand`/`noc` as JSON blobs, so on disk it's roughly the *uncompressed* report — less ~34% from the float truncation, plus SQLite overhead. For the motivating case that's a ~30 MB `.zst` upload turning into something on the order of a 600–900 MB sidecar, and nothing ever deletes it (no NPE delete route today, no cleanup hook).

On a shared host that's a more practical exhaustion vector than transient RSS, since RSS is reclaimed and this isn't. A concrete ratio in #1802 would make the quota decision a lot easier than a qualitative note.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -0,0 +1,388 @@
...
+        # SQLITE_MAX_VARIABLE_NUMBER (999 on sqlite < 3.32), which would otherwise
+        # raise OperationalError and make that window permanently un-viewable.
+        for start in range(0, len(active_ids), _TRANSFER_QUERY_CHUNK):
+            chunk = active_ids[start : start + _TRANSFER_QUERY_CHUNK]
+            placeholders = ",".join("?" * len(chunk))
+            transfers.extend(
+                orjson.loads(blob)
+                for (blob,) in conn.execute(
+                    f"SELECT blob FROM transfer WHERE id IN ({placeholders})",
+                    chunk,
```
smountenay-tt, line 393: Transfers come back in rowid order (ascending id), not source-file order. Rendering is unaffected — `NPEViewComponent` resolves by `.find(tr => tr.id === id)`, and `noc_transfers.filter(...)` preserves whatever order it's handed — but the sidebar transfer list will be ordered differently in the DEV windowed view than in prod's whole-file view whenever the source isn't id-sorted.

Purely cosmetic, but it makes the two paths visibly diverge, which is a bit self-defeating while the DEV gate exists specifically so they can be compared side by side. Re-ordering the assembled list to match `active_ids` would keep them identical for the cost of one sort over a few hundred items.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -0,0 +1,388 @@
...
+        raise
+
+    # Atomic swap so a crashed build never leaves a half-written cache in place.
+    tmp_path.replace(db_path)
+    logger.info(
+        f"NPE index built: {len(transfers)} transfers, {len(timesteps)} timesteps"
+    )
+
+
+def _write_index(conn, obj, transfers, timesteps, source_mtime_ns) -> None:
```
smountenay-tt: Only unannotated signature in the module — mypy lets it through because it isn't running strict, but the rest of the file is fully typed so it stands out. `conn: sqlite3.Connection, obj: dict, transfers: list, timesteps: list, source_mtime_ns: int` would match the surrounding style.

backend/ttnn_visualizer/tests/test_npe_index.py
```diff
@@ -0,0 +1,228 @@
...
+    return {
+        "common_info": {"version": "1.0.0", "arch": "wormhole_b0"},
+        "chips": {"0": {}},
+        "zones": [],
+        "noc_transfers": transfers,
+        "timestep_data": timesteps,
+    }
+
+
+def _write_npe(tmp_path: Path, obj: dict, name: str = "trace.json") -> str:
```
smountenay-tt, line 74: Every fixture in both new test files is plain `.json` — there's no `.zst` case anywhere, so the `zstd.uncompress` branch of `_load_npe_object` is entirely uncovered. That's the production format and the reason the issue exists.

I checked and the happy path does work (`zstd.compress(orjson.dumps(obj))` round-trips through `ensure_index`/`read_window` cleanly), so this is a coverage gap rather than a bug — but it's the branch most worth pinning, and a corrupt-`.zst` companion would have caught the 500/422 mismatch I flagged on views.py:2243.

backend/ttnn_visualizer/tests/test_npe_index.py
```diff
@@ -0,0 +1,228 @@
...
+    obj["timestep_data"][0]["active_transfers"] = list(range(n))
+    obj["timestep_data"][0]["link_demand"] = [[0, 0, 0, "NOC0_EAST", 10.0]]
+    db_path = ensure_index(_write_npe(tmp_path, obj))
+
+    window = read_window(db_path, 0)
+    assert window is not None
+    assert len(window["transfers"]) == n
+
+
+def test_index_rebuilds_on_version_bump(tmp_path, monkeypatch):
```
smountenay-tt, line 243: The PR description lists "freshness/version invalidation", but only the version bump is covered — nothing exercises the `source_mtime_ns` half of `_is_index_fresh`, which is the re-upload path the manual test plan calls out ("re-upload a same-named file and confirm fresh data").

I verified it does work: write a report, `ensure_index`, rewrite the same path with a different timestep count, `ensure_index` again, and the summary reflects the new content. Cheap to add alongside this one, and it locks down the behaviour users actually hit.

backend/ttnn_visualizer/tests/test_npe_routes.py
```diff
@@ -0,0 +1,152 @@
...
+
+def test_window_non_integer_timestep_is_400(client, make_npe_instance):
+    instance_id = make_npe_instance(orjson.dumps(_make_npe_object(n_timesteps=2)))
+    resp = client.get(
+        f"{API}/npe/window", query_string={"instanceId": instance_id, "t": "abc"}
+    )
+    assert resp.status_code == 400
+
+
+def test_summary_malformed_report_is_422(client, make_npe_instance):
```
smountenay-tt, line 156: Companion case worth adding right here: the same assertion for a corrupt `.zst` rather than malformed JSON. As of this commit that returns 500, not 422 (see my note on views.py:2243) — so the test would fail as written, which is rather the point.

smountenay-tt (commented):
Backend-focused review. I checked the branch out, ran the suite (17/17 green, `black`/`isort`/`mypy` clean), and wrote throwaway probe tests for the edges the new tests don't reach — three of the findings below are verified by execution rather than by reading.

**The approach is sound.** The load-bearing question for any windowed design is whether the renderer tolerates a *subset* of `noc_transfers`, and it does: `NPEViewComponent` resolves transfers via `.find(tr => tr.id === id)` (lines 170, 356), never by array index, so serving only the active step's transfers is safe and ordering doesn't matter. A timestep-keyed SQLite sidecar with an mtime+version freshness key is the conventional shape for this. The build mechanics are genuinely well done — atomic temp-file-plus-`replace`, two-layer single-flight (thread lock + `flock`) with the re-check *under* the guard, and an 8-thread stampede test that actually exercises it. The IN-list chunking with its bind-var rationale, and adding `instanceId` to the new query keys (which the existing `useNpe` key lacks), are both the right calls.

**One item I don't think should ship as a deferral: the missing `@local_only`.** Detail inline, but the short version is that the rebuttal's premise doesn't hold — `import.meta.env.DEV` gates the client, not the Flask blueprint, and `/local/upload/npe` isn't `@local_only` either, so the hosted attack path is complete. The fix is free today and #1802 can undo it later. I'd like that one addressed before merge rather than tracked.

The rest are ordinary review points. Two are real bugs (corrupt `.zst` status mapping; transfers with missing/null `id`), the rest are coverage and polish.

For what it's worth on the other bot comments: both Cycode findings are false positives and your rebuttals hold up, and the two substantive Copilot findings were real and correctly fixed — the note about the fixtures fabricating 6-tuples and thereby "proving" the null path safe is good root-causing, and bumping `INDEX_VERSION` alongside it was the right instinct.

One placement nit not worth its own thread: `test_npe_routes.py` is route-contract testing sitting at `tests/` root, while the repo keeps that kind of thing in `tests/views/` (`test_remote_reports.py`, `test_tensors.py`, …). There's precedent both ways (`test_mlir.py`), so take it or leave it.

backend/ttnn_visualizer/views.py
aidemsined: Fixed in 6bb5288d — and you're right, my earlier reasoning was wrong. `import.meta.env.DEV` gates the *client* only; the blueprint registers `/npe/summary` and `/npe/window` unconditionally, so a hosted `curl` reaches `ensure_index()` regardless of the SPA. Both routes now carry `@local_only` (→ 403 under `SERVER_MODE`), which matches the feature's local-only scope and closes the DoS/disk-growth surface. Added a `test_summary_forbidden_under_server_mode` case, and #1802 now tracks the quota work needed before the gate can be lifted for hosted.

backend/ttnn_visualizer/views.py
aidemsined: Fixed in 6bb5288d. `zstd.Error` isn't a `ValueError`, so it slipped past the `orjson.JSONDecodeError` arm into the generic 500 — exactly your repro. Both routes now catch `(orjson.JSONDecodeError, zstd.Error)` → 422, so a corrupt/truncated `.zst` matches the whole-file path's mapping. Added `test_summary_corrupt_zst_is_422` and `test_window_corrupt_zst_is_422`.

backend/ttnn_visualizer/npe_index.py
aidemsined: Fixed in 6bb5288d. `transfer["id"]` is now `transfer.get("id")`; a transfer with no id is skipped (with a warning) rather than `KeyError`-ing the whole build — which, as you note, was cached and 500'd the report permanently on both routes while the whole-file path renders it. Skipping matches that path (an id-less transfer can't be resolved by `active_transfers` anyway). Added `test_build_skips_transfer_without_id`.

backend/ttnn_visualizer/npe_index.py
aidemsined: Fixed in 6bb5288d — `_write_index` now carries `conn: sqlite3.Connection, obj: dict, transfers: list, timesteps: list, source_mtime_ns: int`, matching the rest of the module.

backend/ttnn_visualizer/tests/test_npe_index.py
aidemsined: Fixed in 6bb5288d. Added `test_zst_source_round_trips`, which builds and reads from an actual `zstd.compress`'d `.npeviz.zst` fixture — covering the `zstd.uncompress` branch of `_load_npe_object` that the all-`.json` fixtures skipped. Good catch that this was the production format left uncovered.

backend/ttnn_visualizer/tests/test_npe_index.py
aidemsined: Fixed in 6bb5288d. Added `test_index_rebuilds_on_source_mtime_change`: build, rewrite the same path with a different timestep count, bump mtime via `os.utime` (avoids same-tick flake), and assert the summary reflects the new count — the re-upload path the manual plan called out.

backend/ttnn_visualizer/tests/test_npe_routes.py
aidemsined: Fixed in 6bb5288d. `test_window_corrupt_zst_is_422` (and the summary companion) assert 422 on a corrupt `.zst` right alongside the malformed-JSON case — passing now that `zstd.Error` maps to 422.

backend/ttnn_visualizer/npe_index.py
aidemsined: Not fixing in this PR, by design: the `.lock` is a POSIX `flock` file, and unlinking a held lock creates a fresh inode that defeats the cross-process serialisation. It's meant to persist. I've logged it (plus the sidecar and any orphaned `.tmp` from a killed build) as an explicit artifact for the #1805 cleanup to sweep.

backend/ttnn_visualizer/npe_index.py
aidemsined: Agreed the "large sidecar" phrasing undersells it. Captured your figure — sidecar ≈ uncompressed report minus ~34% (float truncation) plus SQLite overhead, i.e. ~600–900 MB for a ~30 MB `.zst` — on #1802 (hosted-safety/quota decision) and #1805 (delete/quota UX should surface the footprint).

backend/ttnn_visualizer/npe_index.py
aidemsined: Acknowledged, leaving as-is — purely cosmetic as you say (rendering resolves by `.find(id === …)`; only the sidebar list order differs in the DEV windowed view vs prod's whole-file view, and only when the source isn't id-sorted). Source order isn't stored in the index, so preserving it would mean an extra column for a DEV-only cosmetic delta — not worth it at this stage. Noted for whenever the windowed path is promoted.

devtools/npe_render_probe.py
aidemsined: Done in 6bb5288d — moved to `devtools/npe_render_probe.py` and repointed `.cycodeignore`. It also self-aborts under `SERVER_MODE` now (local-only dev tool).

devtools/npe_render_probe.py
```diff
@@ -0,0 +1,81 @@
...
+if _server_mode_enabled():
+    sys.exit(
+        "npe_render_probe is a local-only dev tool; it refuses to run under SERVER_MODE."
+    )
+
+if len(sys.argv) < 2:
+    sys.exit("usage: python devtools/npe_render_probe.py <report.(npeviz.)zst>")
+
+path = sys.argv[1]
+raw = zstd.uncompress(open(path, "rb").read())
```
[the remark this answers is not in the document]
aidemsined: False positive, and re-anchored after the `git mv` into `devtools/`. `open(sys.argv[1])` here is a **local-only CLI profiling tool**: the "user input" is a path the engineer types in their own shell, so there's no privilege boundary to cross (they can already read any file directly). It isn't a route, isn't imported by the app, and isn't run by gunicorn/requests — there's no web-reachable path to it, and it self-aborts under `SERVER_MODE` regardless. It only decompresses + parses locally and prints stats; contents never flow back over a network surface.

Covered by `.cycodeignore` (`devtools/npe_render_probe.py`), which is why the **SAST check passes with 0 findings**; this inline annotation is the bot re-posting on the moved path independent of the gate. Resolving.

src/functions/validateNpeWindow.ts
```diff
@@ -0,0 +1,73 @@
...
+        if (!transfer || typeof transfer !== 'object') {
+            return 'NPE window has a malformed transfer entry.';
+        }
+        if (typeof transfer.id !== 'number') {
+            return 'NPE window transfer is missing a numeric id.';
+        }
+        if (!Array.isArray(transfer.route)) {
+            return 'NPE window transfer is missing its route array.';
+        }
+    }
```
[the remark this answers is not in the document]
aidemsined: Good catch — fixed in 476df454. `validateNpeWindow` now walks each route entry and rejects it unless it's an object with a `links` array, a numeric `injection_rate`, and `src`/`dst` arrays — the exact fields `NPEViewComponent` (`route.links.forEach`) and `ActiveTransferDetails` (`route[0].injection_rate.toFixed(2)`, `route.src/dst.join`) read without guards. Added positive + malformed-route test cases.

src/components/npe/NpeWindowedView.tsx
```diff
@@ -0,0 +1,135 @@
...
+                </Callout>
+            );
+        }
+        return (
+            <div className='npe-windowed-loading'>
+                <Spinner size={20} />
+                <span>{isLoadingSummary ? 'Processing…' : 'Loading timestep…'}</span>
+            </div>
+        );
+    }
```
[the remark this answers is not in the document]
aidemsined: Fixed in 476df454. The window-error branch is now gated on `!isLoadingSummary`, so while the index is still building we always show "Processing…" and only surface a window error once the summary has resolved. Note the empty-trace (`n_timesteps===0`) case is additionally short-circuited to the "Empty NPE report" callout before any window-error branch. Added a regression test for the racing-window-during-load case.

backend/ttnn_visualizer/views.py
```diff
@@ -2216,6 +2221,83 @@ def get_npe_data(instance: Instance):
     return Response(npe_data, mimetype="application/json")
 
 
+# @local_only: ensure_index() parses the whole report and writes a large sidecar
+# DB. The SPA only calls these routes from a DEV build, but the blueprint registers
+# them unconditionally, so on a hosted (SERVER_MODE) deploy an untrusted caller
+# could still reach ensure_index() directly — a DoS + disk-growth vector. The gate
+# returns 403 under SERVER_MODE, matching the feature's local-only scope; hosted
+# promotion (build/RSS/disk quotas) is tracked in #1802.
```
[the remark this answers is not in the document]
aidemsined: Fixed in 476df454. Comment now reads that the SPA calls these routes from any local (non-SERVER_MODE) build — dev **and** local prod — behind the same SERVER_MODE gate, matching the frontend change. The @local_only rationale (blueprint registers unconditionally → hosted DoS/disk vector → 403, promotion tracked in #1802) is unchanged.

smountenay-tt (approved):

src/hooks/useAPI.tsx
```diff
@@ -602,6 +604,55 @@ export const useNpe = (fileName: string | null) => {
...
+        params: { t },
+    });
+    const shapeError = validateNpeWindow(data);
+    if (shapeError) {
+        throw new AxiosError(shapeError, 'ERR_INVALID_RESPONSE');
+    }
+    return data;
+};
+
+export const useNpeWindow = (fileName: string | null, t: number | null) => {
```
dcblundell, line 653: **Must fix:** Abort scrub/play window request storm

`useNpeWindow` keys every timestep and does not pass React Query's `AbortSignal` into axios. Play (~10–20 Hz) and continuous scrub start a new `/npe/window` per `t` while abandoned seeks still complete, contend for connections, and fill the Infinity-`staleTime` cache. Wire `signal` into `axiosInstance.get`; consider debounce / keep-latest for scrub commits.

src/routes/NPE.tsx
```diff
@@ -15,13 +15,27 @@ import { NPEData } from '../model/NPEModel';
...
-    const { data: loadedData, isLoading: isLoadingNPE, error: httpError } = useNpe(filepath ? null : npeFileName);
+    const isServerMode = !!getServerConfig()?.SERVER_MODE;
+    // #861: for uploaded reports the windowed view replaces the whole-file path,
+    // skipping `useNpe` (its full /api/npe fetch is exactly what fails on large
+    // files) and rendering NPEView from per-timestep windowed fetches instead.
+    // Enabled in both local dev and local prod, disabled under SERVER_MODE — the
+    // same boundary as the @local_only gate on /api/npe/{summary,window}, whose
+    // sidecar build isn't hosted-safe yet (#1802). Hosted keeps the whole-file
+    // path; exit criterion is deciding hosted-safety, then dropping the fork.
+    const isWindowedView = !isServerMode && !filepath && !!npeFileName;
```
dcblundell, line 33: **Must fix:** Add route wiring tests for `isWindowedView`

This gate (`!SERVER_MODE && !filepath && !!npeFileName`) and the matching `useNpe(... ? null : …)` skip have no route-level test, and already drifted once (`import.meta.env.DEV` → `SERVER_MODE`). Add a route spec: local+upload mounts windowed and calls `useNpe(null)`; `SERVER_MODE` / `filepath` keep the whole-file path.

src/hooks/useAPI.tsx
```diff
@@ -602,6 +604,55 @@ export const useNpe = (fileName: string | null) => {
...
+};
+
+// #861 windowed loading: per-step aggregates for the whole trace, small enough
+// to load once; transfers/link_demand come from useNpeWindow per visited step.
+// instanceId is part of the key because staleTime is Infinity — a bare basename
+// key would serve one instance's cached report to another on a name collision.
+export const useNpeSummary = (fileName: string | null) => {
+    return useQuery<NpeSummary, AxiosError>({
+        queryFn: fetchNpeSummary,
+        queryKey: ['npe-summary', getOrCreateInstanceId(), fileName],
```
dcblundell: **Should fix:** Export `*_QUERY_KEY` constants for cross-module invalidation

`'npe-summary'` / `'npe-window'` (and `'fetch-npe'`) are magic strings here and again in `NPEFileLoader` `removeQueries`. AGENTS.md requires keys invalidated from another module to be exported (same pattern as `PROFILER_FOLDER_QUERY_KEY`). Export constants from `useAPI` and import them at the invalidation site.

backend/ttnn_visualizer/tests/test_npe_routes.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+    assert resp.status_code == 422
+
+
+def test_summary_missing_instance_id_is_400(client):
+    resp = client.get(f"{API}/npe/summary")
+    assert resp.status_code == 400
+
+
+def test_summary_forbidden_under_server_mode(client, make_npe_instance, app):
```
dcblundell, line 183: **Should fix:** Add SERVER_MODE 403 regression for `/api/npe/window`

Only summary asserts `@local_only` under `SERVER_MODE`. Window uses the same decorator stack and is half of the index-build DoS/disk surface. Add a symmetric `test_window_forbidden_under_server_mode`.

dcblundell (commented):
Conductor five-lens review — Must fix (2) and Should fix (9) as line-anchored comments.

src/components/npe/NPETimelineComponent.tsx
```diff
@@ -116,7 +116,15 @@ const NPETimelineComponent = ({
...
                 : timestep.link_demand;
 
-            const worst = Math.max(-1, ...links.map((linkData) => linkData[NPE_LINK.DEMAND]));
+            // Windowed loading (#861) supplies `link_demand` only for the visited
+            // step; other steps fall back to the per-step scalar from the summary.
+            // Known limitation: that scalar is the global (both-NOC) worst, and
+            // non-visited `noc` aggregates are stubbed to 0, so with a NOC filter
+            // active the heat rows for non-visited steps are approximate until the
+            // summary carries per-NOC aggregates. Unfiltered view is exact.
+            const worst = links.length
```
dcblundell: **Should fix:** Test empty-`link_demand` → `max_link_demand` heat-bar fallback

Non-visited windowed steps rely on `timestep.max_link_demand ?? -1`. No timeline coverage; reverting to `Math.max(-1, ...[])` paints idle steps wrong. Assert worst for: empty links + scalar, non-empty rows, and missing scalar → −1.

backend/ttnn_visualizer/tests/test_npe_index.py
```diff
@@ -0,0 +1,282 @@
...
+    npe_path = _write_npe(tmp_path, _make_npe_object(n_timesteps=4))
+    summary = read_summary(ensure_index(npe_path))
+
+    assert summary["n_timesteps"] == 4
+    assert summary["common_info"]["version"] == "1.0.0"
+    # Columnar: parallel arrays, not one dict per step, and `t` is not sent.
+    timesteps = summary["timesteps"]
+    assert isinstance(timesteps, dict)
+    assert "t" not in timesteps
+    for key in ("start_cycle", "avg_link_demand", "max_link_demand", "active_count"):
```
dcblundell, line 103: **Should fix:** Assert `max_link_demand` values, not just key presence

Summary tests check lengths for `max_link_demand` but not idle `None` vs active ≈ truncated demand (derivation is in `npe_index` when the source omits the field). A broken derivation silently breaks the timeline heat-bar fallback.

src/components/npe/NPEFileLoader.tsx
```diff
@@ -55,6 +57,13 @@ const NPEFileLoader = () => {
                 setErrorMessage(response?.data?.message ?? 'Upload failed');
             } else {
                 const fileName = file.name;
+                // Re-uploading a same-named file reuses the NPE query keys, and the
+                // windowed hooks are staleTime: Infinity — drop the cached summary /
+                // windows so the freshly-rebuilt server index is refetched instead of
+                // serving the previous report's data.
+                queryClient.removeQueries({ queryKey: ['npe-summary'] });
```
dcblundell: **Should fix:** Test cache bust on successful re-upload

Same-name re-upload relies on `removeQueries` for `npe-summary` / `npe-window` / `fetch-npe` because those hooks use `staleTime: Infinity`. Untested — dropping these lines keeps the old report after upload. Assert `queryClient.removeQueries` for those three keys on OK upload.

src/hooks/useAPI.tsx
```diff
@@ -602,6 +604,55 @@ export const useNpe = (fileName: string | null) => {
     });
 };
 
+const fetchNpeSummary = async (): Promise<NpeSummary> => {
```
dcblundell: **Should fix:** Test fetch-boundary validator wiring

`validateNpeSummary` / `validateNpeWindow` are unit-tested, but `fetchNpeSummary` / `fetchNpeWindow` throwing `AxiosError` on shape failure is not. Removing the validate calls would still pass leaf specs while serving bad JSON into `NPEView`. Add axios-mocked tests for valid body vs length-mismatched / bad window.

src/components/npe/NPEViewComponent.tsx
```diff
@@ -232,16 +254,21 @@ const NPEView = ({ npeData }: NPEViewProps) => {
...
 
     useEffect(() => {
+        /* eslint-disable react-hooks/set-state-in-effect */
         // eslint-disable-next-line react-hooks/immutability
         stopAnimation();
-        // eslint-disable-next-line react-hooks/set-state-in-effect
-        setSelectedTimestep(0);
+        // In controlled mode the container owns the timestep and resets it on
+        // report switch; resetting here would snap every windowed refetch to 0.
+        if (!isTimestepControlled) {
```
dcblundell, line 262: **Should fix:** Test controlled mode does not reset timestep on `npeData` identity change

Controlled scrubbing must skip this reset when `isTimestepControlled` (deps use `reportKey ?? npeData`). `NpeWindowedView` mocks `NPEView`, so this is untested. Mount real `NPEView` with `reportKey` + controlled props: new `npeData` (same key) keeps timestep; `reportKey` change resets as intended.

backend/ttnn_visualizer/views.py
```diff
@@ -2216,6 +2221,84 @@ def get_npe_data(instance: Instance):
...
+        mimetype="application/json",
+        headers=_NOSNIFF_HEADERS,
+    )
+
+
+@api.route("/npe/window", methods=["GET"])
+@with_instance
+@local_only
+@timer
+def get_npe_window(instance: Instance):
```
dcblundell, line 2287: **Should fix:** Extract shared NPE index resolve + exception map

`get_npe_summary` and `get_npe_window` duplicate path checks, `ensure_index`, and the three-arm except map (window even says “See get_npe_summary”). They already diverged once on `zstd.Error`. Extract something like `_resolve_npe_index` so status mapping lives in one place.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -0,0 +1,405 @@
...
+            _build_index(source, db_path, source_mtime_ns)
+    return db_path
+
+
+def read_summary(db_path: Path) -> dict:
+    conn = sqlite3.connect(db_path)
+    try:
+        meta = dict(conn.execute("SELECT key, value FROM meta").fetchall())
+        rows = conn.execute(
+            "SELECT start_cycle, end_cycle, avg_link_demand, avg_link_util, "
```
dcblundell: **Should fix:** Single SSOT for summary columnar column order

Frontend `NPE_SUMMARY_COLUMN_KEYS` is the wire SSOT; backend hardcodes the same seven names in this `SELECT` and again in the positional `columns[i]` dict below. A reorder on either side silently corrupts the timeline. Drive SELECT + dict from one backend tuple and note it must match the frontend keys.

src/hooks/useAPI.tsx
aidemsined: Done in 89f242a7 — `useNpeSummary`/`useNpeWindow` now pass React Query's `signal` into `axiosInstance.get`, so abandoned play/scrub seeks are aborted instead of storming `/npe/window` and churning the Infinity-staleTime cache. Covered by a signal-threading assertion in `tests/useNpeApi.spec.tsx`.

src/routes/NPE.tsx
aidemsined: Done in 89f242a7 — added `tests/NPERoute.spec.tsx`: local upload → windowed view mounts and `useNpe` is gated to `null`; SERVER_MODE and `?filepath` → whole-file path (no windowed view), with SERVER_MODE still fetching by name.

src/hooks/useAPI.tsx
aidemsined: Done in 89f242a7 — exported `NPE_QUERY_KEY` / `NPE_SUMMARY_QUERY_KEY` / `NPE_WINDOW_QUERY_KEY` and consumed them from `NPEFileLoader`'s `removeQueries`, so the hooks and the re-upload cache-bust reference one source.

backend/ttnn_visualizer/tests/test_npe_routes.py
aidemsined: Done in 89f242a7 — added `test_window_forbidden_under_server_mode`, mirroring the summary case: `/npe/window` returns 403 under SERVER_MODE via the shared `@local_only` stack.

src/components/npe/NPETimelineComponent.tsx
aidemsined: Done in 89f242a7 — extracted the fallback into `getWorstLinkDemand(links, maxLinkDemand)` and unit-tested it: max across non-empty rows, scalar fallback when the window is absent, `-1` when both are missing, and a guard that an empty row set never yields `-Infinity`.

backend/ttnn_visualizer/tests/test_npe_index.py
aidemsined: Done in 89f242a7 — added `test_summary_max_link_demand_values`, asserting values not just array lengths: idle step 0 → `None` (no rows, no source scalar); active step → derived-from-`link_demand` and truncated to 3 dp (`224.8`).

src/components/npe/NPEFileLoader.tsx
aidemsined: Done in 89f242a7 — added `tests/NPEFileLoader.spec.tsx`: a successful upload evicts the `npe-summary` / `npe-window` / `fetch-npe` caches (guards the same-name re-upload + Infinity-staleTime trap); a failed upload evicts nothing.

src/hooks/useAPI.tsx
aidemsined: Done in 89f242a7 — added `tests/useNpeApi.spec.tsx`: a well-shaped body resolves, while a length-mismatched summary and a malformed window each reject as an `ERR_INVALID_RESPONSE` `AxiosError` from the fetch boundary.

src/components/npe/NPEViewComponent.tsx
aidemsined: Done in 89f242a7 — added `tests/NPEViewControlledTimestep.spec.tsx`: same `reportKey` + new `npeData` object keeps the timestep (no `onSelectedTimestepChange` call); a `reportKey` change does not self-reset in controlled mode (container owns it); uncontrolled mode still resets internal timestep to 0 on `npeData` identity change.

backend/ttnn_visualizer/views.py
aidemsined: Done in 89f242a7 — extracted `_resolve_npe_index(instance, reader)` + `_npe_index_json_response(payload)`. Both routes now share the path check, `ensure_index`, and the 404/422/500 mapping, so `/summary` and `/window` can't drift on error handling again (they already diverged once on `zstd.Error`).

backend/ttnn_visualizer/npe_index.py
aidemsined: Done in 89f242a7 — added the module-level `_SUMMARY_COLUMNS` tuple that drives both the `SELECT` and the positional dict in `read_summary`, with a comment that it must stay in lock-step with the frontend `NPE_SUMMARY_COLUMN_KEYS`.

dcblundell (approved):

PR #1808, opened 2026-07-28
Align report tooltip contents

dcblundell:
Added missing "Report path:" to performance tooltip.

Screenshot 2026-07-28 at 11 27 10

### DIFF
--- a/src/components/FooterInfobar.tsx
+++ b/src/components/FooterInfobar.tsx
@@ -171,7 +171,11 @@ function FooterInfobar() {
                     {activePerformanceReportPath && (
                         <Tooltip
                             disabled={!activePerformanceReportPath}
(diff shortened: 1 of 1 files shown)


aidemsined (commented):
Reviewed — clean and merge-ready. The Performance tooltip now renders `<strong>Report path:</strong> {formatPath(...)}`, matching the profiler (`Memory:`) tooltip precedent exactly, so the contents are genuinely aligned. `Tooltip.content` takes a `ReactNode`, so the string→fragment swap is fine. No blocking issues and nothing to discuss.

One pre-existing note (not this PR — it just brings Performance to parity with the profiler tooltip): `formatPath()` returns only the last path segment, so the label `Report path:` actually shows the report folder name rather than the full path. Worth a separate follow-up only if the wording matters.

PR #1813, opened 2026-07-29
Sync and discover multihost performance reports

### LINKED ISSUE #1801: Remote sync: find multihost performance reports in per-rank subdirectories
Follow-up to #1243, which concluded that multihost Tracy reports need no cross-rank merging: `tt-run --tracy` writes one ordinary Tracy report per rank, just nested one level deeper.

```
tt-metal/generated/profiler/
  reports/<report_dir>/profile_log_device.csv        <- single host, found today
  ttrun/rank0/<report_dir>/profile_log_device.csv    <- multihost, not found today
  ttrun/rank1/<report_dir>/profile_log_device.csv
```

Remote discovery runs `find '<root>' -maxdepth 1 -type d`, so nothing under `ttrun/` is listed and a user cannot select a multihost performance report over a remote connection.

### Goal

Minimal change to claim support for loading multihost performance reports remotely. Reuse the existing `performancePath` / `performance_path` plumbing everywhere; do not introduce a second path that has to be tracked through the app.

### Approach

Add one boolean to the remote connection that changes only the *search depth* used to discover performance reports. The user points `performancePath` at the directory holding the per-rank subdirectories (e.g. `.../generated/profiler/ttrun`) and ticks the flag; everything downstream (listing, sync, mount, `performance_path`, all report views) is untouched.

Backend:
- `RemoteConnection` in `backend/ttnn_visualizer/models.py`: add `multihostPerformance: bool = False`. Default keeps existing saved connections and stored instances valid; `remote_connection` is a JSON column so no migration is required.
- `find_folders_by_files` in `backend/ttnn_visualizer/sftp_operations.py`: add a `depth: int = 1` parameter and change the remote command to `find '<root>' -mindepth <depth> -maxdepth <depth> -type d` (the explicit `-not -path '<root>'` becomes redundant). The existing per-directory `test -f profile_log_device.csv` check is unchanged, so only real report dirs match.
- `get_remote_performance_folders` and the performance branch of `check_remote_path_for_reports`: pass `depth=2` when `multihostPerformance` is set, `depth=1` otherwise. Nothing keys off the literal name `ttrun`.

Frontend:
- `RemoteConnection` in `src/definitions/RemoteConnection.ts`: add `multihostPerformance?: boolean`. Saved connections live in `localStorage` as JSON, so the optional field round-trips and older entries read as falsy.
- `src/components/report-selection/RemoteConnectionDialog.tsx`: add a checkbox/switch directly below the "Performance report folder path" field, labelled for multihost, with sub-label text explaining that reports are expected in per-rank subdirectories such as `.../generated/profiler/ttrun/rank0/<report>`.
- `src/components/report-selection/RemoteSyncConfigurator.tsx`: when an edit flips the flag, the cached per-connection performance folder list is stale, so clear or refetch it in the `onEditConnection` handler alongside `updateSavedRemoteFoldersConnection`.

No display work is expected in the folder picker: `formatRemoteFolderPath` in `src/components/report-selection/RemoteFolderSelector.tsx` already strips the configured path prefix, so a report at `<ttrun>/rank0/<report>` renders as `/rank0/<report>`.

### Explicitly out of scope

- Merging or grouping reports across ranks. Reports from the same run cannot be matched by timestamp (see #1243) and each rank report is viewed independently.
- Local/uploaded multihost performance reports; this issue is remote sync only.
- Prefixing the synced local folder with the rank. Sync destination and mount segments are both the remote report dir basename, and per-rank report dirs are independently timestamped, so names collide only in the unlikely case that two ranks produce identically named report dirs. Track separately if it shows up in practice.

### Acceptance criteria

- With the flag off, remote performance report discovery behaves exactly as today.
- With the flag on and `performancePath` set to a directory of per-rank subdirs, the picker lists every rank's report, labelled so the rank is visible (e.g. `/rank0/<report>`).
- Selecting one syncs and opens it through the existing single-report performance flow.
- The connection test in the add/edit dialog passes for a multihost path.
- Coverage added to `backend/ttnn_visualizer/tests/views/test_remote_reports.py` and `tests/RemoteConnectionDialog.spec.tsx`.

smountenay-tt:
Closes #1801

`tt-run --tracy` writes one performance report per rank, nested a level deeper than a single-host run puts them, so multihost reports were invisible to remote sync. This adds a per-connection flag that searches the per-rank layout, and keeps the resulting reports distinguishable end to end.

## Discovery

Tick **Multihost performance reports** on a remote connection and point the performance path at the folder that directly contains the rank directories (e.g. `.../generated/profiler/ttrun/`). Reports are then found at `rank<N>/reports/<report>`.

The search is pinned to that exact layout rather than just searching deeper: min and max depth are equal and the intervening directories must match, so a path that is not the per-rank parent no longer matches by accident, and each rank's `.logs` directory (which holds a raw `profile_log_device.csv` and otherwise looks like a report) is skipped. Because glob syntax cannot express "digits only", candidates the glob admits are filtered through the same rank definition used to name them locally, before they cost an SSH round trip.

The connection test now reports how many reports it found, and says so in per-rank terms when the flag is on, so "the path exists" isn't mistaken for "reports are there".

## Keeping ranks apart locally

Every rank names its report after its own start time to the second, so ranks of one launch routinely produce **the same report name**. Sync destinations and mount lookups both reduced a remote path to its basename, which meant those reports downloaded on top of each other into one local folder, and the `instances` row could not say which rank it held.

Synced copies are now qualified with their rank (`<report>_rank0`), applied in `folder_segment_from_remote_path` — the single seam the write and read paths already share, with a test pinning that agreement. The last-synced probe was re-deriving the basename itself, so syncing rank 0 lit up every other rank's badge; it now goes through the same helper. Single-host paths keep byte-identical names, so no existing synced folder or stored path churns.

The folder stays one flat segment: nesting a rank directory would break the single-segment sanitisation contract and the sibling assumption three view handlers rely on.

## Two consequences of that rename, also fixed

- **`?name=` swaps 404'd.** Three performance routes override the instance's stored path when the request carries `?name=`, so the Compare selector can switch reports without re-mounting. They rebuilt the directory from the report's *display* name, which omits the new qualifier. The name is now resolved against the reports directory — exact match first, then the active report's own rank — so a bare name never borrows a *different* rank's data, and unresolvable names keep the existing not-found behaviour. Since this is the only choke point for the value, it also collapses it to a single segment.
- **The remote picker highlighted the wrong row.** It matched the active report by report name, which every rank of a launch shares — so it could select the wrong rank immediately after a pick, and none at all after a reload (restore identifies the report by its synced folder name). It now matches on whichever identity the active report actually carries.

The dropdown labels these reports `Rank 0: 2026_07_28_18_04_24` instead of showing raw paths, for both live listings and the offline listing built from synced copies.

## Testing

Backend 453 passed, frontend 1155 passed; mypy, black, isort, ESLint and `tsc --noEmit` clean.

New coverage includes the collision itself (two ranks, one timestamp, now distinct), the sync/mount segment agreement, discovery against a real `find` over a temporary tree mirroring the rank layout, `.logs` and single-host near-misses being excluded, the rank filter dropping a glob-admitted candidate, every branch of the `?name=` resolution, and the picker's select/restore matching per rank.

Manual: synced both ranks of a multihost run from a remote host and confirmed two distinct local folders, the expected `performance_path` in the `instances` table, and the report loading per rank.

## Notes for review

- `docs/src/remote-sync.md` documents the layout and the rank suffix.
- Qualification triggers whenever a `rank<N>` directory appears in the remote path, including without the flag. Pointing a single-host connection *directly* at one rank's reports would therefore start appending the suffix and re-sync those reports once under the new name.


backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1049,6 +1152,9 @@ def find_folders_by_files(
             if not directory:
                 continue
 
+            if directory_filter is not None and not directory_filter(directory):
+                continue
```
dcblundell: **Must fix:** SSH connection fan-out now scales with rank count

The `directory_filter` added here helpfully drops candidates "before they cost an SSH round trip", but the surviving candidates still cost one each: the loop just below spawns a fresh `ssh` process per directory to run `test -f` (~line 1160), and `get_remote_performance_folders` then spawns another per folder for `stat -c %Y` (~line 1235). Multihost changes the candidate count from *reports* to *ranks x reports*, and `_ssh_cmd_prefix` sets no `ControlMaster`/`ControlPersist`, so each is a full TCP + auth handshake. A 32-rank run with 20 launches is roughly 1280 sequential connections, putting the listing endpoint in the minutes range — and `POST /api/remote/test` pays it again independently.

Suggested direction: fold the existence test into the `find` that already runs (`-exec test -f '{}/profile_log_device.csv' \; -print`) and take mtime from the same command (`-printf '%T@ %p\n'`, or one batched remote `stat`), so discovery is one round trip regardless of rank count.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1018,18 +1082,57 @@ def check_remote_path_exists(remote_connection: RemoteConnection, path_key: str)
...
+    if subdirectory_glob is None:
+        return f"find '{search_root}' -mindepth 1 -maxdepth 1 -type d"
+
+    # `*` matches `/` in a `-path` pattern, so pinning min and max depth to the
+    # same value is what stops the glob spanning segments and matching a deeper
+    # accidental layout.
+    depth = 1 + len(subdirectory_glob.split("/"))
+    return (
+        f"find '{search_root}' -mindepth {depth} -maxdepth {depth} -type d "
+        f"-path '{root}/{subdirectory_glob}/*'"
```
dcblundell: **Must fix:** unquoted path interpolation into a remote shell command

`search_root` and the `-path` pattern are wrapped in hand-written single quotes with no escaping, so a `'` anywhere in `performancePath` terminates the quoting and the remainder is run by the remote shell. That input is self-inflicted (same local user owns the SSH credentials, route is `@local_only`), so on its own it's hardening.

The sharper edge is the pre-existing `test -f '{directory}/{file_name}'` at ~line 1160, where `directory` is *remote* `find` output rather than local input — a remote directory whose name contains `'` injects commands that run as the connecting user. This change enlarges the subtree feeding that loop from `<root>/*` to `<root>/rank*/reports/*`, which on a shared lab machine can include directories the connecting user doesn't own.

`shlex.quote` is already the established pattern in `ssh_client.py`, `mlir.py` and `stack_trace_source.py`. Quoting the whole `-path` glob is correct here — `find` wants it unexpanded by the shell.

src/components/report-selection/RemoteFolderSelector.tsx
```diff
@@ -165,6 +166,22 @@ const RemoteFolderSelector = ({
...
+ * Returns null when no rank is present, leaving the plain path formatting.
+ */
+const formatMultihostPerformanceLabel = (folder: RemoteFolder): string | null => {
+    const rank = getReportRank(folder.remotePath);
+
+    if (rank === null) {
+        return null;
+    }
+
+    return `Rank ${rank}: ${getReportBaseName(folder.reportName || folder.remotePath)}`;
```
dcblundell: **Must fix:** the rank label never reaches the collapsed selector button

This label is used for the menu items and the filter predicate, but the `Select` trigger `Button` still renders `remoteFolder?.reportName` (line 159, outside the diff). As the PR description says, every rank of one launch names its report from its own start time at second granularity, so ranks routinely share a basename — once a folder is selected the button reads identically for every rank and the user can't see which one is loaded. That's the ambiguity the feature exists to remove.

Suggested direction: derive the button text from the same helper rather than maintaining two ways to name a folder.

src/functions/reportRank.ts
```diff
@@ -0,0 +1,53 @@
...
+ * The local folder name sync writes for a remote report. Mirrors the backend's
+ * `folder_segment_from_remote_path`, including the `_rank<N>` qualifier that keeps
+ * identically named reports from different ranks apart.
+ */
+export const getSyncedFolderName = (remotePath: string): string => {
+    const segments = getSegments(remotePath);
+    const reportName = segments.at(-1) ?? '';
+    const rankDirectory = getRankDirectory(segments);
+
+    return rankDirectory && rankDirectory !== reportName ? `${reportName}_${rankDirectory}` : reportName;
```
dcblundell: **Should fix:** this doesn't fully mirror `folder_segment_from_remote_path`

The backend passes the basename through `sanitise_path_segment` (`models.py:246`), which trims surrounding whitespace and collapses `\` to `/`; `getSegments` only splits on `/`. For a report directory like `" spaced "` — a case the backend suite explicitly covers — sync writes `spaced_rank11` while the client computes `" spaced _rank11"`, so `isPerformanceFolderActive` silently fails to highlight the active row.

Smallest fix: trim and normalise `\` in `getSegments`/`getSyncedFolderName` so the two agree. If you'd rather not chase full parity, soften the doc comment so the next reader doesn't trust it as an exact mirror.

backend/ttnn_visualizer/models.py
```diff
@@ -248,19 +249,54 @@ def sanitise_path_segment(value: object) -> str:
     return safe_segment
 
 
+RANK_DIRECTORY_RE = re.compile(r"^rank\d+$", re.IGNORECASE)
+RANK_SUFFIX_RE = re.compile(r"_(rank\d+)$", re.IGNORECASE)
```
dcblundell: **Should fix:** add `re.ASCII` so the client and backend agree on what a rank is

Python's `\d` matches Unicode decimal digits, so a remote directory named e.g. `rank٣` is treated as a rank here, while the frontend twins in `src/functions/reportRank.ts:5-8` use JavaScript `\d`, which is ASCII-only. Containment still holds — the suffix lands as literal filename characters inside the target directory — but the two sides then disagree about the synced folder name, which is exactly the write/read alignment `folder_segment_from_remote_path` exists to guarantee.

backend/ttnn_visualizer/models.py
```diff
@@ -248,19 +249,54 @@ def sanitise_path_segment(value: object) -> str:
...
+    return match.group(1) if match else None
+
+
+def rank_directory_from_remote_path(remote_path: str) -> Optional[str]:
+    """The rank folder a multihost report sits under, verbatim, or None.
+
+    Authoritative definition of a rank directory: discovery filters candidates
+    through this so nothing can be listed that cannot then be rank-qualified.
+    """
+    for part in reversed(Path(remote_path).parts):
```
dcblundell: **Should fix:** the reversed scan includes the report basename, so a report named `rank5` still collides

`Path(remote_path).parts` covers the report directory itself, so `/root/rank0/reports/rank5` matches its own name before its rank ancestor. `folder_segment_from_remote_path` then sees `segment == rank_directory` and returns the unqualified `rank5`, meaning rank0 and rank1 sync into one folder and overwrite each other — the precise collision this change was written to prevent. Discovery admits both paths (the `-path` glob matches and `directory_filter` passes), so the docstring's "nothing can be listed that cannot then be rank-qualified" doesn't quite hold.

Contrived naming, hence not a blocker. Scanning `Path(remote_path).parent.parts` fixes it and keeps `test_a_rank_directory_itself_is_not_doubled` green (that case has no rank ancestor); `getSyncedFolderName` needs the mirror change.

tests/remoteFolderSelector.spec.tsx
```diff
@@ -1215,6 +1215,118 @@ const selectRemoteFolder = async (type: 'profiler' | 'performance', remotePath:
...
+];
+
+const multihostFolders: RemoteFolder[] = [
+    {
+        reportName: '2026_07_28_18_04_24',
+        remotePath: `${MULTIHOST_ROOT}/rank0/reports/2026_07_28_18_04_24`,
+        lastModified: 1,
+    },
+    {
+        reportName: '2026_07_28_18_04_31',
```
dcblundell: **Should fix:** the fixture sidesteps the collision the feature exists for

rank0 and rank1 get *different* timestamps (`18_04_24` / `18_04_31`), so the picker is never tested against the scenario the PR description leads with: every rank naming its report from its own start time at second granularity, i.e. the same basename. That's also where the React `key` in `remoteFolderRenderer` (`label + lastSynced ?? lastModified`) would collapse.

Related: all five new selector tests pass `onSelectFolder={() => undefined}` and assert only rendered text, so nothing verifies that clicking `Rank 1: ...` hands back rank1's folder. With two ranks sharing a display name that's a plausible bug which would pass every current test.

Suggested: add a shared-basename fixture asserting both rows render distinctly, plus one test that clicks the `Rank 1:` item and asserts `onSelectFolder` was called with rank1's `remotePath`.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -661,6 +661,11 @@ const RemoteSyncConfigurator = () => {
                         setPersistentSavedConnectionList(updatedConnections);
                         remote.persistentState.updateSavedRemoteFoldersConnection(oldConnection, updatedConnection);
 
+                        if (hasPerformanceDiscoveryChanged(oldConnection, updatedConnection)) {
+                            remote.persistentState.deleteSavedPerformanceFolders(updatedConnection);
```
dcblundell: **Should fix:** the predicate is unit-tested but this wiring isn't

`hasPerformanceDiscoveryChanged` has good coverage as a pure function, but the consumer — `deleteSavedPerformanceFolders(updatedConnection)` plus `setSelectedPerformanceFolder(undefined)` on connection edit — has none, and the spec's own TODO block records "No test for clicking Edit button". Stale cached remote paths after toggling the multihost flag is the bug this fixes, so only the leaf is protected.

Suggested: a `RemoteSyncConfigurator` test that edits a connection to flip `multihostPerformance` and asserts the cached performance folders are dropped and the selection reset.

backend/ttnn_visualizer/views.py
```diff
@@ -1182,8 +1215,9 @@ def get_performance_results_report(instance: Instance):
         raise PerformanceReportNotLoadedException()
 
     if name and not current_app.config["SERVER_MODE"]:
-        performance_path = Path(instance.performance_path).parent / name
-        instance.performance_path = str(performance_path)
+        instance.performance_path = _performance_path_for_requested_name(
+            instance.performance_path, name
+        )
```
dcblundell: **Should fix:** the `?name=` swap boilerplate is still copy-pasted three times

Path resolution is now correctly delegated to `_performance_path_for_requested_name` (which also closes the old traversal — nice), but the surrounding block is identical at three sites (here, ~1252 and ~1294): the `request.args.get("name", None)`, the `if name and not current_app.config["SERVER_MODE"]` gate, the assignment, and the same `************`-prefixed f-string log.

Since all three were touched anyway, folding the whole block into one helper (e.g. `_apply_requested_performance_name(instance)`) gives the `SERVER_MODE` gate a single definition rather than three places to keep in step.

src/components/report-selection/RemoteConnectionDialog.tsx
```diff
@@ -243,6 +243,21 @@ const RemoteConnectionDialog = ({
...
                 </FormGroup>
 
+                <FormGroup
+                    label='Multihost performance reports'
+                    subLabel='The path above must be the folder that directly contains the rank folders (e.g. "/<PATH TO TT METAL>/generated/profiler/ttrun/"), so that each report is found at "ttrun/rank0/reports/<REPORT>"'
+                    labelFor='remote-performance-multihost'
+                >
+                    <Checkbox
+                        id='remote-performance-multihost'
+                        label='Search per-rank subdirectories'
```
dcblundell: **Should fix:** the checkbox ends up with two labels

The `FormGroup` supplies `label='Multihost performance reports'` + `labelFor`, and the `Checkbox` supplies its own `label`, so the computed accessible name is `"Multihost performance reports Search per-rank subdirectories"`. The three new tests use `getByLabelText`, which matches either and hides the duplication entirely.

Every other `FormGroup` in this file pairs a label with an unlabelled `InputGroup`; dropping one of the two would match that pattern (the long guidance already lives in `subLabel`). Worth switching the specs to `getByRole('checkbox', { name: ... })` so the announced name is what's pinned.

Minor, same area: `'Search per-rank subdirectories'` is declared independently here and as `MULTIHOST_CHECKBOX_LABEL` in the spec — CONVENTIONS.md asks for one exported constant imported on both sides.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -954,8 +971,39 @@ def read_remote_file(
...
+class RemoteReportCounts(NamedTuple):
+    """Report folders found per kind; ``None`` when that path is not configured."""
+
+    profiler: Optional[int]
+    performance: Optional[int]
+
+
 @remote_exception_handler
-def check_remote_path_for_reports(remote_connection):
+def check_remote_path_for_reports(remote_connection) -> RemoteReportCounts:
```
dcblundell: **Should fix:** annotate `remote_connection` while the signature is being touched

The return type was added but the parameter is still unannotated, so the body type-checks against an implicit `Any` — including the new `remote_connection.multihostPerformance` access. Its sibling a few lines below is fully annotated (`check_remote_path_exists(remote_connection: RemoteConnection, path_key: str)`).

dcblundell (commented):
Combined review across maintainability, tests, performance, conventions and security lenses.

Structurally this is in good shape: the extractions are genuine, all gates pass locally, and `_performance_path_for_requested_name` actually closes a pre-existing `?name=` traversal. The main risk is operational rather than correctness — multihost discovery multiplies an existing per-directory SSH fan-out by the rank count, so a rank-heavy tree turns a report listing into hundreds of serial handshakes.

11 line comments below: 3 must-fix (SSH fan-out, unquoted remote shell interpolation, rank missing from the collapsed selector button) and 8 should-fix. Lower-severity nits are omitted.

backend/ttnn_visualizer/views.py
```diff
@@ -1155,6 +1156,38 @@ def delete_performance_report(performance_name, instance: Instance):
...
+    try:
+        requested_name = sanitise_path_segment(name)
+    except (TypeError, ValueError):
+        logger.warning("Ignoring unusable performance report name: %r", name)
+        return active_path
+
+    requested = reports_directory / requested_name
+    if requested.is_dir():
+        return str(requested)
+
```
aidemsined: > [!CAUTION]
> An unqualified sibling wins over the active report's own rank, so one rank can silently serve another rank's data.

The exact-match branch (`requested = reports_directory / requested_name; if requested.is_dir(): return ...`) runs *before* the rank fallback. Every `?name=` value reaching these three routes is the report's **display** name — `get_remote_performance_folder` sets `reportName = profile_folder.split("/")[-1]` (`sftp_operations.py:915`), i.e. the bare `2026_07_28_18_04_24`, not `..._rank1`.

So with the active report at `<perf>/<TS>_rank1` and `?name=<TS>`, if a directory literally named `<TS>` also exists, this returns **that** one: rank1's report renders rank0's numbers, with no error and nothing in the UI to indicate it.

That precondition is created by this PR rather than being hypothetical. Before it, the way to reach a multihost report was to point `performancePath` directly at `<root>/rank0/reports`, which synced rank0 under the bare `<TS>`. After upgrading, `<TS>`, `<TS>_rank0` and `<TS>_rank1` coexist in the same `<host>/performance/` directory — exactly the shape that trips this.

`test_exact_folder_names_still_win` (`tests/views/test_remote_reports.py:725`) pins the current ordering, and no test covers a bare name colliding with a stale unqualified folder.

Suggested direction: when the active report carries a rank, try `f"{requested_name}_{rank}"` **first** and fall back to the exact match, so a bare name can never resolve to a different rank than the one that's loaded.

src/functions/reportRank.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    const suffixed = (segments.at(-1) ?? '').match(RANK_SUFFIX_PATTERN);
+
+    return suffixed ? Number(suffixed[1]) : null;
+};
+
+/** Report name without the `_rank<N>` qualifier sync appends, for display. */
+export const getReportBaseName = (remotePathOrName: string): string =>
+    (getSegments(remotePathOrName).at(-1) ?? '').replace(RANK_SUFFIX_PATTERN, '');
```
aidemsined: > [!CAUTION]
> Rank reaches the picker's active-row match but not the app's canonical report identity, which stays non-unique across ranks and changes across a reload.

`getReportId` in `src/functions/reportLinks.ts:13-31` is documented as "Stable report identity: final path segment". Under the multihost layout the final segment of `remotePath` is the same timestamp for **every** rank of a launch, so the declared identity function is non-unique in precisely the case this feature exists for.

Consumers one level out: `src/hooks/useReportLinkBadgeIds.ts:38` and `src/components/ReportLinkStatus.tsx:55` both compute `getReportId(activePerformanceReport.path, reportName)`, and `RemoteFolderSelector.tsx:51,121` key the link-status badge and the link-state sort on the same id.

The identity also **flips over the report's lifecycle**. On fresh remote selection `activePerformanceReport.path` is the remote path (`RemoteSyncConfigurator.tsx:334`), giving `<TS>`; after a reload it's `performance_name` (`useRestoreInstance.tsx:53`), giving `<TS>_rank0`. Two consequences: a profiler↔performance link recorded before a reload isn't found after it, and while the remote listing is showing, linking one rank lights the badge on every rank of that launch.

Either qualify `getReportId` the same way, or make the synced name a field on the wire so there's one identity (see the related comment on `RemoteConnection.ts`).

backend/ttnn_visualizer/models.py
```diff
@@ -248,19 +249,54 @@ def sanitise_path_segment(value: object) -> str:
...
     """
     if remote_path is None:
         return None
     try:
-        return sanitise_path_segment(Path(remote_path).name)
+        segment = sanitise_path_segment(Path(remote_path).name)
     except (TypeError, ValueError):
         return None
 
+    rank_directory = rank_directory_from_remote_path(remote_path)
```
aidemsined: > [!CAUTION]
> The offline synced listing round-trips through this helper and double-qualifies, which breaks mount permanently rather than once.

`backend/ttnn_visualizer/local_remote_reports.py` isn't touched by this PR but is the third encoder of the rank convention, and it owns the *inverse* mapping (local segment → remote path). `_synthetic_remote_path` (`:27-32`) rebuilds `remotePath = f"{configured_path.rstrip('/')}/{entry.name}"` from the **local** folder name, which is already rank-qualified.

When the configured path itself contains a rank component, that synthetic path carries the rank twice:

```
folder_segment_from_remote_path("/home/u/runs/rank0/reports/2025_01_01_rank0")
  -> "2025_01_01_rank0_rank0"
```

`RemoteSyncConfigurator.tsx:509,522` posts that folder to `/remote/use`, which resolves the segment via `_safe_report_folder_name` (`views.py:2112`) and 404s with `_REPORT_NOT_SYNCED_LOCALLY`. Unlike a one-time migration this recurs on **every** mount of an offline row.

The client twin diverges here too: `getSyncedFolderName` returns `2025_01_01_rank0` for the same input, because `2025_01_01_rank0` doesn't match `^rank\d+$` — so the picker and the backend disagree about which folder that row even is.

`list_local_synced_report_folders` also emits `reportName = <local dir name>` (rank-qualified) while the SSH path emits the unqualified basename, so the same report appears under two different `reportName`s depending on which listing produced it.

This file needs to be part of the same change.

src/definitions/RemoteConnection.ts
```diff
@@ -14,6 +14,8 @@ export interface RemoteConnection {
     profilerPath: string;
     performancePath?: string;
     identityFile?: string; // Optional path to SSH private key.
+    // Performance reports sit in per-rank subdirectories of performancePath.
+    multihostPerformance?: boolean;
```
aidemsined, line 18: > [!CAUTION]
> The synced segment is computed in two languages that must agree forever, with nothing that fails when they drift — and the server already computes the answer it could just send.

The maintenance contract this introduces is that `models.py:253-298` (`RANK_DIRECTORY_RE`, `RANK_SUFFIX_RE`, `rank_directory_from_remote_path`, `folder_segment_from_remote_path`) and `src/functions/reportRank.ts:5-53` stay in lockstep across four regexes and two independent segment-splitting implementations. There is no test that compares the two, so drift is silent — and the separate comments on `re.ASCII` and on whitespace/backslash normalisation are two instances of that drift already present on arrival.

The mirror doesn't need to exist. `RemoteReportFolder` (`models.py:388`) is the wire type, and `_annotate_last_synced` (`views.py:1573-1581`) **already computes the exact synced segment per folder** and discards it after using it for `lastSynced`.

Emitting it — `syncedName: str`, plus `rank: Optional[int]` — collapses a lot:

- `reportRank.ts` becomes a display formatter with no parsing
- `isPerformanceFolderActive.ts` loses its second clause
- the rank-guessing fallback in `_performance_path_for_requested_name` becomes unnecessary, because the client would send the real folder name

One-language ownership is the right factoring here specifically because the server is the only side that **writes** the folder; the client only ever reads back a decision the server already made.

backend/ttnn_visualizer/models.py
```diff
@@ -248,19 +249,54 @@ def sanitise_path_segment(value: object) -> str:
...
+    ``tt-run --tracy`` launch names its report from its own start time at second
+    granularity, so ranks routinely produce the same basename and would
+    otherwise sync on top of each other. Stays a single flat segment because
+    callers join it straight onto the report directory.
     """
     if remote_path is None:
         return None
     try:
-        return sanitise_path_segment(Path(remote_path).name)
+        segment = sanitise_path_segment(Path(remote_path).name)
```
aidemsined, line 318: > [!WARNING]
> Qualification is keyed on the path shape rather than the flag, so it silently renames synced folders for existing connections — and applies to profiler reports too.

The suffix is appended whenever `rank_directory_from_remote_path` finds a `^rank\d+$` component anywhere in `Path(remote_path).parts`, with no reference to `multihostPerformance`. The PR description notes this, but the downstream consequences are worth spelling out because `_safe_sync_destination_segment` (`sftp_operations.py:1273`) is shared by **both** `sync_remote_profiler_folders:1294` and `sync_remote_performance_folders:1318` — so memory reports get qualified as well, even though the feature is performance-only.

```
/home/u/runs/rank0/reports/2025_01_01            -> 2025_01_01_rank0
/home/u/tt-metal/generated/rank3/ttnn/reports/x  -> x_rank3
```

For anyone whose configured path already contains a rank component — including the pre-feature workaround of pointing straight at one rank — that means:

- `_annotate_last_synced` (`views.py:1576`) probes the new name, so an already-synced report shows as never synced
- the next Sync writes a **second** folder, orphaning the first, while the existing `instances.performance_path` row still points at the old one (`utils.py:479` only rewrites path *prefixes*, so there's no cleanup)
- `/remote/use` (`views.py:2173`) computes the new segment, fails `is_valid_performance_report_dir`, and returns `_REPORT_NOT_SYNCED_LOCALLY` for a report that is in fact synced

The profiler side also didn't get the matching read-path fix: `RemoteSyncConfigurator.tsx:592` still does `folder.remotePath?.includes(activeProfilerReport.path)`, and `activeProfilerReport.path` is `profiler_name` = `X_rank0`, which is not a substring of `<root>/rank0/reports/X` — so the memory dropdown loses its selection on reload. Only the performance side gained `isPerformanceFolderActive`.

Gating qualification on the flag, or handling the migration explicitly, would contain this.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1018,18 +1082,57 @@ def check_remote_path_exists(remote_connection: RemoteConnection, path_key: str)
...
+    """
     if not root_folder:
         return []
 
     matched_folders: List[str] = []
 
     # Build SSH command to find directories in root_folder (never prompts for password)
     ssh_cmd = _ssh_cmd_prefix(remote_connection) + [
-        f"find '{root_folder}' -maxdepth 1 -type d -not -path '{root_folder}'",
+        _report_search_find_expression(root_folder, subdirectory_glob),
```
aidemsined: > [!WARNING]
> One unreadable subdirectory now discards every report that was found, and reports it as a permission error on the root.

The new expression at `:1108` uses `-mindepth 3 -maxdepth 3 … -path '<root>/rank[0-9]*/reports/*'`, so `find` descends through every depth-1 and depth-2 directory under the root rather than just the root's children as `-maxdepth 1` did.

Running that expression against a tree containing one `chmod 000` sibling: stdout contains both real report paths **and** the process exits 1. `subprocess.run(..., check=True)` therefore raises, `result.stdout` is thrown away, and `:1180-1191` matches `"permission denied" in stderr` and raises

```
Permission denied accessing '<root>'. The user '<u>' does not have read access to this directory.
```

Both halves are wrong: the root *is* readable, and discovery returns nothing despite having successfully found the reports. On a shared multihost profiler root — ranks running under different users, or a rank's `.logs` at mode 700 — this is a likely rather than exotic configuration.

There's no test for `find_folders_by_files` on a nonzero exit; the nearest tests (`tests/test_sftp_operations.py:157-198`) cover `get_remote_file_list`, a different function.

Treating a partial-success exit as partial success (`find` exits nonzero for unreadable subtrees even when it printed matches) would keep discovery working, and would stop the error being attributed to the wrong directory.

src/components/report-selection/RemoteFolderSelector.tsx
```diff
@@ -188,7 +213,14 @@ const formatRemoteFolderPath = (
 
 const filterFolders =
     (type: FolderTypes, connection?: RemoteConnection): ItemPredicate<RemoteFolder> =>
-    (query, folder) =>
-        formatRemoteFolderPath(folder, type, connection).toLowerCase().includes(query.toLowerCase());
+    (query, folder) => {
+        const normalisedQuery = query.toLowerCase();
+
+        // Match the raw path too, so a query like `rank0` still finds a folder
+        // labelled `Rank 0: ...`.
```
aidemsined, line 208: > [!WARNING]
> The cache merge is keyed on `reportName`, which this PR deliberately makes non-unique.

`src/functions/mergeRemoteFolders.ts:16` matches cached folders with

```ts
savedFolders?.find((f) => f.reportName === updatedFolder.reportName)
```

For a multihost listing every rank of one launch shares `reportName` (`sftp_operations.py:915`), so all rank rows match the **first** cached row.

It happens to be harmless today only because `RemoteReportFolder.model_dump()` always emits all four fields, so the `...updatedFolder` spread overwrites everything the stale match contributed. That makes it one added optional field away from cross-contaminating rank rows — a field that's absent on the incoming folder would be inherited from a different rank's cache entry.

There's no test covering merge with duplicate `reportName`s. Keying on `remotePath` (unique per rank by construction) would make the collision structurally impossible rather than incidentally survivable.

backend/ttnn_visualizer/views.py
```diff
@@ -1155,6 +1156,38 @@ def delete_performance_report(performance_name, instance: Instance):
     )
 
 
+def _performance_path_for_requested_name(active_path: str, name: str) -> str:
+    """Resolve a ``?name=`` report swap against the synced performance directory.
```
aidemsined: > [!WARNING]
> The flat-segment constraint is a symptom of sibling arithmetic here, not a requirement — and `_rank<N>` isn't collision-safe as a naming scheme.

This helper resolves `?name=` as `Path(active_path).parent / name`, and the stated reason rank must be encoded into the folder *name* is that this sibling assumption has to keep holding. But the repo already has the canonical way to build a report directory: `create_path_resolver(...).get_base_report_path("performance")` (used at `:1029`), plus `local_synced_report_path` / `remote_host_report_path` (`local_remote_reports.py:24`).

Resolving `?name=` through the resolver instead of through *the active report's parent* removes the constraint that forced name-encoding, and removes the "fall back to the active report's own rank" heuristic — which is a guess that appends one report's rank to a different report's name. That matters because the comparison flow passes other reports' names through the same query param (`ComparisonReportSelector.tsx:52`).

On the naming scheme itself: a single-host report genuinely named `2026_07_28_rank3` is indistinguishable from rank 3 of `2026_07_28`, the suffix is unescaped, and adding a second dimension later (host, job id) needs a second suffix with an ambiguous parse.

If the name must carry rank, carrying a normalised `int` rather than a verbatim directory string would at least make it single-valued.

backend/ttnn_visualizer/models.py
```diff
@@ -248,19 +249,54 @@ def sanitise_path_segment(value: object) -> str:
...
+
 def folder_segment_from_remote_path(remote_path: Optional[str]) -> Optional[str]:
-    """Sanitised basename of a remote report path, or None if missing/invalid.
+    """Sanitised local folder segment for a remote report, or None if invalid.
 
-    Sync destinations and mount lookups must use this so write and read segments
-    stay aligned (``sftp_operations`` ↔ ``views``).
+    Sync destinations, mount lookups and last-synced probes must all use this so
+    write and read segments stay aligned (``sftp_operations`` ↔ ``views``).
+
```
aidemsined, line 304: > [!WARNING]
> Rank is introduced as an opaque case-preserving string, while the rest of the codebase already models it as `int`.

`models.py` declares `rank: int = 0` on roughly a dozen models (`:37, 59, 66, 84, 99, 122, 130, 157, 168, 176, 193, 210, 221`), `queries.py:235-262` has a full `rank` column story, and the API already exposes `?rank=N` params (`useAPI.tsx:555, 563`).

The new code instead keeps rank as a string: `rank_directory_from_remote_path` returns the directory verbatim, `rank_suffix_from_segment` returns `rank<N>`, and only the client recovers an integer via `Number(...)`.

Two consequences of the string choice:

- `Rank0/` and `rank0/` become two distinct local folders for the same rank
- the case-insensitivity has to be restated in four separate regexes across two languages

Parsing once to `int` at the discovery boundary and deriving every name from it (`f"rank{rank}"`) gives one spelling and one type, and would also remove the `Number(rank)` mismatch where `rank0`, `rank00` and `rank000` all render as `Rank 0` while the backend keeps them as three distinct folders.

docs/src/remote-sync.md
```diff
@@ -117,6 +117,19 @@ either be any folder on the remote machine where you have a sub-folders with rep
 stored there yourself, or you can point it at the generated directory in `tt-metal`:
 `/home/username/tt-metal/generated/profiler/reports/`.
 
+#### Multihost performance reports
+
+`tt-run --tracy` writes one report per rank, each under its own `rank<N>` directory, so the
+reports are one level deeper than a single-host run puts them. To pick these up, tick
+_Multihost performance reports_ and point the _Performance report folder path_ at the folder
```
aidemsined, line 125: > [!WARNING]
> The flag is an exclusive mode switch, but the docs read as if it were additive.

With `multihostPerformance` on, discovery looks **only** at `rank[0-9]*/reports` (`sftp_operations.py:975-993`), so single-host reports sitting directly under the same path disappear from the list — and toggling the flag wipes the cached list (`RemoteSyncConfigurator.tsx:664`).

This section describes ticking the box and repointing the path, but never states that the two layouts are mutually exclusive per connection. A user with both layouts under one root will read this and reasonably conclude the flag widens the search rather than replacing it.

Either say so explicitly here, or model the field as a layout enum (`PerformanceLayout.SINGLE_HOST | PER_RANK`) so the mode semantics are explicit on the wire and a future third layout doesn't need a second boolean.

Worth checking the same paragraph doesn't commit the project to a guarantee the code doesn't enforce — the rank suffix is described as the naming rule, but nothing validates that a report directory isn't *already* named `..._rank<N>`.

src/functions/reportRank.ts
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+const RANK_DIRECTORY_PATTERN = /^rank(\d+)$/i;
+// Synced copies carry the rank as a name suffix, since the local report folders
+// are siblings and cannot nest a rank directory.
+const RANK_SUFFIX_PATTERN = /_rank(\d+)$/i;
+
+const getSegments = (remotePath: string): string[] => remotePath.split('/').filter(Boolean);
```
aidemsined: > [!WARNING]
> Third path-basename implementation in the codebase, and the exports don't distinguish which input shape each one accepts.

`getSegments` + `.at(-1)` duplicates `src/functions/getFolderNameFromPath.ts` and `pathBasename` in `src/functions/reportLinks.ts:26-31`. That's three sites — past the point where a shared helper is warranted — and `reportLinks.pathBasename` is the better of the two, since it normalises `\` and rejects `.` / `..`. Reusing it would also close the normalisation gap raised separately on line 52.

On the interface: `getReportRank` / `getReportBaseName` take a "remotePathOrName" that in practice covers three genuinely different shapes — a true remote path, the synthetic remote path built by `local_remote_reports._synthetic_remote_path`, and a bare local folder name — and infer which one they got. Meanwhile the sibling `getSyncedFolderName` accepts only a true remote path, and applied to an already-qualified name produces `x_rank0_rank0`.

A caller can't tell from the signatures which spelling is legal where. Splitting "parse an identity" from "format a label", and typing the inputs distinctly, would make the misuse unrepresentable rather than a convention.

src/components/report-selection/RemoteFolderSelector.tsx
```diff
@@ -165,6 +166,22 @@ const RemoteFolderSelector = ({
     );
 };
 
+/**
+ * Label a multihost report by its rank rather than its raw path, so both
+ * `<ttrun>/rank0/reports/2026_07_28_18_04_24` and its synced copy
+ * `2026_07_28_18_04_24_rank0` read `Rank 0: 2026_07_28_18_04_24`.
```
aidemsined: > [!WARNING]
> Rank-aware labelling is gated on remote-connection state, so every other place the report is named still shows the raw `x_rank0`.

`formatMultihostPerformanceLabel` is reached only when `type === 'performance' && selectedConnection.multihostPerformance` — i.e. the condition is shaped by which connection is selected, not by what the report is.

Separately from the trigger-button case already raised on line 182, that leaves the raw qualified name rendered in:

- `LocalFolderPicker.tsx:88-97`
- the comparison dropdown, `ComparisonReportSelector.tsx:52`
- `PerfReport.tsx:614`
- `FooterInfobar.tsx:69`
- sync toasts, `RemoteSyncConfigurator.tsx:346`

So the same report reads as `Rank 0: 2026_07_28_18_04_24` in one picker and `2026_07_28_18_04_24_rank0` everywhere else, including after a reload when no remote connection is selected at all.

Rank display belongs at the shared report-label layer rather than behind a connection-shaped condition in one component — and it falls out for free if rank becomes a field on the wire.

backend/ttnn_visualizer/views.py
```diff
@@ -1155,6 +1156,38 @@ def delete_performance_report(performance_name, instance: Instance):
...
+
+    rank = rank_suffix_from_segment(Path(active_path).name)
+    if rank:
+        qualified = reports_directory / f"{requested_name}_{rank}"
+        if qualified.is_dir():
+            return str(qualified)
+
+    return str(requested)
+
+
```
aidemsined, line 1176: > [!WARNING]
> `CONVENTIONS.md:379` now describes code this PR deleted, and it's a normative rule rather than prose.

That section pins the exact expression this helper replaces — computing `Path(instance.performance_path).parent / name` from a `?name=` swap — and names `get_performance_results_report`, `get_performance_data_raw` and `get_performance_device_meta` as "the live examples" of the rule.

`docs/src/remote-sync.md` was updated, but the convention doc still points at the old shape, so the next reader following CONVENTIONS.md will reimplement what this helper exists to centralise. By the repo's own rule that stale comments are bugs (`CONVENTIONS.md:101`), this helper is where that guard requirement now lives and the doc should say so.

Also worth a look while you're here, though outside this diff: the remote branch of `delete_performance_report` (`:1115-1150`) resolves a report by name **without** going through the new helper, so it doesn't get the traversal guard or the rank resolution. Same question for `create_path_resolver`'s report-mode handling — both answer "resolve a performance report by name", which this PR has now centralised for three routes but not for those.

backend/ttnn_visualizer/models.py
smountenay-tt: Done in `01d13efe`. Rank is parsed once, at the discovery boundary, and every name is derived from that number: `folder_segment_from_remote_path` builds `f"{segment}_rank{rank}"` rather than echoing the directory, so `rank0/`, `Rank0/` and `rank00/` collapse to one local folder instead of three.

`rank_directory_from_remote_path` is folded into `rank_from_remote_path` (`-> Optional[int]`), the unused `rank_suffix_from_segment` is deleted, and `split_rank_suffix` now reads the number straight out of the suffix — two regexes on the backend, one display-only regex on the client. Reads stay case-insensitive so folders synced by earlier builds keep their rank, with a test pinning that.

`rank: int` also travels on the wire on `RemoteReportFolder`, so the client no longer recovers it with `Number(...)`.

src/definitions/RemoteConnection.ts
smountenay-tt: Agreed, and the computation moved to the server. `RemoteReportFolder` now carries `syncedName` and `rank`, populated by the side that actually writes the folder, for both the SSH listing and the offline synced listing. The client no longer parses a path or recomputes a folder name — `reportRank.ts` is display-only formatting over the wire fields, so there is no second implementation left to drift.

src/functions/reportRank.ts
smountenay-tt: Fixed, and this one had teeth. `ReportFolder` carries `syncedName`, and `getReportId(syncedName, path)` is the identity everywhere now — link badges, both pickers, and the active report — so it is unique per rank and stable across a reload.

It also turned out the `?name=` callers were passing `activePerformanceReport.reportName`, which is the report name on the remote host and so drops the rank qualifier. A freshly selected multihost report therefore resolved to a directory that does not exist and overwrote a correct `instance.performance_path` with the unqualified sibling. That is `faffd7db`: the on-disk folder is derived once in `activePerformanceReportFolderNameAtom` and read at every call site.

backend/ttnn_visualizer/views.py
smountenay-tt: `CONVENTIONS.md` is updated — the section now names `_apply_requested_performance_name` as the owner of the `?name=` block and tells new routes to call it rather than rebuild `Path(instance.performance_path).parent / name`.

On the out-of-diff pointer, thank you: `delete_performance_report` was worse than a missing guard. The remote branch never joined the report name onto the directory, so deleting one report `rmtree`d the entire per-host performance directory — every rank of the launch — and the branch was chosen by the *active* report's location rather than the location of the report being deleted, so a remote report being open sent a local delete into the synced tree. `delete_profiler_report` had the same shape. Both now resolve through one helper against the local data directory the paired listing reads, with the name sanitised (`60951ab7`, regression tests in `tests/views/test_report_deletion.py`).

`create_path_resolver`'s report-mode handling is filed as #1821 rather than fixed here: in direct-report mode the listings read the tt-metal tree while deletes target our data directory, so the delete button silently 404s. Making those agree means `rmtree` inside the user's tt-metal checkout, which wants a deliberate decision rather than a quiet refactor.

backend/ttnn_visualizer/views.py
smountenay-tt: Two of the three are done; pushing back on the resolver.

The rank guess is gone. `_apply_requested_performance_name` resolves `?name=` exactly as given and never appends the active report's rank to another report's name, so the comparison flow can no longer be answered with the wrong rank's data. And `01d13efe` takes the fallback you offer at the end: the suffix is built from the parsed `int`, so it is single-valued.

On routing `?name=` through `create_path_resolver` — `get_base_report_path` picks its tree from config, and in direct-report mode (`TT_METAL_HOME` set) it returns the tt-metal tree, which is the wrong base for a report synced into the remote data directory. Deriving a sibling of the path that is actually mounted keeps that combination working, where a config-derived base would not. The flat-segment constraint is not a symptom of the sibling arithmetic either: any base-plus-name join needs the name to be one segment, and `sanitise_path_segment` is wanted there as a traversal guard regardless of which base is chosen.

Rank-in-the-name is likewise not a consequence of the sibling assumption — it is forced by the on-disk collision, since every rank of one launch names its report from its own start time at second granularity and they routinely match. The resolver would not remove that.

Agreed that `<name>_rank3` is ambiguous against a single-host report genuinely named that. That is why `split_rank_suffix` is gated on the same multihost flag the write path uses and documented as unsound otherwise, and why the client treats a parsed rank as a label that must never choose a directory. A second dimension later would want a real encoding rather than a second suffix; noted for when there is one.

dcblundell (approved):

PR #1820, opened 2026-07-30
Fix version check flagging updates when the local app is ahead

### LINKED ISSUE #1818: Version check reports an update when the local app is newer than the published version
The footer version indicator reports "App update available" when the running app is **newer** than the published one. Observed running `v0.96.0` against a published `v0.95.1`:

### Cause

`getVersionOutdatedLevel` in `src/components/AppVersionStatus.tsx` compares major/minor/patch as three independent deltas instead of establishing an ordering, so a later component can fire even when an earlier one already proves the local build is ahead.

For `0.96.0` vs `0.95.1`: `majorDiff = 0`, `minorDiff = -1` (matches none of the `=== 1` / `=== 2` / `> 2` branches), so it falls through to the final `return patchDiff > 0 ? ONE : NONE` — and `patchDiff = 1`. The patch comparison is only meaningful when major *and* minor are equal, and nothing enforces that.

The minor branches have the same shape of bug: they don't require `majorDiff === 0` either.

Simulating the current implementation:

| local | published | reported | correct? |
|---|---|---|---|
| 0.96.0 | 0.95.1 | ONE | **wrong** — observed above |
| 1.0.0 | 0.99.5 | THREE | **wrong** — a local major bump reports the most severe level |
| 2.5.0 | 1.6.0 | ONE | **wrong** |
| 0.96.0 | 0.96.0 | NONE | ok |
| 0.95.0 | 0.96.0 | ONE | ok |
| 0.96.0 | 0.96.1 | ONE | ok |
| 0.96.1 | 0.96.0 | NONE | ok |

### Why it shows up in normal use

"Latest" comes from PyPI's release feed (`backend/ttnn_visualizer/views.py`, `pypi.org/rss/project/ttnn-visualizer/releases.xml`), which legitimately lags a freshly tagged GitHub release. So "local is ahead of published" is a routine state for anyone on a dev build or a just-released version — not an edge case.

### Suggested fix

Compare `(major, minor, patch)` as an ordered triple, return `NONE` unless the published version sorts strictly after the local one, and only then derive the severity level from the distance. Worth a unit test over the table above, including the local-ahead rows.

dcblundell:
## Summary

`getVersionOutdatedLevel` compared major, minor and patch as three independent deltas, so a lower-precedence component could fire even when a higher-precedence one already proved the local build was newer than the published release. Running `v0.96.0` against a published `v0.95.1` fell through to `patchDiff > 0` and reported "App update available". Because PyPI's release feed legitimately lags a fresh tag, "local is ahead of published" is a routine state rather than an edge case.

- Compare `(major, minor, patch)` as an ordered triple, returning `NONE` unless the published version sorts strictly after the local one, and only then deriving the severity level.
- Extract the comparison into the pure `src/functions/getVersionOutdatedLevel.ts`, with the `OutdatedLevel` enum moved to the existing `src/definitions/Versions.ts` alongside `DBVersionValidation`. The enum stays numeric because its ordering is load-bearing (`> OutdatedLevel.NONE`) and it keys `OUTDATED_CLASS_MAP`.
- Drop the dead `if (!current || !latest)` guard: `semverParse` always returns a `SemVer`, falling back to `0.0.0`.

No other version comparison in the frontend shares the bug — `evaluateDbVersion`, `validateNpeData` and the `schema_version` check all compare `major` only.

Prerelease ordering is deliberately unchanged: `semverParse` captures `prerelease` but the comparison ignores it, so a local `0.97.0-dev` against a published `0.97.0` still reports up to date. Flagging every dev build seems worse than the status quo, but happy to revisit.

## Test plan

- [x] New `tests/getVersionOutdatedLevel.spec.ts` covers the table from the issue, including every local-ahead row (`0.96.0`/`0.95.1`, `1.0.0`/`0.99.5`, `2.5.0`/`1.6.0`, `0.96.1`/`0.96.0`), the level one/two/three ladder, missing versions, and omitted minor/patch components.
- [x] `pnpm test` — 139 files, 1192 tests passing.
- [x] `pnpm lint`, `pnpm lint:ts`, `pnpm lint:spdx`, Prettier check on touched files.
- [ ] `/styleguide` still shows up to date plus levels one to three (samples are `0.80.0`/`0.79.0`/`0.78.0`/`0.77.0` against `0.80.0`, all local-behind-or-equal, so unaffected by the fix).

Fixes #1818

### DIFF
--- a/src/functions/getVersionOutdatedLevel.ts
+++ b/src/functions/getVersionOutdatedLevel.ts
@@ -5,37 +5,42 @@
 import { OutdatedLevel } from '../definitions/Versions';
 import { semverParse } from './semverParse';
 
-const MINOR_DIFF_LEVEL_ONE = 1;
-const MINOR_DIFF_LEVEL_TWO = 2;
-
 /**
  * Compares (major, minor, patch) as an ordered triple. PyPI's release feed lags a freshly tagged
  * release, so a local build ahead of the published one is routine and must report NONE.
+ *
+ * Prereleases compare equal to their release: a local 1.0.0-rc1 is not prompted towards 1.0.0,
+ * because prompting every dev build to "update" to the version it already contains is worse noise
+ * than the missed prompt.
  */
-export const getVersionOutdatedLevel = (appVersion: string, latestAppVersion: string): OutdatedLevel => {
+export const getVersionOutdatedLevel = (
+    appVersion: string | undefined,
+    latestAppVersion: string | undefined,
+): OutdatedLevel => {
     if (!appVersion || !latestAppVersion) {
         return OutdatedLevel.NONE;
     }
 
     const current = semverParse(appVersion);
     const latest = semverParse(latestAppVersion);
+    const majorDiff = latest.major - current.major;
 
-    if (latest.major !== current.major) {
-        return latest.major > current.major ? OutdatedLevel.THREE : OutdatedLevel.NONE;
+    if (majorDiff !== 0) {
+        return majorDiff > 0 ? OutdatedLevel.THREE : OutdatedLevel.NONE;
     }
 
-    if (latest.minor !== current.minor) {
-        if (latest.minor < current.minor) {
+    const minorDiff = latest.minor - current.minor;
+
+    if (minorDiff !== 0) {
+        if (minorDiff < 0) {
             return OutdatedLevel.NONE;
         }
 
-        const minorDiff = latest.minor - current.minor;
-
-        if (minorDiff === MINOR_DIFF_LEVEL_ONE) {
+        if (minorDiff === 1) {
             return OutdatedLevel.ONE;
         }
 
-        return minorDiff === MINOR_DIFF_LEVEL_TWO ? OutdatedLevel.TWO : OutdatedLevel.THREE;
+        return minorDiff === 2 ? OutdatedLevel.TWO : OutdatedLevel.THREE;
     }
 
     return latest.patch > current.patch ? OutdatedLevel.ONE : OutdatedLevel.NONE;
--- a/src/functions/getVersionOutdatedLevel.ts
+++ b/src/functions/getVersionOutdatedLevel.ts
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { OutdatedLevel } from '../definitions/Versions';
+import { semverParse } from './semverParse';
+
+const MINOR_DIFF_LEVEL_ONE = 1;
+const MINOR_DIFF_LEVEL_TWO = 2;
+
+/**
+ * Compares (major, minor, patch) as an ordered triple. PyPI's release feed lags a freshly tagged
+ * release, so a local build ahead of the published one is routine and must report NONE.
+ */
+export const getVersionOutdatedLevel = (appVersion: string, latestAppVersion: string): OutdatedLevel => {
+    if (!appVersion || !latestAppVersion) {
+        return OutdatedLevel.NONE;
+    }
+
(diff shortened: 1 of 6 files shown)


src/functions/getVersionOutdatedLevel.ts
```diff
@@ -0,0 +1,47 @@
...
+    appVersion: string | undefined,
+    latestAppVersion: string | undefined,
+): OutdatedLevel => {
+    if (!appVersion || !latestAppVersion) {
+        return OutdatedLevel.NONE;
+    }
+
+    const current = semverParse(appVersion);
+    const latest = semverParse(latestAppVersion);
+    const majorDiff = latest.major - current.major;
```
aidemsined, line 30: > [!WARNING]
> Dropping the guard rests on `semverParse` being a total function, and nothing in the suite pins that.

The reasoning is correct today — `src/functions/semverParse.ts:12-32` has exactly three exits, two of them the `0.0.0` fallback and one an object literal, with no path returning nullish. But there is no `tests/semverParse.spec.ts` anywhere in the 140-file suite, so if someone later tightens the signature to `SemVer | null`, `latest.major` here starts throwing and no test catches it.

The codebase is already inconsistent about whether it trusts this contract: `src/functions/validateNpeData.ts:38,44` carries `if (!parsedVersion)` / `!minSupportedVersion` checks that are dead for exactly the same reason. So the invariant is load-bearing in two places and asserted in none.

A small spec covering `semverParse(undefined)`, `semverParse('')`, `semverParse('garbage')` and `semverParse('v1.2.3')` all returning `{ major: 0, minor: 0, patch: 0 }` would pin it, and would also document that `v`-prefixed and PEP 440 inputs degrade to `0.0.0` rather than erroring — which is the boundary the next reader will wonder about.

Separately, worth confirming the "139 files / 1192 tests" figure in the test plan: `tests/*.spec.ts*` counts 140 files at this head, and a two-file addition should move the count by two.

src/functions/getVersionOutdatedLevel.ts
```diff
@@ -0,0 +1,47 @@
...
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { OutdatedLevel } from '../definitions/Versions';
+import { semverParse } from './semverParse';
+
+/**
+ * Compares (major, minor, patch) as an ordered triple. PyPI's release feed lags a freshly tagged
+ * release, so a local build ahead of the published one is routine and must report NONE.
+ *
+ * Prereleases compare equal to their release: a local 1.0.0-rc1 is not prompted towards 1.0.0,
```
aidemsined: > [!WARNING]
> The prerelease docstring asserts more than the code delivers — it holds for hyphenated semver, not for the PEP 440 spellings a Python-packaged app actually produces.

`1.0.0-rc1` does behave as documented, and the two tests prove it. But `semverParse`'s regex doesn't match the non-hyphenated forms — `1.0.0rc1`, `1.0.0.dev0`, `1.0.0.post1` all fail it and degrade to `0.0.0`. On the *local* side that inverts the intent: a dev build would compare as `0.0.0` against any published release and get a spurious level-`THREE` "major update available" prompt, which is the opposite of "not prompted towards".

Not reachable today, and worth saying why explicitly because it strengthens your case rather than weakening it: `appVersion` comes from `package.json` via `import.meta.env.APP_VERSION` (`FooterInfobar.tsx:67`), and the published side is regex-constrained to `\d+\.\d+\.\d+` by `views.py:2411`. So a prerelease can only ever appear on the local side — which is exactly where "don't nag a dev build" is unambiguously the right policy.

On the open question you raised in the description: current behaviour is right, and that asymmetry is the reason. Suggest narrowing the comment to say *hyphenated* prereleases and folding in the `views.py` constraint, so the next reader doesn't have to re-derive why ignoring `prerelease` is safe rather than merely convenient.

PR #1822, opened 2026-07-30
Read ssh config to prepopulate remote connection

### LINKED ISSUE #1724: Remote connections: Use ssh config
I think there should be an option to select from a dropdown, prepopulated with the configured hosts in ~/.ssh/config, and then either a "create new host" button / link or just show both a dropdown and the individual fields.

### LINKED ISSUE #1823: Align report select/edit/delete UI
The UI for selecting local/remote reports and servers should be unified so we don't have a different interaction model for each, e.g. deleting reports is within a dropdown and has a confirmation dialog but deleting a server is a button outside the UI and does not have a dialog.

dcblundell:
This pull request introduces several important improvements to the handling of SSH remote connections and CORS configuration in the TT-NN Visualizer backend and frontend. The main changes include a new endpoint and backend logic for parsing SSH config files, a more robust and secure CORS allowlist implementation, updates to the documentation to reflect the new SSH config host picker feature, and comprehensive tests for both CORS and SSH client behavior.

Screenshot 2026-07-30 at 15 16 12

Screenshot 2026-07-30 at 15 16 16

**SSH config host picker feature:**

* Added a backend parser (`ssh_config.py`) that reads concrete `Host` aliases from the user's `~/.ssh/config`, supporting `Include` directives and ignoring patterns/wildcards. This enables the frontend to offer a dropdown of available SSH config hosts for remote connections.
* Introduced a new API endpoint `/remote/ssh-config-hosts` in `views.py` to expose parsed SSH config hosts to the frontend, and integrated the parser into the backend. [[1]](diffhunk://#diff-34e6863077d645b69bcb1a182649fa531a6fc979eff410b026cdb9158d70405aR99) [[2]](diffhunk://#diff-34e6863077d645b69bcb1a182649fa531a6fc979eff410b026cdb9158d70405aR1683-R1689)
* Updated the frontend to import and use the SSH config host picker, allowing users to prefill connection details from their SSH config.

**CORS allowlist improvements:**

* Refactored CORS origin handling in `settings.py` to use a helper function that defaults to the app’s own origin and, in non-production, the Vite dev server. This narrows the trust boundary and makes configuration clearer and more secure. [[1]](diffhunk://#diff-4da46363434000d16aa4d53b8ca4b45d42820f8a8e6c016d4effc0655c6e6c63R22-R47) [[2]](diffhunk://#diff-4da46363434000d16aa4d53b8ca4b45d42820f8a8e6c016d4effc0655c6e6c63L29-L35) [[3]](diffhunk://#diff-4da46363434000d16aa4d53b8ca4b45d42820f8a8e6c016d4effc0655c6e6c63R123-R130) [[4]](diffhunk://#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049cL26-R28)
* Added tests in `test_settings.py` to ensure CORS defaults and configuration behave as intended.

**SSH identity file and config behavior:**

* Added tests in `test_ssh_client.py` to pin and document the contract: leaving the identity file empty uses `~/.ssh/config`, while setting it isolates the connection from config.

**Documentation updates:**

* Updated `remote-sync.md` to document the new SSH config host picker, clarify how identity files and SSH config interact, and explain the user experience in more detail. [[1]](diffhunk://#diff-e892fcd447813c394ad3895aa37934c6cacd0e8fd958c4396271157d786fe43dL23-R23) [[2]](diffhunk://#diff-e892fcd447813c394ad3895aa37934c6cacd0e8fd958c4396271157d786fe43dL99-R103)

These changes collectively improve security, user experience, and maintainability around SSH remote connections and CORS configuration.

---

**SSH config host picker:**
- Added `ssh_config.py` to parse `~/.ssh/config` for concrete Host aliases, supporting `Include` and exposing a structured API for the frontend.
- Exposed a new `/remote/ssh-config-hosts` endpoint and integrated the parser into the backend API. [[1]](diffhunk://#diff-34e6863077d645b69bcb1a182649fa531a6fc979eff410b026cdb9158d70405aR99) [[2]](diffhunk://#diff-34e6863077d645b69bcb1a182649fa531a6fc979eff410b026cdb9158d70405aR1683-R1689)
- Updated frontend to use the SSH config host picker for prefilling remote connection details.

**CORS allowlist improvements:**
- Refactored CORS origin resolution to default to the app’s own origin and the Vite dev server in non-production, and added comprehensive tests for the new logic. [[1]](diffhunk://#diff-4da46363434000d16aa4d53b8ca4b45d42820f8a8e6c016d4effc0655c6e6c63R22-R47) [[2]](diffhunk://#diff-4da46363434000d16aa4d53b8ca4b45d42820f8a8e6c016d4effc0655c6e6c63L29-L35) [[3]](diffhunk://#diff-4da46363434000d16aa4d53b8ca4b45d42820f8a8e6c016d4effc0655c6e6c63R123-R130) [[4]](diffhunk://#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049cL26-R28) [[5]](diffhunk://#diff-c9bfa2b0969f0e20a6e261b4ffe4d7ec000452135b173e916ad87f35c04bd563R1-R76)

**SSH identity/config contract:**
- Added tests to ensure SSH client commands correctly honor or isolate `~/.ssh/config` depending on whether an identity file is set.

**Documentation:**
- Updated remote sync documentation to describe the SSH config host picker and clarify SSH config and identity file interactions. [[1]](diffhunk://#diff-e892fcd447813c394ad3895aa37934c6cacd0e8fd958c4396271157d786fe43dL23-R23) [[2]](diffhunk://#diff-e892fcd447813c394ad3895aa37934c6cacd0e8fd958c4396271157d786fe43dL99-R103)

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1724.
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1823.


docs/src/remote-sync.md
```diff
@@ -96,10 +96,11 @@ sync the individual reports you would like to use.
...
+4. Enter or adjust SSH connection details (hostname or Host alias, username, and report paths). The app always connects as `user@host` and passes the form’s port explicitly, so both override any `User` or `Port` in SSH config (including `Host *`). That can differ from bare `ssh my-alias` in a terminal when the stanza has no `User`/`Port` and the form uses a different default. Editing any field the test exercises — host, username, port, identity file, or either report path — discards an earlier test result; renaming the connection does not.
```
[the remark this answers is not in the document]
dcblundell: Agreed, and it's wrong in the default case specifically. `SSHClient._build_base_ssh_cmd` only appends `-p` when `port != 22` (`ssh_client.py:122`, same for `-P` in the SFTP and SCP builders at 136 and 387), and `SSH_DEFAULT_PORT` defaults to 22 — so out of the box the port is *not* passed and the alias's `Port` from SSH config applies.

Reworded to split the two halves: the username always overrides because `user@host` is always appended, while the port only overrides when it isn't 22.

backend/ttnn_visualizer/tests/test_ssh_config.py
```diff
@@ -0,0 +1,507 @@
...
+    monkeypatch.setattr(
+        "ttnn_visualizer.views.load_ssh_config_hosts",
+        lambda: load_ssh_config_hosts(config),
+    )
+
+    response = client.get("/api/remote/ssh-config-hosts")
+
+    assert response.status_code == 200
+    # configExists true with no hosts is what tells the UI to show an empty picker
+    # rather than hide it entirely.
```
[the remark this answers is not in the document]
dcblundell: Correct — `SshConfigHostPicker` returns null on `!configExists || hosts.length === 0`, so this payload hides the picker exactly like a missing config does. The comment described a contract that doesn't exist.

Replaced with what the test actually pins: an unreadable config still answers 200 rather than erroring, and the empty host list is what hides the picker.

Separately worth a look: `configExists` currently makes no observable difference to the UI, since both `false` and `true`-with-no-hosts hide the picker. Either the field is redundant in the response, or the picker should distinguish "no `~/.ssh/config`" from "config present but only wildcard stanzas". Leaving that out of this PR.

backend/ttnn_visualizer/models.py
```diff
@@ -262,9 +262,31 @@ def folder_segment_from_remote_path(remote_path: Optional[str]) -> Optional[str]
...
+    ``username@host`` is passed to ``ssh``/``sftp``/``scp`` in option position, so a
+    leading ``-`` makes the whole token an option — ``-oProxyCommand=…`` is then run
+    through a shell. No POSIX username or DNS label starts with ``-``, so refusing one
+    costs nothing.
+    """
+    if not isinstance(value, str):
+        return value  # type: ignore[return-value]
+    if value.startswith("-"):
+        raise ValueError("must not start with '-'")
+    return value
```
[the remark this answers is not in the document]
dcblundell: Right conclusion, wrong mechanism — fixed at all three sites in d628ef95.

A JSON number can't get through. Pydantic v2 lax mode doesn't coerce numerics to `str` (that was v1); I checked every JSON-representable non-string against `RemoteConnection.model_validate(..., strict=False)` on the installed 2.12.5, and `int`, `float`, `bool`, `list`, `dict` and `None` all raise `ValidationError` before reaching argv.

What *is* coerced in lax mode is `bytes` and `bytearray`, and that does bypass the check: `username=b"-oProxyCommand=id"` was accepted and landed as the string `-oProxyCommand=id`. Not reachable today — these models are only validated from `request.get_json()` and from SQLAlchemy `JSON` columns, neither of which yields `bytes` — but the function documented an invariant it didn't actually hold, so it's worth closing.

One detail worth flagging for anyone reading later: hardening `reject_ssh_option_like` alone would only have covered the `host` path. `sanitise_ssh_username` and `MlirServerConnection._strip_required_strings` returned early on non-string input and never reached it, so the three locations you listed all genuinely needed changing. Each now raises `ValueError("must be a string")`, with regression tests over `bytes`/`bytearray` for both the `RemoteConnection` and `MlirServerConnection` targets.

backend/ttnn_visualizer/app.py
```diff
@@ -185,7 +185,7 @@ def extensions(app: flask.Flask):
     """
     flask_static_digest.init_app(app)
     if app.config["USE_WEBSOCKETS"]:
-        socketio.init_app(app)
+        socketio.init_app(app, cors_allowed_origins=app.config["ALLOWED_ORIGINS"])
```
aidemsined: > [!CAUTION]
> Passing an explicit list here removes engine.io's same-origin auto-allow, so socket connections break for every origin that isn't literally `http://localhost:{PORT}`.

`cors_allowed_origins is None` is the **only** branch in `engineio/base_server.py:284-305` that derives the allowed origin from `wsgi.url_scheme` + `HTTP_HOST` (and from `X-Forwarded-Proto`/`X-Forwarded-Host`). An explicit list is used verbatim, and `engineio/server.py:206-215` returns `400 "Not an accepted origin."` for any request whose `Origin` falls outside it.

Combined with the new default of `["http://localhost:{PORT}"]`, these first-party paths break:

- `--server` sets `HOST=0.0.0.0` (`app.py:406`) and `open_browser` opens `http://0.0.0.0:{PORT}` (`app.py:544-547`) — so the browser's `Origin` is `http://0.0.0.0:{PORT}`, not `localhost`
- `--host 127.0.0.1`, which the bind-failure message at `app.py:537` explicitly recommends
- any container or VM reached by IP or DNS name
- any `SERVER_MODE` deployment behind a reverse proxy

Before this PR `extensions.py` passed `cors_allowed_origins="*"`, so none of these were affected. The HTTP allowlist narrowing is sound and worth keeping — it's specifically the socketio side that needs the same-origin case back.

Two ways out: pass `cors_allowed_origins=None` when `ALLOWED_ORIGINS` wasn't explicitly configured (restoring the auto-derive), or append `f"http://{HOST}:{PORT}"` to the default set so the app's real serving origin is always present.

**This is also the only line in the CORS change with no test.** `grep cors_allowed_origins backend/ttnn_visualizer/tests/` returns nothing, and `test_settings.py` has no assertion that the default allowlist contains the origin the app actually serves on. A test parametrised over `HOST` in `{localhost, 0.0.0.0, 127.0.0.1}` would have caught this and would pin the invariant going forward.

One caveat on my own claim: I confirmed the library code path and the config values by reading them, but did not execute a request, so I can't say whether the polling POST or the websocket upgrade fails first — the observable symptom may be a degraded transport with console errors rather than a hard disconnect.

backend/ttnn_visualizer/models.py
```diff
@@ -262,9 +262,36 @@ def folder_segment_from_remote_path(remote_path: Optional[str]) -> Optional[str]
...
+
 def sanitise_remote_host_segment(value: object) -> str:
     """Normalise a user-provided host to a single safe path segment."""
-    return sanitise_path_segment(value)
+    return reject_ssh_option_like(sanitise_path_segment(value))
+
+
+def sanitise_ssh_username(value: object) -> str:
+    """Normalise a user-provided SSH username for use in an ``ssh`` argv."""
+    if not isinstance(value, str):
```
aidemsined, line 294: > [!WARNING]
> The two username validators this commit touched now disagree about the empty string.

`sanitise_ssh_username` strips and rejects option-like values, but never rejects empty — so `RemoteConnection(username="")` validates and produces the argv token `@host`. Its counterpart `MlirServerConnection._strip_required_strings` (`models.py:334-339`) *does* reject empty for the same conceptual field.

Since `d628ef95` already brought both functions into line on the non-string case, worth closing the remaining asymmetry in the same pass: `raise ValueError("must not be empty")` after the strip. Neither direction is currently tested, so there's no existing behaviour depending on the lenient path.

Unrelated but adjacent, since you're in this file: `sanitise_path_segment` (`models.py:243-244`) still returns non-strings unchanged. That's the right call — it's also used for report names at `views.py:2069` where raising would be a behaviour change — and the composition through `sanitise_remote_host_segment` → `reject_ssh_option_like` closes the hole for the SSH path. Just noting it so the inner helper's leniency reads as deliberate rather than missed.

backend/ttnn_visualizer/ssh_config.py
```diff
@@ -0,0 +1,300 @@
...
+        if keyword is None:
+            continue
+
+        keyword_lower = keyword.lower()
+
+        if keyword_lower == "include":
+            # OpenSSH applies Include where it appears and the surrounding Host
+            # context continues afterwards, so the open stanza is left pending —
+            # flushing here would silently drop the keywords that follow. The
+            # included file parses with its own stanza state, so it cannot inherit
```
aidemsined: > [!WARNING]
> The mid-stanza `Include` comment implies OpenSSH parity, but this deliberately differs from it.

OpenSSH passes `activep` *by pointer* into the included file, so an included fragment both inherits the currently-open stanza and can leave a *different* stanza active when control returns to the parent. This parser gives the included file fresh state and then resumes the parent's stanza afterwards.

`test_load_ssh_config_hosts_include_inside_host_keeps_later_keywords` pins that divergence as the intended behaviour, which is fine — the blast radius is bounded to a wrong `HostName` in a dropdown label or a wrong `Port` prefill, and only for configs that `Include` mid-stanza. But the comment should say "we deliberately differ from OpenSSH here, because X" rather than reading as a description of OpenSSH's semantics, so nobody later "fixes" the parser toward true parity and breaks that test without understanding why it existed.

Same file, related: `MAX_INCLUDE_DEPTH = 5` against OpenSSH's 16. A real 6-deep chain is silently truncated with only a `logger.warning` the user never sees — worth either matching 16 or noting in the docstring why 5 is enough.

Credit where due: the `Include` handling is otherwise careful. Filtering glob matches to regular files (`ssh_config.py:275-283`) correctly defuses `/dev/zero` and FIFOs, and the visited-set-before-depth-cap ordering (`:88-91`) is the right way round with a comment explaining why.

docs/src/remote-sync.md
```diff
@@ -96,10 +96,11 @@ sync the individual reports you would like to use.
...
-5. Passphrase: The app never prompts for a passphrase. If your key has one, run `ssh-add /path/to/your_private_key` once (in the same terminal you use to start the app, or in an environment the app can see), then start the app.
-6. Click "Test Connection". If it succeeds, click "Add connection".
+3. Optionally choose an **SSH config host** from the dropdown (populated from your local `~/.ssh/config`; local installs only — not available when `SERVER_MODE` is enabled). This prefills the Host alias, username, and port, and names the connection after the alias if you haven’t typed a name yourself. Selecting a config host clears the identity file so OpenSSH can still apply ProxyJump, IdentityFile, and other config for that alias, and discards any earlier connection-test result, since it was run against a different target. Saved connections live in browser localStorage separately — the dropdown is only a prefiller, not a second connection list.
```
aidemsined, line 99: > [!WARNING]
> `ALLOWED_ORIGINS` is now operationally load-bearing but appears nowhere in the docs.

`grep ALLOWED_ORIGINS docs/ README.md AGENTS.md` returns nothing — the only mention anywhere is `.env.sample:26`. After this PR, an operator running `--server`, binding a non-default host, or deploying behind a reverse proxy has a variable they must set for the app to work, and no documentation telling them it exists or what to set it to.

Given the socketio finding on `app.py:188`, this matters more than it would as a standalone doc gap: whoever hits the broken-socket symptom will have no documented lever to fix it with. A short subsection here or in `running-from-source.md` covering the default (the app's own origin, plus the Vite dev origin outside production), when you need to override it, and the comma-separated format would close both.

The rest of the docs changes in this PR are good — the picker steps read well, and the port sentence you corrected in `0f969256` now matches `ssh_client.py:126-127` exactly and is pinned by `test_base_commands_omit_the_port_flag_for_the_default_port`.

backend/ttnn_visualizer/app.py
dcblundell: Confirmed and fixed in `1a11403c`. Taking the second of your two options, but as a callable rather than a static list.

`extensions.py` no longer names origins at import time; `app.py:194` now hands `socketio.init_app` the result of `build_socketio_origin_check(ALLOWED_ORIGINS)` (`settings.py:65-89`), which returns `True` for anything in the configured allowlist and otherwise falls back to the request's own origin derived from `wsgi.url_scheme`/`HTTP_HOST` plus `X-Forwarded-Proto`/`X-Forwarded-Host` (`_request_own_origins`). Same-origin is therefore accepted unconditionally under every binding, which is what your `--server`, `--host 127.0.0.1`, container-by-IP and reverse-proxy cases all needed.

A callable rather than appending `f"http://{HOST}:{PORT}"` to the defaults for two reasons: `HOST` at config time isn't necessarily the host the browser used (a proxy or container port mapping breaks that equality), and an explicit empty list is not inert to engine.io — it means "skip the origin check entirely", so `ALLOWED_ORIGINS=""` to trust nothing would have silently opened the socket to every origin. A callable is always consulted, so that footgun is closed too; `test_socketio_still_accepts_same_origin_when_nothing_is_configured` pins it.

On the missing test: `test_settings.py:117-159` now covers the socket path, including the parametrisation you suggested over `localhost`/`0.0.0.0`/`127.0.0.1`, plus the forwarded-header proxy case, a configured cross-origin, and rejection of an unrelated origin.

Your caveat was the right one to flag. I didn't execute a handshake either, so I can't say which transport fails first on the pre-fix code — but that's moot for the fix, since the check is now unit-tested at the boundary engine.io consults.

backend/ttnn_visualizer/models.py
dcblundell: Agreed and closed in `1a11403c`. `sanitise_ssh_username` (`models.py:292-301`) now strips first, then raises `ValueError("must not be empty")` before delegating to `reject_ssh_option_like`, so it and `MlirServerConnection._strip_required_strings` agree on empty as well as on non-strings. Stripping before the check matters for the same reason it does for the option-like case: `"   "` would otherwise have reached argv as the bare target `@host`.

Covered by `test_connection_rejects_an_empty_username` and `test_mlir_server_connection_rejects_an_empty_username`, both parametrised over `""` and `"   "`.

One consequence worth recording, since "no existing behaviour depends on the lenient path" held for request validation but not for what was already on disk: a `username` of `"  "` was storable before this, and `InstanceTable.to_pydantic()` validates the persisted `remote_connection` JSON column on read. Tightening the validator alone would have turned any such stored row into a 500 on every request against that instance, with no UI path to clear it. `to_pydantic` now drops an unvalidatable connection with a warning instead of raising — see `test_to_pydantic_drops_a_connection_the_validators_now_reject`.

And thanks for pre-empting the `sanitise_path_segment` question. Agreed it should stay lenient for the `views.py:2069` report-name path; I've left the leniency in place and the composition through `sanitise_remote_host_segment` is what closes the SSH side.

backend/ttnn_visualizer/ssh_config.py
dcblundell: Both points taken, fixed in `1a11403c`.

The comment now states the divergence rather than implying parity (`ssh_config.py:139-150`): it separates the part that *does* match OpenSSH (the surrounding Host context continues after the `Include`, so the open stanza is left pending and flushing there would drop the keywords that follow) from the part that doesn't (the included file gets fresh state and the parent stanza resumes, instead of OpenSSH threading `activep` in by pointer). It also says why — a mid-stanza `Include` can't silently rewrite the alias being read, and the worst case is a lost `HostName`/`Port` prefill — and names `test_load_ssh_config_hosts_include_inside_host_keeps_later_keywords` with an explicit "don't fix this toward parity without replacing that test".

On the depth cap: raised to match OpenSSH rather than documenting 5, since "a config OpenSSH accepts never loses hosts from the picker" is a cheaper invariant to hold than to explain. `MAX_INCLUDE_DEPTH = 16` (`ssh_config.py:27-29`) with a comment noting it's `MAX_READCONF_DEPTH`. The visited-set check still runs before the depth cap, so the fan-out bound you were worried about doesn't loosen with the higher number — each file is parsed at most once regardless of depth.

docs/src/remote-sync.md
dcblundell: Documented in `1a11403c`, as an `ALLOWED_ORIGINS` subsection under Environment variables in `running-from-source.md` rather than here — it's a server-configuration lever rather than part of the remote-sync flow, and that page already lists `--host`/`--port`/`--server`, which are the flags an operator would be holding when they need it.

It covers the default (the app's own origin, plus the Vite dev server outside production), the comma-separated format with an example, when you actually need to set it, and why the list is deliberately narrow — no authentication, and local-only endpoints hand out SSH hosts, usernames and report paths, so each added origin is another page that can read them.

One thing the doc can now say that it couldn't when you wrote this: the app's own serving origin is accepted regardless of the setting, so `--host`, `--server` and containers reached by IP need no configuration. That's a consequence of the `app.py:188` fix, so the two findings closed together as you predicted — the difference being that there's no broken-socket symptom left to need a lever for. What remains genuinely configuration-only is a separately hosted front end, or a proxy rewriting scheme/host without setting `X-Forwarded-Proto`/`X-Forwarded-Host`, and both are called out.

src/components/report-selection/LocalFolderPicker.tsx
```diff
@@ -110,83 +118,70 @@ const LocalFolderPicker = ({
...
+                    isOpen
+                    entity={ManagedEntity.REPORT}
+                    entityName={folderToDelete.reportName}
+                    onCancel={() => setFolderToDelete(null)}
+                    onConfirm={() => {
+                        if (!loading) {
+                            handleDelete(folderToDelete);
+                        }
+                        setFolderToDelete(null);
+                    }}
```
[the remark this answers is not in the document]
dcblundell: Not doing this one — neither half of the claim holds, and the change would make the code read worse.

On type-checking: `pnpm lint:ts` passes with `strict: true` (`tsconfig.json:19`), so `strictNullChecks` is already on. TypeScript's aliased-condition narrowing resolves `canDeleteReports` back to its definition at `LocalFolderPicker.tsx:67`, so the `canDeleteReports && folderToDelete &&` guard at `:170` narrows `handleDelete` to defined, and the narrowing survives into the `onConfirm` arrow because `handleDelete` is a never-reassigned parameter binding.

On the crash: the alert cannot exist without `handleDelete`. `canDeleteReports = !!handleDelete && !isServerMode` gates the trash button and the `ConfirmDeleteAlert` together, which is what the comment above it is for — rendering one without the other is the state being ruled out. A parent that later passes `handleDelete` as `undefined` flips the flag false and unmounts the alert, and the already-created closure captured the defined function anyway.

So `handleDelete?.(...)` would silence nothing real, and it would actively weaken the invariant: an optional call implies the alert can legitimately open with no delete handler, which is exactly what the gating exists to prevent. If that guarantee ever needs to be enforced rather than inferred, the honest fix is two prop shapes (deletable vs not), not an optional call at the use site.

backend/ttnn_visualizer/app.py
smountenay-tt: This is the change I'd most want to re-check before merge, because it turns a permissive default into a **hard failure** rather than a browser-side one.

`socketio` was previously built with `cors_allowed_origins="*"`, so the handshake accepted anything. It now receives `ALLOWED_ORIGINS`, which in production defaults to exactly `["http://localhost:{PORT}"]`. Engine.IO rejects a non-matching origin outright:

```python
# engineio/server.py (v4, installed)
if self.cors_allowed_origins != []:
    # Validate the origin header if present
    # This is important for WebSocket more than for HTTP, since
    # browsers only apply CORS controls to HTTP.
    origin = environ.get('HTTP_ORIGIN')
    if origin:
        allowed_origins = self._cors_allowed_origins(environ)
        if allowed_origins is not None and origin not in allowed_origins:
            r = self._bad_request('Not an accepted origin.')
```

The library's own comment is the crux: unlike the HTTP API — where a missing `Access-Control-Allow-Origin` is harmless same-origin — browsers send `Origin` on **every** WebSocket handshake, same-origin included. So the handshake 400s whenever the browser's origin isn't the literal string `http://localhost:<PORT>`:

- `ttnn-visualizer --server` binds `0.0.0.0` and sets `SERVER_MODE`; users reach it as `http://<machine>:8000`.
- `--host 127.0.0.1` produces origin `http://127.0.0.1:8000`, which is a *different* origin string from `http://localhost:8000`. The app's own bind-failure message recommends this flag (`"Try ... a different host with --host (e.g. --host 127.0.0.1)"`).
- `HOST` already defaults to `0.0.0.0` when `is_running_in_container()`.
- Hosted `https://ttnn-visualizer.tenstorrent.com` now requires `ALLOWED_ORIGINS` to be set explicitly; it worked before purely because of `"*"`.

The failure mode is also unpleasant to diagnose: HTTP works fine, only sockets die, and `_log_error_once` means one log line.

**Suggestion:** for the socket server specifically, `cors_allowed_origins=None` already means precisely what the `settings.py` docstring says it wants. Engine.IO derives the allowlist per request from `wsgi.url_scheme` + `HTTP_HOST`, and additionally honours `X-Forwarded-Proto`/`X-Forwarded-Host` — which matters for the proxied hosted deployment, and note `ProxyFix` is only applied when `USE_WEBSOCKETS` is *false*, so Flask isn't normalising that for you on this path.

So something like: pass `None` when no explicit `ALLOWED_ORIGINS` is configured (true same-origin, correct for every deployment shape without guessing a hostname), and only add the Vite dev-server origin when the SPA is actually served cross-origin in development. That keeps the security win — dropping `"*"` — without hardcoding a hostname the app doesn't necessarily serve on.

backend/ttnn_visualizer/settings.py
```diff
@@ -18,6 +19,56 @@
...
+    Only the app's own origin is trusted by default: local-only endpoints hand out SSH
+    host, user and path metadata, and with no authentication CORS is what stops another
+    page served from a different localhost port reading it. Production serves the built
+    SPA same-origin, while ``pnpm dev`` serves it from Vite's own origin, so only
+    non-production adds the dev server.
+    """
+    if configured is None:
+        configured = f"http://localhost:{app_port}"
+        if flask_env.lower() != "production":
+            configured += f",http://{dev_server_host}:{dev_server_port}"
```
smountenay-tt, line 47: Narrower framing of the `app.py` comment, since this is where the value originates: the default is derived from `PORT` but **never from `HOST`**, and `localhost` is hardcoded.

That's the gap. `HOST` defaults to `"0.0.0.0" if is_running_in_container()`, `--server` sets it to `0.0.0.0`, and `main()` then launches the browser at the host it bound to:

```python
port = config.PORT if flask_env == "production" else config.DEV_SERVER_PORT
host = config.HOST if flask_env == "production" else config.DEV_SERVER_HOST
threading.Thread(target=open_browser, args=[host, port, instance_id]).start()
```

So on those paths the app opens the browser at an origin its own allowlist doesn't contain.

The docstring says "defaults to the narrowest set that still works" and "Only the app's own origin is trusted by default" — I agree with both goals, but `http://localhost:{PORT}` is a *guess* at the app's own origin that's wrong for any access path other than literally `localhost`. Resolving it per request (see the `app.py` comment) makes the docstring true rather than aspirational.

Also worth noting for coverage: `test_settings.py` pins the `PORT` and `FLASK_ENV` behaviours nicely, but nothing exercises `HOST`, so this regression passes green. A case asserting the allowlist for `HOST=0.0.0.0` would have caught it.

Separately — and this part I like — thanks for the `hasattr(value, "__get__")` guard in `override_with_env_variables`. Worth knowing it's largely belt-and-braces in practice: that loop walks `self.__class__.__dict__`, and for `ProductionConfig`/`DevelopmentConfig` that dict doesn't contain `ALLOWED_ORIGINS` (it lives on `DefaultConfig`), so the descriptor is only reachable there when `DefaultConfig` is instantiated directly. Still correct to have; just don't count on that loop as the reason the descriptor survives.

backend/ttnn_visualizer/ssh_config.py
```diff
@@ -0,0 +1,300 @@
...
+            # the pending values.
+            for included in _expand_include_paths(argument, path.parent):
```
smountenay-tt: `path.parent` is the right base directory only for the top-level config file. From `ssh_config(5)`:

> **Include** — ... Files without absolute paths are assumed to be in `~/.ssh` if included in a user configuration file or `/etc/ssh` if included from the system configuration file.

The base is the user's SSH directory, fixed — not the directory of whichever file contains the `Include`. Because this recurses with `path.parent` recomputed per file, the two agree at depth 1 and diverge below it:

- `~/.ssh/config` containing `Include config.d/*` → both resolve `~/.ssh/config.d/*`. Correct, and this is what the tests cover.
- `~/.ssh/config.d/work` containing `Include work-extra` → OpenSSH reads `~/.ssh/work-extra`; this reads `~/.ssh/config.d/work-extra`.

Practical effect is a host silently missing from the picker (or, less likely, a stanza picked up that OpenSSH ignores), which is hard to attribute to the parser.

Every relative-`Include` test (`..._include_relative_glob`, `..._include_expands_a_home_relative_glob`, `..._include_glob_skips_directories`) puts the `Include` in the entry file, so the divergent case isn't covered — worth adding alongside the fix.

The fix is small: resolve the base directory once from the entry file and thread it through the recursion as a parameter, rather than deriving it from the current file:

```python
def _parse_ssh_config_file(path, host_by_alias, depth, visited, base_dir):
    ...
    for included in _expand_include_paths(argument, base_dir):
        _parse_ssh_config_file(included, host_by_alias, depth + 1, visited, base_dir)
```

with `load_ssh_config_hosts` seeding `base_dir=path.parent`. That also keeps the `config_path` injection the tests rely on working unchanged, since `tmp_path` configs get `tmp_path` as their base.

(Unrelated and genuinely minor: `MAX_INCLUDE_DEPTH = 5` where OpenSSH allows 16. Fine as a picker limit — the warning log makes it debuggable — just noting it's a deliberate divergence in case you'd rather match.)

backend/ttnn_visualizer/models.py
```diff
@@ -262,9 +262,36 @@ def folder_segment_from_remote_path(remote_path: Optional[str]) -> Optional[str]
         return None
 
 
+def reject_ssh_option_like(value: object) -> str:
```
smountenay-tt, line 268: This validator is a good addition and the docstring reasoning about Pydantic lax-mode `bytes` coercion running *after* `mode="before"` is a genuinely subtle catch.

Flagging something adjacent rather than asking you to fix it here: it closes option injection on `username`/`host`, but the argv-safety story stays only partly closed, one module over. `ssh_client.read_file` / `check_path_exists` correctly use `shlex.quote`, while `sftp_operations.find_folders_by_files` hand-rolls quoting:

```python
ssh_cmd = _ssh_cmd_prefix(remote_connection) + [
    f"find '{root_folder}' -maxdepth 1 -type d -not -path '{root_folder}'",
]
...
file_checks = [f"test -f '{directory}/{file_name}'" for file_name in file_names]
```

`root_folder` is `RemoteConnection.profilerPath` / `performancePath`, and unlike `host` and `username` those two fields have no validator at all — plain `str` / `Optional[str]`. A single quote in either escapes the quoting and executes arbitrary commands on the remote host. Same pattern in `get_file_stat`, `list_directory` (`f"ls -1 '{path}'"`) and `download_file`'s SFTP batch.

It's the same threat model this PR is hardening, and `shlex.quote` is already imported in `ssh_client`, so the mitigation is mechanical. Reachability is bounded by `@local_only` — which the CORS half of this PR narrows further — so I don't think it blocks anything. But given the PR's framing, it'd be good to either sweep these in or open a follow-up so the remaining gap is tracked rather than implicitly considered handled.

smountenay-tt (commented):
Reviewed with a backend focus.

Overall this is strong work. The SSH config parser is deliberately scoped (concrete `Host` aliases + `Include`, no `Match`/pattern merging) and says so up front, the decision to keep `IdentityFile` out of the wire payload is the right call, and the test suite is better than most — the `Include` cycle/self-matching-glob termination cases and the `bytes`/`bytearray` coercion case in `test_ssh_client.py` are the kind of thing that usually gets found in production instead.

Three things worth acting on:

1. **Likely release blocker** — moving Socket.IO off `cors_allowed_origins="*"` onto a default allowlist of `["http://localhost:{PORT}"]` will hard-fail the handshake for `--server`, container, `127.0.0.1` and hosted deployments. Details on `app.py` and `settings.py`.
2. **Correctness in the new parser** — relative `Include` paths resolve against the wrong base directory for nested includes. Details on `ssh_config.py`.
3. **Non-blocking** — the option-injection hardening leaves an adjacent shell-quoting hole in the same threat model. Details on `models.py`; happy for it to be a follow-up issue.

Nothing to say about the docs changes beyond that they're unusually thorough — the note about the form's username overriding `User` in SSH config (including `Host *`) is exactly the thing users would otherwise file a bug about.

backend/ttnn_visualizer/ssh_config.py
```diff
@@ -0,0 +1,300 @@
...
+    """Payload for the SSH config host picker."""
+
+    configExists: bool
+    hosts: List[SshConfigHost]
+
+    def to_dict(self) -> Dict[str, object]:
+        return {
+            "configExists": self.configExists,
+            "hosts": [host.to_dict() for host in self.hosts],
+        }
```
smountenay-tt: These two are the HTTP response shape for `GET /api/remote/ssh-config-hosts`, and I think they should go through Pydantic rather than being dataclasses — and once they do, the field names should be snake_case.

**Why Pydantic:** the split in this codebase is by transport, and it's pretty consistent. HTTP request/response shapes are `SerializeableModel` and serialize with `model_dump()` — `RemoteConnection`, `MlirServerConnection`, `RemoteReportFolder`, `StatusMessage`. Dataclasses serve socket event payloads (`FileProgress`, `ReportGenerated` on `SerializeableDataclass`) and the internal report/DB models (`Operation`, `Buffer`, `Tensor`, …), and every one of those is snake_case. This is an HTTP payload, so it lands in the first bucket.

**Why snake_case:** I'll be upfront that there's precedent for camelCase attributes — `RemoteReportFolder.reportName`, `RemoteConnection.profilerPath`, `MlirServerConnection.sshPort`. But that precedent only exists because there's currently no aliasing anywhere in the backend, so `model_dump()` emits attribute names verbatim and the wire format has to be spelled out in Python identifiers. That's a serialization concern leaking into the Python API of a Black/isort/mypy-configured package. Pydantic is precisely the tool that removes the tradeoff — you get PEP 8 identifiers *and* an unchanged camelCase contract, which a dataclass can't give you.

I checked this runs against the installed Pydantic (2.12.5) and emits output byte-identical to the current `to_dict()`:

```python
class SshConfigHost(SerializeableModel):
    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True, frozen=True)

    host: str
    host_name: Optional[str] = None
    user: Optional[str] = None
    port: Optional[int] = None


class SshConfigHostsResult(SerializeableModel):
    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True, frozen=True)

    config_exists: bool
    hosts: List[SshConfigHost]
```

with `to_camel` from `pydantic.alias_generators`, and the view becoming:

```python
return jsonify(load_ssh_config_hosts().model_dump(by_alias=True, exclude_none=True))
```

Verified output:

```
{'configExists': True, 'hosts': [{'host': 'a', 'hostName': 'a.example', 'user': 'alice'}, {'host': 'b'}]}
{'configExists': False, 'hosts': []}
```

**`by_alias=True` is load-bearing, not cosmetic.** `src/model/SshConfigHost.ts` declares `hostName?` and `SshConfigHostsResponse.configExists`, and `useSshConfigHosts` reads `data.configExists`, so the JSON keys must not change. Dropping `by_alias` would silently ship snake_case keys and the picker would render hosts with no hostname.

What this buys beyond convention: both hand-rolled `to_dict()` methods disappear. That also resolves that they're currently inconsistent with each other — `SshConfigHost.to_dict` filters `None` via `asdict()`, while `SshConfigHostsResult.to_dict` spells its two keys out literally, so adding a third field means remembering to touch it. `exclude_none=True` covers the `None`-filtering that presumably motivated hand-rolling in the first place (it's the one thing `SerializeableDataclass.to_dict()` wouldn't have given you, so I get why you didn't just use that base).

Three caveats so this isn't a drive-by:

- `frozen=True` in `ConfigDict` preserves the immutability you have now — worth keeping deliberately rather than losing it in the move.
- `Field(ge=1, le=65535)` is **not** a drop-in for the port handling. `_store_host_block` deliberately *discards* an out-of-range port and keeps the host, whereas a `Field` constraint would raise and lose the whole stanza. Keep the existing `int()`/range logic in the parser and leave the model's `port` unconstrained.
- This would be the first alias usage in the backend, so it is a precedent-setting decision rather than a cleanup. I think it's the right precedent and new code is the right place to start, but flagging it as a call to make rather than something the codebase already does. If you'd rather not open that door in this PR, the fallback that still deletes the hand-rolled serialization is `SerializeableModel` with the existing camelCase field names — consistent with `RemoteReportFolder`, no aliasing, though it keeps the casing wart.

(Tangent, definitely not for this PR: `SerializeableModel` still uses the v1-style `class Config`, which raises `PydanticDeprecatedSince20` and is slated for removal in Pydantic v3. Inheriting from it and setting `model_config` in the subclass works fine today — I confirmed that — but the base will need migrating eventually.)

backend/ttnn_visualizer/app.py
dcblundell: Agreed, and already closed — you reviewed `22fee825`, two commits before the fix. `1a11403c` moved `socketio.init_app` off the raw allowlist onto `build_socketio_origin_check(ALLOWED_ORIGINS, bind_host=HOST)` (`app.py:194-197`, `settings.py:119-160`), so every case you list works without configuration.

Worth saying where the implementation lands relative to your suggestion, because it isn't quite `cors_allowed_origins=None`. A callable rather than `None` for two reasons:

- An explicit empty list is not inert to engine.io — it means "skip the origin check entirely". So `ALLOWED_ORIGINS=""` to trust nothing would have silently opened the socket to every origin. A callable is always consulted, which closes that. `test_socketio_still_accepts_same_origin_when_nothing_is_configured` pins it.
- Plain self-derivation trusts any `Host` header, and a `Host` header is attacker-controlled. DNS rebinding forges a same-origin claim for a *name*: a page on `attacker.example` repoints that name at 127.0.0.1, and both its origin and its `Host` become `attacker.example`, which passes a naive self-check and can then read instance-scoped socket traffic — SSH host, username, report paths — off a local install.

So `_names_this_machine` only self-trusts what can't be forged that way: IP literals (no name is resolved, so matching the origin means the request really is same-origin), `localhost`, and the address passed to `--host`. Your `--server`, `--host 127.0.0.1` and container-by-IP cases are all covered by that. The one case that still needs `ALLOWED_ORIGINS` is a deployment reached under a DNS name we weren't launched with — including the hosted one behind its proxy — and that's deliberate rather than an oversight.

That does mean the hosted deployment needs `ALLOWED_ORIGINS` set before this ships; it worked previously only because of `"*"`. Documented under Environment variables in `running-from-source.md`.

On the missing test: `test_settings.py:117-215` now covers the socket path, including the `localhost`/`0.0.0.0`/`127.0.0.1` parametrisation you asked for, the forwarded-header proxy case, an IPv6 loopback binding, and rejection of both an unrelated origin and a forged `X-Forwarded-Host`.

backend/ttnn_visualizer/settings.py
dcblundell: The observation is right and the docstring was overclaiming, but the defect it caused is gone as of `1a11403c` — you were reading `22fee825`. Details on the `app.py` thread; short version is that the socket path no longer takes this list verbatim, it goes through `build_socketio_origin_check`, which accepts the app's real serving origin per request and needs no `HOST` guess.

Why the default still says `localhost` rather than being derived from `HOST`: for HTTP it doesn't have to be right. A browser at `http://0.0.0.0:8000` fetching `http://0.0.0.0:8000/api/...` is same-origin, and a missing `Access-Control-Allow-Origin` on a same-origin response changes nothing. The allowlist only governs which *other* pages may read us, so an entry that names an access path nobody uses is inert rather than wrong. It was only the socket handshake that turned the wrong guess into a hard failure, because engine.io 400s on an unlisted `Origin` and browsers send `Origin` on every WebSocket handshake including same-origin ones.

The docstring now says that directly instead of claiming the default is the app's own origin.

On coverage: `HOST` is exercised now, in the place it actually matters. `test_settings.py:120` parametrises the socket check over `localhost`/`0.0.0.0`/`127.0.0.1`, with further cases for an IPv6 loopback binding, a `--host`-supplied name, and being reached by address.

And noted on `override_with_env_variables` — you're right that the loop walks `self.__class__.__dict__`, so the descriptor is only reachable there when `DefaultConfig` is instantiated directly. Keeping the guard as belt-and-braces rather than as load-bearing.

backend/ttnn_visualizer/ssh_config.py
dcblundell: Correct, and fixed in `941f85b3` the way you suggested: `base_dir` is a parameter threaded through the recursion, seeded from the entry file's parent in `load_ssh_config_hosts`, rather than recomputed as `path.parent` per file. `tmp_path` configs get `tmp_path` as their base, so the `config_path` injection the tests rely on is unchanged.

You were also right that no existing test discriminated the two behaviours. `test_load_ssh_config_hosts_nested_include_resolves_against_the_entry_directory` places a file named `sibling` in *both* candidate directories — one beside the entry config, one beside the nested include — and asserts which one is read, so it fails on the old code rather than passing either way. Verified it does before the fix.

On the depth cap: already raised to 16 in `1a11403c` after @aidemsined raised the same point, so this one had crossed with your review. `MAX_INCLUDE_DEPTH` now cites `MAX_READCONF_DEPTH` as its reason. The visited-set check still runs before the depth cap, so each file is parsed at most once and the higher number doesn't loosen the `Include` fan-out bound.

backend/ttnn_visualizer/models.py
dcblundell: Good catch, and swept here rather than deferred — `941f85b3`. It came out mechanical enough that leaving it open would have cost more in tracking than in doing.

Quoted: `find_folders_by_files`' `find` and its per-directory `test -f` (`sftp_operations.py:1046`, `:1068`), `get_remote_performance_folder`'s `stat` (`:917`), and `resolve_file_path`'s `ls` (`:197`).

Went one module wider than you flagged, because `ssh_client` had three siblings of the same pattern that `read_file`/`check_path_exists` being correct rather hid: `get_file_stat` (`:442`), `list_directory` (`:472`) and `download_file`'s SFTP batch (`:339`). Every remote command in the backend now goes through `shlex.quote`, so a future call site is the only way back in — which is what the test in the follow-up issue is for.

One site needed more than `shlex.quote`. `resolve_file_path` exists *because* the path contains a `*`, and quoting the pattern wholesale would turn the glob into a literal filename and break the caller. `_quote_keeping_globs` (`sftp_operations.py:66-75`) quotes each side of the wildcards and joins on a bare `*`, so quote removal returns the pattern verbatim while the wildcard still expands. `test_resolve_file_path_quotes_around_the_wildcard` asserts both halves — `shlex.split(cmd) == ["ls", "-1", pattern]` for the first, and no `'*'` in the command for the second.

Four tests added to the existing `TestRemoteFindShellQuoting`. The `test -f` one is worth a mention: the directory there comes back from the remote `find`, not from anything we validated on the way out, so it's quoted independently of `root_folder`.

What I did *not* do here is your underlying point about the fields themselves — `profilerPath` and `performancePath` still have no validator, so quoting is the only thing between them and the remote shell. Tracked in #1839, along with a test to pin that no remote command interpolates a path unquoted.

backend/ttnn_visualizer/ssh_config.py
dcblundell: Taken, with the fallback rather than the aliasing — `941f85b3`. Both are now `SerializeableModel` with `frozen=True` in `model_config`, keeping the camelCase field names, and both hand-rolled `to_dict()` methods are gone. The view is `jsonify(load_ssh_config_hosts().model_dump(exclude_none=True))`.

The transport argument is what decided it: an HTTP response shape belonging with `RemoteConnection` and `RemoteReportFolder` rather than with the socket dataclasses is straightforwardly right, and it deletes the inconsistency you spotted between the two `to_dict()` implementations — one filtering `None` via `asdict()`, the other spelling its keys out so a third field would need remembering.

Where I landed differently is the alias generator. You framed it as precedent-setting and I think that's the reason to leave it out of this PR rather than the reason to do it here. The value only materialises once it's the *rule* — one aliased model beside a dozen camelCase ones is a third convention, not a migration, and the next person writing a response model has two examples pointing opposite ways with nothing saying which wins. That's a decision that wants an AGENTS.md line and a sweep behind it, not a new endpoint. Happy to be the one to open it as its own change if you think it's worth doing.

Your `Field(ge=1, le=65535)` caveat was the useful one — `_store_host_block` deliberately discards an out-of-range port and keeps the alias, so a field constraint would lose the whole stanza. Left `port` unconstrained and said why in the model docstring, since the temptation to "tighten" it later is obvious.

`exclude_none` is load-bearing for the same reason you gave for `by_alias`: `src/model/SshConfigHost.ts` declares the optional fields absent, not null. Noted on the docstring, and the endpoint tests already asserted the exact JSON — `test_ssh_config_hosts_endpoint_omits_absent_fields` and `..._returns_hosts` — so the wire format is pinned rather than assumed.

On the tangent: yes, `SerializeableModel`'s v1-style `class Config` still emits `PydanticDeprecatedSince20`, and setting `model_config` in a subclass of it works fine (confirmed here too — the error only fires when both appear on the same class). Not touching it in this PR.

PR #1824, opened 2026-07-31
Memory tag style and formatting

dcblundell:
Fixes the copy for the L1 Small memory tag: `MemoryTag` no longer hyphenates the display label, so it reads `L1 Small` rather than `L1-Small`, while still producing the `tag-l1-small` class the styles already target.

Also corrects two call sites that passed a raw enum *key* instead of a display label, so they rendered `L1_SMALL` with a `tag-l1_small` class that matched no style rule:

- `OperationGraphComponent.tsx` now passes `BufferTypeLabel[tensor.buffer_type]`
- `DeviceOperationsFullRender.tsx` now passes `StringBufferTypeLabel[type]`

Screenshot 2026-08-04 at 11 07 38

### DIFF
--- a/src/components/MemoryTag.tsx
+++ b/src/components/MemoryTag.tsx
@@ -13,9 +13,12 @@ const MemoryTag = ({ memory }: MemoryTagProps) => {
     if (memory === undefined) {
         return null;
     }
-    const memoryLabel = stripEnum(memory).replace(' ', '-');
-    const memoryType = memoryLabel?.toLowerCase();
 
-    return <Tag className={`memory-tag tag-${memoryType}`}>{memoryLabel}</Tag>;
+    const memoryLabel = stripEnum(memory);
+    const memoryType = memoryLabel?.toLowerCase();
+    const memoryClass = `memory-tag tag-${memoryType.replace(/ /g, '-')}`;
+
+    return <Tag className={memoryClass}>{memoryLabel}</Tag>;
 };
+
 export default MemoryTag;
--- a/src/components/MemoryTag.tsx
+++ b/src/components/MemoryTag.tsx
@@ -3,9 +3,9 @@
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
 import { Tag } from '@blueprintjs/core';
-import { stripEnum } from '../functions/formatting';
 
 interface MemoryTagProps {
+    /** A display label from `BufferTypeLabel` / `StringBufferTypeLabel`, not a raw enum key. */
     memory: string | undefined;
 }
 
@@ -14,11 +14,9 @@ const MemoryTag = ({ memory }: MemoryTagProps) => {
         return null;
     }
 
-    const memoryLabel = stripEnum(memory);
-    const memoryType = memoryLabel?.toLowerCase();
-    const memoryClass = `memory-tag tag-${memoryType.replace(/ /g, '-')}`;
+    const memoryClass = `memory-tag tag-${memory.toLowerCase().replaceAll(' ', '-')}`;
 
-    return <Tag className={memoryClass}>{memoryLabel}</Tag>;
+    return <Tag className={memoryClass}>{memory}</Tag>;
 };
 
 export default MemoryTag;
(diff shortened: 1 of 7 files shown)


src/components/MemoryTag.tsx
```diff
@@ -13,9 +13,12 @@ const MemoryTag = ({ memory }: MemoryTagProps) => {
     if (memory === undefined) {
         return null;
     }
-    const memoryLabel = stripEnum(memory).replace(' ', '-');
+
+    const memoryLabel = stripEnum(memory);
     const memoryType = memoryLabel?.toLowerCase();
+    const memoryClass = `memory-tag tag-${memoryType.replace(/ /g, '-')}`;
```
aidemsined: > [!WARNING]
> Two of the six call sites feed this component an enum *key*, not a label, so they still render `L1_SMALL` in grey — both halves of the fix miss them.

The `/ /g` normalisation only handles spaces, and `stripEnum` (`src/functions/formatting.ts:29-37`) splits on `::` and `.` only — underscores survive both. Four call sites pass a label from `BufferTypeLabel` (`TensorDetailsComponent.tsx:127`, `PerfTensorRow.tsx:36`, `perfFunctions.tsx:106`, `ListItem.tsx:56`) and are fixed by this change. The other two pass the raw key:

- `src/components/OperationGraphComponent.tsx:942` — `BufferType[tensor.buffer_type]` → `"L1_SMALL"`
- `src/components/operation-details/DeviceOperationsFullRender.tsx:82,104` — `BufferTypeToStringBufferType[...]` → `StringBufferType.L1_SMALL` = `"L1_SMALL"`

Those render the text `L1_SMALL` and the class `tag-l1_small`, which matches no rule in `_common.scss` — so they fall through to the bare `.memory-tag { background-color: $tt-grey-6 }` and never pick up the `$tt-yellow-tint-1` treatment at `_common.scss:198`. The operation-graph tooltip and the device-operations buffer rows still show the unfixed tag.

The more principled fix is at the call sites rather than here — `src/definitions/BufferType.ts:32-38` already exists to map keys to display labels, so passing `BufferTypeLabel[tensor.buffer_type]` and `StringBufferTypeLabel[type]` respectively would make both surfaces correct without widening this component's contract. If you'd rather fix it centrally, `.replace(/[ _]/g, '-')` on the class plus the same normalisation on the label does it.

Worth noting the styling half of the PR description is already true on `dev`: the class string this produces is byte-identical before and after for a space-separated input (`"L1-Small".toLowerCase()` and `"L1 Small".toLowerCase().replace(/ /g,'-')` both give `tag-l1-small`), and `.tag-l1-small` landed in `354221c4`. So this diff fixes the copy; the styling was already working.

src/components/MemoryTag.tsx
```diff
@@ -13,9 +13,12 @@ const MemoryTag = ({ memory }: MemoryTagProps) => {
...
 
-    return <Tag className={`memory-tag tag-${memoryType}`}>{memoryLabel}</Tag>;
+    return <Tag className={memoryClass}>{memoryLabel}</Tag>;
```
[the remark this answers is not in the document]
dcblundell: Good catch — confirmed. Fixed at the call sites rather than in `MemoryTag`, since `BufferTypeLabel`/`StringBufferTypeLabel` already exist to map keys to display labels: `OperationGraphComponent.tsx` now passes `BufferTypeLabel[tensor.buffer_type]` and `DeviceOperationsFullRender.tsx` passes `StringBufferTypeLabel[type]`. That corrects both the text and the class, and keeps the component's contract as "give me a display label".

src/components/MemoryTag.tsx
dcblundell: Verified and fixed — thanks, the call-site analysis was accurate. Went with the call-site fix you recommended rather than widening `MemoryTag`: `OperationGraphComponent.tsx:942` now passes `BufferTypeLabel[tensor.buffer_type]`, and `DeviceOperationsFullRender.tsx:104` passes `StringBufferTypeLabel[type]`. Both surfaces now render `L1 Small` with `tag-l1-small`.

Agreed on the last point too: `.tag-l1-small` was already on `dev` (`0e552cea` rather than `354221c4`), so the original diff was a copy fix only. PR description updated accordingly.

PR #1840, opened 2026-08-05
Render Cluster without a baked SoC descriptor

### LINKED ISSUE #1772: Make the baked SoC descriptor optional for Cluster (decouple from device arch) BOS request, quasar support
Follow-up from the device-arch dependency assessment. Cluster's chip **placement** and full **link graph** already come entirely from the *cluster descriptor* (`ethernet_connections[_to_remote_devices]`, mesh coords, `chip_unique_ids`, `chips_with_mmio`). The baked SoC/arch descriptor (`ChipDesign` via `useArchitecture`) is used only for **enrichment**:

- translating an eth **channel index → NoC core-id** for the port uid key, the coordinate label, and which ports to draw (`ClusterRenderer.tsx`: `a.eth[link.a.chan]`, `chipDesign.eth.forEach`);
- PCIe marker dots on MMIO chips (`chipDesign.pcie` + `chipDesign.grid`).

Cluster never reads `functional_workers` / `dram` / `router_only`, and ETH **port placement is adjacency-based, not hardware-based**. So a baked per-arch descriptor is **not** required to render a cluster — which means we should *not* need to bake `arch-quasar.json` (or any licensee arch) just to render its cluster.

> Rescoped from "Add Quasar (emulation) SoC descriptor": baking an emulation descriptor is the wrong lever — decoupling is. The NPE topology path (which genuinely needs the grid) is handled separately in #1776.

## Scope

- Render Cluster from the cluster descriptor alone when the arch descriptor is unknown/absent:
  - Key eth ports by **channel** (`ethUid(rank, chipId, chan)`) so links resolve from `ethernet_connections` without an arch core-id lookup.
  - Treat PCIe markers + physical-coordinate labels as optional enrichment that no-ops when arch is absent.
  - Loosen the `!chipDesign?.eth → 'unsupported'` guard (`ClusterRenderer.tsx:159`) so an unknown arch no longer blanks the whole view; reserve `'unsupported'` for a genuinely missing cluster descriptor.
- Keep WH/BH enrichment (pcie dots, coordinate labels) working — no regression.
- Sweep "and such": confirm no other surface hard-fails on unknown arch. Performance already degrades to a Wormhole core-count default and prefers the device-log `max_cores`, so it's fine — flag anything else found.

## Acceptance criteria

- A cluster report with an unknown/unsupported arch string (e.g. `quasar`, `grendel`, a licensee name) renders chip placement + eth links from the cluster descriptor — no "unsupported" panel solely due to arch.
- PCIe markers and coordinate labels render when the arch descriptor is known; are cleanly omitted when it isn't.
- Wormhole/Blackhole cluster reports are visually unchanged.

## Notes

- NPE is the surface that genuinely needs a full grid (to land NoC transfers on nodes); that path is the report-supplied override in #1776.

aidemsined:
Closes #1772

## Summary

Cluster's chip placement and link graph already come entirely from the **cluster descriptor**. The baked SoC/arch descriptor was only ever used for *enrichment* — translating an eth channel index to a NoC core-id for coordinate labels, and placing PCIe marker dots — yet it was load-bearing: an unrecognised arch blanked the entire view behind the "not supported" panel. This decouples them, so a cluster renders from its own descriptor alone and we don't need to bake an arch JSON per licensee just to draw its topology.

- Eth ports are keyed by **channel** (`ethUid(rank, chipId, chan)`), so links resolve from `ethernet_connections` without an arch core-id lookup.
- Coordinate labels and PCIe markers became optional enrichment, resolved **per chip** from that chip's own `arch` entry — a heterogeneous cluster no longer labels every chip from chip 0 of host 0, and an unresolved chip doesn't suppress its neighbour's label.
- The `!chipDesign?.eth → unsupported` guard is gone. An unknown arch now degrades to a fully drawn topology without labels or PCIe dots.

A cluster report with an arch of `quasar`, `grendel` or any licensee name renders its placement and links normally.

## Fixed along the way

**Every cluster report was rendering Wormhole eth coordinates and PCIe positions regardless of its actual architecture.** `arch` in `cluster_descriptor.yaml` is a map keyed by chip id, but the model typed it `string[]` and the component gated on `firstHostArch.length` — `undefined` on an object — so `|| DEFAULT_ARCHITECTURE` won unconditionally and the arch string was never read at all. Wormhole's eth list has 16 entries against Blackhole's 14, so every Blackhole channel indexed successfully into the wrong list: coordinate labels were wrong, but no links were dropped, which is why it went unnoticed.

This **deliberately deviates from the ticket's "Wormhole/Blackhole cluster reports are visually unchanged" criterion**. Blackhole labels do change, because they were previously wrong. Wormhole is unchanged, and placement and links are unchanged for both.

Arch matching is also now prefix-based rather than substring. Descriptors carry a revision suffix (`wormhole_b0`) that has to resolve, but a name merely *containing* an arch must not — now that the resolved arch drives per-chip enrichment, a licensee name containing "wormhole" would mislabel that chip instead of cleanly omitting its labels.

## Test plan

Covered by tests (`tests/ClusterUnknownArch.spec.tsx`, `tests/stringToArchitecture.spec.ts`):

- [x] Recognised Wormhole arch labels ports, draws a PCIe marker, draws the link segment
- [x] Blackhole arch labels from Blackhole's channel map, not Wormhole's — the two are disjoint at the linked channels, so this pins the per-chip resolution
- [x] Unrecognised arch still draws ports and link segments, omits labels and PCIe markers, and does not fall back to the unsupported panel
- [x] Heterogeneous arch enriches each chip independently, in either ordering
- [x] Arch resolver handles the revision suffix and case, and rejects a name that merely contains an arch

Worth checking when reviewing:

- [ ] A Wormhole cluster report looks identical to `dev`
- [ ] A Blackhole cluster report now shows Blackhole eth coordinates (this is the fix above — labels change)
- [ ] A multi-host report still stitches and draws inter-host links

## Notes for the reviewer

The port uid is now built in one place. Previously the segment pass rebuilt it independently, and a drift between the two constructions rendered **zero links while the whole suite passed** — ports, labels and PCIe markers never touch `portPixelByUid`. The uids are now carried from `recordConnection` to the segment pass, and both remaining constructions are immediately validated by a map lookup the tests exercise.

## Follow-ups (not in scope here)

- The `unsupported` panel is now unreachable on the real fetch path — `fetchClusterTopology` always yields at least one host. A genuinely missing cluster descriptor leaves the view on "Loading…" indefinitely, because there's no error branch. Pre-existing; this change makes it the only remaining path.
- Inter-host link rendering has no test coverage; the fixture here is single-host.
- `getChipDesign` signals "no descriptor" via a frozen sentinel compared by identity. Returning `ChipDesign | null` would put that in the type system.
- A channel beyond a known arch's eth list renders one blank label among labelled siblings. Currently unspecified behaviour.

### DIFF
--- a/src/components/cluster/ClusterRenderer.tsx
+++ b/src/components/cluster/ClusterRenderer.tsx
@@ -127,6 +127,18 @@ interface EthPort {
     chan: EthChannel;
 }
 
+// Endpoint uids are built once, when connections are recorded, and carried to the segment
+// pass. Rebuilding them there instead would let the two constructions drift, which renders
+// no links at all rather than failing loudly. #1772
+interface ResolvedLink {
+    a: RenderChip;
+    b: RenderChip;
+    uidA: string;
+    uidB: string;
+    interHost: boolean;
+    index: number;
+}
+
 // Cheap deterministic non-negative hash for distributing ports across a small
 // candidate edge list. Quality only needs to be uniform over 2-3 buckets. #1510
 const hashLinkKey = (key: string): number => {
@@ -270,21 +282,27 @@ function buildClusterRenderModel(
         canonicalLinkKeyByUid.set(uidB, key);
     };
 
-    for (const link of intraHostLinks) {
+    const resolvedLinks: ResolvedLink[] = [];
+
+    intraHostLinks.forEach((link, index) => {
         const a = chipsByKey.get(chipKey(link.rank, link.a.chip));
         const b = chipsByKey.get(chipKey(link.rank, link.b.chip));
         if (a && b) {
-            rememberLinkKey(...recordConnection(a, link.a.chan, b, link.b.chan));
+            const [uidA, uidB] = recordConnection(a, link.a.chan, b, link.b.chan);
+            rememberLinkKey(uidA, uidB);
+            resolvedLinks.push({ a, b, uidA, uidB, interHost: false, index });
         }
-    }
+    });
 
-    for (const link of interHostLinks) {
+    interHostLinks.forEach((link, index) => {
         const a = chipsByKey.get(chipKey(link.a.rank, link.a.chip));
         const b = chipsByKey.get(chipKey(link.b.rank, link.b.chip));
         if (a && b) {
-            rememberLinkKey(...recordConnection(a, link.a.chan, b, link.b.chan));
+            const [uidA, uidB] = recordConnection(a, link.a.chan, b, link.b.chan);
+            rememberLinkKey(uidA, uidB);
+            resolvedLinks.push({ a, b, uidA, uidB, interHost: true, index });
         }
-    }
+    });
 
     // Ascending to match the previous arch-list order, which was channel order; dedup
     // guards a channel appearing on two links.
@@ -529,19 +547,20 @@ function buildClusterRenderModel(
     const pushSegment = (
         a: RenderChip,
         b: RenderChip,
-        uidA: string | undefined,
-        uidB: string | undefined,
+        uidA: string,
+        uidB: string,
         interHost: boolean,
         prefix: string,
         index: number,
     ) => {
-        const portA = uidA ? portPixelByUid.get(uidA) : undefined;
-        const portB = uidB ? portPixelByUid.get(uidB) : undefined;
+        const portA = portPixelByUid.get(uidA);
+        const portB = portPixelByUid.get(uidB);
+        // A uid can exist without a placed port when its chord direction is degenerate.
         if (!portA || !portB) {
             return;
         }
         // Inter-host always curves; intra-host curves only when long-haul.
-        const longHaul = !interHost && (longHaulUids.has(uidA ?? '') || longHaulUids.has(uidB ?? ''));
+        const longHaul = !interHost && (longHaulUids.has(uidA) || longHaulUids.has(uidB));
         const useCurve = interHost || longHaul;
         linkSegments.push({
             key: `${prefix}__${uidA}__${uidB}__${index}`,
@@ -556,36 +575,8 @@ function buildClusterRenderModel(
         });
     };
 
-    intraHostLinks.forEach((link, index) => {
-        const a = chipsByKey.get(chipKey(link.rank, link.a.chip));
-        const b = chipsByKey.get(chipKey(link.rank, link.b.chip));
-        if (a && b) {
-            pushSegment(
-                a,
-                b,
-                ethUid(a.rank, a.id, link.a.chan),
-                ethUid(b.rank, b.id, link.b.chan),
-                false,
-                'intra',
-                index,
-            );
-        }
-    });
-
-    interHostLinks.forEach((link, index) => {
-        const a = chipsByKey.get(chipKey(link.a.rank, link.a.chip));
-        const b = chipsByKey.get(chipKey(link.b.rank, link.b.chip));
-        if (a && b) {
-            pushSegment(
-                a,
-                b,
-                ethUid(a.rank, a.id, link.a.chan),
-                ethUid(b.rank, b.id, link.b.chan),
-                true,
-                'inter',
-                index,
-            );
-        }
+    resolvedLinks.forEach(({ a, b, uidA, uidB, interHost, index }) => {
+        pushSegment(a, b, uidA, uidB, interHost, interHost ? 'inter' : 'intra', index);
     });
 
     const neighboursByChip = new Map<string, Set<string>>();
--- a/src/components/cluster/ClusterRenderer.tsx
+++ b/src/components/cluster/ClusterRenderer.tsx
@@ -23,6 +23,7 @@ import {
     ClusterCoordinates,
     ClusterTopology,
     DEFAULT_ARCHITECTURE,
+    EthChannel,
 } from '../../model/ClusterModel';
 import {
     CHIP_GAP,
@@ -95,7 +96,7 @@ interface ClusterRenderModel {
     totalRows: number;
     contentWidth: number;
     contentHeight: number;
-    ethPositionsByChip: Map<string, Map<CLUSTER_ETH_POSITION, string[]>>;
+    ethPositionsByChip: Map<string, Map<CLUSTER_ETH_POSITION, EthPort[]>>;
     linkSegments: LinkSegment[];
     idFontSize: number;
     isMultiHost: boolean;
@@ -112,7 +113,20 @@ interface ClusterRenderModel {
 type ClusterRenderResult = { status: 'ready'; model: ClusterRenderModel } | { status: 'unsupported' };
 
 const chipKey = (rank: number, id: number) => `${rank}-${id}`;
-const ethUid = (rank: number, chipId: number, coreId: string) => `${rank}-${chipId}-${coreId}`;
+// Keyed by channel, not arch core-id, so links resolve from the cluster descriptor alone. #1772
+const ethUid = (rank: number, chipId: number, chan: EthChannel) => `${rank}-${chipId}-ch${chan}`;
+
+// `chipDesign.eth` is channel-indexed; undefined when no descriptor is baked for the arch.
+const ethCoordLabel = (chip: ClusterChip, chan: EthChannel): string | undefined => {
+    const coreId = chip.design?.eth?.[chan];
+    return coreId === undefined ? undefined : `${chip.rank ?? 0}-${chip.id}-${coreId}`;
+};
+
+// Channel travels with the uid so the render needn't parse it back out for the label.
+interface EthPort {
+    uid: string;
+    chan: EthChannel;
+}
 
 // Cheap deterministic non-negative hash for distributing ports across a small
 // candidate edge list. Quality only needs to be uniform over 2-3 buckets. #1510
@@ -156,14 +170,14 @@ function buildClusterRenderModel(
     chipDesign: ChipDesign,
     requestedLayoutMode: ClusterLayoutMode = 'mesh',
 ): ClusterRenderResult {
-    if (!chipDesign?.eth) {
-        return { status: 'unsupported' };
-    }
-
+    // A missing cluster descriptor is unrenderable; a missing *arch* one only costs the
+    // optional enrichment, since placement and links come from the cluster descriptor. #1772
     const { hosts, isMultiHost, worldSize, intraHostLinks, interHostLinks } = topology;
     if (hosts.length === 0) {
         return { status: 'unsupported' };
     }
+
+    const hasArchDescriptor = Boolean(chipDesign?.eth?.length);
 
     // Step 1: place every chip on a global virtual grid. `mesh` honours the
     // report's mesh coords (1D or 2D); `condensed` tiles each host into a
@@ -216,9 +230,9 @@ function buildClusterRenderModel(
                 chipUniqueId: host.descriptor.chip_unique_ids?.[chipId],
                 coords: [x, y, 0, 0] satisfies ClusterCoordinates,
                 mmio: mmioChipIds.has(chipId),
-                eth: chipDesign.eth.map((coreId) => ethUid(host.rank, chipId, coreId)),
+                ethChannels: [],
                 connectedChipsByEthId: new Map(),
-                design: chipDesign,
+                design: hasArchDescriptor ? chipDesign : undefined,
             });
         }
 
@@ -233,26 +247,21 @@ function buildClusterRenderModel(
     // Step 2: wire intra-host and inter-host eth connections into each chip's
     // `connectedChipsByEthId` map so the port-positioning step below can compute
     // edges relative to neighbours.
-    const recordConnection = (
-        chipA: RenderChip,
-        chipB: RenderChip,
-        uidA: string | undefined,
-        uidB: string | undefined,
-    ) => {
-        if (uidA === undefined || uidB === undefined) {
-            return;
-        }
+    // Also collects live channels, which replace the arch eth list as the ports to draw.
+    const recordConnection = (chipA: RenderChip, chanA: EthChannel, chipB: RenderChip, chanB: EthChannel) => {
+        const uidA = ethUid(chipA.rank, chipA.id, chanA);
+        const uidB = ethUid(chipB.rank, chipB.id, chanB);
         chipA.connectedChipsByEthId.set(uidA, chipB);
         chipB.connectedChipsByEthId.set(uidB, chipA);
+        chipA.ethChannels.push(chanA);
+        chipB.ethChannels.push(chanB);
+        return [uidA, uidB] as const;
     };
 
     // Canonical link key (lex-min of endpoint uids) shared by both ends, so
     // ordering ports by it on each edge keeps facing edges in sync.
     const canonicalLinkKeyByUid = new Map<string, string>();
-    const rememberLinkKey = (uidA: string | undefined, uidB: string | undefined) => {
-        if (uidA === undefined || uidB === undefined) {
-            return;
-        }
+    const rememberLinkKey = (uidA: string, uidB: string) => {
         const key = uidA < uidB ? uidA : uidB;
         canonicalLinkKeyByUid.set(uidA, key);
         canonicalLinkKeyByUid.set(uidB, key);
@@ -262,8 +271,7 @@ function buildClusterRenderModel(
         const a = chipsByKey.get(chipKey(link.rank, link.a.chip));
         const b = chipsByKey.get(chipKey(link.rank, link.b.chip));
         if (a && b) {
-            recordConnection(a, b, a.eth[link.a.chan], b.eth[link.b.chan]);
-            rememberLinkKey(a.eth[link.a.chan], b.eth[link.b.chan]);
+            rememberLinkKey(...recordConnection(a, link.a.chan, b, link.b.chan));
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/functions/stringToArchitecture.ts
```diff
@@ -4,11 +4,15 @@
 
 import { DeviceArchitecture } from '../definitions/DeviceArchitecture';
 
+// Prefixed, not substring: descriptors carry a revision suffix (`wormhole_b0`) that has to
+// resolve, but a name merely *containing* an arch must not. Cluster now enriches per chip
+// from this result, so a licensee arch silently resolving to Wormhole would mislabel it. #1772
 export const stringToArchitecture = (arch: string): DeviceArchitecture => {
-    if (arch.toLowerCase().includes('wormhole')) {
+    const name = arch.toLowerCase();
```
dcblundell, line 17: **Must fix:** a non-string `arch` in an uploaded descriptor throws during render

The type says `string`, but this value now comes straight from `yaml.safe_load` of an uploaded cluster descriptor with no schema validation, and YAML readily yields an int or a nested map (`arch: {0: 1}`). `?? ''` at `ClusterRenderer.tsx:219` only defends against `null`/`undefined`, so `.toLowerCase()` throws — and the nearest boundary is the router's root `errorElement`, which replaces the whole app shell with no in-place recovery (the same failure mode backstopped at `NPEViewComponent.tsx:191-195` for #1803).

This is newly reachable: on `dev`, `(firstHostArch.length && firstHostArch[0]) || DEFAULT_ARCHITECTURE` meant a map-shaped `arch` never reached this function. Guarding here also matches the PR's own "unresolved arch just loses enrichment" posture:

```ts
if (typeof arch !== 'string') {
    return DeviceArchitecture.UNKNOWN;
}
```

src/hooks/useAPI.tsx
```diff
@@ -1261,18 +1261,31 @@ export const useInstance = () => {
...
+        default:
+            return NO_CHIP_DESIGN;
+    }
+};
+
+export const useArchitecture = (arch: DeviceArchitecture): ChipDesign => {
+    const design = getChipDesign(arch);
+    if (design === NO_CHIP_DESIGN) {
+        // eslint-disable-next-line no-console
+        console.error(`Unsupported arch: ${arch}`);
```
dcblundell: **Should fix:** logs once per render, not once per arch

This sits in the hook body, so it fires on every render rather than when the arch changes. `useNodeType` → `NPEViewComponent:305` is a consumer and playback drives a re-render per `setInterval` tick, so an unresolved arch emits a `console.error` per frame for the duration of playback — and this PR is precisely what makes "arch doesn't resolve" a routine state rather than a bug. Move it into a `useEffect` keyed on `arch`, or gate it behind a module-level set of already-warned archs.

The `eslint-disable-next-line no-console` moved in this diff, so it also needs the explanatory comment AGENTS.md requires for suppressions.

src/hooks/useAPI.tsx
```diff
@@ -1261,18 +1261,31 @@ export const useInstance = () => {
     });
 };
 
-export const useArchitecture = (arch: DeviceArchitecture): ChipDesign => {
+// Stable identity across renders: a fresh `{}` made consumers memoizing on the
+// descriptor recompute every render, which now matters on the unknown-arch path. #1772
+const NO_CHIP_DESIGN = Object.freeze({}) as ChipDesign;
```
dcblundell: **Should fix:** the sentinel is a typed lie, and two predicates now answer "did the arch resolve?"

`Object.freeze({}) as ChipDesign` claims `arch_name`, `grid`, `eth` and `pcie` as required fields it doesn't have, which forces every reader into defensive optional chaining against a non-optional type (`chip.design?.eth?.[chan]`, `design.eth?.length`). Separately, `useArchitecture` tests identity against this sentinel while `designForChip` (`ClusterRenderer.tsx:220`) tests `design.eth?.length` — they agree today only because every baked descriptor happens to have a non-empty `eth`.

Returning `ChipDesign | null` from `getChipDesign` collapses both: one predicate, honest types, and it matches the repo's prefer-`null`-for-absence convention. `null` is as stable an identity as a frozen object, so the memo-stability rationale in the comment above still holds.

src/hooks/useAPI.tsx
```diff
@@ -1261,18 +1261,31 @@ export const useInstance = () => {
...
+
+// Pure so callers that need a descriptor per chip (Cluster, for heterogeneous reports)
+// can resolve one outside a hook. Returns `NO_CHIP_DESIGN` rather than logging, leaving
+// the caller to decide whether an unresolved arch is degraded-but-fine or an error.
+export const getChipDesign = (arch: DeviceArchitecture): ChipDesign => {
```
dcblundell: **Should fix:** pure accessor living in `hooks/`

`getChipDesign` has no query, no network and no hook usage — it's a switch over two bundled JSON assets, now imported straight into a component. AGENTS.md puts pure helpers in `src/functions/`; the non-hook exports in this file are `fetch*` axios wrappers, which this isn't. Its natural companion `stringToArchitecture` is already in `src/functions/`, so as it stands the cluster resolves arch from two different layers. Suggest moving it there and re-pointing `useArchitecture` at it.

src/components/cluster/ClusterRenderer.tsx
```diff
@@ -374,15 +408,15 @@ function buildClusterRenderModel(
...
         const directNeighbourEdges = new Set<CLUSTER_ETH_POSITION>();
         // Long-haul/inter-host placements deferred until pass 2.
-        const deferred: { uid: string; chord: CLUSTER_ETH_POSITION }[] = [];
+        const deferred: (EthPort & { chord: CLUSTER_ETH_POSITION })[] = [];
 
         // Pass 1: direct cardinal neighbours go on their chord edge.
-        clusterChip.design?.eth.forEach((coreId) => {
-            const uid = ethUid(clusterChip.rank, clusterChip.id, coreId);
+        clusterChip.ethChannels.forEach((chan) => {
+            const uid = ethUid(clusterChip.rank, clusterChip.id, chan);
```
dcblundell: **Should fix:** second `ethUid` construction site

The uid is rebuilt from `(rank, id, chan)` here, having already been built in `recordConnection` at :267-268 — exactly the drift hazard the `ResolvedLink` comment at :130-132 describes, just in the port pass rather than the segment pass that the last commit fixed.

The `EthPort { uid, chan }` carrier already exists at :125-128 but the chip only stores `ethChannels`. Storing `EthPort[]` instead lets `recordConnection` push `{ uid, chan }`, the dedupe at :309-311 key on `uid`, and this pass read `port.uid` directly — leaving `ethUid` with exactly one call site. Behaviour is identical, since the uid is a pure function of the channel.

src/components/cluster/ClusterRenderer.tsx
```diff
@@ -994,15 +1012,15 @@ function ClusterRenderer() {
...
+                                    {[...ethPosition.entries()].map(([position, ports]) => {
+                                        return ports.map(({ uid, chan }, index) => {
                                             const { x, y } = getEthGridPosition(position, index);
                                             const size = clusterChipSize / CLUSTER_NODE_GRID_SIZE - CHIP_GAP;
                                             // scale the label with the port so it doesn't overflow on dense meshes
                                             const ethFontSize = Math.max(4, Math.round(size / 4));
                                             return (
                                                 <div
-                                                    title={`${uid}`}
+                                                    title={ethCoordLabel(clusterChip, chan) ?? uid}
```
dcblundell: **Should fix:** `ethCoordLabel` called twice per port in the render body, and the two call sites disagree

This line falls back to `uid`, while :1034 renders blank — so an unenriched port shows a tooltip but no text. Plausibly intentional, but nothing pins it, and any change to one has to be mirrored in the other.

`ClusterRenderer` re-renders on every `setHoveredChip` (:984-985), so on a 32-chip cluster this is a few hundred optional-chain lookups plus template-string allocations per hover. The label is a pure function of `(chip.design, chan)`, both fixed at model-build time — computing it once onto `EthPort` inside the memoised `buildClusterRenderModel` reduces the render to a property read and removes the divergence as a side effect.

tests/ClusterUnknownArch.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+// Cluster placement and links come from the cluster descriptor; the arch descriptor is
+// only enrichment. An unrecognised arch used to blank the whole view because port uids
+// were keyed off the arch eth list. #1772
+//
+// The real `getChipDesign` is used deliberately — these assert against the baked
+// wormhole json, so a change to its channel ordering surfaces here.
+
+// Two chips side by side, linked on channels 4 and 6. Channel 6 sits past the end of a
+// short eth list, so a regression to arch-indexed lookup drops that port rather than
```
dcblundell: **Should fix:** this comment describes a guard the test doesn't provide

Channel 6 isn't past the end of a short eth list — `arch-wormhole.json` has 16 eth entries and `arch-blackhole.json` has 14, and the test itself asserts channel 6 resolves to `6-0` / `7-1` (:86, :117). So the out-of-range path is unexercised, and that's the one that genuinely regressed on `dev`: `a.eth[link.a.chan]` was `undefined`, `recordConnection` returned early, and the link vanished with no diagnostic.

Worth adding a real case — a link on a channel ≥ 16 under a wormhole arch, asserting both ports render, the label is empty and the title falls back to `0-0-ch20`. Fix the comment either way, since it currently claims protection that isn't there.

tests/ClusterUnknownArch.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+        expect(portLabels().sort()).toEqual(['', '0-0-7-0']);
+    });
+
+    it('does not let an unresolved first chip suppress the rest', () => {
+        clusterArch = { 0: 'quasar', 1: 'wormhole_b0' };
+        render(<ClusterRenderer />);
+
+        expect(portLabels().sort()).toEqual(['', '0-1-6-0']);
+    });
+});
```
dcblundell, line 230: **Should fix:** three branches this PR introduced or narrowed are uncovered

The existing cases are good — these are the gaps around them:

- **Inter-host / multi-rank.** Every fixture is single-host, rank 0, so `interHost: true` is never exercised. The rank component of the uid is exactly what stops chip ids colliding across ranks, and inter-host links take the curve branch (`pathD`) rather than `<line>`. A two-rank topology with chips sharing local id `0` would pin both.
- **The `unsupported` panel.** This PR removed one of its two triggers, leaving only `hosts.length === 0`. The spec asserts the panel is *absent* for an unknown arch, but nothing asserts it still appears when it should, so a further narrowing would go unnoticed. A `hosts: []` case covers it.
- **The `ethChannels` dedupe** at `ClusterRenderer.tsx:309-311` is load-bearing (ports are keyed `key={uid}`, so a channel on two links would emit duplicate React keys) and untested. A descriptor listing the same `(chip, chan)` on two links, asserting one port and two segments, covers it.

src/model/ClusterModel.ts
```diff
@@ -57,7 +60,9 @@ export type RemoteEthernetConnectionRaw = [
 ];
 
 export interface ClusterModel {
-    arch: string[];
+    // Keyed by chip id, as the YAML writes it. Optional because a descriptor may omit it
+    // entirely, in which case chips render without arch enrichment.
+    arch?: Record<ChipId, string>;
```
dcblundell, line 76: **Should fix:** existing fixtures still encode the old array shape (comment anchored here; the fixtures are in `tests/clusterTopology.spec.ts:19,343,415,465`)

Now that `arch` is `Record<ChipId, string>`, those fixtures still write `arch: ['blackhole']` / `arch: []`. They only still compile because `string[]` is structurally assignable to `Record<number, string>`, so nothing forces them into line with the shape this PR redefines — and `makeDescriptor` is the fixture other cluster tests are most likely to copy. Worth updating to the chip-id-keyed form as on-touch cleanup, since this branch owns the change.

dcblundell (commented):
Combined review (maintainability, tests, performance, conventions, security). Full suite, `tsc`, ESLint and SPDX all pass, and the core change is a good one — per-chip enrichment, channel-keyed port uids, graceful degradation on an unrecognised arch.

One must-fix: a non-string `arch` value in an uploaded descriptor now reaches `String.prototype.toLowerCase` and throws during render, which is newly reachable because the old code never passed descriptor values into that function. The rest are should-fixes: a sentinel that defeats strict typing, a `console.error` in a render body that NPE playback hits per tick, one duplicated uid construction the last commit didn't quite finish, and a fixture comment that describes a guard the test doesn't provide.

Consider/nit-level findings (double-cast fixture, `TEST_IDS` vs class selectors, duplicated panel copy, orphaned tombstone comment in `ClusterModel.ts`, enum literals in `stringToArchitecture`, a11y names on eth ports) are left out of the inline comments — happy to add them if useful.

src/functions/stringToArchitecture.ts
aidemsined: Fixed in `5419fca7`. `stringToArchitecture` now takes `unknown` and returns `UNKNOWN` for anything that isn't a string, so the call site drops its `?? ''` coercion and the function owns the question.

Widening the parameter rather than adding a `typeof` guard under a `string` annotation was deliberate — the guard only means something if the type admits the values that actually arrive.

Confirmed the blast radius: `main.tsx` has a single root `errorElement`, so the throw replaces the whole app shell rather than the cluster view. `tests/stringToArchitecture.spec.ts` now covers `undefined`, `null`, an int, a map and an array.

src/hooks/useAPI.tsx
aidemsined: Fixed in `5419fca7` — moved into a `useEffect` keyed on `[arch, design]`. `design` is a stable reference either way (module-level JSON or `null`), so it fires once per arch value.

The frequency was worse than once per render: `useNodeType` is the consumer and NPE playback re-renders per interval tick, so it was a line per frame for the length of the run.

src/hooks/useAPI.tsx
aidemsined: Fixed in `5419fca7` — `getChipDesign` now returns `ChipDesign | null`.

There was a third predicate outside the diff: `NPEViewComponent.tsx:310` tested `architecture.arch_name === undefined`, which only worked *because* the sentinel omitted a field the type declares required. That's now `architecture === null` too, and `designForChip` drops its `.eth?.length` check entirely — one predicate across all three sites.

Kept the identity-stability property from the original rationale; `tests/getChipDesign.spec.ts` pins it, along with the non-empty eth list that made two of those predicates agree by accident.

src/hooks/useAPI.tsx
aidemsined: Fixed in `5419fca7` — moved to `src/functions/getChipDesign.ts`. `useAPI.tsx` and `ClusterRenderer.tsx` both import it from there; `useArchitecture` stays in `hooks/`.

src/components/cluster/ClusterRenderer.tsx
aidemsined: Fixed in `5419fca7`. Chips now carry `EthPort[]` (uid, chan, coordLabel) instead of a bare channel list, built in `recordConnection`. The port pass and the render read `port.uid` off the object, so `ethUid` has one call site and the drift is structurally impossible rather than merely tested for.

Dedupe moved from channel to uid at the same time, and that branch now has a test — mutation-verified by removing the dedupe, which produces a duplicate port on the shared channel.

src/components/cluster/ClusterRenderer.tsx
aidemsined: Fixed in `5419fca7`. `coordLabel` is resolved once in `recordConnection` and carried on the `EthPort`, so both render reads come from the same value.

Preserved the divergence rather than flattening it, because it is intentional: the tooltip falls back to the uid so an unlabelled port stays identifiable, while the inline text stays empty because at 4px on dense meshes a raw uid is unreadable. Added a comment saying so, and a test pinning both halves — `0-0-7-0` in the tooltip vs `''` inline for an out-of-range channel.

tests/ClusterUnknownArch.spec.tsx
aidemsined: Fixed in `5419fca7`. You're right — wormhole carries 16 entries and blackhole 14, so both 4 and 6 resolve. The comment now says what the fixture actually establishes: both channels are in range for both lists, so the label assertions compare two real lookups rather than one lookup and one miss.

Added the guard the old comment claimed, as its own suite: channel 20 against wormhole, asserting the port renders, the link draws, the label is blank, and the tooltip falls back to the uid.

tests/ClusterUnknownArch.spec.tsx
aidemsined: All three covered in `5419fca7`, plus the panel case:

- **Inter-host** — two hosts, both with local chip id 0 on channel 4, linked across ranks. Asserts two ports rather than one, so an unqualified uid fails it; also the link and the rank badges.
- **Dedupe** — same channel on two links, asserting one port per channel. Mutation-verified.
- **Unsupported panel** — `hosts: []`, asserting the copy is *present*. That's what makes the existing negative assertion meaningful.
- Plus the out-of-range-channel suite from the thread above.

Fixtures are parameterised now — `twoChipTopology(arch, links)`, `twoHostTopology(arch)`, and a whole-topology override for shapes the arch knob can't express. 60 tests across the four cluster specs, up from 45.

src/model/ClusterModel.ts
aidemsined: Fixed in `5419fca7` — `tests/clusterTopology.spec.ts:19` is now `{ 0: 'blackhole' }`, and the three `arch: []` are `arch: {}`.

They typechecked because an array satisfies `Record<number, string>`, which is why the shape drift went unnoticed.

dcblundell (approved):

Issue #1844, opened 2026-08-05
Operation Details: per-device CB fan-out is summed, inflating per-core CB memory by the device count


aidemsined:
## Symptom

On a multi-device report, the DeviceOp **CBs** list repeats every circular buffer once per device, with no device shown, and the running CB total climbs as though the repeats stacked on the same cores:

```
0111360  4 KiB x 120 cores    4 KiB   0 B    4 KiB
0111360  4 KiB x 120 cores    8 KiB   0 B    8 KiB
0111360  4 KiB x 120 cores   12 KiB   0 B   12 KiB
...
0111360  4 KiB x 120 cores   32 KiB   0 B   32 KiB
0115456  2 KiB x 120 cores   34 KiB   0 B   34 KiB
```

## Root cause

The repeats are the devices of the mesh, and the emitted graph is correct. Every `circular_buffer_allocate` node in `captured_graph` carries a `device_id`; a mesh op allocates the CB on each device, and mesh addressing is uniform, so address, size and core range legitimately match. From `multihost_poc_jun24_2321`, operation 2 rank 0 has 24 CB nodes = 3 CBs x 8 devices:

```
node id=10  device_id=6  addr=111360  size=4096  cores={[0-0 - 11-9]}
node id=11  device_id=4  addr=111360  size=4096  cores={[0-0 - 11-9]}
node id=16  device_id=0  addr=111360  size=4096  cores={[0-0 - 11-9]}
...
device_ids per (addr, size): [6, 4, 5, 7, 3, 2, 0, 1]
```

The visualizer drops the device dimension. `src/functions/processMemoryAllocations.ts` keys the per-core accumulator by position only:

```ts
for (const { x, y } of cores) {
    const k = `${x},${y}`;
    cbBytesByCore.set(k, (cbBytesByCore.get(k) ?? 0) + size);
}
```

So all 8 devices' 4 KiB CBs land in bucket `"0,0"` and `maxCbPerCore()` returns 32 KiB where the true per-core figure is 4 KiB. The function already reads `node.params.device_id` (around lines 79-81) but only to label the enclosing op.

## Impact

Everything derived from `maxCbPerCore()` is inflated by the device count:

- the `total_cb` column and the `total_memory` sum in the DeviceOp rows
- the **Peak L1 memory load per core** heading
- the per-core CB pressure modal's `byCore` values and `maxBytes`
- `liveCBs` gets one push per device, so the modal lists every CB N times

[...]

Issue #1841, opened 2026-08-05
Operation Details unions every rank: duplicate buffer rows and arbitrary collapse in the memory legend


aidemsined:
Split out of #1594, which was closed as invalid — its premise (multi-*device* address collisions) turned out to be structurally impossible in the report format. This is the defect that actually reproduces.

## Problem

Operation Details requests buffers without a `rank` parameter, so the backend returns **every rank's rows unioned into one response**. The client can't tell them apart, because `rank` is on the wire but not declared in the TS type. Duplicate rows then flow into the memory legend and are collapsed arbitrarily.

Reproduces today in `multihost_poc_jun24_2321` (15 duplicated groups) and `borys-visualizer-db` (4):

```
operation_id|device_id|address|max_size_per_bank|buffer_type|buffer_layout|rank
1|0|5848704|2048|0|0|0
1|0|5848704|2048|0|0|1
```

Identical except `rank`. Note both rows carry `device_id = 0` — see "Why `device_id` can't disambiguate" below.

## Mechanism

1. `views.py:434` applies `db.merge_rank_filter("buffers", {...}, rank)`, but that is a documented no-op when `rank is None`.
2. `src/libs/axiosInstance.ts:41-50` injects only `instanceId`. `rank` is passed explicitly only for `CLUSTER_DESCRIPTOR` / `MESH_DESCRIPTOR` (`useAPI.tsx:555-566`), so operation-scoped requests never carry it.
3. The backend `Buffer` dataclass carries `rank` and serialises it (`models.py:76-83`), but the TS `BufferData` interface (`src/model/APIData.ts`) doesn't declare it — so the discriminator is dropped client-side. `BufferChunk` (`:407`) and `BufferPage` (`:444`) *do* declare `rank?: number`; only the operation-details buffer path is missing it.
4. There is no active-rank concept anywhere in `src/store/`.

## Symptoms

- `getGroupedMemoryReport` groups `deviceBuffers` by address, so a duplicated address yields `group.length > 1` and `MemoryLegendGroup` renders N visually identical rows.
- `buffersSummary` (`useAPI.tsx:758-779`) dedups by `address`, collapsing ranks arbitrarily (keeps the largest `max_size_per_bank`).
- The user cannot tell which rank a row belongs to, nor that a collapse happened.

## Why `device_id` can't disambiguate

This is the part that made #1594 look plausible. Confirmed by reading the tt-metal writer:

[...]

PR #1845, opened 2026-08-05
Scope report reads to rank 0 so multi-host reports stop rendering a colliding union

### LINKED ISSUE #1842: Architectural: rank is not part of the frontend's identity model, so multi-host reports render a colliding union
## Summary

The report DB's primary key for an operation is **`(rank, operation_id)`** — tt-metal's writer says so explicitly. The visualizer treats **`operation_id`** as a global key everywhere: API params, routes, React Query keys, and expand/collapse state. On a multi-host report every rank-bearing table is therefore unioned and every id collides.

Reproduces on `multihost_poc_jun24_2321` (`world_size = 2`): the Operations list renders **8 rows for 4 operations** — `1 ttnn.to_device`, `2 ttnn.multiply`, `3 ttnn.add`, `4 ttnn.gelu`, each twice, with nothing distinguishing the copies.

## Root cause: an asymmetric default

The backend is already rank-aware end to end — `_optional_rank_query_param()` is wired into ~15 endpoints (`/operations`, `/operations/<id>`, `/tensors`, `/devices`, `/buffers`, `/buffer-chunks`, `/errors`, …), each guarded by `merge_rank_filter` and `_reject_nonzero_rank_on_legacy_db`. The problem is what `rank=None` *means*, which differs by route family:

| Route family | `rank` absent means | Result on a 2-rank report |
|---|---|---|
| File-backed (`/cluster-descriptor`, `/mesh-descriptor`, `/profiler-config`) | `logical_rank = 0` | One host. Correct. |
| DB-backed (`/operations`, `/tensors`, `/buffers`, …) | `merge_rank_filter` no-ops → **no filter** | Every rank unioned. Colliding ids. |

And `src/libs/axiosInstance.ts:41-50` injects only `instanceId`. `rank` is passed explicitly *only* for the cluster and mesh descriptors (`useAPI.tsx:555-566`), i.e. exactly the family that already defaults safely. No DB-backed request ever carries it, and there is no rank concept anywhere in `src/store/`.

## Why the ids collide by design

From `graph_report.py`:

- `_OPERATION_ID_STRIDE_PER_RANK_FILE = 10000` is applied only to *additional files within the same rank* — "First JSON file per rank uses operation ids 1..N" (`:1969-1991`). Ranks deliberately reuse `1..N`.
- `import_graph`'s docstring: *"rank: Distributed / SPMD rank for this capture (**disambiguates tensor_id and ops when merging**)"* (`:1012`).
- `device_id` is likewise re-normalized to 0-based **per report file** (`:1997-2003`), so every rank's first mesh is `device 0`.

So `(rank, operation_id)` and `(rank, tensor_id)` are the real keys. Confirmed on the report:

```
rank|operation_id|name          tensors:  20 rows
0|1|ttnn.to_device                      10 distinct tensor_id
1|1|ttnn.to_device                      20 distinct (tensor_id, rank)
0|2|ttnn.multiply
1|2|ttnn.multiply     ... every operation_id has rows=2, ranks=2
```

Every rank-bearing table is doubled: `operations` 8, `tensors` 20, `buffers` 30, `input_tensors` 10, `output_tensors` 8, `devices` 16, `captured_graph` 8, `stack_traces` 8, `operation_arguments` 16.

## Symptoms

1. **Duplicated Operations list** — 2N rows, copies indistinguishable (screenshot above).
2. **Coupled expand/collapse** — `handleToggleCollapsible(operationId)` tracks bare ids (`OperationList.tsx:164-170`), so expanding op 2 expands *both* rank rows at once.
3. **React Query cache collisions** — `['get-operation-detail', operationId, activeProfilerReport?.path]` (`useAPI.tsx:752`) is a single entry for two distinct operations; whichever rank resolves first is served to both. Same for `get-operation-buffers` (`:355`) and `get-buffer-chunks` (`:970`). This is precisely the #1674 bug class one level deeper — the comment at `useAPI.tsx:737-738` documents the lesson for *reports* but it was never extended to *ranks*.
4. **Ambiguous deep links** — `/operations/:operationId` and `/graphtree/:operationId?` (`routeObjectList.tsx:32,48`) cannot express "rank 1's op 2".
5. **Memory legend duplication** — tracked separately in #1841; same root cause.
6. **`report_metadata` is last-write-wins** — the DB carries `world_size = 2` *and* `rank = 1`, so the report's own metadata reports the last-imported rank as its identity.
7. **The app can't tell it's multi-host** — `worldSize` exists only inside the Cluster topology model (`ClusterModel.ts:118`, derived by probing descriptor files), not from `report_metadata`. Every other view is unaware.

## Proposed fix, in two stages

**Stage 1 — stop rendering a corrupted union (small, safe).** Make DB-backed routes default to rank 0 when `rank` is absent, matching the file-backed family. On legacy DBs this is a no-op (`merge_rank_filter` already skips tables without a `rank` column, and `_reject_nonzero_rank_on_legacy_db` only rejects *non-zero*), so single-host reports are unaffected. Surface `world_size` through `/report-metadata` and show an explicit "showing rank 0 of N" affordance so the scoping is honest rather than silent.

Showing one rank correctly is strictly better than showing every rank with colliding keys.

**Stage 2 — make rank a first-class dimension (the actual architecture work).** Add a `selectedRankAtom` in `src/store/app.ts`, thread `rank` through operation-scoped requests, include it in the affected React Query keys, add it to the operation routes, and add a rank selector gated on `world_size > 1`.

Stage 2 needs a product decision that Stage 1 does not: should Operation Details be *scoped to* one rank, or *display* all ranks side by side? Stage 1 is correct under either answer.

## Acceptance criteria

- Open `multihost_poc_jun24_2321`. The Operations list shows **4 rows, not 8**. Expanding one row does not expand another. Operation detail, buffers, and tensors all resolve to a single unambiguous rank.
- Single-rank reports (resnet50) and legacy DBs without a `rank` column render identically to today.
- A regression test using a two-rank fixture shaped like the real rows above — same `operation_id` and `tensor_id` on both ranks — rather than synthetic unique ids.

## Out of scope

- Changing how tt-metal writes ids. `(rank, id)` is the intended key; the visualizer should honour it.
- Multi-rank *comparison* UI (showing ranks side by side). Separate feature.

## Context

- #1841 — memory-legend duplication; a symptom of this root cause.
- #1594 — closed as invalid; the investigation there established that `device_id` is a rank-local mesh index and cannot disambiguate.
- #1674 — the report-scoped query-key precedent this extends.

aidemsined:
## Summary

A multi-host report rendered every rank at once with colliding ids — the Operations list showed 8 rows for 4 operations, indistinguishable from one another (screenshot in #1842). The cause was an asymmetric default: an absent `?rank` meant "rank 0" on file-backed routes but "no filter" on DB-backed ones, and the report writer restarts `operation_id` and `tensor_id` at 1 for every rank, so the unfiltered read unioned the ranks and collided on both.

This is Stage 1 of the agreed two-stage plan: make the scoping correct and visible. Showing one rank correctly beats showing every rank with colliding keys. Rank *selection* is Stage 2 and is not in here.

Closes #1842

### Backend

- `views.py` — `_optional_rank_query_param() -> Optional[int]` becomes `_rank_query_param() -> int`, returning `_DEFAULT_RANK` (0) when `rank` is absent or empty, and still 400ing on unparseable input. All 11 DB-backed report routes now pass an explicit rank; the three file-backed routes drop their hand-rolled `0 if rank_param is None else rank_param` in favour of the shared helper. `_reject_nonzero_rank_on_legacy_db` narrows to `rank: int`, so its `None` arm goes away.
- `views.py` — the tensor-comparison branch in `tensors_list` kept a `rank is not None` guard that can no longer be false. Removed, with a note that the comparison tables have no `rank` column and can only be narrowed through the rank-scoped tensor ids.

Legacy reports are unaffected: `merge_rank_filter` already skips tables without a `rank` column, and the legacy guard only rejects *non-zero* ranks, so the new default never reaches it.

### Frontend

- `useAPI.tsx` / `APIData.ts` — `ReportMetadata` gains `worldSize`, parsed from the `world_size` key already present in `/report-metadata` and clamped to at least one rank so single-host and legacy reports degrade to 1.
- `FooterInfobar.tsx` — a warning-intent `rank 0 of N` tag and a matching tooltip line appear beside the report name when `worldSize > 1`, so a partial view reads as partial instead of looking like a complete run. Nothing branches on `worldSize`; it is purely an honesty affordance.
- `definitions/Reports.ts` — `SINGLE_HOST_WORLD_SIZE` lives here now that both the hook and the footer read it.

- **No rank selection.** Ranks other than 0 are reachable only by passing `?rank=` by hand. Stage 2 (#1842) adds the atom, the query-key dimension, the route segment, and the selector.
- **`/report-metadata` is still report-global** and last-write-wins across ranks, so the capture timestamp and git SHA beside the new tag may belong to a different rank. Called out in #1842 item 6.
- **#1844** — per-device CB fan-out is summed into one core bucket, inflating per-core CB memory by the device count. Same duplication *symptom*, different cause (devices within a rank, not ranks), and unaffected by this change.
- **#1841** — memory-legend duplication, a child symptom of the same root cause.

## Test plan

- [x] Backend: `pytest backend/ttnn_visualizer/tests` — 522 passed
- [x] Frontend: `vitest run` — 1355 passed across 149 files
- [x] `tsc`, `eslint --max-warnings 0`, `black`, `isort`, `mypy` clean
- [x] Manual: `multihost_poc_jun24_2321` (`world_size = 2`) — Operations list drops from 8 rows to 4, and the footer shows `rank 0 of 2`
- [x] Manual: single-device reports and a legacy DB without a `rank` column render unchanged, with no tag

New coverage worth noting: a second fixture whose two ranks reuse `operation_id = 1` and `tensor_id = 1`, matching how a real capture numbers them, since the pre-existing fixture needs ids unique per rank for its mismatch-404 cases. The equivalence test now asserts rank 0 is the only rank anywhere in the payload — comparing an omitted-rank response to an explicit `rank=0` response is otherwise satisfied by two identical unions. Confirmed by stubbing `merge_rank_filter` to a no-op, which fails nine tests.

## Follow-ups

- #1842 — Stage 2, rank as a first-class dimension
- #1844 — per-device CB fan-out summed across devices
- #1841 — memory-legend duplication

### DIFF
--- a/backend/ttnn_visualizer/tests/views/test_rank_query_param.py
+++ b/backend/ttnn_visualizer/tests/views/test_rank_query_param.py
@@ -3,13 +3,14 @@
 # SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
 """
-API tests for optional ``?rank=`` filtering on multi-host report databases.
+API tests for ``?rank=`` filtering on multi-host report databases.
 """
 
 import sqlite3
 import tempfile
 from http import HTTPStatus
 from pathlib import Path
+from typing import Any, List
 
 import pytest
 from ttnn_visualizer.extensions import db
@@ -245,9 +246,157 @@ CREATE TABLE global_tensor_comparison_records (
 """
 
 
+# Shaped like a real multi-host capture: the writer restarts `operation_id` and
+# `tensor_id` at 1 for every rank, so both ranks reuse the SAME ids and only
+# `(rank, id)` is unique. `_RANKED_REPORT_SQL` above keeps ids distinct per rank
+# because its mismatch tests need an id that exists on one rank only; this
+# fixture is the one that reproduces the collision from #1842.
+_COLLIDING_RANK_REPORT_SQL = """
+CREATE TABLE operations (
+    operation_id int,
+    name text,
+    duration float,
+    rank int NOT NULL DEFAULT 0,
+    UNIQUE(operation_id, rank)
+);
+INSERT INTO operations VALUES (1, 'ttnn.to_device', 1.0, 0), (1, 'ttnn.to_device', 2.0, 1);
+
+CREATE TABLE operation_arguments (
+    operation_id int,
+    name text,
+    value text,
+    rank int NOT NULL DEFAULT 0
+);
+
+CREATE TABLE stack_traces (
+    operation_id int,
+    stack_trace text,
+    rank int NOT NULL DEFAULT 0
+);
+
+CREATE TABLE input_tensors (
+    operation_id int,
+    input_index int,
+    tensor_id int,
+    rank int NOT NULL DEFAULT 0
+);
+
+CREATE TABLE output_tensors (
+    operation_id int,
+    output_index int,
+    tensor_id int,
+    rank int NOT NULL DEFAULT 0
+);
+INSERT INTO output_tensors VALUES (1, 0, 1, 0), (1, 0, 1, 1);
+
+CREATE TABLE tensors (
+    tensor_id int,
+    shape text,
+    dtype text,
+    layout text,
+    memory_config text,
+    device_id int,
+    address int,
+    buffer_type int,
+    rank int NOT NULL DEFAULT 0,
+    size int,
+    UNIQUE(tensor_id, rank)
+);
+INSERT INTO tensors VALUES
+(1, '(1,)', 'float32', 'TILE', '{}', 0, 100, 0, 0, 4096),
+(1, '(1,)', 'float32', 'TILE', '{}', 0, 100, 0, 1, 4096);
+
+CREATE TABLE buffers (
+    operation_id int,
+    device_id int,
+    address int,
+    max_size_per_bank int,
+    buffer_type int,
+    buffer_layout int,
+    rank int NOT NULL DEFAULT 0
+);
+INSERT INTO buffers VALUES (1, 0, 100, 512, 0, 0, 0), (1, 0, 100, 512, 0, 0, 1);
+
+CREATE TABLE buffer_pages (
+    operation_id int,
+    device_id int,
+    address int,
+    core_y int,
+    core_x int,
+    bank_id int,
+    page_index int,
+    page_address int,
+    page_size int,
+    buffer_type int,
+    rank int NOT NULL DEFAULT 0
+);
+INSERT INTO buffer_pages VALUES
+(1, 0, 100, 0, 0, 0, 0, 100, 4096, 0, 0),
+(1, 0, 100, 0, 0, 0, 0, 100, 4096, 0, 1);
+
+-- `device_id` is re-normalised to 0-based per rank by the writer, so both ranks
+-- report `device 0`. That is exactly why device_id cannot disambiguate a rank.
+CREATE TABLE devices (
+    device_id int,
+    num_y_cores int,
+    num_x_cores int,
+    num_y_compute_cores int,
+    num_x_compute_cores int,
+    worker_l1_size int,
+    l1_num_banks int,
+    l1_bank_size int,
+    address_at_first_l1_bank int,
+    address_at_first_l1_cb_buffer int,
+    num_banks_per_storage_core int,
+    num_compute_cores int,
+    total_l1_memory int,
+    total_l1_for_tensors int,
+    total_l1_for_interleaved_buffers int,
+    total_l1_for_sharded_buffers int,
+    cb_limit int,
+    rank int NOT NULL DEFAULT 0
+);
+INSERT INTO devices VALUES
+(0, 4, 4, 2, 2, 1024, 4, 256, 0, 0, 1, 2, 4096, 2048, 2048, 2048, 256, 0),
+(0, 4, 4, 2, 2, 1024, 4, 256, 0, 0, 1, 2, 4096, 2048, 2048, 2048, 256, 1);
+
+CREATE TABLE errors (
+    operation_id int,
+    operation_name text,
+    error_type text,
+    error_message text,
+    stack_trace text,
+    timestamp text,
+    rank int NOT NULL DEFAULT 0
+);
+
+CREATE TABLE local_tensor_comparison_records (
+    tensor_id int,
+    golden_tensor_id int,
+    matches int,
+    desired_pcc float,
+    actual_pcc float
+);
+CREATE TABLE global_tensor_comparison_records (
+    tensor_id int,
+    golden_tensor_id int,
+    matches int,
+    desired_pcc float,
+    actual_pcc float
+);
+"""
+
+
 def _write_ranked_report_db(path: str) -> None:
     conn = sqlite3.connect(path)
     conn.executescript(_RANKED_REPORT_SQL)
+    conn.commit()
+    conn.close()
+
+
+def _write_colliding_rank_report_db(path: str) -> None:
+    conn = sqlite3.connect(path)
+    conn.executescript(_COLLIDING_RANK_REPORT_SQL)
     conn.commit()
     conn.close()
 
@@ -302,6 +451,35 @@ def test_operations_list_without_rank_defaults_to_rank_zero(app, client):
         Path(path).unlink(missing_ok=True)
 
 
+def _collect_ranks(payload: Any) -> List[int]:
+    """Every ``rank`` value anywhere in a response, however deeply nested."""
+    found: List[int] = []
+    if isinstance(payload, dict):
+        for key, value in payload.items():
+            if key == "rank" and isinstance(value, int):
+                found.append(value)
+            else:
+                found.extend(_collect_ranks(value))
+    elif isinstance(payload, list):
+        for item in payload:
+            found.extend(_collect_ranks(item))
+    return found
+
+
+def _assert_scoped_to_rank_zero(payload: Any, path: str) -> None:
+    """
+    Assert a response carries rank-0 rows only.
+
+    A union would surface rank 1 somewhere in the payload, so this is the
+    assertion that can actually fail if a route stops filtering. The
+    non-empty check keeps it from passing vacuously on a route whose
+    serializer drops the ``rank`` field.
+    """
+    ranks = _collect_ranks(payload)
+    assert ranks, f"{path} serialized no rank field, so scoping is unverifiable"
+    assert set(ranks) == {0}, f"{path} leaked non-zero ranks: {sorted(set(ranks))}"
+
+
 @pytest.mark.parametrize(
     "path,extra_query",
     [
@@ -312,6 +490,13 @@ def test_operations_list_without_rank_defaults_to_rank_zero(app, client):
         ("/api/operation-buffers", {}),
         ("/api/errors", {}),
         ("/api/operations/1", {}),
+        ("/api/operation-buffers/1", {}),
+        ("/api/tensors/100", {}),
+        ("/api/buffer-pages", {"operation_id": "1", "address": "100"}),
+        # `/api/buffer` is deliberately absent: `query_next_buffer` looks for the
+        # address in a *later* operation, so it 404s on a one-operation-per-rank
+        # fixture for reasons unrelated to rank. Its rank handling is covered by
+        # `test_legacy_nonzero_rank_returns_422_not_rank_zero_payload`.
     ],
 )
 def test_omitted_rank_is_equivalent_to_explicit_rank_zero(
@@ -321,6 +506,9 @@ def test_omitted_rank_is_equivalent_to_explicit_rank_zero(
     The two route families used to disagree on what an absent rank meant:
     file-backed routes defaulted to 0, DB-backed routes returned every rank.
     Pin the unified contract so the asymmetry can't come back. #1842
+
+    Payload equality alone is not enough: two identical unions would satisfy it.
+    The rank-0 assertion below is what makes this test able to fail.
     """
     with tempfile.NamedTemporaryFile(suffix=".sqlite", delete=False) as f:
         db_path = f.name
@@ -342,6 +530,7 @@ def test_omitted_rank_is_equivalent_to_explicit_rank_zero(
         assert (
             omitted.get_json() == explicit.get_json()
         ), f"{path} returns a different payload when rank is omitted"
+        _assert_scoped_to_rank_zero(omitted.get_json(), path)
     finally:
         Path(db_path).unlink(missing_ok=True)
 
@@ -613,6 +802,96 @@ def test_legacy_nonzero_rank_returns_422_not_rank_zero_payload(
         assert not isinstance(err, list)
     finally:
         Path(path_db).unlink(missing_ok=True)
+
+
+def test_colliding_ids_do_not_union_without_rank(app, client):
+    """
+    The #1842 symptom: with both ranks reusing ``operation_id = 1`` and
+    ``tensor_id = 1``, an unfiltered read returned two indistinguishable rows
+    per entity. Scoped to rank 0 there must be exactly one of each.
+    """
+    with tempfile.NamedTemporaryFile(suffix=".sqlite", delete=False) as f:
+        path = f.name
+    try:
+        _write_colliding_rank_report_db(path)
+        _register_profiler_instance(app, path)
+
+        operations = client.get(
+            "/api/operations",
+            query_string={"instanceId": INSTANCE_ID},
+        )
+        assert operations.status_code == HTTPStatus.OK
+        ops = operations.get_json()
+        assert len(ops) == 1, f"unioned ranks: {ops}"
+        assert ops[0]["id"] == 1
+        assert ops[0]["rank"] == 0
+        assert len(ops[0]["outputs"]) == 1
+
+        tensors = client.get(
+            "/api/tensors",
+            query_string={"instanceId": INSTANCE_ID},
+        )
+        assert tensors.status_code == HTTPStatus.OK
+        tensor_rows = tensors.get_json()
+        assert len(tensor_rows) == 1, f"unioned ranks: {tensor_rows}"
+        assert tensor_rows[0]["id"] == 1
+        assert tensor_rows[0]["rank"] == 0
+
+        buffers = client.get(
+            "/api/buffers",
+            query_string={"instanceId": INSTANCE_ID},
+        )
+        assert buffers.status_code == HTTPStatus.OK
+        buffer_rows = buffers.get_json()
+        assert len(buffer_rows) == 1, f"unioned ranks: {buffer_rows}"
+        assert buffer_rows[0]["rank"] == 0
+    finally:
+        Path(path).unlink(missing_ok=True)
+
+
+def test_colliding_ids_resolve_per_rank_on_detail_routes(app, client):
+    """
+    A shared id must still address both rows: ``/operations/1`` is rank 0's
+    operation by default and rank 1's when asked, never an arbitrary winner.
+    """
+    with tempfile.NamedTemporaryFile(suffix=".sqlite", delete=False) as f:
+        path = f.name
+    try:
+        _write_colliding_rank_report_db(path)
+        _register_profiler_instance(app, path)
+
+        default = client.get(
+            "/api/operations/1",
+            query_string={"instanceId": INSTANCE_ID},
+        )
+        assert default.status_code == HTTPStatus.OK
+        assert default.get_json()["rank"] == 0
+        assert default.get_json()["duration"] == 1.0
+
+        rank_one = client.get(
+            "/api/operations/1",
+            query_string={"instanceId": INSTANCE_ID, "rank": "1"},
+        )
+        assert rank_one.status_code == HTTPStatus.OK
+        assert rank_one.get_json()["rank"] == 1
+        assert rank_one.get_json()["duration"] == 2.0
+
+        # Same colliding tensor_id, disambiguated only by rank.
+        tensor_zero = client.get(
+            "/api/tensors/1",
+            query_string={"instanceId": INSTANCE_ID},
+        )
+        assert tensor_zero.status_code == HTTPStatus.OK
+        assert tensor_zero.get_json()["rank"] == 0
+
+        tensor_one = client.get(
+            "/api/tensors/1",
+            query_string={"instanceId": INSTANCE_ID, "rank": "1"},
+        )
+        assert tensor_one.status_code == HTTPStatus.OK
+        assert tensor_one.get_json()["rank"] == 1
+    finally:
+        Path(path).unlink(missing_ok=True)
 
 
 def test_rank_zero_explicit_allowed_on_legacy_db(app, client):
--- a/backend/ttnn_visualizer/tests/views/test_rank_query_param.py
+++ b/backend/ttnn_visualizer/tests/views/test_rank_query_param.py
@@ -276,7 +276,12 @@ def _register_profiler_instance(
         db.session.commit()
 
 
-def test_operations_list_without_rank_returns_all_ranks(app, client):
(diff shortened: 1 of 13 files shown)


backend/ttnn_visualizer/views.py
```diff
@@ -691,7 +699,9 @@ def tensors_list(instance: Instance):
         tensors = list(
             db.query_tensors(db.merge_rank_filter("tensors", tensor_filters, rank))
         )
-        if rank is not None and "rank" in db._get_table_columns("tensors"):
+        # The comparison tables carry no `rank` column, so they can only be
+        # narrowed indirectly, via the rank-scoped tensor ids.
+        if "rank" in db._get_table_columns("tensors"):
```
dcblundell: **Must fix:** this now materialises every tensor id into an `IN (...)` list on every request.

Nothing in `src/` sends `?rank=` to this endpoint, so on `dev` the guard was never true and the `else` ran two plain scans. Now `_rank_query_param()` always returns 0, so any report with a `rank` column on `tensors` takes the id-list path, and `_query_table` emits one bind parameter per id, twice. The bundled SQLite 3.47 caps that at 32766 variables — above it the request 500s with `OperationalError: too many SQL variables`, and below it the query still grows linearly with the report.

Suggest pushing the scoping into SQL rather than round-tripping the ids: `tensor_id IN (SELECT tensor_id FROM tensors WHERE rank = ?)` via a rank-aware sibling of `query_tensor_comparisons` keeps the parameter count at one and lets SQLite use the `tensors` index.

backend/ttnn_visualizer/views.py
```diff
@@ -175,24 +175,33 @@ def _stack_source_request_params():
...
+    passed None and returned the union. #1842
     """
-    if "rank" not in request.args:
-        return None
     raw = request.args.get("rank")
     if raw is None or raw == "":
-        return None
+        return _DEFAULT_RANK
     try:
         return int(raw)
```
dcblundell: **Should fix:** `int(raw)` accepts arbitrary-precision integers.

`?rank=99999999999999999999` on a rank-schema report gets past this parse and reaches SQLite binding, where it raises `OverflowError: Python int too large to convert to SQLite INTEGER` — `execute_query` doesn't catch it. These routes aren't `@local_only`, so that's an unauthenticated 500 under `SERVER_MODE`. Bound the parse to a sane range (e.g. `0 <= rank < 2**31`) and `abort(400)` outside it, alongside the existing non-integer branch.

backend/ttnn_visualizer/views.py
```diff
@@ -691,7 +699,9 @@ def tensors_list(instance: Instance):
         tensors = list(
             db.query_tensors(db.merge_rank_filter("tensors", tensor_filters, rank))
```
dcblundell, line 714: **Should fix:** `merge_rank_filter` now runs two schema probes on every call.

It used to return immediately on `rank is None`, which was the real-world case. With rank always 0 it now runs `_check_table_exists` (a `sqlite_master` query) *and* `_get_table_columns` (a `PRAGMA table_info`) every time, and `_get_table_columns` is uncached (`queries.py:160-166`). On this endpoint it compounds: six calls in `query_tensors`, three more in `query_producers_consumers`, plus the new one just below. That's a dozen-plus schema round-trips per request for a schema that can't change within the lifetime of one `DatabaseQueries`. Memoising both behind a dict on `__init__` removes the regression and a pre-existing cost together.

backend/ttnn_visualizer/views.py
```diff
@@ -175,24 +175,33 @@ def _stack_source_request_params():
...
+        return _DEFAULT_RANK
     try:
         return int(raw)
     except (TypeError, ValueError):
         abort(
             400,
             description="Invalid query parameter 'rank': expected an integer.",
         )
-        return None
+        # Unreachable: `abort` raises. Flask's stubs don't type it `NoReturn`,
```
dcblundell: **Should fix:** the stated reason for this `return` isn't accurate.

Flask 3.1.3 (the pinned version) does annotate `abort` as `-> t.NoReturn`. The `return` is genuinely still required, but because `follow_imports = "skip"` under `[tool.mypy]` in `pyproject.toml` stops mypy reading Flask's annotations at all. As written, a reader who checks Flask's typing concludes the comment is obsolete and deletes a line CI still needs — worth pointing at the mypy setting instead.

src/components/FooterInfobar.tsx
```diff
@@ -221,6 +236,21 @@ function FooterInfobar() {
...
 
+interface ReportRankTagProps {
+    worldSize: number;
+}
+
+const ReportRankTag = ({ worldSize }: ReportRankTagProps) => (
+    <Tag
+        minimal
+        className='report-rank-tag'
+        aria-label={`Showing rank ${SCOPED_RANK} of ${worldSize}`}
```
dcblundell: **Should fix:** the rank sentence is written inline three times.

The same user-visible copy appears in the tooltip (line 163), in this `aria-label`, and in the visible tag text (line 250). The tests assert against the accessible name, so the visible text can drift from it without anything failing. A single `getScopedRankLabel(worldSize)` helper — with the tooltip appending the "Other ranks are not yet selectable" clause — gives the tag text, the accessible name and the assertions one source of truth, which is also what the naming rule on user-visible copy asks for.

backend/ttnn_visualizer/tests/views/test_rank_query_param.py
```diff
@@ -245,13 +246,161 @@
...
+    operation_id int,
+    operation_name text,
+    error_type text,
+    error_message text,
+    stack_trace text,
+    timestamp text,
+    rank int NOT NULL DEFAULT 0
+);
+
+CREATE TABLE local_tensor_comparison_records (
```
dcblundell, line 381: **Should fix:** the comparison-narrowing branch is never exercised.

All three fixtures declare `local_`/`global_tensor_comparison_records` but insert no rows, so the new `if "rank" in db._get_table_columns("tensors")` branch in `tensors_list` and its `else` both produce an empty result — the change is unobservable, and stubbing it out wouldn't fail anything. This colliding fixture is the natural home for it: insert comparison rows for a tensor id that exists on both ranks and assert `/api/tensors` attaches only rank 0's comparison to the rank-0 tensor.

That same path is the one that breaks at scale (see the comment on `views.py:704`), so it's currently untested for both correctness and volume.

dcblundell (commented):
Combined review across maintainability, tests, performance, conventions and security. One regression worth fixing before merge — `/api/tensors` now builds an `IN (...)` list of every tensor id per request, which hard-fails above SQLite's 32766-variable cap — plus five smaller items. Lower-severity nits are in chat rather than here.

backend/ttnn_visualizer/views.py
```diff
@@ -175,24 +175,33 @@ def _stack_source_request_params():
...
-    Returns None if the parameter is absent or empty.
+    Parse ``?rank=`` for multi-host report DBs, defaulting to rank 0.
+
+    An absent rank must mean rank 0, never "every rank". The report writer
+    restarts ``operation_id`` and ``tensor_id`` at 1 for each rank and
+    re-normalises ``device_id`` per rank, so an unfiltered read unions the
+    ranks and collides on all three. File-backed routes (cluster/mesh
+    descriptor, profiler config) already defaulted to 0; DB-backed routes
+    passed None and returned the union. #1842
     """
```
smountenay-tt, line 201: Clear why-comment — this is exactly the asymmetry that bit us.

One Stage-2 hygiene note while you're here: `DatabaseQueries.merge_rank_filter` still treats `rank is None` as "do not filter" (`queries.py`). Every route in this PR now goes through `_rank_query_param()`, so that branch is dead from `views.py`, but it's still the easy way to reintroduce the union if a new endpoint forgets the helper and passes `None`. Worth either making `rank` non-optional on `merge_rank_filter`, or at least updating its docstring so "None = every rank" is called out as the footgun, not the default.

backend/ttnn_visualizer/views.py
```diff
@@ -651,8 +660,7 @@ def get_config(instance: Instance):
     to read another host's file (debugging).
     """
     report_dir = Path(str(instance.profiler_path)).parent
-    rank_param = _optional_rank_query_param()
-    logical_rank = 0 if rank_param is None else rank_param
+    logical_rank = _rank_query_param()
```
smountenay-tt, line 677: Nice cleanup. Dropping the local `0 if rank_param is None else rank_param` in favour of the shared helper is what makes the "unified contract" claim true for the file-backed side, not just the DB routes.

backend/ttnn_visualizer/views.py
```diff
@@ -691,7 +699,9 @@ def tensors_list(instance: Instance):
         tensors = list(
             db.query_tensors(db.merge_rank_filter("tensors", tensor_filters, rank))
         )
-        if rank is not None and "rank" in db._get_table_columns("tensors"):
+        # The comparison tables carry no `rank` column, so they can only be
+        # narrowed indirectly, via the rank-scoped tensor ids.
+        if "rank" in db._get_table_columns("tensors"):
             tensor_ids = [t.tensor_id for t in tensors]
```
smountenay-tt: Agree with the other review here — this isn't just a cleanup of a now-unreachable `rank is not None` guard. On any report with a `rank` column this branch is now the *default* path (nothing in the FE sends `?rank=` yet), so the `IN (...)` of every tensor id stops being optional and becomes the hot path for multi-host reports.

The comment about narrowing via rank-scoped ids is correct for the comparison tables' schema; the implementation just wants to stay in SQL rather than round-tripping the id list. Happy for that to land in this PR or as a fast follow before Stage 2 starts sending rank from the client.

backend/ttnn_visualizer/tests/views/test_rank_query_param.py
```diff
@@ -245,13 +246,161 @@
...
 
+# Shaped like a real multi-host capture: the writer restarts `operation_id` and
+# `tensor_id` at 1 for every rank, so both ranks reuse the SAME ids and only
+# `(rank, id)` is unique. `_RANKED_REPORT_SQL` above keeps ids distinct per rank
+# because its mismatch tests need an id that exists on one rank only; this
+# fixture is the one that reproduces the collision from #1842.
+_COLLIDING_RANK_REPORT_SQL = """
+CREATE TABLE operations (
+    operation_id int,
+    name text,
```
smountenay-tt, line 258: This fixture is the right characterisation of the bug. Keeping `_RANKED_REPORT_SQL` with unique-per-rank ids for the mismatch-404 cases, and adding a second fixture that reuses `operation_id`/`tensor_id` (and re-normalised `device_id`) the way a real capture does, is exactly how you avoid tests that can't fail. The `device_id` comment at the bottom of the SQL is a good trap for anyone tempted to "fix" collisions via device.

backend/ttnn_visualizer/tests/views/test_rank_query_param.py
```diff
@@ -290,13 +444,97 @@ def test_operations_list_without_rank_returns_all_ranks(app, client):
...
+def test_omitted_rank_is_equivalent_to_explicit_rank_zero(
+    app, client, path, extra_query
+):
+    """
+    The two route families used to disagree on what an absent rank meant:
+    file-backed routes defaulted to 0, DB-backed routes returned every rank.
+    Pin the unified contract so the asymmetry can't come back. #1842
+
+    Payload equality alone is not enough: two identical unions would satisfy it.
+    The rank-0 assertion below is what makes this test able to fail.
```
smountenay-tt, line 538: Appreciate this being explicit. Payload equality of omitted vs `rank=0` is necessary but not sufficient; `_assert_scoped_to_rank_zero` is the assertion that actually encodes "no union". The PR note about stubbing `merge_rank_filter` to a no-op and watching nine tests fail is the kind of mutation check I wish more of these characterisation suites had.

backend/ttnn_visualizer/tests/views/test_rank_query_param.py
```diff
@@ -566,6 +804,96 @@ def test_legacy_nonzero_rank_returns_422_not_rank_zero_payload(
...
+        )
+        assert buffers.status_code == HTTPStatus.OK
+        buffer_rows = buffers.get_json()
+        assert len(buffer_rows) == 1, f"unioned ranks: {buffer_rows}"
+        assert buffer_rows[0]["rank"] == 0
+    finally:
+        Path(path).unlink(missing_ok=True)
+
+
+def test_colliding_ids_resolve_per_rank_on_detail_routes(app, client):
```
smountenay-tt, line 891: Good companion to the list-route collision test. On a real multi-host capture `GET /operations/1` without a rank used to be ambiguous (two rows, same id); pinning "default = rank 0's row, `?rank=1` = the other" is the detail-route half of the contract and is what Stage 2's selector will depend on.

smountenay-tt (commented):
Backend-focused pass on Stage 1.

The contract change is right: absent `?rank` meaning "every rank" on DB routes was the actual bug once ids restart per rank, and collapsing both route families onto `_rank_query_param()` → `_DEFAULT_RANK` is the cleanest fix. Legacy behaviour looks correct (`merge_rank_filter` still no-ops without a `rank` column; `_reject_nonzero_rank_on_legacy_db` only fires for non-zero). The colliding-id fixture + rank-leak assertion are what make the regression coverage real — comparing omitted vs `rank=0` alone would have greenlit the old union.

I'm aligned with the existing note on `/api/tensors` now taking the id-list comparison path by default on ranked schemas; that feels like the one thing worth landing before this ships. Everything else below is observation / Stage-2 hygiene, not blockers for the Stage-1 goal.

backend/ttnn_visualizer/views.py
aidemsined: Fixed in 93ab15ad — reproduced first, since this was worth pinning rather than reasoning about.

Restoring the id-list against a 40k-tensor ranked report gives exactly what you predicted: `sqlite3.OperationalError: too many SQL variables` → 500. That's now `test_comparison_scoping_survives_a_report_larger_than_the_sqlite_var_cap`, which fails on the old implementation and passes on the new one.

Took your subquery, on `query_tensor_comparisons(rank=...)` rather than a sibling method so the local and global tables get it from one place:

```sql
AND tensor_id IN (SELECT tensor_id FROM tensors WHERE rank = ?)
```

One parameter regardless of report size, and `tensors_list` drops the branch entirely.

Two things the fix does not cover, both documented on the method:

- Comparison rows still cannot be disambiguated *between* ranks. With ids restarting per rank, a row for `tensor_id = 1` belongs to whichever rank wrote it and there is no column to say which. Narrowing by rank-scoped ids is as far as this schema allows.
- `operation_detail` (`views.py:539`) has the same id-list shape, but bounded by one operation's inputs and outputs rather than the whole report, so it is nowhere near the cap. Left alone because it predates this PR — say the word if you would rather it were consistent.

backend/ttnn_visualizer/views.py
aidemsined: Fixed in 93ab15ad. The parse is bounded to `0..2**31-1`, with one 400 covering both the non-integer and out-of-range branches.

Negatives are in the same check, which also closes a split that exists today: `?rank=-1` returns `200 []` on the DB routes while the file-backed routes return 400 `rank_out_of_range` for the same value.

`test_invalid_rank_query_returns_400` is now parametrised over the three cases — non-integer, `99999999999999999999`, and `-1`.

backend/ttnn_visualizer/views.py
aidemsined: Fixed in 93ab15ad. Both probes are memoised on dicts built in `DatabaseQueries.__init__`; the report DB is read-only for the lifetime of the context manager, so the schema cannot change under the cache.

Agreed on the framing — the regression and the pre-existing cost are the same fix, and it is the compounding that makes it worth doing rather than the single extra probe.

backend/ttnn_visualizer/views.py
aidemsined: You are right, and the comment was actively misleading rather than merely imprecise. Verified both halves on the pinned version: `inspect.signature(flask.abort)` on Flask 3.1.3 reports `-> 't.NoReturn'`, and dropping the `return` still fails CI with `Missing return statement` because `follow_imports = "skip"` under `[tool.mypy]` stops mypy reading Flask's annotations at all.

Comment now points at the mypy setting, so the line reads as load-bearing for the reason it actually is. Fixed in 93ab15ad.

src/components/FooterInfobar.tsx
aidemsined: Fixed in 93ab15ad. `getScopedRankLabel(scopedRank, worldSize)` now backs the tag text, the `aria-label` and the tooltip, so the accessible name the tests read cannot drift from what is on screen.

Put it in `src/functions/reportRank.ts` rather than a new module: dev landed that file for the synced-folder rank labels while this branch was open, so rank copy now has one home. Direct unit test in `tests/reportRank.spec.ts`.

backend/ttnn_visualizer/tests/views/test_rank_query_param.py
aidemsined: Fixed in 93ab15ad, and this went further than expected — the obvious version of the test is *also* unfalsifiable.

I first wrote it exactly where you suggested: comparison rows for a rank-1-only tensor, assert `/api/tensors` never returns them. It passes with the rank scoping removed. `serialize_tensors` keys comparisons by the tensors already in the response, so surplus rows for another rank's tensors are dropped during serialization and never reach the payload. The narrowing is not observable at the route boundary at all.

So it is split, with each half asserting only what it can:

- `test_query_tensor_comparisons_rank` covers the narrowing at the query layer, where it is observable. Tensor 2 exists on rank 1 only, so `rank=0` must not see its comparison rows (golden 901/801). Inverting the guard in `query_tensor_comparisons` fails it: `[900, 901] == [900]`.
- `test_scoped_comparisons_still_attach_to_their_tensor` keeps a route-level assertion for the part that *is* observable — that scoping does not cost the rank its own comparison — with a docstring saying why it cannot carry the narrowing claim.

Fixtures now insert comparison rows, so the branch is no longer dead whichever way it is read. And you were right that this is the same path as the cap issue; the 40k-tensor test covers the volume half.

backend/ttnn_visualizer/views.py
aidemsined: Went with the docstring, and it now names `rank=None` as the footgun rather than a mode.

I looked at making it non-optional first. It is mechanically easy — `merge_rank_filter` is only called from `views.py`, all of those now via `_rank_query_param()`, so nothing passes `None` — but it would not close the hole. `query_producers_consumers` (`queries.py:801`) and `query_next_buffer` (`queries.py:869`) each carry their own `rank: Optional[int] = None` and their own `rank is not None` branch, without routing through `merge_rank_filter`. Tightening one of three doors would read as closed while two stayed open.

Tightening all three is small, but it touches the rank surface Stage 2 revisits anyway, so I would rather it land there deliberately than partially here. Noted on #1842 — happy to pull it into this PR if you would rather not carry it.

backend/ttnn_visualizer/views.py
aidemsined: Landed in this PR rather than as a fast follow — on any ranked report above the cap it is a hard 500, not just a hot path.

Reproduced the old behaviour before changing it: 40k-tensor ranked report → `too many SQL variables` → 500, now a regression test that fails on the previous implementation. Stayed in SQL as you both suggested (`tensor_id IN (SELECT tensor_id FROM tensors WHERE rank = ?)`), one bind parameter regardless of report size. 93ab15ad.

Also agreed on the `device_id` trap comment being worth keeping — it is the next thing someone would reach for.

dcblundell (approved):

smountenay-tt (approved):

PR #1849, opened 2026-08-06
Improve remote connection UX

dcblundell:
This pull request refactors how remote report folders are discovered and handled, making the distinction between "no reports found" and "report path does not exist" explicit. It introduces a new `RemoteFolderSearch` type to carry both the list of found folders and whether the search root was missing, and updates the codebase and tests to use this new structure. Error handling is improved to provide clearer feedback when remote paths are missing.

**Remote report folder discovery and error handling:**

* Introduced the `RemoteFolderSearch` named tuple, which returns both the list of matched folders and a boolean indicating if the search root was missing, replacing previous list-only returns. (`[[1]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076R994-R1012)`, `[[2]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1207-R1217)`)
* Updated the report folder search logic to use a sentinel exit code (`_MISSING_ROOT_EXIT_CODE`) to distinguish between an empty directory and a missing/inaccessible root, and propagate this distinction through the code. (`[[1]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076R54-R59)`, `[[2]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076R1132-R1135)`, `[[3]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1235-R1237)`)
* Refactored error handling to raise a `RemoteConnectionException` when a report path is missing, and to treat empty but present paths as a count of zero, improving user feedback. (`[backend/ttnn_visualizer/sftp_operations.pyL1014-R1104](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1014-R1104)`)
* Removed the `NoReportsException` and the separate `check_remote_path_exists` code, consolidating path existence checks into the search logic. (`[[1]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L24)`, `[[2]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1014-R1104)`)

**API and test updates:**

* Updated all usages and tests to work with the new `RemoteFolderSearch` return type, including adjusting assertions and helper functions to check both `folders` and `is_root_missing`. (`[[1]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1292-R1294)`, `[[2]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1326-R1328)`, `[[3]](diffhunk://#diff-cf1b48187622774638f1cea2d004c0f133483e20a95795aa5d450309d77fddb4L168-R169)`, `[[4]](diffhunk://#diff-cf1b48187622774638f1cea2d004c0f133483e20a95795aa5d450309d77fddb4L187-R227)`, `[[5]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L566-R579)`, `[[6]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L594-R607)`, `[[7]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L646-R659)`, `[[8]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L667-R680)`, `[[9]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L710-R723)`)
* Replaced the internal `_report_search_find_expression` with `_report_search_command` to include the explicit root existence check in the remote command. (`[[1]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1014-R1104)`, `[[2]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L543-R556)`)
* Added and updated tests to verify the new behavior, including distinguishing empty directories from missing roots and checking the correct propagation of the sentinel exit code. (`[backend/ttnn_visualizer/tests/test_sftp_operations.pyL187-R227](diffhunk://#diff-cf1b48187622774638f1cea2d004c0f133483e20a95795aa5d450309d77fddb4L187-R227)`)

These changes make the handling of remote report directories more robust and user-friendly, providing clearer diagnostics and ensuring all code paths consistently distinguish between missing and empty report paths.


tests/helpers/multihostCheckbox.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Accessible name of the multihost checkbox, for `getByRole('checkbox', { name })`.
+ *
+ * The multihost FormGroup's label points at the checkbox as well as heading the group, so the
+ * checkbox's accessible name is that heading followed by its own label. Matching on the half
+ * that identifies the control is what makes this work, which is too easy to get subtly wrong
+ * for each spec that needs it to derive its own.
+ */
+export const MULTIHOST_CHECKBOX_NAME = new RegExp(MULTIHOST_CHECKBOX_LABEL);
```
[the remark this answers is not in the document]
dcblundell: Agreed, and this has since been addressed a different way: `tests/helpers/multihostCheckbox.ts` was removed in 837b60b. Both `RemoteConnectionDialog.spec.tsx` and `remoteFolderSelector.spec.tsx` now query the checkbox with `screen.getByRole('checkbox', { name: MULTIHOST_CHECKBOX_LABEL })`, passing the label as a plain string so Testing Library matches it exactly rather than as a pattern. No regex construction left to escape, so resolving.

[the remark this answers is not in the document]
dcblundell: Correction to my last reply: dropping the helper for a plain `MULTIHOST_CHECKBOX_LABEL` match broke four tests, because the FormGroup's `labelFor` points at the checkbox, so its accessible name is the group heading plus its own label (`Multihost Performance Reports Search per-rank subdirectories`). The helper is back, but as a plain string composed from the two exported copy constants rather than a `RegExp`, so there's still nothing to escape and the specs share one definition.

smountenay-tt (commented):
Reviewed with a backend focus. Backend tests pass locally (113), `tsc --noEmit` is clean, CI is green. **Nothing here blocks merge** — the core idea is right and the execution is clean. Comments below are a mix of things worth addressing and one flow/docs gap.

## What's good

**Collapsing `check_remote_path_exists` into the search closes a real transport inconsistency, not just a round trip.** The old existence check went through `SSHClient.check_path_exists` (paramiko), while `find_folders_by_files` goes through the `ssh` subprocess via `_ssh_cmd_prefix`. Those two resolve `~/.ssh/config`, `ProxyJump`, and identity selection differently — especially given the `-F /dev/null` behaviour documented in `docs/src/remote-sync.md` when a custom identity file is set. So a path could plausibly "exist" for the check and be unreachable for the search, or vice versa. One command over one transport removes that whole class of confusion. That's the strongest argument for this change and it isn't in the description.

**`RemoteFolderSearch` puts the distinction at the boundary and leaves the meaning to the caller.** The listing flows keep collapsing missing/empty via `.folders`; only the connection test acts on `is_root_missing`. I checked all four non-test call sites were updated and nothing reads the tuple positionally.

**Moving `_report_search_status` into `views.py` is the right home.** Response-message formatting belongs in the view, not the SFTP layer. The payoff shows up in the tests: `test_connection_test_warning_names_the_expected_multihost_layout` dropped its `app_context()`, its `find_folders_by_files` patch, and its `pytest.raises` in favour of a direct call on a pure function.

**Test tightening.** The view tests moved from `assert "…" in messages` to full-list equality. Exactly right for a PR whose point is *removing* the "folder path exists" lines — the old style wouldn't have caught a leftover.

## Docs are stale on the flow this PR changed

`docs/src/remote-sync.md` got the multihost label update but not the behaviour changes, in the "Add SSH Connection" steps at lines 99–103:

- Steps 3 and 4 both state that editing a field or picking a config host *"discards any earlier connection-test result"*. It no longer discards — `hasStaleTestResults` keeps the results on screen and marks them stale. (That change is an improvement; the docs just describe the old behaviour precisely enough that they're now wrong.)
- Step 3 describes the picker as *"Optionally choose an SSH config host"*, but `isAwaitingHostChoice` now hides the entire form until the user picks an alias or the add-new option (when adding, and when `~/.ssh/config` has hosts). That's a bigger change than "optionally".
- Step 7 doesn't mention that a unique connection name is now required before "Add connection" enables.

## Frontend (brief — not my main focus)

Nothing egregious. The staleness model is a genuine improvement over the old `setConnectionTests([])`, which wiped the record of what had failed. Both dialogs get `existingConnections` (add via `AddRemoteConnection`, edit via `RemoteConnectionSelector`), and the edit case excludes itself by identity rather than by name — that's the subtle case and it's handled. Dropping `formatRemoteTestPreview` was also right; it would now be showing a command the backend no longer runs.

Minor, non-blocking: `updateTarget` sets `hasStaleTestResults = true` even before any test has run. Harmless today because `isConnectionSaveable` also requires `tests.length > 0`, but "stale" ends up carrying two meanings.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -21,7 +21,6 @@
 from ttnn_visualizer.exceptions import (
     AuthenticationException,
     HostKeyVerificationException,
-    NoReportsException,
     NoValidConnectionsError,
```
smountenay-tt, line 24: Dropping this import leaves `NoReportsException` with no remaining raise site anywhere in the codebase — but it's still defined at `exceptions.py:118` and still imported and handled at `decorators.py:16` and `decorators.py:103`. That dangling `except NoReportsException` branch is the misleading part, since it implies the `"No remote reports found: …"` wrapping is still something that can happen.

`SSHClient.check_path_exists` (`ssh_client.py:297`) is also left with zero callers now that `check_remote_path_exists` is gone.

Both are cheap deletions while the context is fresh.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1011,103 +1028,80 @@ def _find_performance_report_folders(
...
+    """
+    profiler_search = RemoteFolderSearch(folders=[], is_root_missing=False)
     if remote_connection.profilerPath:
-        remote_profiler_paths = find_folders_by_files(
+        profiler_search = find_folders_by_files(
             remote_connection, remote_connection.profilerPath, [TEST_DB_FILE]
         )
+        if profiler_search.is_root_missing:
+            _raise_missing_report_path(
+                "Profiler directory does not exist or cannot be accessed"
```
smountenay-tt: Vocabulary drift: for this one field a user can now see three different nouns depending on which branch they hit.

- failure (here): `"Profiler directory does not exist or cannot be accessed"`
- warning (`views.py`, via `label.capitalize()`): `"Memory path exists but no reports found"`
- success: `"Found 3 memory reports"`
- form label (`src/definitions/RemoteConnection.ts:47`): `"Remote Memory Report Folder Path"`

Given this PR is specifically about connection-test messaging, making the failure say "Memory" too seems in scope, and it's a one-word change.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1011,103 +1028,80 @@ def _find_performance_report_folders(
...
         logger.info("No profiler path configured; skipping check")
 
-    remote_performance_paths = []
+    performance_search = RemoteFolderSearch(folders=[], is_root_missing=False)
     if remote_connection.performancePath:
-        remote_performance_paths = _find_performance_report_folders(remote_connection)
+        performance_search = _find_performance_report_folders(remote_connection)
+        if performance_search.is_root_missing:
+            _raise_missing_report_path(
+                "Performance directory does not exist or cannot be accessed"
```
smountenay-tt: Raising from the middle discards a profiler count that was already successfully computed above — `test_a_missing_performance_path_names_performance_not_the_profiler` encodes this deliberately, so I assume it was considered.

Worth confirming it's still the preference: it matches the old `has_failures()` short-circuit so it isn't a regression, but the reason the old code behaved that way was that the two checks were separate SSH round trips. Now that both searches happen inside one call, returning a status per path instead of raising would be cheap and strictly more useful — the user would learn their memory path is fine in the same run that tells them their performance path is wrong.

Optional, and fine to leave as-is if the simpler control flow is worth more.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1232,7 +1230,11 @@ def find_folders_by_files(
         )
     except subprocess.TimeoutExpired:
         logger.error(f"Timeout finding folders in: {root_folder}")
-        return []
+        return RemoteFolderSearch(folders=[], is_root_missing=False)
```
smountenay-tt: A timeout now reports as a positive claim about the path. This returns `is_root_missing=False`, which the view renders as *"Memory path exists but no reports found"* — but the search never reached the root, so that message asserts something the code doesn't know. The old copy (`"No remote reports found: Profiler folder path: …"`) was vaguer and therefore accidentally more honest.

The whole point of this PR is that "does not exist" and "exists but empty" are different answers the user can act on differently. "We never got an answer" is a third one, and right now it's being reported as the second.

No strong opinion on the fix — a third state on `RemoteFolderSearch`, or just letting the timeout surface as FAILED, would both work. Not a blocker.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1135,7 +1129,10 @@ def _report_search_find_expression(
         for file_name in file_names
     )
 
+    # `test -e`, matching what the standalone path check used to run, so a path
+    # whose parent is unreadable still reports as absent rather than as empty.
     return (
+        f"test -e {shlex.quote(search_root)} || exit {_MISSING_ROOT_EXIT_CODE}; "
```
smountenay-tt: Nice — this is the load-bearing line of the PR and the `test -e` / `|| exit` ordering is right. Worth noting for anyone reading later that the sentinel check in `find_folders_by_files` correctly runs *before* the `returncode == 255` SSH-error branch, and 87 is documented against the codes the remote side actually produces (`find`'s 1, the shell's 126/127, ssh's 255). Good to see that reasoning captured rather than just a magic number.

One minor thought: `test -e` means a path that's a regular file passes and then reports as "exists but no reports found". `test -d` would be more accurate for something the UI labels a folder path. The comment justifies `-e` as matching the old check, which is a fair reason to leave it — just flagging it since you're in here.

backend/ttnn_visualizer/tests/test_sftp_operations.py
```diff
@@ -184,23 +185,46 @@ def test_blank_lines_in_the_listing_are_dropped(self, connection):
...
+        with patch(
+            "subprocess.run",
+            return_value=_completed("", returncode=_MISSING_ROOT_EXIT_CODE),
+        ):
+            matched = find_folders_by_files(
+                connection, "/remote/missing", ["config.json"]
             )
 
+        assert matched.is_root_missing is True
+        assert matched.folders == []
```
smountenay-tt, line 306: Good test — it pins the one thing actually carrying the distinction back across the wire, and the `_found()` / `_missing()` helpers in the view tests keep intent legible at each call site rather than making readers decode a bare tuple.

Two gaps I noticed in the new coverage:

1. Nothing covers the timeout path returning `is_root_missing=False` (see my comment on `find_folders_by_files`). If the current behaviour is intentional, a test asserting it would make it a decision rather than a default.
2. No view-level test for a root that exists but is unreadable *with a readable parent*. That case goes down the `RemoteConnectionException(FAILED)` route inside `find_folders_by_files` rather than the sentinel — two different code paths that produce similar-looking output from the outside, so one test each would be worth having.

backend/ttnn_visualizer/sftp_operations.py
dcblundell: Both gone in 00d95c1. `NoReportsException` is deleted along with the `except NoReportsException` branch in `decorators.py` and its import — you're right that the dangling handler was the misleading part, since the `"No remote reports found: …"` wrapping it implies can no longer happen. `SSHClient.check_path_exists` is deleted too; `shlex` and `SSHException` are still used elsewhere in that module so the imports stay.

Also dropped the `NoReportsException` mention from `CONVENTIONS.md`, which listed it under "specific not-found cases".

backend/ttnn_visualizer/sftp_operations.py
dcblundell: Agreed and fixed in 00d95c1 — it's "Memory directory does not exist or cannot be accessed" now, so all three lines and the form label use the same noun.

Rather than hardcoding the word at the raise site, the label is now a parameter of a `_count_reports_found(label, search)` helper that both paths go through, called with `"Memory"` and `"Performance"` — the same words the view already passes to `_report_search_status`. That also removed the duplicated `is_root_missing` check and the placeholder `RemoteFolderSearch` values the two branches needed.

backend/ttnn_visualizer/sftp_operations.py
dcblundell: Confirming the preference: keeping the raise for now, so yes, a failing performance path still discards a profiler count that was already computed.

Your reasoning is right that the original justification no longer applies — the two checks were separate round trips, and now they aren't, so "report a status per path" is genuinely cheap in a way it wasn't before. The reason I'm not doing it here is shape rather than cost: it means `RemoteReportCounts` stops being `Optional[int]` per kind and starts carrying a per-path outcome, which the view then has to render for the failed case as well as the found and empty ones. That's a wider change than anything else on this branch, and this PR is already the one moving the missing/empty distinction.

Happy to raise it as a follow-up so it doesn't get lost — the user-facing win you describe (learning your memory path is fine in the same run that tells you the performance path is wrong) is real.

backend/ttnn_visualizer/sftp_operations.py
dcblundell: You're right, and this was the best catch of the review — a timeout was being reported as the exact answer the PR exists to separate out. Fixed in 00d95c1 with your first suggestion, a third state.

`RemoteFolderSearch.is_root_missing` is now `root_state: RemoteSearchRootState`, one of `PRESENT`, `MISSING` or `UNKNOWN`. The timeout returns `UNKNOWN`, and so does the `not root_folder or not file_names` early return, whose comment already said "nothing is known about the root" — that comment is now true of the type rather than just aspirational.

I went with the third state over letting the timeout surface as `FAILED` because `find_folders_by_files` is shared with the listing flows (`get_remote_profiler_folders` and friends), where a raise on timeout would be a behaviour change nobody asked for in this PR. The enum keeps the decision with the caller: the connection test raises `FAILED` for anything that isn't `PRESENT` ("Memory directory could not be checked because the search did not complete"), the listing flows still see an empty result.

Covered by `test_a_timed_out_search_settles_nothing_about_its_root` and a view-level test asserting the message and `FAILED` status.

backend/ttnn_visualizer/sftp_operations.py
dcblundell: Thanks — and you've read the sentinel ordering exactly as intended: it has to be checked before the `255` branch, or an SSH-error interpretation would swallow it.

On `test -e` versus `test -d`: you're right that a regular file at the configured path currently passes and then reports as a folder that exists but holds no reports. I'm leaving `-e` for now because switching the test alone makes the copy wrong in the other direction — a file *does* exist, so "does not exist or cannot be accessed" would be a worse description of it than the empty-folder warning is. Doing this properly means `-d` plus a distinct message along the lines of "is not a directory", which is a third failure case rather than a one-character change, and it's a fairly unlikely misconfiguration compared with the missing-path one this PR is about.

I'll fold it into the same follow-up as the per-path statuses so the copy work happens in one place.

backend/ttnn_visualizer/tests/test_sftp_operations.py
dcblundell: Both gaps closed in 00d95c1.

1. The timeout no longer returns `is_root_missing=False` at all — see my reply on `find_folders_by_files`, it's a third `RemoteSearchRootState.UNKNOWN` state now. `test_a_timed_out_search_settles_nothing_about_its_root` pins the unit-level behaviour, and `test_a_search_that_never_answered_is_not_reported_as_an_empty_path` pins what the user actually sees, so it's a decision with two tests behind it rather than a default.
2. Added `test_an_unreadable_root_fails_under_the_path_it_could_not_read` at view level, covering the readable-parent case that goes down the `RemoteConnectionException(FAILED)` route inside `find_folders_by_files` rather than through the sentinel. Agreed on why it was worth having: the two paths produce similar-looking output from outside, so nothing else would have caught them diverging.

The `_found()` / `_missing()` helpers gained a `_unsettled()` sibling for the new state.

backend/ttnn_visualizer/sftp_operations.py
dcblundell: Filed as #1856 so it doesn't get lost, and resolving this here.

The issue carries your framing of the win — learning your memory path is fine in the same run that tells you the performance path is wrong — along with the point that the original short-circuit justification lapsed once both searches moved into one call. It also folds in the `test -e` / `test -d` thread, since a "path is not a directory" message is the same per-path outcome work and the copy is better written once.

One detail I hadn't spotted when I replied: the loss is visible in the UI, not just internal. `RemoteConnectionDialog` seeds a `PROGRESS` placeholder per configured path and then swaps in whatever the server returns, so on a performance failure the memory row doesn't stay pending — it disappears from the list. The per-path placeholders are already there; the backend just can't fill them.

PR #1851, opened 2026-08-07
Basic local event log

### LINKED ISSUE #1828: Usage signals: backend local event log, fixed path, off switch, app_start
Sub-issue of #1819. The foundation everything else writes through. Serves Q1.

No backend module writes an append log today — the only file writes outside the database and reports are `.last-synced` markers ([`utils.py`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/backend/ttnn_visualizer/utils.py) around line 1069) — so this is new infrastructure rather than an extension of something.

## Scope

**New module `backend/ttnn_visualizer/usage.py`**

- logfmt writer, one `key=value` line per event, appended with `O_APPEND`. Short-line `O_APPEND` writes are atomic, so several instances share one log with no locking.
- Line fields: `ts` (**UTC**, fixed-width `Z` form — see the comment below; this originally said "true local timestamp", which contradicted #1819's own worked example), `event`, `schema_version=1`, plus closed-enum detail fields. Version the format from day one so the collector does not break silently when an event is added or renamed.
- A random **per-launch** `run_id` (see the comment below; originally specified as per-process), regenerated every launch, never persisted across runs and never exported. A stable identifier would turn this into per-user tracking.
- The `app_start` event: `version`, `deployment_mode`, `os`, `python_version`.
- Module-level `logger = logging.getLogger(__name__)`, per AGENTS.md.

**Fixed path, deliberately outside the app data directory**

Write to `~/.ttnn-visualizer/usage/`, not `get_app_data_directory()` ([`utils.py:245-283`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/backend/ttnn_visualizer/utils.py#L245-L283)). Two of that function's four branches are environment-derived (`TT_METAL_HOME`, `$APP_DATA_DIRECTORY`) and a root collector has no access to the user's shell environment, so it could only crawl home directories guessing at `*/generated/ttnn-visualizer`. See caveat 8 on #1819 for why scanning forces a choice between non-monotonic counters and path labels.

This file is an interface to another team: narrow, fixed, documented. The app data directory is an implementation detail that has already moved once — `migrate_old_data_directory()` exists for that reason.

**Settings** in [`settings.py`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/backend/ttnn_visualizer/settings.py)

- `USAGE_RECORDING_ENABLED`, default true, read via `str_to_bool(os.getenv(...))` — never `bool(os.getenv(...))`, which is truthy for `"false"`.
- The off switch must also be readable from a file at the fixed usage path, since an env var is per-shell and therefore easy to set and lose.
- Recording is **disabled outright when `SERVER_MODE`** (`settings.py:122`). The hosted instance writes no usage log at all.
- Startup log line stating that usage is being recorded locally and where.

**Retention**

Cap the log generously — a heavy user is roughly 27 KB/day, so about 10 MB a year. If it must be trimmed, **compact rather than delete**: replace the trimmed span with a summary line carrying its totals, so cumulative counts stay monotonic. Deleting old lines makes totals go down, Prometheus reads the decrease as a counter reset and extrapolates the recovery, inventing activity on top of the lost history. The cap is a privacy control as much as a disk one; agree the number here.

## Decisions to pin in this issue

Both settled — see the comment below.

- **The counting unit** (caveat 17). Proposal: the log is machine-scoped, `app_start` counts processes and is therefore never used for reach, and activity is machine-level.
- **`deployment_mode` needs a `container` member.** #1819 offers `local_upload` / `tt_metal_home`, but `get_app_data_directory()` has a container branch (`utils.py:271-280`) and caveat 1 names containers as a distinct population. Folding them into `local_upload` would misreport them. Decide also whether an explicit `$APP_DATA_DIRECTORY` deployment is worth its own member.

## Tests

`backend/ttnn_visualizer/tests/test_usage.py`:

- path resolves identically regardless of `TT_METAL_HOME` and `$APP_DATA_DIRECTORY`
- off switch honoured from the environment and from the file
- nothing is written under `SERVER_MODE` — note the shared `app` fixture sets `SERVER_MODE: True` (`conftest.py:34`), so the enabled cases must override it
- logfmt round-trip, including that no field can contain a newline or an unescaped separator
- concurrent appends from two writers interleave without truncating a line
- compaction keeps cumulative totals monotonic

## Out of scope

The ingest endpoint, any frontend event, the collector, the docs page.

## Verification

`pnpm flask:test`, `pnpm flask:lint`, `pnpm flask:mypy`. New files need an SPDX header with the current year (`pnpm lint:spdx`).

## Sequencing

~~[#1822](https://github.com/tenstorrent/ttnn-visualizer/pull/1822) is editing `settings.py`. Branch after it merges and re-read the cited line numbers, which will have drifted.~~ #1822 merged on 5 Aug.

dcblundell:
This pull request introduces a new local usage recording feature for the application, allowing usage events to be logged locally and providing users with clear controls to enable or disable this functionality. The implementation ensures that usage recording is opt-out, writes only to the user's machine, and is easily discoverable and testable. The most significant changes are grouped below.

Example log entry
`ts=2026-08-10T18:57:36Z event=app_start schema_version=1 run_id=939fc9b3 version=0.97.0 deployment_mode=local_upload os=darwin python_version=3.10`

**Usage Recording Feature:**

* Added support for local usage recording, controlled by the `USAGE_RECORDING_ENABLED` environment variable and a marker file, with clear instructions in `.env.sample`. Usage events are logged to `~/.ttnn-visualizer/usage/events.log` and are not sent anywhere.
* Introduced the `_UsageRecordingEnabled` descriptor in `settings.py` to correctly resolve the environment variable at runtime, avoiding issues with environment variable overrides and ensuring the feature can be toggled reliably.
* Updated the `DefaultConfig` class to include `USAGE_RECORDING_ENABLED`, defaulting to enabled, and documented its behavior.

**Application Lifecycle Integration:**

* Added the `_record_launch(config)` function to `app.py`, which records each application launch (when enabled), ensures all gunicorn workers share a run ID, and prints a clear message to the user about local logging and how to disable it.
* Integrated usage recording into the application startup sequence, ensuring launches are only recorded after successful startup (i.e., after avoiding early exits).

**Imports and Wiring:**

* Updated imports in `app.py` and `settings.py` to bring in the necessary usage recording utilities and constants from the new `ttnn_visualizer.usage` module. [[1]](diffhunk://#diff-14674be812616fc4545d50f7661dafc8f85a2727687bda0471eda8b372f6af3eR41-R50) [[2]](diffhunk://#diff-4da46363434000d16aa4d53b8ca4b45d42820f8a8e6c016d4effc0655c6e6c63R12)

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1828.


backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,461 @@
...
+    global _run_id
+
+    if _run_id is None:
+        inherited = os.environ.get(RUN_ID_ENV_VAR, "")
+        # An inherited value is attacker-controllable in the same sense any
+        # environment variable is; an unchecked one carrying a newline would forge
+        # log lines. Fall back rather than write a line that fails validation later.
+        _run_id = (
+            inherited if _is_safe_value(inherited) else uuid.uuid4().hex[:RUN_ID_LENGTH]
+        )
```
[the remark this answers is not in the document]
dcblundell: Leaving this as is.

The race can't be reached on any current path. `record_event` has exactly one non-test caller, `_record_launch`, which runs once on the main thread from `main()` and exports `TTNN_VISUALIZER_RUN_ID` before gunicorn is spawned. Every worker therefore starts with the variable already set, so concurrent `get_run_id()` calls there all read the same inherited string and the assignment is idempotent — a race would be unobservable. Nothing records events from a request path yet.

The suggested fix also wouldn't close the window it describes: `os.environ` is a Python-level `MutableMapping`, so its `setdefault` is a get followed by a set and is interruptible under the GIL, same as the current code.

If we later record events from request threads in a process that skipped `main()` (running the Flask dev server directly), the worst case is two run ids in one dev session rather than a crash. Happy to revisit then.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,461 @@
...
+            try:
+                _compact(log_path)
+            finally:
+                fcntl.flock(lock_file, fcntl.LOCK_UN)
+    except OSError as error:
+        logger.warning("Unable to compact the usage log: %s", error)
+
+
+def _compact(log_path: Path) -> None:
+    lines = log_path.read_text(encoding="utf-8").splitlines()
```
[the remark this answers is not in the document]
dcblundell: Already addressed in e77e632e, which is why this thread is marked outdated. `_compact()` now reads with `errors="replace"`, and the comment there records the reason the strict read was a startup risk rather than a lost line: `UnicodeDecodeError` is a `ValueError`, so the `OSError` handler in `compact_if_needed()` would not have caught it, and compaction runs from `main()` before gunicorn is spawned. Covered by `test_compaction_survives_a_log_that_is_not_valid_utf_8`.

Resolving.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,461 @@
...
+            continue
+
+        timestamp = fields.get(TIMESTAMP_FIELD, "")
+        key = tuple(
+            sorted(
+                (name, value)
+                for name, value in fields.items()
+                if name not in _UNSUMMARISABLE_FIELDS
+            )
+        )
```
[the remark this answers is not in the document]
dcblundell: Good catch — fixed in b737d098.

The reachable case is an NFS interleave that severs a line on a key boundary, leaving a tail like `schema_version=1 run_id=aaaaaaaa deployment_mode=local_upload`. That parses cleanly, so it was being summarised into `ts= event=unknown deployment_mode=local_upload schema_version=1 count=1` — an empty timestamp and an event name the line never had. The problem is less that the fields are missing than that the result no longer *looks* malformed, which cuts against the contract in the module docstring that readers skip garbled lines.

`_summarise()` now sends any parsed line missing `ts`, `event` or `schema_version` down the existing verbatim path instead, so a fragment stays visibly broken and still costs at most its own event. Every line the module writes carries all three, and so does every summary line, so re-compaction stays idempotent.

Regression test is `test_compaction_does_not_dress_up_a_fragment_as_a_summary`; it fails on the previous code.

Resolving.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+
+def record_app_start(config: Any, server_mode: Optional[Any] = None) -> None:
+    record_event(
+        UsageEvent.APP_START,
+        server_mode=server_mode,
+        version=get_application_version(),
+        deployment_mode=get_deployment_mode(getattr(config, "TT_METAL_HOME", None)),
+        os=get_operating_system(),
+        python_version=get_python_version(),
+    )
```
aidemsined: > [!CAUTION]
> These are evaluated as arguments, in this frame, before `record_event` enters its `try` on line 311 — so the `except OSError` on line 323 protects none of them, and `record_app_start` has no guard of its own.

`get_deployment_mode` reaches `is_running_in_container()` (`utils.py:823`), which opens `/proc/self/cgroup` catching only `FileNotFoundError` and `PermissionError`; an `IsADirectoryError` or an EIO `OSError` propagates. `get_application_version` catches `PackageNotFoundError` around `distribution_version`, but a damaged `dist-info` can raise other errors out of `importlib.metadata`.

Any of those escapes `_record_launch` (`app.py:434`) into `main()`. Because this runs before `subprocess.run(gunicorn_args)`, the result is that the server doesn't start — the same failure shape `e77e632e` closed on the compaction path, still open on the write path.

Two ways out: wrap this body in its own `except Exception`, or compute the details inside `record_event` behind the `is_recording_enabled` check. The second is the better one — it also stops three file reads building a payload that gets discarded whenever recording is off, which stops being free the moment anything records from a request path.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+
+        if not is_recording_enabled(server_mode):
+            return
+
+        line = _format_line(event, details)
+        if line is None:
+            return
+
+        _append_line(line)
+    except OSError as error:
```
aidemsined: > [!CAUTION]
> `except OSError` is the right handler for `_append_line`, but it doesn't cover everything this `try` can raise.

The reasoning recorded in `e77e632e` — that `UnicodeDecodeError` is a `ValueError` and therefore slips an `OSError` handler — applies inside this block too. `line.encode("utf-8")` raises `UnicodeEncodeError`, also a `ValueError`, if a detail value ever carries a lone surrogate. And `_as_bool` → `str_to_bool` raises `AttributeError` on a `server_mode` that is neither a string nor a bool, which is reachable because the parameter is typed `Any` precisely so it can absorb whatever the config hands it.

For a module whose stated contract is that it never breaks the application it measures, `except Exception` with the existing `logger.warning` is the posture that actually delivers that contract. Narrow handlers are the right instinct almost everywhere else, but here the cost of guessing the exception set wrong is the thing the module promises can't happen.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+    if not split_at:
+        return
+
+    compacted = _summarise(lines[:split_at]) + lines[split_at:]
+
+    # An append landing between the read above and the replace below goes to the old
+    # inode and is lost. The window is a few milliseconds once a year, which is a
+    # better trade than taking a lock on every append.
+    temporary_path = log_path.with_name(f"{log_path.name}.compacting")
+    temporary_path.write_text("\n".join(compacted) + "\n", encoding="utf-8")
```
smountenay-tt, line 519: **Compaction silently drops the log's `0o600` permission.**

`write_text` creates `.compacting` at `0o666 & ~umask`, and `os.replace` then carries that mode onto the log, undoing the `0o600` that `_append_line`'s `os.open` establishes. Ran it against this branch:

```
after append     log mode: 0o600
after compaction log mode: 0o644   <- group/world readable
```

`test_the_log_is_not_world_readable` asserts exactly this invariant, but only against a freshly appended log, so it stays green while the guarantee lapses after the first compaction.

The `0o700` parent directory mitigates it in the common case, but not reliably: `mkdir(mode=0o700, ..., exist_ok=True)` does not tighten a directory that already exists, which is a plausible state here given #1828 puts another team's collector in this path. And the file mode is the control the test names.

One line before the replace:

```python
os.chmod(temporary_path, 0o600)
os.replace(temporary_path, log_path)
```

Worth extending that test to force a compaction first, otherwise the same regression can land again unnoticed.

backend/ttnn_visualizer/settings.py
```diff
@@ -184,6 +186,28 @@ def __get__(self, instance: object, owner: type) -> List[str]:
...
+    stay in step with :func:`usage.is_recording_enabled`, which is the gate that
+    actually decides whether a line is written. Keeping a default this reports but the
+    gate ignores would publish "recording on" while nothing records.
+
+    This is only the environment half of the switch; :func:`usage.is_recording_enabled`
+    combines it with ``SERVER_MODE`` and the marker file at the usage path.
+    """
+
+    def __get__(self, instance: object, owner: type) -> bool:
+        return str_to_bool(os.getenv(USAGE_RECORDING_ENV_VAR, "true"))
```
smountenay-tt: **This reports "recording on" for deployments that record nothing — the failure the docstring above says it exists to prevent.**

The descriptor consults only the env var, but `is_recording_enabled` gates on three things. The one production consumer of `USAGE_RECORDING_ENABLED` is the `PRINT_ENV` dump at `app.py:550`, so the mismatch is user-visible at launch. Ran both cases on this branch:

```
marker file present ->  dump says True,  is_recording_enabled()          = False
SERVER_MODE=True    ->  dump says True,  is_recording_enabled(True)      = False
```

So a hosted instance prints `USAGE_RECORDING_ENABLED=True` in its startup banner while writing no log at all — which is the "publish 'recording on' while nothing records" case the docstring calls out, one gate over from where it was guarded.

`__get__` already receives `instance`, so it can answer honestly:

```python
def __get__(self, instance: object, owner: type) -> bool:
    return is_recording_enabled(getattr(instance, "SERVER_MODE", False))
```

That also removes the duplicated env parsing, and the docstring's note about staying in step with `is_recording_enabled` becomes structural rather than a convention to maintain.

The alternative is to drop the attribute entirely — nothing but the dump and one test reads it, and `is_recording_enabled()` is already the real gate.

backend/ttnn_visualizer/app.py
```diff
@@ -407,6 +417,33 @@ def display_mode_info_without_db(config):
...
+
+    compact_if_needed()
+    record_app_start(config, server_mode=config.SERVER_MODE)
+
+    # `print` rather than `logger.info`: nothing has configured logging at this point
+    # in `main()` — `create_app()` does that — and the last-resort handler drops
+    # anything below WARNING, so an info line here would never be seen.
+    print(
+        f"📊 Recording usage locally to {get_usage_log_path()}\n"
+        f"   Nothing is sent anywhere. Switch it off with "
```
smountenay-tt: **"Nothing is sent anywhere" is true of this code but not of the system it's part of.**

#1828 describes this file as "an interface to another team," read by "a root collector" — so on a managed machine the data does leave, by a mechanism this PR simply doesn't contain. The sentence is accurate about the application and misleading about the outcome, which is the wrong way round for the single line a user reads about data collection. Same wording in `.env.sample` ("sent nowhere").

Suggest scoping the claim to what this code actually guarantees, e.g. *"Written to this machine only; the application transmits nothing."* That stays honest if a collector is later deployed alongside it, and doesn't need revisiting when one is.

Team call rather than a code defect — but better settled now, while this is the text establishing the privacy commitment, than after it ships.

smountenay-tt (commented):
Reviewed at `0a85826`. Ran the gates from #1828 in a clean worktree: `pytest` 847 passed (33 new in `test_usage.py`), `isort`, `black` and `mypy` all clean.

Good PR. The non-obvious decisions are the ones that are commented — the fixed path outside `get_app_data_directory()`, compaction-instead-of-deletion for counter monotonicity, and the two stringified-bool traps, which are subtle enough that the tests pinning them are the most valuable thing here. Coverage goes beyond what the issue asked for (fragment handling, non-UTF-8 survival).

I've skipped the threads already settled above (the strict-read `UnicodeDecodeError`, the fragment-dressed-as-summary bug, the `get_run_id` race) and Copilot's suppressed note about `**details` not enforcing the closed-enum contract — I agree with that one, but it's already on the PR so I won't restate it.

Two findings below I verified by running the code rather than reading it (comments on `usage.py` L472 and `settings.py` L208); both are narrow bugs, not design objections. The third (`app.py` L442) is a wording decision for the team. The rest are nits.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+
+SCHEMA_VERSION = 1
+
+# Deliberately not get_app_data_directory(): two of its four branches are derived
+# from the environment (TT_METAL_HOME, $APP_DATA_DIRECTORY), and a collector running
+# as root has no access to the user's shell environment, so it could only find those
+# by crawling home directories. This path is an interface to another team — narrow,
+# fixed and documented — while the app data directory is an implementation detail
+# that has already moved once.
+USAGE_DIRECTORY = Path.home() / ".ttnn-visualizer" / "usage"
```
smountenay-tt: Nit / hardening: this resolves at import time, and since `settings.py` imports this module, `Path.home()` is now on the import path of the whole application.

`Path.home()` raises `RuntimeError` when home can't be resolved — unset `HOME` plus a uid absent from `/etc/passwd`, i.e. the arbitrary-uid container pattern, and `container` is a population you're explicitly targeting with `DeploymentMode`. An unresolvable home would fail app import rather than cost a usage line.

Resolving inside `get_usage_directory()` instead would confine the blast radius to the feature. I did not reproduce this — it needs a passwd-less uid — so treat it as hardening rather than a confirmed break.

backend/ttnn_visualizer/settings.py
```diff
@@ -10,6 +10,7 @@
 
 from dotenv import load_dotenv
 from sqlalchemy.pool import NullPool
+from ttnn_visualizer.usage import USAGE_RECORDING_ENV_VAR
```
smountenay-tt: Nit: this inverts the layering for one string constant — config now depends on a feature module, which puts `usage.py` (and its import-time `Path.home()`) on the import path of all config.

Cleaner the other way round: define `USAGE_RECORDING_ENV_VAR` in `utils.py` alongside the other shared env-parsing pieces this module already pulls from, and have `usage.py` import it from there.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+
+    log_path = get_usage_log_path()
+
+    try:
+        if not log_path.exists() or log_path.stat().st_size <= MAX_LOG_BYTES:
+            return
+
+        with open(
+            get_usage_directory() / COMPACTION_LOCK_NAME, "w", encoding="utf-8"
+        ) as lock_file:
```
smountenay-tt, line 484: Nit: `.compaction.lock` is left behind in the usage directory afterwards, and created with default permissions rather than the `0o600` used for the log.

Harmless in itself, but this directory is a documented interface to another team, so a collector globbing it will find a file that isn't an event log. Worth agreeing the collector matches `*.log` only, or naming that constraint in the module docstring next to the other consumer-facing properties.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+            (TIMESTAMP_FIELD, latest_timestamps[key]),
+            (EVENT_FIELD, remaining.pop(EVENT_FIELD, UNKNOWN_VALUE)),
+        ]
+        summary_fields += sorted(remaining.items())
+        summary_fields.append((COUNT_FIELD, str(count)))
+        summaries.append(_render_line(summary_fields).rstrip("\n"))
+
+    # Lines we could not parse are kept verbatim rather than dropped, so a garbled
+    # line costs its own event at worst and never the totals around it.
+    return sorted(summaries) + unparsed
```
smountenay-tt, line 458: Nit, following on from the fragment fix: `sorted(summaries) + unparsed` moves the kept-verbatim lines to the end of the compacted span, so they land out of time order relative to the summaries they sat between.

Fine if the collector makes no ordering assumption — but that's now a third property it depends on, and the module docstring deliberately states the other two (`count` defaults to 1; malformed lines are skippable) precisely because they aren't obvious from the reading code. Worth adding a line there.

Unrelated and lower still: if the older half is mostly unparseable lines kept verbatim, compaction can't get the file under the cap, so it rewrites the full 10 MB on every launch. Bounded and harmless, just noting it isn't self-limiting.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+    would be a shared server file mixing many users, including external ones, which
+    has a different privacy profile entirely.
+
+    The file half of the off switch exists because an environment variable is
+    per-shell, and so easy to set in one terminal and lose in the next.
+    """
+    if _as_bool(server_mode):
+        return False
+
+    if not str_to_bool(os.getenv(USAGE_RECORDING_ENV_VAR, "true")):
```
aidemsined, line 188: > [!WARNING]
> This expression appears verbatim at `settings.py:208` as well, and `_UsageRecordingEnabled`'s docstring asks the next editor to keep the two in step.

A duplicated default on a privacy switch is exactly the kind of thing that drifts: one copy picks up an alias or a different fallback, and then the `PRINT_ENV` dump starts disagreeing with the writer about whether recording is on. Exporting one helper from here and having the descriptor call it makes the invariant the code's problem rather than the reader's. It also removes the only reason `settings.py` needed the `str_to_bool` import this PR added to it.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+    # `errors="replace"` rather than a strict read: a `UnicodeDecodeError` is a
+    # `ValueError`, so the `OSError` handler around this would not catch one, and
+    # compaction runs from `main()` before gunicorn is spawned — a corrupted log
+    # would stop the server starting rather than cost us a line.
+    lines = log_path.read_text(encoding="utf-8", errors="replace").splitlines()
+    split_at = len(lines) // 2
+    if not split_at:
+        return
+
+    compacted = _summarise(lines[:split_at]) + lines[split_at:]
```
aidemsined, line 508: > [!WARNING]
> When the older half holds nothing summarisable this returns a list no shorter than the input, the file stays above `MAX_LOG_BYTES`, and every subsequent launch pays the full read, summarise and rewrite for no reduction.

The condition is sticky rather than transient: lines that fail the `_REQUIRED_FIELDS` guard are carried forward verbatim, and next time they land in the older half again. An NFS home accumulating severed fragments is the realistic way in — the log then grows without bound *and* pays compaction cost at every start, which is the one outcome the retention design was meant to rule out.

Cheapest guard is to return without rewriting when `len(compacted) >= len(lines)`; that also skips a pointless `os.replace` and the append-loss window documented just below. Bounding the carried-forward unparsed list would address the unbounded growth too.

backend/ttnn_visualizer/usage.py
```diff
@@ -0,0 +1,473 @@
...
+    That guarantee does not hold over NFS, where two simultaneous writers may
+    interleave; readers of this file are expected to skip malformed lines.
+    """
+    get_usage_directory().mkdir(mode=0o700, parents=True, exist_ok=True)
+
+    descriptor = os.open(
+        get_usage_log_path(), os.O_WRONLY | os.O_APPEND | os.O_CREAT, 0o600
+    )
+    try:
+        os.write(descriptor, line.encode("utf-8"))
```
aidemsined: > [!WARNING]
> `os.write` can return a short count without raising — on a full filesystem it typically does — which leaves a partial line in the log.

That makes this the one place the writer can produce the malformed line the module docstring tells readers to expect, so a truncated record becomes indistinguishable from an NFS interleave and the collector can't tell a disk-full local install from a shared-mount artefact. Either loop until the buffer is drained, or compare the return against `len(encoded)` and log a warning.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -0,0 +1,412 @@
...
+    assert get_usage_log_path().stat().st_mode & 0o077 == 0
+    assert usage_directory.stat().st_mode & 0o077 == 0
+
+
+def test_record_launch_records_one_line_and_exports_the_run_id(
+    usage_directory, monkeypatch
+):
+    from ttnn_visualizer.app import _record_launch
+
+    monkeypatch.delenv(RUN_ID_ENV_VAR, raising=False)
```
aidemsined: > [!WARNING]
> `delenv(..., raising=False)` has nothing to record when the variable is absent, so the value `_record_launch` writes into the real `os.environ` isn't rolled back at teardown and persists for the rest of the session.

`test_record_launch_records_nothing_in_server_mode` leaks it the same way on line 407. Inert today, since nothing outside this module calls `get_run_id()` — but it makes test order significant the moment something does, and `test_the_run_id_is_inherited_when_it_is_safe` on line 195 is precisely the shape that would silently start passing for the wrong reason.

`monkeypatch.setenv(RUN_ID_ENV_VAR, "")` before the call gives the fixture something to restore.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -0,0 +1,412 @@
...
+
+    assert parse(read_lines(usage_directory)[0])["run_id"] == "abc12345"
+
+
+def test_concurrent_writers_never_truncate_a_line(usage_directory):
+    writes = 200
+
+    with ThreadPoolExecutor(max_workers=8) as executor:
+        for _ in range(writes):
+            executor.submit(record_event, UsageEvent.APP_START)
```
aidemsined, line 230: > [!WARNING]
> Threads in a single process, each doing its own open/write/close under the GIL, don't meaningfully exercise the `O_APPEND` atomicity this test is named for — it would pass on an implementation that funnelled every writer through one shared buffered file object.

`_append_line`'s docstring makes its claim about separate *instances* sharing the log with no coordination, so two `subprocess` writers are the case worth pinning, which is also how the ticket phrases it. Failing that, writing lines long enough to make a non-atomic write visible would at least give the assertion something to catch.

backend/ttnn_visualizer/tests/test_usage.py
aidemsined, line 1: > [!WARNING]
> Nothing in here makes a write fail, so the guarantee the module leads with — that instrumentation never breaks the application it measures — is the one behaviour with no test behind it.

The suite is otherwise thorough, which is what makes the gap stand out: the off switch is covered from both the environment and the marker file, `SERVER_MODE` is covered three ways including the `conftest` override, and compaction has monotonicity, idempotency and two regression tests. The fail-open path has none.

A test that makes the usage directory read-only, or monkeypatches `os.write` to raise `OSError`, and asserts `record_event` returns normally and logs a warning, would pin it — and would have caught the `record_app_start` argument-evaluation problem, since that escape happens before the handler is ever reached. Worth pairing with one that asserts the disabled and `SERVER_MODE` paths don't build the `app_start` payload at all.

Two smaller gaps in the same area: `compact_if_needed`'s early return when another process holds the `flock` (line 446) is uncovered, as is the `ImportError` fallback for `fcntl`.

backend/ttnn_visualizer/usage.py
dcblundell: Fixed. `record_app_start` now checks `is_recording_enabled` before building details, and wraps the whole body in `except Exception` so a failure in `get_deployment_mode` / `get_application_version` cannot escape into `main()`.

That also skips the three file reads when recording is off. Covered by `test_disabled_recording_does_not_build_the_app_start_payload` and `test_an_app_start_detail_failure_does_not_break_the_caller`.

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Agreed — switched to `except Exception` with the existing warning. The `UnicodeEncodeError` case is real; `_as_bool` itself only calls `str_to_bool` on strings, but the broader handler is the right posture for the module contract regardless.

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Fixed — `os.chmod(temporary_path, 0o600)` before the replace. `test_the_log_is_not_world_readable` now forces a compaction after setting `0o600` on the log so the regression stays visible.

Resolving.

backend/ttnn_visualizer/settings.py
dcblundell: Fixed — `__get__` now returns `is_recording_enabled(instance.SERVER_MODE)`, so the `PRINT_ENV` dump matches the writer. That also removes the duplicated env parse and the `str_to_bool` import this PR had added to `settings.py`.

Covered by `test_usage_recording_config_reflects_server_mode` and `test_usage_recording_config_reflects_the_marker_file`.

Resolving.

backend/ttnn_visualizer/app.py
dcblundell: Agreed — reworded the launch banner, `.env.sample`, and the module / config comments to: *"Written on this machine only; the application transmits nothing."*

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Fixed — `Path.home()` is resolved inside `get_usage_directory()` on each call. `USAGE_DIRECTORY` is now an optional test override (`None` in production) so import of `settings` no longer depends on home being resolvable.

Resolving.

backend/ttnn_visualizer/settings.py
dcblundell: Leaving the `settings` → `usage` edge on purpose: the descriptor now calls `is_recording_enabled`, so moving `USAGE_RECORDING_ENV_VAR` into `utils` alone would not remove the dependency. The import-time blast radius (`Path.home()`) is gone via the lazy resolve above.

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Documented in the module docstring: collectors should match `*.log` only; `.compaction.lock` is not event data.

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Documented the ordering property in the module docstring. Also added the no-op rewrite skip from the related warning (`len(compacted) >= len(lines)`), which covers the sticky unparseable-half case you noted at the end.

Not bounding the verbatim carry-forward in this pass — the skip already stops the rewrite-every-launch cost; growth from NFS fragments remains a known limit.

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Fixed by having the descriptor call `is_recording_enabled` directly — one gate, no duplicated default. The `str_to_bool` import is gone from `settings.py`.

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Fixed — `_compact` returns without rewriting when `len(compacted) >= len(lines)`. Covered by `test_compaction_skips_rewrite_when_nothing_is_summarisable`.

Not bounding the unparsed carry-forward here; the skip removes the every-launch rewrite cost. Happy to revisit a hard bound if the collector starts seeing sticky growth in the wild.

Resolving.

backend/ttnn_visualizer/usage.py
dcblundell: Fixed — `_append_line` now loops until the buffer is drained, and treats a zero-byte write as `OSError` so it still fails open through `record_event`.

Resolving.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Fixed — both launch tests now `monkeypatch.setenv(RUN_ID_ENV_VAR, "")` before calling `_record_launch`, so teardown restores the prior environment.

Resolving.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Added `test_a_write_failure_does_not_break_the_caller` (`os.write` → `OSError`), `test_an_app_start_detail_failure_does_not_break_the_caller`, and `test_disabled_recording_does_not_build_the_app_start_payload`.

Leaving the flock-held early return and `fcntl` `ImportError` fallback uncovered for now — lower risk than the fail-open path that had nothing behind it.

Resolving.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Agreed the ThreadPool version does not pin cross-process `O_APPEND` atomicity. Leaving it for a follow-up rather than expanding this review round — happy to revisit with two subprocess writers.

Leaving open.

dcblundell: Tracked in #1871. Resolving.

PR #1861, opened 2026-08-11
Refuse report deletion where the app does not own the reports

### LINKED ISSUE #1821: Deleting a report silently does nothing in direct-report (TT_METAL_HOME) mode
## Problem

In direct-report mode — the app running on the same machine as TT-Metal, with `TT_METAL_HOME` set — the report listings and the report deletes resolve against different trees, so deleting a listed report can never succeed.

`GET /profiler` and `GET /performance` both resolve their base path through `PathResolver.get_base_report_path()` (`backend/ttnn_visualizer/utils.py`), which in direct mode points into the TT-Metal checkout:

- profiler → `$TT_METAL_HOME/generated/ttnn/reports`
- performance → `$TT_METAL_HOME/generated/profiler/reports`

The matching `DELETE /profiler/<name>` and `DELETE /performance/<name>` routes (`backend/ttnn_visualizer/views.py`) instead resolve against the app's own data directory, `LOCAL_DATA_DIRECTORY/{profiler-reports,performance-reports}/<name>`. In direct mode nothing the listing returned exists there, so every delete falls through to `response_not_found`.

The UI doesn't hide the control. `LocalFolderSelector` gates only the upload `FileInput`s on `isDirectReportMode`; the trash button in `LocalFolderPicker` renders whenever a `handleDelete` prop is supplied, which both pickers do. `handleDeleteProfiler` and `handleDeletePerformance` don't catch, so the rejected request produces no toast either — the user confirms the delete alert and nothing observable happens.

## Why it matters

Minor in impact, but it presents as a broken app rather than an unsupported operation: a delete button that is enabled, asks for confirmation, and then does nothing at all.

## Possible directions

1. **Hide the delete button in direct-report mode.** `isDirectReportMode` is already computed in `LocalFolderSelector`; passing no `handleDelete` in that mode drops the button. Honest about the fact that the app doesn't manage reports it didn't create.
2. **Surface the failure.** Cheapest, but leaves a control that never works.
3. **Resolve deletes through the same `PathResolver` the listing uses,** so a delete removes what was listed. Coherent, but it means `shutil.rmtree` inside the user's TT-Metal checkout against reports the app neither created nor owns. That deserves a deliberate decision — and probably a distinct confirmation — rather than falling out of a refactor.

Option 1 looks right unless we decide the app should own reports in the TT-Metal tree.

## Related

- #1813 confined both delete routes to the local data directory and to a single named report (they previously chose their tree from the *active* report's location, which sent deletes into the synced remote directory and dropped the report-name segment entirely). The direct-mode mismatch described here predates that PR and is unchanged by it: the non-remote branch already resolved to `LOCAL_DATA_DIRECTORY`.

smountenay-tt:
Closes #1821

In direct-report mode (`TT_METAL_HOME` set) the report listings and the report deletes resolved against different trees, so deleting a listed report could never succeed — and nothing said so.

`GET /profiler` and `GET /performance` resolve their base path through `PathResolver.get_base_report_path()`, which in direct mode points into the TT-Metal checkout. Both `DELETE` routes were pinned to `LOCAL_DATA_DIRECTORY/{profiler-reports,performance-reports}/<name>` instead. Nothing the listing returned exists there, so every delete fell through to `response_not_found`.

The UI didn't hide the control either: `LocalFolderSelector` gated only the upload `FileInput`s on `isDirectReportMode`, while `LocalFolderPicker` renders its trash button whenever a `handleDelete` prop is supplied — which both pickers did unconditionally. The handlers didn't catch, so the rejected request produced no toast. The user confirmed the delete alert and nothing observable happened.

## What this does

**Hides the control.** The gate goes in `LocalFolderPicker` alongside the `SERVER_MODE` gate it already owns, rather than dropping `handleDelete` at the two call sites. That component's existing comment records why the trash button and its confirmation alert must be gated *together* — one gate keeps that invariant, covers both pickers, and can't be forgotten by a future call site. `ComparisonReportSelector` passes no `handleDelete`, so it is unaffected.

**Refuses the routes.** `_direct_report_mode_refusal()` returns 403 from both `DELETE` routes in direct mode, reusing the `PathResolver.is_direct_report_mode` predicate the listings already use rather than adding a second `TT_METAL_HOME` check. A 404 against a tree the client never saw listed is a misleading answer to an API caller; per `AGENTS.md`, local-only behaviour is gated on both sides.

This deliberately does **not** point deletes at the TT-Metal tree. That would mean `shutil.rmtree` inside the user's checkout against reports the app neither created nor owns — a product decision to take on its own terms, not one to fall out of a path fix. `docs/src/running-the-application.md` already documents that sync and upload are disabled in this mode; deletion is now named there too.

**Surfaces failures.** `handleDeleteProfiler` and `handleDeletePerformance` now catch, emit the response error via the existing `getResponseError` + `ToastType.ERROR` convention (as `notifyFolderSyncError.ts` does), and return before claiming success or clearing the active report. So a delete that fails for any other reason is visible rather than silent.

## Testing

Backend 817 passed, frontend 1618 passed; eslint, `tsc --noEmit`, isort, black and mypy clean.

New coverage: the picker offering no delete action or confirmation in direct-report mode (asserting the rows still render, so it's the gate under test rather than an empty dropdown); a rejected delete surfacing an error toast and leaving the report listed; and both `DELETE` routes refusing with 403 in direct mode, with the report still on disk — including a same-named report in the app's own data directory, which must survive the refusal too.

I confirmed the new tests fail without the source changes rather than merely passing alongside them.

## Notes for review

- `_report_directory_to_delete`'s docstring asserted that the paired listings "only ever read the local data directory". That was false in direct mode; it now states the precondition the new guard establishes.
- Branched off `origin/dev` at `58a0166e`.

### DIFF
--- a/backend/ttnn_visualizer/tests/views/test_report_deletion.py
+++ b/backend/ttnn_visualizer/tests/views/test_report_deletion.py
@@ -16,6 +16,7 @@ from pathlib import Path
 
 import pytest
 from ttnn_visualizer.app import create_app
+from ttnn_visualizer.decorators import DIRECT_REPORT_MODE_DELETE_REFUSAL
 from ttnn_visualizer.extensions import db
 from ttnn_visualizer.instances import (
     KEY_PERFORMANCE_LOCATION,
@@ -25,7 +26,6 @@ from ttnn_visualizer.instances import (
 )
 from ttnn_visualizer.models import InstanceTable, RemoteConnection, ReportLocation
 from ttnn_visualizer.utils import create_path_resolver
-from ttnn_visualizer.views import _DIRECT_REPORT_MODE_DELETE_REFUSAL
 
 API = "/api"
 HOST = "yyzc-wh-05"
@@ -33,13 +33,21 @@ HOST = "yyzc-wh-05"
 
 @pytest.fixture
 def app():
-    """A local-mode app: ``@local_only`` refuses these routes under SERVER_MODE."""
+    """A local-mode app: ``@local_only`` refuses these routes under SERVER_MODE.
+
+    ``TT_METAL_HOME`` is pinned off because it is otherwise read from the developer's own
+    environment (via ``DefaultConfig`` and ``create_app``'s ``load_dotenv``), and it is
+    exported on any machine that profiles TT-Metal. Left unpinned, every upload/sync-mode
+    case here meets the direct-report-mode refusal and fails in a way indistinguishable
+    from a real regression.
+    """
     tmpdir = tempfile.mkdtemp()
     try:
         settings = {
             "TESTING": True,
             "SQLALCHEMY_DATABASE_URI": f"sqlite:///{Path(tmpdir) / 'app.db'}",
             "SERVER_MODE": False,
+            "TT_METAL_HOME": None,
             "APP_DATA_DIRECTORY": tmpdir,
             "REPORT_DATA_DIRECTORY": tmpdir,
             "LOCAL_DATA_DIRECTORY": str(Path(tmpdir) / "local"),
@@ -306,7 +314,7 @@ class TestDirectReportModeDeletion:
         assert response.status_code == 403
         # The UI renders this through getResponseError, so a bare "Forbidden" would leave
         # the user with a refusal that says nothing about why.
-        assert response.get_json()["error"] == _DIRECT_REPORT_MODE_DELETE_REFUSAL
+        assert response.get_json()["error"] == DIRECT_REPORT_MODE_DELETE_REFUSAL
         assert (parent / "listed_report").is_dir()
 
     def test_refuses_to_delete_a_listed_performance_report(
@@ -321,7 +329,7 @@ class TestDirectReportModeDeletion:
         )
 
         assert response.status_code == 403
-        assert response.get_json()["error"] == _DIRECT_REPORT_MODE_DELETE_REFUSAL
+        assert response.get_json()["error"] == DIRECT_REPORT_MODE_DELETE_REFUSAL
         assert (parent / "listed_report").is_dir()
 
     def test_refusal_does_not_reach_the_local_data_directory(
--- a/backend/ttnn_visualizer/tests/views/test_report_deletion.py
+++ b/backend/ttnn_visualizer/tests/views/test_report_deletion.py
@@ -4,9 +4,10 @@
 
 """Tests for the profiler and performance report ``DELETE`` routes.
 
-The listings these routes are paired with only ever read the local data
-directory, so every case here pins the blast radius of a delete to the single
-directory the client asked for.
+In upload/sync mode the listings these routes are paired with read the local data
+directory, so those cases pin the blast radius of a delete to the single directory
+the client asked for. In direct-report mode the listings read the TT-Metal tree
+instead, which these routes do not manage, so the delete is refused outright.
 """
 
 import shutil
@@ -23,6 +24,8 @@ from ttnn_visualizer.instances import (
     KEY_PROFILER_NAME,
 )
 from ttnn_visualizer.models import InstanceTable, RemoteConnection, ReportLocation
+from ttnn_visualizer.utils import create_path_resolver
+from ttnn_visualizer.views import _DIRECT_REPORT_MODE_DELETE_REFUSAL
 
 API = "/api"
 HOST = "yyzc-wh-05"
@@ -65,6 +68,39 @@ def _synced_remote_reports(app, directory_key, *names):
     parent = (
         Path(app.config["REMOTE_DATA_DIRECTORY"]) / HOST / app.config[directory_key]
     )
+    for name in names:
+        (parent / name).mkdir(parents=True)
+    return parent
+
+
+@pytest.fixture
+def direct_mode_app():
+    """A direct-report-mode app: the listings read ``$TT_METAL_HOME/generated``."""
+    tmpdir = tempfile.mkdtemp()
+    try:
+        settings = {
+            "TESTING": True,
+            "SQLALCHEMY_DATABASE_URI": f"sqlite:///{Path(tmpdir) / 'app.db'}",
+            "SERVER_MODE": False,
+            "TT_METAL_HOME": str(Path(tmpdir) / "tt-metal"),
+            "APP_DATA_DIRECTORY": tmpdir,
+            "REPORT_DATA_DIRECTORY": tmpdir,
+            "LOCAL_DATA_DIRECTORY": str(Path(tmpdir) / "local"),
+            "REMOTE_DATA_DIRECTORY": str(Path(tmpdir) / "remote"),
+        }
+        yield create_app(settings_override=settings)
+    finally:
+        shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+@pytest.fixture
+def direct_mode_client(direct_mode_app):
+    return direct_mode_app.test_client()
+
+
+def _tt_metal_reports(app, report_type, *names):
+    """Create reports where direct-report mode lists them, and return their parent."""
+    parent = create_path_resolver(app).get_base_report_path(report_type)
     for name in names:
         (parent / name).mkdir(parents=True)
     return parent
@@ -246,3 +282,61 @@ class TestProfilerReportDeletion:
         assert response.status_code in {400, 404}
         assert parent.is_dir()
         assert (parent / "keep_me").is_dir()
+
+
+class TestDirectReportModeDeletion:
+    """With ``TT_METAL_HOME`` set, the listings read a tree these routes cannot delete from.
+
+    Deleting resolved against the app's own data directory regardless, so every delete of
+    a listed report 404'd and the UI showed nothing at all. Refuse instead of 404ing
+    against a tree the client never saw listed.
+    """
+
+    def test_refuses_to_delete_a_listed_profiler_report(
+        self, direct_mode_app, direct_mode_client
+    ):
+        parent = _tt_metal_reports(direct_mode_app, "profiler", "listed_report")
+        _register_instance(direct_mode_app, "profiler-direct", {})
+
+        response = direct_mode_client.delete(
+            f"{API}/profiler/listed_report",
+            query_string={"instanceId": "profiler-direct"},
+        )
+
+        assert response.status_code == 403
+        # The UI renders this through getResponseError, so a bare "Forbidden" would leave
+        # the user with a refusal that says nothing about why.
+        assert response.get_json()["error"] == _DIRECT_REPORT_MODE_DELETE_REFUSAL
+        assert (parent / "listed_report").is_dir()
+
+    def test_refuses_to_delete_a_listed_performance_report(
+        self, direct_mode_app, direct_mode_client
+    ):
+        parent = _tt_metal_reports(direct_mode_app, "performance", "listed_report")
+        _register_instance(direct_mode_app, "perf-direct", {})
+
+        response = direct_mode_client.delete(
+            f"{API}/performance/listed_report",
+            query_string={"instanceId": "perf-direct"},
+        )
+
+        assert response.status_code == 403
+        assert response.get_json()["error"] == _DIRECT_REPORT_MODE_DELETE_REFUSAL
+        assert (parent / "listed_report").is_dir()
+
+    def test_refusal_does_not_reach_the_local_data_directory(
+        self, direct_mode_app, direct_mode_client
+    ):
+        """A same-named report in the app's own tree must survive the refusal too."""
+        local_parent = _local_reports(
+            direct_mode_app, "PROFILER_DIRECTORY_NAME", "listed_report"
+        )
+        _register_instance(direct_mode_app, "profiler-direct-local", {})
+
+        response = direct_mode_client.delete(
+            f"{API}/profiler/listed_report",
+            query_string={"instanceId": "profiler-direct-local"},
+        )
+
+        assert response.status_code == 403
+        assert (local_parent / "listed_report").is_dir()
--- a/backend/ttnn_visualizer/views.py
+++ b/backend/ttnn_visualizer/views.py
@@ -26,7 +26,11 @@ from ttnn_visualizer.csv_queries import (
     OpsPerformanceQueries,
     OpsPerformanceReportQueries,
 )
-from ttnn_visualizer.decorators import local_only, with_instance
+from ttnn_visualizer.decorators import (
+    local_only,
+    refuse_in_direct_report_mode,
+    with_instance,
+)
 from ttnn_visualizer.enums import ConnectionTestStates, StackSourceOrigin
 from ttnn_visualizer.exceptions import (
     AuthenticationFailedException,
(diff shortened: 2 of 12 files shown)


backend/ttnn_visualizer/tests/views/test_report_deletion.py
```diff
@@ -70,6 +73,39 @@ def _synced_remote_reports(app, directory_key, *names):
...
+@pytest.fixture
+def direct_mode_app():
+    """A direct-report-mode app: the listings read ``$TT_METAL_HOME/generated``."""
+    tmpdir = tempfile.mkdtemp()
+    try:
+        settings = {
+            "TESTING": True,
+            "SQLALCHEMY_DATABASE_URI": f"sqlite:///{Path(tmpdir) / 'app.db'}",
+            "SERVER_MODE": False,
+            "TT_METAL_HOME": str(Path(tmpdir) / "tt-metal"),
```
dcblundell, line 93: **Must fix:** the local-mode `app` fixture needs `"TT_METAL_HOME": None`

Commenting here because the affected fixture (`app`, lines 39–47) is unchanged by this PR. It doesn't pin `TT_METAL_HOME`, so `DefaultConfig` reads it from the environment (plus `create_app`'s `load_dotenv`) and the new refusal fires. I verified that 11 of the 14 tests in this file fail with `TT_METAL_HOME=/tmp/fake-metal` exported — including both traversal tests and `test_clears_the_active_report_when_it_is_the_one_deleted` — while the same 11 pass on `dev` with the variable set. That variable is exported on essentially every machine that profiles TT-Metal, so the branch turns the suite red for the repo's own developers with failures indistinguishable from a real regression. `test_settings.py` already applies this discipline for the same reason.

src/components/report-selection/LocalFolderPicker.tsx
```diff
@@ -62,13 +62,19 @@ const LocalFolderPicker = ({
     const isDisabled = disabled || loading || !items || items.length === 0 || !instance;
     const activePath = value;
     const activeName = value ? (valueLabel ?? value) : null;
-    const isServerMode = !!getServerConfig()?.SERVER_MODE;
+    const serverConfig = getServerConfig();
+    const isServerMode = !!serverConfig?.SERVER_MODE;
+    // Direct-report mode lists reports out of the TT-Metal tree, which the app neither created nor
+    // manages — the delete routes only ever reach the app's own data directory, so the control could
```
dcblundell: **Should fix:** this clause describes the behaviour the PR replaced

"the delete routes only ever reach the app's own data directory, so the control could never remove the report the row shows" was the reason before this change; the routes now refuse with 403 before resolving any path in this mode. The first clause (the app neither created nor manages the TT-Metal tree) is the durable reason and is still true. Reword the second to name the refusal, so the two sides of the gate read consistently.

src/components/report-selection/LocalFolderPicker.tsx
```diff
@@ -62,13 +62,19 @@ const LocalFolderPicker = ({
...
+    // never remove the report the row shows.
+    const isDirectReportMode = !!serverConfig?.TT_METAL_HOME;
```
dcblundell: **Should fix:** third inline copy of the direct-report-mode check

`!!getServerConfig()?.TT_METAL_HOME` now appears here, in `LocalFolderSelector.tsx:151` (the parent that renders this component), and in `routes/Home.tsx:19`. The backend names the concept once as `PathResolver.is_direct_report_mode`; the frontend names it nowhere, so the rule "`TT_METAL_HOME` present means no upload, no sync, no delete" is re-derived per call site and the next gate can easily land in two of three. Export an `isDirectReportMode()` predicate from `src/functions/getServerConfig.ts` beside `isServerModeEnabled` and use it at all three — the picker keeps owning the gate, it just stops re-deriving the condition.

tests/localFolderSelector.spec.tsx
```diff
@@ -439,3 +442,77 @@ it('deletes memory report and updates state', async () => {
...
+
+    // The failure is what the user sees — without this the delete is silent.
+    await waitFor(
+        () => expect(screen.getByTestId(TEST_IDS.TOAST_FILENAME).textContent).to.contain(DELETE_REFUSAL_MESSAGE),
+        WAIT_FOR_OPTIONS,
+    );
+}
+
+it('surfaces an error toast and keeps the report when the memory delete fails', async () => {
+    mockDeleteProfiler.mockRejectedValueOnce(new Error(DELETE_REFUSAL_MESSAGE));
```
dcblundell: **Should fix:** reject with an AxiosError shape, not a bare `Error`

Both new failure tests (here and line 494) exercise `getResponseError`'s `instanceof Error` branch, not the AxiosError-with-`{ error: … }` branch a real 403 from `_direct_report_mode_refusal` takes — so the contract that matters, backend `response_forbidden` body reaching the toast, is unpinned. Reject with `{ isAxiosError: true, response: { status: 403, data: { error: … } } }` instead.

Related: `DELETE_REFUSAL_MESSAGE` (line 445) is a hand-copied duplicate of the backend string and its comment claims it is what the backend returns, but nothing enforces the parity — the Python side correctly imports the constant, whereas this spec only needs *some* rejection message. It becomes a lie the first time the message is reworded.

src/components/report-selection/LocalFolderSelector.tsx
```diff
@@ -275,7 +284,13 @@ const LocalFolderOptions = () => {
     };
 
     const handleDeletePerformance = async (folder: ReportFolder) => {
-        await deletePerformance(folder.path);
+        try {
+            await deletePerformance(folder.path);
+        } catch (err: unknown) {
+            createToastNotification(PERFORMANCE_REPORT_DELETE_FAILED_TITLE, getResponseError(err), ToastType.ERROR);
```
dcblundell: **Should fix:** no success-path test for the performance delete

The memory delete has one (`localFolderSelector.spec.tsx:400`, "deletes memory report and updates state"), but nothing exercises `handleDeletePerformance` resolving through `invalidateQueries`, the `'Performance report deleted'` toast, and the active-report clear. This change added a new early `return` inside the `try`/`catch` on both handlers, so a misplaced `return` would only be caught on the memory side. Mirror the memory test.

backend/ttnn_visualizer/views.py
```diff
@@ -1014,6 +1032,10 @@ def _report_directory_to_delete(directory_name_key: str, report_name: str) -> Pa
 @with_instance
 @local_only
 def delete_profiler_report(profiler_name, instance: Instance):
+    refusal = _direct_report_mode_refusal()
```
dcblundell: **Should fix:** express the refusal as a decorator rather than a copied prologue

This three-line prologue is duplicated at line 1151, making it the third thing both delete routes must remember to include identically (alongside `@local_only` and `_report_directory_to_delete`). A `refuse_in_direct_report_mode` decorator beside `local_only` in `decorators.py` says it once, can't be half-applied, and fits the documented `@api.route → @with_instance → @local_only` stack.

While there: `_direct_report_mode_refusal` (line 1004) has no return annotation, unlike the `-> Path` sibling added in the same block, so mypy leaves its body unchecked. `-> Optional[Response]`.

src/components/report-selection/LocalFolderSelector.tsx
```diff
@@ -83,6 +83,9 @@ const directoryErrorStatus: ConnectionStatus = {
     message: 'Selected directory does not contain a valid report',
 };
 
+const MEMORY_REPORT_DELETE_FAILED_TITLE = 'Unable to delete memory report';
+const PERFORMANCE_REPORT_DELETE_FAILED_TITLE = 'Unable to delete performance report';
```
dcblundell: **Should fix:** export these titles so the specs stop hardcoding the copy

Both are module-private, so `localFolderSelector.spec.tsx:482` and `:508` assert on hand-copied duplicates of the same strings. The house pattern is an exported constant consumed by both — `FOLDER_SYNC_ERROR_TOAST_TITLE` (`src/functions/notifyFolderSyncError.ts`) and `ACTIVE_MEMORY_REPORT_TOAST_TITLE` (`src/definitions/notifyActiveReport.ts`), which this file already imports. Worth promoting the two success titles a few lines below (`'Memory report deleted'`, `` `Performance report deleted` ``) in the same pass so the pair is consistent.

While you're in here: `handleDeleteProfiler` and `handleDeletePerformance` are now near-identical 15-line bodies (try/catch → error toast → return → invalidate → success toast → clear active report). This change added the same new branch to both, which is the drift signal — the next fix will land in one. A small shared helper taking the varying parts would collapse them.

dcblundell (commented):
Combined review (DRY, TEST, PERF, STANDARDS, SECURITY, REQUIREMENTS lenses). All five acceptance criteria in #1821 are implemented, the gating is correctly mirrored on both sides, and no security or performance findings came up in scope.

One must-fix, and it's a one-line change: the local-mode `app` fixture in `test_report_deletion.py` doesn't pin `TT_METAL_HOME`, so the new 403 refusal makes 11 of that file's 14 tests fail on any machine with the variable exported. I verified those same 11 pass on `dev` with it set. The remaining comments are should-fix: a comment that now describes the pre-change behaviour, a third inline copy of the direct-report-mode check, two test-coverage gaps, and two pieces of duplication the change introduced.

Lenses that reported findings outside this diff (`settings.py`, `app.py`, `getServerConfig.ts`, `utils.py`) were reviewing already-merged `dev` work, so those are not raised here.

backend/ttnn_visualizer/tests/views/test_report_deletion.py
smountenay-tt: Confirmed and fixed in 69ba6da7 — thank you, this was a real regression.

Reproduced exactly as you described: `TT_METAL_HOME=/tmp/fake-metal uv run pytest .../test_report_deletion.py` gave 11 failed, 3 passed on this branch, and 14 passed on `dev`. The fixture now pins `"TT_METAL_HOME": None`, with the reason recorded in its docstring so the next person adding a case here knows why it is there. Passes both with and without the variable exported.

While verifying I ran the whole backend suite with it set: 7 other tests fail the same way (`test_instances_paths.py` ×2, `test_mlir.py` ×1, `test_remote_reports.py` ×4). Those fail identically on `dev`, so they predate this branch and I have left them alone rather than widen this PR — worth a follow-up issue if you agree.

src/components/report-selection/LocalFolderPicker.tsx
smountenay-tt: Fixed in 69ba6da7. You are right that the second clause described the behaviour this PR replaced. Now:

```
// Direct-report mode lists reports out of the TT-Metal tree, which the app neither created nor
// manages — the delete routes refuse it outright, so the control could never act on the row.
```

src/components/report-selection/LocalFolderPicker.tsx
smountenay-tt: Agreed on the substance, done in 69ba6da7 — with one deliberate deviation on placement I would like to flag.

The predicate exists and all three call sites use it. But I put it in `src/functions/isDirectReportMode.ts` rather than in `getServerConfig.ts`, because exporting it from `getServerConfig.ts` makes it worse than the status quo: that module is mocked wholesale by 15 specs (`vi.mock('../src/functions/getServerConfig', () => ({ default: ... }))`), so a named export from it is `undefined` inside every one of them. I tried it that way first and `serverMode.spec.tsx` broke immediately; the fix was to write `isDirectReportMode: () => !!getServerConfigMock()?.TT_METAL_HOME` into the mock — a fourth copy of the rule, now in test code, which is the thing your comment is about. Any future spec rendering a gated component would hit the same trap with a confusing "not a function".

In its own module it composes with whatever config a spec has already mocked, so no spec needs to restate the rule. This also matches the house shape for a single predicate (`isRemoteFolderActive.ts` — one file, default export). The module docstring records why it is not next to `isServerModeEnabled`. Happy to move it if you would still rather have it there.

tests/localFolderSelector.spec.tsx
smountenay-tt: Both fixed in 69ba6da7.

The rejections are now AxiosError-shaped, so the tests exercise the branch a real 403 takes:

```ts
const refusedDelete = () => ({
    isAxiosError: true,
    response: { status: HttpStatusCode.Forbidden, data: { error: SERVER_ERROR_DETAIL } },
});
```

This does discriminate: the object has no `.message`, so if `axios.isAxiosError` did not match it, `getResponseError` would fall through to `'An unexpected error occurred'` and the assertion would fail.

On the duplication — you are right that nothing enforced the parity the comment claimed. Rather than pretend to enforce it across the language boundary, I made the string an explicit stand-in (`SERVER_ERROR_DETAIL = 'Reports in the TT-Metal tree are not managed here'`) and said so: the backend's exact wording is pinned where it is defined, and what this spec pins is only the plumbing — whatever the server puts in `error` reaches the toast. It can no longer become a lie, because it no longer claims to be a copy.

src/components/report-selection/LocalFolderSelector.tsx
smountenay-tt: Fixed in 69ba6da7 — added 'deletes performance report and updates state', mirroring the memory test: the delete call, the `'Performance report deleted'` toast, and the row disappearing after `invalidateQueries`.

Two things fell out of writing it. `deletePerformance` was an inline `vi.fn()` in the module mock with no handle to drive, so it is now a hoisted mock that removes from the folder list like its profiler counterpart; and the performance list is now restored in `beforeEach` (the profiler one already was), which makes these tests order-independent — an earlier upload test pushes into that array.

backend/ttnn_visualizer/views.py
smountenay-tt: Fixed in 69ba6da7. `refuse_in_direct_report_mode` now lives beside `local_only` in `decorators.py` and both routes read `@api.route → @with_instance → @local_only → @refuse_in_direct_report_mode`.

It returns `response_forbidden(...)` rather than calling `abort` like `local_only` does, so the body keeps carrying the reason as `{"error": ...}` — the shape `getResponseError` reads on the client. That is noted in the docstring so nobody "fixes" it into an `abort` for symmetry.

On the annotation: the function you flagged is gone with the prologue, so I applied the point to the decorator instead — `Callable[..., Any] -> Callable[..., Any]` with an annotated wrapper, which gets the body type-checked. It is the only annotated decorator in the file; I took your comment as pointing in that direction rather than toward matching the unannotated ones.

src/components/report-selection/LocalFolderSelector.tsx
smountenay-tt: Both done in 69ba6da7.

All four titles are now exported from `src/definitions/notifyActiveReport.ts` — the two failure titles and the two success ones — and the specs assert against the imports instead of hand-copied copy. (`'Choose directory...'` was also sitting in the file four times, so it is a constant now too.)

On the duplication: agreed, and collapsed. There is now one `deleteReport(folder, options)` carrying the whole sequence, with the two handlers supplying the six values they differ by (`sendDelete`, `folderQueryKey`, `failedTitle`, `deletedTitle`, `isActive`, `clearActive`). The new early `return` you pointed at now exists once. Both paths are covered end to end — success and failure, memory and performance — which is what made the refactor safe to do.

src/components/report-selection/LocalFolderPicker.tsx
dcblundell: Agreed — keep it where you put it.

I checked the trap you describe and it's real: 17 specs mock this module as `{ default: … }` only, so a named export would be `undefined` in every one of them, and each spec rendering a gated component would have to restate the rule in its mock. That's the fourth copy the comment was trying to prevent, just moved into test code. `LocalFolderPicker.spec.tsx:200` setting `TT_METAL_HOME` on the config it already mocks and having the predicate compose with it is the outcome I was after.

My suggestion was also weaker than I framed it. `isServerModeEnabled` isn't a peer for this — it's a pure value normaliser taking an argument, not a config reader, and no component consumes it, which is why it never hit the mock trap. A single-predicate module matches the house shape anyway.

On the seven pre-existing `TT_METAL_HOME` failures: reproduced exactly those, and none of the three files are touched here or exercise the delete routes, so I agree they're out of scope for this PR. Filing a follow-up.

Resolving.

backend/ttnn_visualizer/tests/views/test_report_deletion.py
dcblundell: Filed the follow-up as #1869. I reproduced the same seven on `dev` at `cf6c86f9` in a clean worktree, so they're confirmed pre-existing and rightly out of scope here. The issue suggests an autouse fixture in `conftest.py` as the durable version, with direct-report mode opted into explicitly.

dcblundell (approved):

Issue #1862, opened 2026-08-11
Late deallocation in Operation Details communicates nothing without markers


dcblundell:
Follow-up from #963, which asked for the late-deallocation feature to be made noticeable and was resolved for the **Buffer Summary** only.

### Problem

`Mark late tensor deallocations` is a global switch, so it also applies in Operation Details, where PR #960 shipped the same pattern (`getChartData.ts`). Turning it on there still produces only an unexplained `/` Plotly hatch plus hover text: no marker, no count, nothing that says which tensor is being held or what it was last used by. That is exactly the complaint #963 opened with, so the feature now reads as inconsistent — one view explains itself, the other does not.

### What Buffer Summary got (#963)

- A glyph badge in the y-axis gutter on every row still holding a stale tensor, naming the tensors and their last consumer.
- A navigation rail of clickable dots, one per row where a tensor *starts* being held past its last use, scrolling the list to that row.
- A count beside the toggle, shown whether or not the overlay is on, so the switch advertises whether it is worth flipping.

### Suggested scope

Decide which of those readings make sense on the Operation Details memory plot (the count and per-tensor naming look most transferable; a navigation rail probably does not), then apply them so the global switch means the same thing in both views.

Issue #1517, opened 2026-05-20
Memory chart: Annotate top-N slowest ops on the timeline


aidemsined:
## Context

The memory chart shows L1 / DRAM usage across the op timeline. Today there's no perf signal on it, so correlating "memory peak" with "slow op" requires switching to the perf table and cross-referencing by op ID.

## Ask

When a perf report is loaded alongside the profiler report, annotate the top-N slowest ops on the memory chart timeline — a marker, vertical line, or labeled tick at each highlighted op position.

Lets users visually correlate cost peaks with memory pressure in one view.

## Suggested behaviour

- N is configurable (default ~10 or top 5% by duration)
- Marker on hover shows op ID, name, duration
- Click marker = jump to that op (select / scroll memory chart)

## Open questions

- Annotation style: vertical line, labeled tick, dot above the chart, colored bar segment?
- N default value and whether to expose a slider/input
- Should we also (or instead) annotate top-N highest-memory ops, or is that already visible enough?
- Behaviour when perf report isn't loaded

## Related

- #1512 Integrate perf data into device operations rendering
- #1186 Recommending improvements based on the data

PR #1864, opened 2026-08-11
Add late deallocation axis markers to the buffer summary

### LINKED ISSUE #963: I think we need to make this feature more noticeable. The pattern we're adding doesn't really communicate anything.
I think we need to make this feature more noticeable. The pattern we're adding doesn't really communicate anything. 

The below is an extremely rough concept, but I've added warning icons on the axis to indicate "Hey, there's something to look at here". Just an example of more helpful information appearing when this feature is engaged.

Screenshot 2025-11-07 at 10 56 10 AM

_Originally posted by @dcblundell in https://github.com/tenstorrent/ttnn-visualizer/issues/960#issuecomment-3503356155_

dcblundell:
## Summary

Makes late-deallocated tensors noticeable in the buffer summary. The hatch pattern alone didn't communicate anything, so this adds a gutter badge beside every affected row and a navigation rail that plots the rows worth jumping to, plus a count on the toggle.

- `LateDeallocationBadge` marks each row still holding a tensor past its last use, glyph pinned after the operation name so the icons align in a scannable column.
- `NavigationRail` is a memoised minimap of clickable dots down the side of the virtualized list, plotting run starts — the rows where a tensor *starts* being held late.
- `useLateDeallocationOverlay` is the single owner of the data: the per-row report the canvas hatches from, the rail's rows, and the count. It is deliberately the only caller of `useGetTensorDeallocationReportByOperation` in this subtree, because that hook's per-operation derivation isn't cached by React Query and a second call pays the full cost again.

Scope is the buffer summary only. The same global switch drives Operation Details, which gains nothing here — tracked separately in [#1862](https://github.com/tenstorrent/ttnn-visualizer/issues/1862).

Screenshot 2026-08-11 at 13 31 06

## Decisions worth reviewing

- **Markers follow the hatching, the rail doesn't.** Dotting every continuation row would draw a solid bar down the gutter and bury the rows where something actually changed, but leaving a hatched row with an empty gutter reads as a marker that went missing. So badges cover every affected row and the rail plots run starts only.
- **The count shows whether or not the overlay is engaged.** A switch that would reveal nothing is worth saying so before it's flipped, matching the existing Tensor List precedent.

## Provenance

This is a re-land. The work was pushed to `dev` unreviewed because its branch tracked `origin/dev`, then reverted in [#1863](https://github.com/tenstorrent/ttnn-visualizer/pull/1863). The single commit here reverts that revert; the tree is identical to the original `6e2ec102`.

## Test plan

- [x] `pnpm exec vitest run tests/lateDeallocation.spec.ts tests/useLateDeallocationOverlay.spec.tsx tests/BufferSummaryVirtualizedList.spec.tsx tests/BufferSummaryPlotControls.spec.tsx` — 91 passing
- [ ] Load a memory report, open Buffer Summary, toggle the deallocation report: badges appear beside hatched rows and the rail dots scroll to the right operations
- [ ] Rail tooltips stay anchored to their dots while scrolling the virtualized list

Fixes #963


src/functions/lateDeallocation.ts
```diff
@@ -0,0 +1,347 @@
...
+ * The caller owns the cache so it lives no longer than one derivation pass.
+ */
+export const getLateDeallocationReport = (
+    { tensorId, address, operationId, consumers }: LateDeallocationCandidate,
+    operationNamesById: Map<number, string>,
+    lastConsumerByTensorId?: Map<number, LastValidConsumer>,
+): TensorDeallocationReport | null => {
+    if (!tensorId || !consumers || consumers.length === 0) {
+        return null;
+    }
```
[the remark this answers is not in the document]
aidemsined: > [!CAUTION]
> Confirmed, and still unaddressed at `c5da210b` — `src/functions/lateDeallocation.ts:76` reads `if (!tensorId || !consumers || consumers.length === 0)`, so a tensor with id `0` is discarded as "missing".

`c5da210b` touches this file only from line 249 down (`getConsumerLabel`, `getLateDeallocationSummary`), so this line is untouched.

The consequence is worse than a dropped row. A tensor with id `0` genuinely held past its last use produces no hatch, no gutter badge, no rail dot, and no contribution to the count — it fails silently and indistinguishably from "there is nothing here", in a feature that exists precisely because the hatch pattern "doesn't really communicate anything" (#963). In the degenerate case where it is a report's only late deallocation, `hasLateDeallocations` at `BufferSummaryPlotControls.tsx:123` is false, so the switch is also disabled with a `0` badge over a report that does have a finding.

Suggested fix, and a narrowing that removes the question permanently:

```ts
if (tensorId === null || !consumers || consumers.length === 0) {
```

`LateDeallocationCandidate.tensorId` is `number | null` at `:56`, but the only producer (`:145`) passes `tensor.id`, which `src/model/APIData.ts:33` declares as a non-nullable `number` — so the nullable half of that type isn't earning its keep and could be dropped, making the guard unnecessary rather than merely correct.

One piece of internal evidence that this was an oversight rather than a decision: `tests/lateDeallocation.spec.ts:69` is titled "accepts operation id 0 as a real consumer", so zero-valued ids were reasoned about on the operation axis and pinned with a test. The same reasoning wasn't applied to the tensor axis, and the test at `:204` that looks like it covers this ("reports nothing for a tensor with no id or no consumers") passes `tensorId: null`, never `0` — so a `tensorId: 0` case would fail today.

src/definitions/TopNAnnotations.ts
```diff
@@ -54,9 +55,18 @@ export enum TopNAnnotationStatus {
...
+/**
+ * Inclusive bounds on the numeric input. The ceiling is the rail's dot capacity
+ * rather than a figure to taste: this rail plots ranks, and a rank can't be
+ * pooled into a neighbouring dot the way the late-deallocation rail's tensors
+ * are, so capping the input is the only way it avoids drawing dots that cover
+ * each other.
+ */
 export const TOP_N_COUNT_MIN = 1;
-export const TOP_N_COUNT_MAX = 50;
+export const TOP_N_COUNT_MAX = RAIL_MAX_DOTS;
```
aidemsined: > [!CAUTION]
> This drops a shipped feature's user-facing input bound by 40% — `RAIL_MAX_DOTS` is `Math.floor(600 / 20)` = **30**, where this was `50` — and it arrives in a commit titled "Code review round" with no mention in the PR body or the test plan.

The reasoning in the docstring above is sound in isolation: a rank can't be pooled into a neighbouring dot the way `coalesceLateDeallocationRunStarts` pools tensors, so capping the input is the only way the top-N rail avoids drawing dots that cover each other. But that is a behaviour change to #1517, not to #963, and reviewers approving this PR aren't being asked to approve it.

There is also a user-visible consequence that isn't covered anywhere. `topNAnnotationCountAtom` is `atomWithStorage` (`src/store/app.ts:111`), so anyone who already set 50 keeps it: `NumericInput` renders `value={topNCount}` (`BufferSummaryPlotControls.tsx:211`) and shows **50**, while `selectTopNAnnotations` clamps to 30 (`src/functions/topNAnnotations.ts:205`) and draws 30. `clampValueOnBlur` only corrects the field once the user focuses and leaves it, so until they touch it the input states a number the app won't honour.

Either restore `50` and handle dense top-N rails the way the late-deallocation rail does, or keep the cap and (a) say so in the body and on #1517, and (b) clamp the displayed value, e.g. `value={Math.min(topNCount, TOP_N_COUNT_MAX)}`.

Worth noting the existing test doesn't protect this: `tests/topNAnnotations.spec.ts`'s "offers no count the rail cannot draw" asserts `TOP_N_COUNT_MAX <= RAIL_MAX_DOTS`, which is tautological once one is defined as the other.

src/components/buffer-summary/BufferSummaryPlotControls.tsx
```diff
@@ -116,13 +132,37 @@ const BufferSummaryPlotControls = () => {
...
-                    }}
-                />
+                <div className='late-dealloc-control'>
+                    <GlobalSwitch
+                        label='Mark late tensor deallocations'
+                        checked={showDeallocationReport}
+                        onChange={() => {
+                            setShowDeallocationReport(!showDeallocationReport);
+                        }}
+                        disabled={!hasLateDeallocations}
```
aidemsined, line 142: > [!WARNING]
> This can strand the switch checked *and* disabled, leaving no way to turn it off from this surface.

`showDeallocationReportAtom` is shared with Operation Details (`OperationDetailsComponent.tsx:63`). Enable it there, then open a Buffer Summary whose L1 tab has zero late deallocations: `hasLateDeallocations` is false, so this renders disabled, but `checked={showDeallocationReport}` on line 138 is still true. The control shows as on and greyed out.

The top-N switch 60 lines below already solves exactly this, at line 204:

```tsx
checked={topNEnabled && !isTopNDisabled}
disabled={isTopNDisabled}
```

Mirroring that here (`checked={showDeallocationReport && hasLateDeallocations}`) would keep the two controls consistent, or alternatively don't disable while `showDeallocationReport` is true, or clear the atom when the count reaches zero.

The atom is non-persisted (`src/store/app.ts:46`), so this resets on reload — a within-session trap rather than a permanent one, which is why this isn't a blocker. No test covers the disabled-and-checked combination.

src/definitions/NavigationRail.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { CSSProperties, ReactNode } from 'react';
+
+// Rail geometry, mirrored from `$scrollable-height` and `$rail-dot-size` in
+// `BufferSummaryPlot.scss`. The cap below is arithmetic on those numbers and has
+// to be known before anything is laid out, so it can't be read back off the
+// element; the stylesheet stays the source of truth for what is drawn, which
+// means these two move with it.
+const RAIL_HEIGHT = 600;
+const RAIL_DOT_SIZE = 20;
```
aidemsined, line 13: > [!WARNING]
> These hand-mirror `$scrollable-height: 600px` and `$rail-dot-size: 20px` in `src/scss/components/BufferSummaryPlot.scss`, which is the split source of truth the repo's magic-layout-number convention exists to prevent.

The comment above is honest about the duplication and the pre-layout timing argument for why the value can't be read back off the element is fair. What makes it worth acting on now is that `c5da210b` put a user-facing input bound downstream of it: `TOP_N_COUNT_MAX = RAIL_MAX_DOTS = Math.floor(RAIL_HEIGHT / RAIL_DOT_SIZE)`. Editing a length in a stylesheet now silently changes the largest number a user can type into the top-N field, and nothing in the suite would notice — the assertion that looks like it guards this (`TOP_N_COUNT_MAX <= RAIL_MAX_DOTS`) is tautological once one is defined as the other.

Promoting both to CSS custom properties in `src/scss/_base.scss` and reading them through the `cssVar()` helper is the convention-compliant fix. If the pre-layout timing genuinely rules that out, a build-time assertion that the TS constants match the SCSS values would at least make a drift loud.

Separately, and smaller: `RAIL_MAX_DOTS` divides the full 600px, but `.rail-track` spans `top: $height-offset` (15px) to `bottom: 0`, so the usable rail is ~585px. Thirty 20px dots need 600px, so at the cap the last dots overlap slightly — combined with `translateY(-50%)` on `.rail-item`, a few pixels. Subtracting `$height-offset` from `RAIL_HEIGHT` would make the arithmetic match what's drawn. That's reasoned from the stylesheet, not measured in a browser.

aidemsined:
> [!WARNING]
> Two claims in the description are stale, and the description also undersells the change by hiding two real bug fixes.

**Provenance.** "The single commit here reverts that revert; the tree is identical to the original `6e2ec102`." I verified this: `git diff --stat 6e2ec102 b906ea25` is empty, so it is true of the first commit. It is not true of the head — `git diff --stat 6e2ec102 c5da210b` reports 11 files changed, +284/−20. Since the identical-tree claim is the argument for why the re-land is safe to take on trust, it's worth qualifying rather than leaving to a reader to discover that a second commit exists.

**Test count.** "91 passing" is plausible for the `b906ea25` snapshot; the head carries roughly nine more tests across six spec files (39 + 34 + 26 + 16 + 7 + 3 = 125 across the touched specs, of which the late-deallocation subset is smaller).

**Undersold.** Extracting the derivation out of `useAPI.tsx` fixed two live correctness bugs in the old inline implementation, neither of which is mentioned:

- a `sort()` on stringified ids, so operation 2 outranked operation 10 when picking the last consumer
- a missing sentinel guard, so tensors consumed only by deallocates were reported as late-deallocated with `lastConsumerOperationId: -1`

Both now have named regression tests. These are user-visible fixes and they're a better argument for the PR than the refactor framing currently in the body.

**Scope.** Worth adding the `TOP_N_COUNT_MAX` reduction (50 → 30) to the description, per the separate comment on `src/definitions/TopNAnnotations.ts` — it changes behaviour for #1517, which a reader of this body wouldn't expect from a #963 PR.

src/functions/lateDeallocation.ts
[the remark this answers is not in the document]
dcblundell: Fixed in `f79a0f00`: guard is now `tensorId === null`, and there is a test that accepts tensor id `0` as a real tensor (mirroring the operation-id-0 case). Left `number | null` on the candidate for the explicit missing-id path the tests exercise.

src/definitions/TopNAnnotations.ts
dcblundell: Restored `TOP_N_COUNT_MAX = 50` in `f79a0f00` (no longer tied to `RAIL_MAX_DOTS`), and the NumericInput now displays `Math.min(topNCount, TOP_N_COUNT_MAX)` so a persisted out-of-range value cannot lie about what the app will honour. Also kept the `selectTopNAnnotations` clamp to `TOP_N_COUNT_MAX`.

src/definitions/NavigationRail.ts
dcblundell: Deferring for a follow-up. `TOP_N_COUNT_MAX` is back to `50` and no longer derived from these constants, so the user-facing input is not downstream of the SCSS/TS mirror anymore. Happy to take the CSS custom-property (or build-time assertion) cleanup and the `$height-offset` arithmetic separately.

src/components/buffer-summary/BufferSummaryPlotControls.tsx
dcblundell: Fixed in `f79a0f00`: mirrored the top-N pattern with `checked={showDeallocationReport && hasLateDeallocations}` so the switch cannot sit checked-and-disabled.

PR #1866, opened 2026-08-11
Add performance-report coverage to the CI smoke tests (#1859)

### LINKED ISSUE #1859: Smoke tests have no performance-report coverage
## Summary

The CI smoke tests exercise memory (profiler) reports only. No performance report is ever uploaded, so every performance endpoint is unverified on `dev` and in release wheels.

## Detail

`.github/workflows/smoke-tests.yml` runs `scripts/smoke_test.py` across Python 3.10–3.14. That test:

- uploads the three archives in `DEMO_REPORT_ZIPS` (`scripts/demo_reports.py`), all of which are memory reports containing `db.sqlite`;
- walks `MAIN_TAB_NAMES` — Operations, Tensors, Buffers.

There is no `profile_log_device.csv` or `ops_perf_results*` file anywhere in the repository, so `POST /api/local/upload/performance` is never called and none of these routes are covered:

- `GET /api/performance`
- `GET /api/performance/device-log`, `/device-log/raw`, `/device-log/meta`
- `GET /api/performance/perf-results`, `/perf-results/raw`, `/perf-results/report`
- `GET /api/performance/npe/manifest`, `/npe/timeline`

A regression in performance ingestion or any of these handlers would reach a release unnoticed.

## Suggested fix

1. Commit a small demo performance report under `demo-reports/` — a folder with `profile_log_device.csv` plus one `ops_perf_results*.csv` is enough to satisfy `validate_files` (`PERFORMANCE_REPORT_REQUIRED_FILES` / `PERFORMANCE_OPS_PERF_PREFIX` in `backend/ttnn_visualizer/utils.py`).
2. Extend `scripts/smoke_test.py` to upload it and visit the performance views, in the same shape as the existing memory-report walk.

## Related observations

- All three committed demo reports predate the `report_metadata` and `errors` tables, so `GET /api/report-metadata` and `GET /api/errors` answer 422 for every fixture in CI. Those paths are only covered in their degraded form; a fixture with the tables present would be worth adding too.
- While uploading a performance report over HTTP, note that `create_performance_files` derives `performance_path` from `paths[0].parent` — the parent of the *first* file saved. A report containing a subdirectory (for example `npe_viz/`) binds the instance to that subdirectory when the subdirectory's files happen to arrive first, after which every performance endpoint returns 404 or 500. Browser `FileList` ordering is not specified, so this looks worth hardening independently — probably by resolving the report root from the located `profile_log_device.csv` rather than from upload order.

smountenay-tt:
Closes #1859. Follow-up split out as #1867.

## Problem

`scripts/smoke_test.py` only drove the memory-profiler half of the app — upload via `local-profiler-upload`, walk Operations / Tensors / Buffers. `POST /api/local/upload/performance` was never called, so every `/api/performance/*` route went unverified on `dev` and in release wheels.

## Two corrections to the issue

**The demo archives already contain performance reports.** All three `demo-reports/*.zip` carry a full `local/performance-reports/<NAME>/` subtree — `backend/ttnn_visualizer/tests/test_perf_report.py:18` already reads one. What doesn't exist is a *loose* CSV. So no new demo report was needed; but their device logs are 3.5 / 20 / 36 MB, and they're user-facing artifacts rather than CI inputs. This PR adds a purpose-built **~55 KB** fixture instead, so CI cost stays flat.

(The issue also names `scripts/demo_reports.py`; the real location is `scripts/smoke_test_helpers.py`. Both noted on #1859 so they aren't lost when it closes.)

**The `paths[0].parent` observation is a live bug, not a theoretical one.** `performance_path` was derived from whichever file led the multipart body, so a report containing `npe_viz/` binds the instance to that subdirectory whenever one of its files arrives first, after which every performance endpoint 404s or 500s.

The first fix anchored on the located `profile_log_device.csv`, but review showed that was still wrong: `validate_files` skips its depth check when `folderName` is supplied, so a part named `<anything>/profile_log_device.csv` lands the log a level deeper and makes the root's final segment fully caller-chosen — and that segment is the hosted-session scoping key in `get_performance_data_list`. It now comes from the destination the server chose (`target_directory / paths[0].relative_to(target_directory).parts[0]`), which keeps the `SERVER_MODE` timestamp prefix and stays consistent with the `performance_name` written alongside it.

## Coverage

All nine routes #1859 lists are now asserted **in CI**, not by hand:

| Route | How it's covered |
|---|---|
| `/api/performance` | SPA-driven, asserted 2xx via the response tracker |
| `/api/performance/perf-results/report` | SPA-driven |
| `/api/performance/device-log/meta` | SPA-driven |
| `/api/performance/npe/manifest` | SPA-driven |
| `/api/performance/npe/timeline` | SPA-driven (reached by clicking the NPE link) |
| `/api/performance/device-log` | direct request — no SPA caller |
| `/api/performance/device-log/raw` | direct request — no SPA caller |
| `/api/performance/perf-results` | direct request — no SPA caller |
| `/api/performance/perf-results/raw` | direct request — no SPA caller |

The last four have no caller anywhere in `src/` (`views.py` even marks `/device-log/raw` as no longer used), so no browser-driven walk can reach them; they're asserted directly against the uploaded instance rather than quietly left off the list.

`ApiErrorTracker` now records every observed status per path. Flagging only 5xx was not enough — the bug above surfaces as a 404 — but blanket 4xx failure would be wrong, since the perf page legitimately 404s profiler-scoped requests. Instead each scenario asserts the routes it *claims* to cover were seen answering 2xx.

## Assertions that can actually fail

Both new assertions were verified to fail, not just to pass:

- **NPE view.** `assert_no_error_ui` only checks `#error-page`, but every NPE failure mode renders `NPEProcessingStatus` instead — so a 404, a 422, or an unparseable timeline all printed a tick. Corrupting the fixture's `.npeviz.zst` now fails the leg with `npe-processing-unhandled-error`; before, it passed.
- **Endpoint coverage.** Adding a route the SPA never calls makes `assert_answered_ok` fire with `(never requested)`.
- **Report root.** The new upload test is red against the previous derivation (`.../mine/1770000000_victimreport`).
- **Fixture join.** Mutating the manifest's `global_call_count` fails `test_manifest_entries_join_a_report_row`.

## Fixture design notes

- **Rows are trimmed; columns never are.** The report endpoint hands the CSV to the pinned external `tt-perf-report`, which reads far more columns than this repo names.
- **The first two lines of `profile_log_device.csv` are verbatim.** Line 1 is what `/device-log/meta` regexes; `LocalCSVQueryRunner` reads with `skiprows=1` then overwrites the header positionally.
- **`npe_viz/manifest.json` is generated from the retained rows.** `PerfTable` renders the NPE link only when a manifest entry's `global_call_count` matches a report row, and the smoke test *clicks* that link — that click is what proves the join survived ingestion.
- No `tracy_profile_log_host.tracy`, keeping the tracy-less ingestion path covered.

`backend/ttnn_visualizer/tests/test_smoke_perf_fixture.py` asserts these invariants against the committed bytes in about a second, so a bad regeneration fails long before CI reaches the browser. The generator now preserves the fixture README, so the documented regenerate command no longer deletes its own provenance notes.

## Structure

- `scripts/report_fixtures.py` — Playwright-free home for the zip extraction and its traversal guard, which had been duplicated between the smoke helpers and the generator and had already diverged on the guard. Also the single home for the repo-root and fixture-path constants.
- `upload_report` + `UploadKind` replaces the two line-for-line upload clones; `tracked_page` replaces the per-scenario page/tracker boilerplate and owns the tracker assertion, so a new scenario can't forget it. `exact=True` now applies to every nav lookup.
- `scripts/tests/` covers the generator's three deliberate guards (output confinement, zip slip, manifest basename collapse), so the `.cycodeignore` entry rests on tested behaviour. `testpaths` in `pyproject.toml` now owns the suite list — CI passed an explicit path, so a suite outside the backend package would never have run.
- `PerfTable`'s NPE button was icon-only with no accessible name; it now carries an `aria-label` mirroring its tooltip, and `PerfTable.spec.tsx` pins the manifest join in both directions including the deliberate `global_call_count: 0` non-match.

## Verification

- `pytest` — 838 pass (both suites, via `testpaths`).
- `pnpm test` — 1582 pass. `pnpm lint`, `pnpm lint:spdx`, `mypy` clean.
- Full local smoke run against a production build: three memory-report passes plus the performance pass, all nine routes asserted.


backend/ttnn_visualizer/views.py
```diff
@@ -1438,7 +1438,24 @@ def create_performance_files():
...
+    # `validate_files` has already established the device log is in the payload;
+    # the fallback is defensive only.
+    report_root = next(
+        (
+            path.parent
+            for path in paths
+            if path.name in PERFORMANCE_REPORT_REQUIRED_FILES
+        ),
+        paths[0].parent,
+    )
```
dcblundell: **Must fix:** the anchored `report_root` is still caller-influenceable, and its final segment is the `SERVER_MODE` scoping key

Anchoring on the device log's `.parent` is the right direction, but `validate_files` skips its depth check whenever `folderName` is supplied (`file_uploads.py:41`), so a request with `folderName=mine` and a part named `1770000000_victimreport/profile_log_device.csv` lands the log one level deeper and makes `report_root.name` fully caller-chosen. `get_performance_data_list` scopes hosted sessions by `Path(performance_path).name` (`views.py:1061`) and joins it back onto the shared performance directory, so that name becomes an existence oracle over other sessions' reports — and `PUT /api/instance` accepts an arbitrary `performance_path` with no session check, which converts a listed name into a read.

Not newly introduced (`paths[0].parent` was equally controllable, since the caller picks which part leads), but this diff owns the line and the fix satisfies the stated goal anyway: derive the root from the known destination rather than a payload scan, e.g. `report_root = target_directory / paths[0].relative_to(target_directory).parts[0]`. That still skips a leading `npe_viz/` file, keeps the server-chosen timestamp prefix intact, and keeps `performance_path` consistent with the `performance_name` written alongside it. Worth a regression test posting a nested device log with `folderName` set.

scripts/smoke_test.py
```diff
@@ -103,19 +125,118 @@ async def exercise_main_tabs(page: Page) -> None:
...
+            "No NPE link in the performance table. The fixture's npe_viz/manifest.json "
+            "should join a report row by global_call_count — regenerate it with "
+            "scripts/generate_smoke_perf_fixture.py if the report shape changed."
+        ) from exc
+
+    await npe_link.click()
+    await page.wait_for_url("**/npe/**", timeout=TAB_TIMEOUT_MS)
+    await wait_for_view_to_settle(page)
+    await assert_no_error_ui(page)
+    print("✅ NPE timeline loaded without errors")
```
dcblundell: **Must fix:** this leg cannot fail, so it doesn't cover what the fixture was designed to prove

`assert_no_error_ui` only checks `#error-page` (`smoke_test_helpers.py:147`), but `NPE.tsx:96` renders `NPEProcessingStatus` for every failure mode — invalid data, invalid JSON, empty trace — not the router error page. `ApiErrorTracker` also ignores status < 500 (`smoke_test_helpers.py:93`). So a 404 or 422 from `/api/performance/npe/timeline`, or a timeline the SPA rejects as invalid, still prints "✅ NPE timeline loaded without errors".

That's the regression this leg exists to catch, and the whole `global_call_count` join design in the fixture serves it. `TEST_IDS.NPE_VIEW` and `NPE_WINDOWED_VIEW` already exist, so a positive assertion is enough: wait for one of them to be visible, and/or assert no `npe-processing-*` testid is present.

scripts/smoke_test.py
```diff
@@ -103,19 +125,118 @@ async def exercise_main_tabs(page: Page) -> None:
...
+
+
+async def exercise_performance_tab(page: Page) -> None:
+    """Open the Performance view and follow its NPE timeline link.
+
+    Visiting the tab covers the perf-results report, device-log meta and NPE
+    manifest endpoints. The NPE link is rendered only when a manifest entry's
+    `global_call_count` matches a report row, so clicking it — rather than
+    navigating to the timeline URL directly — is what proves that join survived
+    ingestion, on top of covering `/api/performance/npe/timeline`.
```
dcblundell, line 293: **Must fix:** four of the nine routes #1859 enumerates are still uncovered, and the PR's evidence for them is a manual check

This docstring is honest about scope (it claims three endpoints), but the PR body's "all nine endpoints returned real data" rests on a local `curl` pass by hand — which is the thing #1859 wants replaced by CI. Grepping `Endpoints.PERFORMANCE` across `src/` returns only `/api/performance`, `/perf-results/report`, `/device-log/meta`, `/npe/manifest` and `/npe/timeline`, so `/device-log`, `/device-log/raw`, `/perf-results` and `/perf-results/raw` have no SPA caller and no browser-driven test can reach them. `views.py:1229` even carries a pre-existing `# this is no longer used atm` above `/device-log/raw`.

Cheapest fix: after `upload_performance_report`, assert the four directly from the same browser context — `page.request.get(f"{BASE_URL}/api/performance/device-log?instanceId=…")` and friends, checking `status < 400`. Reuses the uploaded instance, needs no UI, four requests.

The alternative is to descope them deliberately: amend #1859 to mark those four as UI-dead (needing direct HTTP assertions or deletion) and say so here. Either is fine, but as it stands `Closes #1859` closes an issue whose route list reads as a checklist with 4/9 unticked.

scripts/smoke_test_helpers.py
```diff
@@ -29,6 +29,13 @@
 
 MAIN_TAB_NAMES = ("Operations", "Tensors", "Buffers")
 
+# Purpose-built performance report, small enough to upload on every matrix leg.
+# The demo archives above carry performance reports too, but they are user-facing
+# artifacts with 3.5-36 MB device logs. See the fixture's README.
+PERFORMANCE_FIXTURE_DIR = (
+    REPO_ROOT / "scripts" / "fixtures" / "smoke-performance-report"
+)
```
dcblundell: **Should fix:** the error tracker is blind to the failure mode this PR is about (true location: `smoke_test_helpers.py:93`, outside the diff — anchored here)

`ApiErrorTracker._on_response` records only status ≥ 500, so the whole performance walk misses 404s and 422s — exactly what the `create_performance_files` fix guards against ("every performance endpoint 404s or 500s").

Tightening to all 4xx would be wrong (the perf page legitimately 404s profiler-scoped requests on a page with no profiler report). Better: have the tracker record observed statuses and assert each claimed endpoint was seen with a 200. That makes the coverage claim checkable rather than aspirational, and gives you the mechanism to cover the four routes with no SPA caller in one move.

scripts/fixtures/smoke-performance-report/README.md
```diff
@@ -0,0 +1,60 @@
...
+- **`npe_viz/manifest.json` is generated from the retained rows.**
+  `PerfTable` only renders the NPE launch button when a manifest entry's
+  `global_call_count` matches a report row's, and the smoke test clicks that
+  button to reach `/api/performance/npe/timeline`. The manifest must also stay
+  schema-clean per `src/schemas/npe-manifest.schema.json`
+  (`additionalProperties: false`) — note the source manifest in the demo report
+  has a typo'd `_global_call_count` key that would fail that schema.
+- **No `tracy_profile_log_host.tracy`.** It is optional per
+  `PERFORMANCE_REPORT_REQUIRED_FILES`, and omitting it keeps the tracy-less
+  ingestion path covered.
```
dcblundell, line 55: **Should fix:** these four invariants are documented as things that rot silently, but nothing asserts them outside the Playwright matrix

The manifest/report `global_call_count` join, the verbatim first two lines of the device log, the retained column set, and manifest schema-cleanliness are all load-bearing, and the only thing checking them needs a wheel build plus a built SPA and runs as five matrix legs. A pytest over the committed fixture would catch a bad regeneration in seconds instead.

`backend/ttnn_visualizer/tests/test_perf_report.py:17` already shows the pattern (`parents[3]` plus `OpsPerformanceReportQueries.generate_report`), and this fixture is a pre-extracted 55 KB version of the same report — so the pytest version is cheaper and faster than the demo-zip test that already exists. Suggested assertions: `is_valid_performance_report_dir` accepts the fixture; `generate_report` yields a non-empty report; the `ARCH:` / `CHIP_FREQ[MHz]:` regexes at `views.py:1255` match line 1; every manifest `global_call_count` equals some report row's; the manifest validates against `src/schemas/npe-manifest.schema.json`.

scripts/fixtures/smoke-performance-report/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Provenance
+
+Distilled from `demo-reports/n300-llama.zip`
+(`local/performance-reports/DEMO_N300-LLAMA/`) by:
+
+```
+uv run python scripts/generate_smoke_perf_fixture.py \
+    --source demo-reports/n300-llama.zip \
+    --output scripts/fixtures/smoke-performance-report
+```
```
dcblundell, line 20: **Should fix:** following this command deletes the README that documents the fixture

`--output` points at the fixture directory itself, and `build_fixture` does an unconditional `shutil.rmtree(output)` before recreating it (`generate_smoke_perf_fixture.py:308-310`). The file half-anticipates this — the sibling-plus-`diff` recipe below excludes `README.md` — but the headline command is the destructive one, so the first person to regenerate loses the provenance notes.

Either make the sibling-plus-`diff` recipe the documented way to regenerate, or have the generator preserve a `README.md` it finds in the output directory.

scripts/generate_smoke_perf_fixture.py
```diff
@@ -0,0 +1,388 @@
...
+    `--output scripts/fixtures` cannot wipe every fixture at once.
+    """
+    resolved = output.expanduser().resolve()
+
+    if resolved == FIXTURE_ROOT or not resolved.is_relative_to(FIXTURE_ROOT):
+        raise ValueError(
+            f"--output must name a directory inside {FIXTURE_ROOT}, got {resolved}"
+        )
+
+    return resolved
```
dcblundell, line 95: **Should fix:** the three deliberate guards in this script have no tests, and a test placed beside the script wouldn't run

`_resolve_output_dir` is what stands between an `--output` typo and the `shutil.rmtree` at line 309; the zip-slip check (line 121) and the manifest basename collapse (line 189) were both added as hardening in "Confine the fixture generator's output path and sanitize manifest filenames". The repo's own convention is that sanitisation fixes get a regression test, and the `.cycodeignore` entry would then rest on tested behaviour rather than on a comment asserting it.

Worth noting as part of the fix: CI runs `pytest backend/ttnn_visualizer` (`.github/workflows/lint-and-test.yml:93`) and `flask:test` is scoped the same way, so the invocation or `testpaths` needs widening. Minimum cases: `--output` = the fixtures root (reject), a `..` escape (reject), a valid child (accept); a crafted zip entry escaping the extraction dir raises; a manifest entry with `"file": "../../etc/passwd"` is skipped rather than copied.

scripts/generate_smoke_perf_fixture.py
```diff
@@ -0,0 +1,388 @@
...
+        for name in archive.namelist():
+            if not name.startswith(prefix) or name.endswith("/"):
+                continue
+            target = (report_dir / name[len(prefix) :]).resolve()
+            if not target.is_relative_to(report_dir_resolved):
+                raise ValueError(f"Zip entry escapes extraction directory: {name}")
+            target.parent.mkdir(parents=True, exist_ok=True)
+            target.write_bytes(archive.read(name))
+
+        return report_dir
```
dcblundell: **Should fix:** this is a second copy of `extract_profiler_report_dir`, and the copies have already diverged on the guard

`smoke_test_helpers.py:100-142` does the same thing end to end — find the single marker entry, derive the prefix with `rsplit('/', 1)[0]`, skip directory entries, resolve each target, `is_relative_to` containment check, `mkdir(parents=True)`, write — differing only in marker filename (`profile_log_device.csv` vs `db.sqlite`). The helper additionally validates the derived directory name against `.`/`..`; this copy doesn't. Duplicating a traversal guard is the worst case for copy-paste, because the next fix lands in one copy only.

`smoke_test_helpers.py` imports Playwright, which is presumably why it wasn't reused here. Moving the zip logic into a Playwright-free module (e.g. `scripts/report_fixtures.py`) taking the marker filename as a parameter lets both callers share it, and would also collapse the separately-computed `REPO_ROOT` / fixture-path constants (here at lines 48-52 vs `smoke_test_helpers.py:21,35`) — today, moving the fixture would silently leave this containment root pointing at the old tree.

scripts/smoke_test.py
```diff
@@ -103,19 +125,118 @@ async def exercise_main_tabs(page: Page) -> None:
...
+        status = page.get_by_test_id("local-performance-status")
+        status_text = (
+            await status.text_content() if await status.count() > 0 else "unknown"
+        )
+        raise TimeoutError(
+            f"Performance tab did not become enabled after upload ({status_text})"
+        ) from exc
+
+    await assert_no_error_ui(page)
+    print(f"✅ Uploaded performance report from {report_dir.name}")
```
dcblundell: **Should fix:** line-for-line clone of `upload_profiler_report`, and the drift has already started

Same Reports click, same `wait_for_url(HOME_URL)`, same `set_input_files`, same `to_be_enabled` wait, same status-test-id fallback, same `TimeoutError` shape, same print (`smoke_test.py:76-97`). Only four values vary: upload test id, status test id, gate button, label. Evidence it will drift: the `exact=True` strict-mode fix in d5962758 landed only on this path — `upload_profiler_report:84` and `exercise_main_tabs:122` still use loose `get_by_role(name=…)` matching, where the same reasoning applies.

Collapse into one `upload_report(page, report_dir, *, upload_test_id, status_test_id, gate_button, label)` (or a small dataclass per upload kind) and apply `exact=True` uniformly.

Same pattern one level up: `smoke_test_performance_report` repeats the per-scenario boilerplate from `smoke_test_report_tabs` — `new_page`, tracker + `attach`, `goto` + `networkidle` with the same two 10 s timeouts, the trailing `if tracker.errors: raise`, and `finally: page.close()`. An `@asynccontextmanager tracked_page(browser, label)` that asserts the tracker on exit removes both copies and means a new scenario can't forget the check.

src/components/performance/PerfTable.tsx
```diff
@@ -223,6 +223,7 @@ const PerformanceTable = ({
                                 onClick={() => navigate(`${ROUTES.NPE}/${manifestRecord.file}`)}
                                 variant={ButtonVariant.MINIMAL}
                                 className='graph-button'
+                                data-testid={TEST_IDS.PERF_NPE_LINK}
```
dcblundell, line 227: **Should fix:** the testid is needed here because the button has no accessible name

This is an icon-only button with a `Tooltip` and no `aria-label`, so screen-reader users get an unlabelled button in the Global Call Count column — and the smoke test has to reach for a testid where the sibling tensor-drawer button (line 191) is findable by `aria-label`. Adding `aria-label={`Launch NPE timeline for ${row.raw_op_code}`}`, mirroring the tooltip text as line 191 does, fixes the affordance and lets both the smoke test and future unit tests use a role query.

While here: this new testid has no unit coverage, and `tests/PerfTable.spec.tsx:78` already mocks `useGetNPEManifest`, so pinning the join is short — a manifest entry matching a row renders the button, a non-matching one doesn't, and `global_call_count: 0` doesn't (the `parseInt(…) || -1` at line 211 means 0 never matches, which is deliberate per the comment but currently unpinned in either direction).

scripts/fixtures/smoke-performance-report/README.md
```diff
@@ -0,0 +1,60 @@
+# Smoke test performance report fixture
+
+The performance report `scripts/smoke_test.py` uploads to cover
+`POST /api/local/upload/performance` and the `/api/performance/*` routes (#1859).
```
dcblundell, line 4: **Should fix:** the deferred `report_metadata` / `errors` observation is untracked (about the PR as a whole — anchored on the #1859 reference)

I searched the repo and no follow-up issue exists for it, while `Closes #1859` will close the only record. The rationale for deferring is sound and the criterion's wording ("would be worth adding too", under *Related observations* rather than *Suggested fix*) is soft enough that it shouldn't block merge — it just shouldn't be lost. Open the follow-up, link it from the PR body, and keep `Closes #1859`.

Two related bits of spec maintenance worth a comment on #1859 before it closes, since the PR body identifies both but the issue will close carrying them: (a) "There is no `profile_log_device.csv` or `ops_perf_results*` file anywhere in the repository" — `demo-reports/n300-llama.zip` contains one at ~20 MB, just inside an archive; (b) the issue names `scripts/demo_reports.py` as the home of `DEMO_REPORT_ZIPS`, which actually lives in `scripts/smoke_test_helpers.py:24`.

dcblundell (commented):
Combined review across six lenses (maintainability, tests, performance, conventions, security, requirements conformance). Comments below cover the Must fix and Should fix findings only; the Consider/nit items are held back.

Three things to decide rather than just fix:

- **Coverage vs. the issue's checklist.** Only 5 of the 9 `/api/performance/*` routes #1859 names are reachable from a browser-driven test, because the other 4 have no caller anywhere in `src/`. Either assert them directly via `page.request.get`, or amend #1859 to descope them as UI-dead. Both are defensible; silently closing with 4/9 unticked isn't.
- **Fixture location.** The `demo-reports/` → `scripts/fixtures/` divergence from the issue's "Suggested fix" is the right call — the issue is what should be amended, not the code. One stated justification doesn't hold up though: `pyproject.toml`'s `where = ["backend"]` excludes `demo-reports/` just as completely, so the wheel argument doesn't distinguish the two locations. The CI-input/artifact distinction and the size delta are the load-bearing reasons.
- **`performance_path` derivation.** The hardening is correct for every path the app itself drives (I traced the plain and `MALWARE_SCANNER` branches, the Safari `folderName` branch, and the Chromium/Firefox depth check), and it now mirrors `create_profiler_files`. The residual hole is narrow and pre-existing, but has a real consequence under `SERVER_MODE` — details inline.

Everything else checked out: the regression test asserts exactly the behaviour #1859 asks for including the triggering order, the fixture design notes are accurate about the `skiprows=1` / positional-header and `tt-perf-report` column constraints, and I found no convention, data-fetching or state-management violations.

dcblundell (approved):

PR #1873, opened 2026-08-13
Strengthen the environment variable override mechanism

### LINKED ISSUE #1857: override_with_env_variables only reaches the concrete config class, so most settings can't be overridden late
Follow-up from #1850, which fixed the *typing* half of this loop. This is the *reach* half.

## The defect

`DefaultConfig.override_with_env_variables()` iterates `self.__class__.__dict__` — the concrete class's **own** dict, which excludes everything inherited. `Config._determine_config()` always returns a subclass instance, and those subclasses declare almost nothing:

```console
DevelopmentConfig.__dict__ public keys: []
ProductionConfig.__dict__  public keys: ['DEBUG', 'TESTING']
TestingConfig.__dict__     public keys: ['DEBUG', 'TESTING']
```

So in development the loop is a **no-op**, and in production/testing it overrides `DEBUG` and `TESTING` only. The other ~46 settings on `DefaultConfig` are never visited.

This mostly doesn't bite, because every environment-backed setting is also read via `os.getenv` in the class body. It bites when the environment changes **after** import, which is exactly what `main()` does:

```python
os.environ["HOST"] = "0.0.0.0"
os.environ["SERVER_MODE"] = "true"   # --server
```

The class body has already run by then, so `Config()` returns a config with `SERVER_MODE` still `False`. `main()` compensates by patching the object by hand:

```python
if args.server:
    config.HOST = "0.0.0.0"
    config.SERVER_MODE = True
```

That patch exists because the loop doesn't work. It also only fixes the launching process — correctness in the workers comes from gunicorn being a fresh subprocess that re-imports the module with the mutated environment. `_AllowedOrigins` is a descriptor for the same underlying reason.

## Why it wasn't fixed in #1850

Widening the reach without fixing the typing would have exposed every boolean and integer setting to the stringification bug at once. With #1850 landed, coercion is in place and widening is now safe to consider on its own.

## What a fix needs to cover

> **Amended 2026-08-10.** A bullet asking for per-setting validation to be re-run has been **struck** — #1850 landed that mechanism rather than leaving it here. `_coerce_env_value` dispatches on the declared type, and `_ENV_PARSERS` registers the two settings that parse more richly than theirs: `SSH_DEFAULT_PORT` (range-checked, and reported rather than silently defaulted on the override path) and `MAX_CONTENT_LENGTH` (empty means no limit). A value a setting can't represent is logged and discarded, including one whose declared type the loop has no coercion for. This issue's risk review therefore scopes to *widening the loop*, not to designing validation it would find already written.

1. **Walk the MRO** (`reversed(type(self).__mro__)`) rather than one class's `__dict__`, so subclass declarations still win.
2. **Decide what should be overridable at all.** The MRO includes things that are derived rather than configured — `APPLICATION_DIR`, `GUNICORN_BIND`, `SQLALCHEMY_DATABASE_URI`, `SQLALCHEMY_ENGINE_OPTIONS` (a dict, with no sensible string coercion), `LOCAL_DATA_DIRECTORY`/`REMOTE_DATA_DIRECTORY` (`Path`). Some of these are computed from other settings and would go stale rather than wrong. `_coerce_env_value` currently declines the `Path` and `dict` ones with a warning, which is a safe default but not the same thing as a decision.
3. **Retire `main()`'s manual patches** (`config.HOST`, `config.SERVER_MODE`, `config.PORT`) once the loop makes them redundant, so there is one mechanism instead of two.

## Risk

This widens what an environment variable can change, in a codebase where `SERVER_MODE` is a security boundary. Worth doing deliberately, with the blast radius enumerated in review rather than discovered afterwards.

One thing #1850 changed here: the boolean vocabulary check now lives at parse time in `_parse_env_bool`, so the class body already reports `SERVER_MODE=yes` instead of quietly selecting the local posture. Widening the loop no longer moves that guard.

## Tests

`backend/ttnn_visualizer/tests/test_settings.py` has the shape to extend — #1850 added coercion cases and a check that `.env.sample`'s documented defaults don't invert their settings. Add: a case asserting an inherited setting is reachable at all, and one asserting `--server` yields `SERVER_MODE is True` without the manual patch.

The per-setting validation cases #1850 landed (`test_an_unusable_ssh_port_is_reported_and_discarded`, `test_an_uncoercible_max_content_length_keeps_the_declared_limit`, `test_a_setting_with_no_coercible_type_is_reported_and_discarded`) exercise `_coerce_env_value` directly today; they start covering the shipped path for free once the loop reaches those settings.

## Verification

`pnpm flask:test`, `pnpm flask:lint`, `pnpm flask:mypy`.

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1857.

Strengthens the environment variable override mechanism in `settings.py` so that settings are consistently and safely overridden, including those inherited from parent classes. Introduces a classified skip list for derived and constant attributes, makes a single method the owner of every derived value, and adds tests that pin the behaviour.

## Configuration override improvements

* `DefaultConfig.override_with_env_variables` now walks the method resolution order (MRO), so inherited settings on `DefaultConfig` are reachable when `Config()` returns a subclass, and subclass declarations still win.

* `_coerce_env_value` parses each environment string back to the declared attribute's type, so an explicitly set variable no longer undoes the class body's parse (`"false"` is truthy). Uncoercible values are reported and discarded rather than applied.

* Boolean parsing is unified so the class body and the override loop share one vocabulary and one strictness rule — `SERVER_MODE` refuses to start on a value it cannot read, everything else warns and keeps its default.

* Attributes the loop must not touch are classified into three frozensets, unioned into `_ENV_OVERRIDE_SKIP` at the point of use, because they have different lifetimes:
  * `_ENV_OVERRIDE_DERIVED` — computed from other settings or the filesystem.
  * `_ENV_OVERRIDE_CONSTANTS` — structural values; changing one is a code change.
  * `_ENV_OVERRIDE_UNCONFIGURED` — deployment knobs whose answer today is "no", not "never" (`SESSION_COOKIE_SECURE`, `SESSION_COOKIE_SAMESITE`, `PRINT_ENV`).

* A variable naming a skipped setting is now reported via `_report_ignored_skip` instead of being dropped silently — an inert variable was otherwise indistinguishable from a typo in its name.

* **`recompute_derived_settings()` is the single owner of the derived group.** It rebuilds `APP_DATA_DIRECTORY`, `REPORT_DATA_DIRECTORY`, `LOCAL_`/`REMOTE_DATA_DIRECTORY`, `SQLALCHEMY_DATABASE_URI` and `GUNICORN_BIND` together, and runs both at the end of the override loop and from `main()`'s `--tt_metal_home` handling. This is what stops a late `TT_METAL_HOME` from serving reports out of `$TT_METAL_HOME/generated` while the database and upload directories stay on the import-time tree. `APP_DATA_DIRECTORY` / `REPORT_DATA_DIRECTORY` keep the precedence the class body gives them and their children follow whichever value wins.

* `DEV_SERVER_HOST` / `DEV_SERVER_PORT` now actually read their environment variables. `.env.sample` has always documented both, but they were declared as bare literals and were unreachable by accident; they feed the dev CORS allowlist and `main()`'s browser-open target.

## Reachability: what this widens

The env-var *names* are not new — 21 of the 24 settings below already called `os.getenv` in the class body. What changes is the **window**: values are now honoured up to `create_app`'s `load_dotenv` rather than only at import, which makes `backend/.env` a config surface for the whole set rather than for `FLASK_DEBUG` / `TESTING` alone.

Measured before this change, the loop reached `['DEBUG', 'TESTING']` on `ProductionConfig` and `[]` on `DevelopmentConfig`. It now reaches 24 settings:

`BASE_PATH`, `DEV_SERVER_HOST`, `DEV_SERVER_PORT`, `FLASK_DEBUG`, `GUNICORN_APP_MODULE`, `GUNICORN_TIMEOUT`, `GUNICORN_WORKERS`, `GUNICORN_WORKER_CLASS`, `HOST`, `LAUNCH_BROWSER_ON_START`, `MALWARE_SCANNER`, `MAX_CONTENT_LENGTH`, `PORT`, `SECRET_KEY`, `SERVER_MODE`, `SESSION_MAX_UPLOADED_REPORTS`, `SSH_DEFAULT_PERFORMANCE_PATH`, `SSH_DEFAULT_PORT`, `SSH_DEFAULT_PROFILER_PATH`, `SSH_REMOTE_CHECK_TIMEOUT`, `SSH_SUBPROCESS_TIMEOUT`, `TESTING`, `TT_METAL_HOME`, `USE_WEBSOCKETS`

Most are late strings. Three are worth naming explicitly:

* **`MALWARE_SCANNER`** declares `None`, so the raw string passes through uncoerced and is `shlex.split()` and executed at `file_uploads.py`.
* **`GUNICORN_APP_MODULE`** is an arbitrary import target handed to the gunicorn subprocess.
* **`MAX_CONTENT_LENGTH`** declares `None` when unset, so an *unreadable* late value warns and keeps no upload limit at all — the outcome `_parse_max_content_length`'s hard raise prevents at import time.

Anyone who can write `backend/.env` can already write `settings.py`, so this is judged acceptable, but it is the blast radius the ticket asked to have enumerated in review.

## CLI handling

* The CLI environment override logic is split into `_apply_cli_env_overrides` and `_config_after_cli_env`, so `--server` / `--host` / `--port` can be asserted without binding a socket or spawning gunicorn.

* `--server` now sets `SERVER_MODE` independently of `--host`. Previously an explicit `--host` won the `if/elif` and the flag was never written, so `--server --host <addr>` bound a reachable interface with `SERVER_MODE` still false — leaving every `@local_only` endpoint and the SSH-metadata branches open.

* `_config_after_cli_env` asserts `Config._instance is None`. `Config` is a singleton, so an instance built during import would make the CLI's environment writes silently ineffective; this fails loudly instead of yielding `SERVER_MODE=False` with no error.

* Manual patching of config attributes is removed for `HOST` / `SERVER_MODE` / `PORT`, which rely solely on the override mechanism. `--tt_metal_home` still writes `os.environ` and `config.TT_METAL_HOME` directly — it runs after `Config()` has been constructed — but its hand-rolled cascade is gone in favour of `recompute_derived_settings()`. The old version recomputed only `APP_DATA_DIRECTORY` and the DB URI, leaving `REPORT_DATA_DIRECTORY` and the local/remote directories stale in the launching process.

## Testing and reliability

* `test_the_settings_inventory_is_pinned` asserts `vars(DefaultConfig)` partitions exactly into `_OVERRIDABLE_SETTINGS` (24) and `_ENV_OVERRIDE_SKIP` (21), descriptors excluded. The skip list alone only fails closed for a *dead* entry; this catches the opposite direction, where a new derived or constant attribute that nobody classified silently becomes env-overridable. It subsumes the earlier `_REQUIRED_ENV_OVERRIDE_SKIPS` list, which is deleted.

* Further tests cover: inherited settings reachable on subclasses; skipped settings genuinely not overridden (using only settings the loop is the sole writer of, so the assertions cannot pass vacuously); the full `TT_METAL_HOME` cascade including the DB URI; explicit `REPORT_DATA_DIRECTORY` still beating the derivation with children following; the ignored-variable warning and its `_RECOMPUTE_HONOURS` exemption; both `DEV_SERVER_*` settings including reaching the CORS allowlist; `--server --host` together; and the singleton assert.

* The test for coercing non-string types (like `Path`) targets the backstop logic directly, since the override loop skips those attributes.

## Documentation

* `AGENTS.md` and `CONVENTIONS.md` describe the three skip sets, `recompute_derived_settings()` as the place to add a derivation, the ignored-variable warning, and the requirement to classify a new setting in the pinned inventory.


backend/ttnn_visualizer/settings.py
```diff
@@ -262,6 +260,44 @@ def _parse_ssh_port(env_value: Optional[str]) -> int:
...
+        "MLIR_DIRECTORY_NAME",
+        "TEST_CONFIG_FILE",
+        "SQLITE_DB_PATH",
+        "SEND_FILE_MAX_AGE_DEFAULT",
+        "DEV_SERVER_PORT",
+        "DEV_SERVER_HOST",
+        "SESSION_COOKIE_SAMESITE",
+        "SESSION_COOKIE_SECURE",
+        "PRINT_ENV",
+    }
```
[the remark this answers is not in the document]
dcblundell: Pre-#1857 the override loop never reached `DefaultConfig.PRINT_ENV`, so it was already always-on. The skip preserves that constant; it doesn’t remove an env knob that existed. Making it env-controllable would be a separate change (`_parse_env_bool` + drop from skip).

backend/ttnn_visualizer/settings.py
```diff
@@ -262,6 +260,44 @@ def _parse_ssh_port(env_value: Optional[str]) -> int:
...
+# ``REPORT_DATA_DIRECTORY`` are skipped so a late ``.env`` cannot update the parent
+# while leaving Path children and the DB URI on the import-time tree — the class body
+# (and ``main()``'s ``--tt_metal_home`` cascade) remain the ways to set them.
+# Constants have no env story. When adding a new constant or derived string attribute,
+# list it here — type decline only catches ``Path`` / ``dict``.
+_ENV_OVERRIDE_SKIP = frozenset(
+    {
+        "APPLICATION_DIR",
+        "APP_DATA_DIRECTORY",
+        "REPORT_DATA_DIRECTORY",
```
aidemsined, line 281: > [!CAUTION]
> `TT_METAL_HOME` is a derived-value parent that isn't skipped, so the desync this set exists to prevent still happens one level up.

`APP_DATA_DIRECTORY` (settings.py:426-429), `REPORT_DATA_DIRECTORY` (431-433), `LOCAL_DATA_DIRECTORY`/`REMOTE_DATA_DIRECTORY` (434-435) and `SQLALCHEMY_DATABASE_URI` (463-464) are all derived from `TT_METAL_HOME` at class-body time. All five children are in this set; the parent isn't, and it's newly reachable. A `TT_METAL_HOME` arriving via `create_app`'s `load_dotenv` flips `PathResolver.is_direct_report_mode` (utils.py:640-641) and serves reports from `$TT_METAL_HOME/generated/...` while the DB and upload directories stay on the import-time tree, and `_build_spa_client_config` publishes the new value alongside the old `REPORT_DATA_DIRECTORY` (app.py:84-85).

This is the same class of bug the second commit fixed by adding `APP_DATA_DIRECTORY`/`REPORT_DATA_DIRECTORY` here — the fix just needs to generalise to the root. `.env.sample` documents `TT_METAL_HOME=` in the same block as the two settings that were skipped, so operators do set it there.

**Suggested action:** add `"TT_METAL_HOME"` to `_ENV_OVERRIDE_SKIP` (consistent with the comment above, which already nominates the class body and `--tt_metal_home` as the sanctioned paths), or recompute the cascade after the loop the way `GUNICORN_BIND` is. Add it to `_REQUIRED_ENV_OVERRIDE_SKIPS` if skipping, and cover it either way.

backend/ttnn_visualizer/tests/test_settings.py
```diff
@@ -476,20 +480,117 @@ def test_every_env_alias_names_a_real_setting():
...
+        "REMOTE_DATA_DIRECTORY",
+        "STATIC_ASSETS_DIR",
+    }
+)
+
+
+def test_every_env_override_skip_names_a_real_setting():
+    # Keyed by string, so a renamed setting leaves a dead skip entry and the loop
+    # starts accepting env strings for a derived attribute again.
+    assert _ENV_OVERRIDE_SKIP <= set(vars(DefaultConfig))
```
aidemsined, line 544: > [!CAUTION]
> This guard is one-way: it catches a dead entry, but not a missing one.

`_ENV_OVERRIDE_SKIP <= set(vars(DefaultConfig))` catches a typo'd or renamed entry, and the `_REQUIRED_ENV_OVERRIDE_SKIPS` assert catches deletion of the nine hand-listed names. Nothing catches the direction that actually bites: add a new derived or constant attribute to `DefaultConfig`, forget the skip entry, and it silently becomes env-overridable — the type-decline backstop only fires for `Path`/`dict`, as the comment at settings.py:271-272 acknowledges. Concretely, 14 of the 23 entries (including `SQLITE_DB_PATH`, `DB_VERSION`, `SESSION_COOKIE_SECURE` and the four `*_DIRECTORY_NAME` constants) can be removed from the skip list with this suite still green.

Worth noting the shape of the tradeoff: there are 23 skipped names against 22 overridable ones, and 21 of those 22 already announce themselves in the class body by calling `os.getenv` / `_parse_env_bool` / a named parser. So an allowlist was available at essentially identical maintenance cost, and is nearly derivable rather than hand-maintained — relevant given the ticket asks for a deliberate decision about what should be overridable at all in a class where `SERVER_MODE` is the security boundary.

**Suggested action:** if inverting the list is too much for this PR, add a pinned-inventory test asserting the exact partition of `vars(DefaultConfig)` into {overridable, skipped, descriptor}. That converts "forgot to classify" from fail-open to a forced decision, and subsumes `_REQUIRED_ENV_OVERRIDE_SKIPS`, which is currently a third copy of the same list.

backend/ttnn_visualizer/settings.py
```diff
@@ -262,6 +260,44 @@ def _parse_ssh_port(env_value: Optional[str]) -> int:
...
+        "DB_VERSION",
+        "PROFILER_DIRECTORY_NAME",
+        "PERFORMANCE_DIRECTORY_NAME",
+        "NPE_DIRECTORY_NAME",
+        "MLIR_DIRECTORY_NAME",
+        "TEST_CONFIG_FILE",
+        "SQLITE_DB_PATH",
+        "SEND_FILE_MAX_AGE_DEFAULT",
+        "DEV_SERVER_PORT",
+        "DEV_SERVER_HOST",
```
aidemsined: > [!CAUTION]
> These two are documented in `.env.sample` as settable, and this PR is the moment they become permanently inert.

`.env.sample` offers `# DEV_SERVER_HOST=localhost` / `# DEV_SERVER_PORT=5173`, but settings.py:482-483 declares both as bare literals with no `os.getenv`, so they were never readable — by accident. The MRO walk would have made them work for free (plain strings, so coercion is identity); adding them here deliberately keeps them dead.

They aren't inconsequential: both feed the dev CORS allowlist through `_build_allowed_origins` (settings.py:53-54, read via the `_AllowedOrigins` descriptor at 182-183) and `main()`'s browser-open target (app.py:604-605). An operator running Vite on a non-default port has a genuine reason to set them and will get silence. `test_the_documented_defaults_match_the_code_defaults` only inspects boolean settings, so nothing flags the drift.

Since the skip list is now the de facto answer to "what is configurable", each entry is worth checking against the documented surface — `SESSION_COOKIE_SECURE` / `SESSION_COOKIE_SAMESITE` deserve the same explicit call even if the answer is "no".

**Suggested action:** either drop these two skip entries and let the class body read them, or delete the two `.env.sample` lines.

backend/ttnn_visualizer/app.py
```diff
@@ -445,18 +445,13 @@ def _record_launch(config):
...
 
-    # Handle host/port CLI overrides
-    # Priority: CLI args > env vars > auto-detection (in settings.py)
-    # Note: We need to set env vars before creating Config, but also
-    # manually update the config object in case it was already instantiated
+    Must run before ``Config()`` so ``override_with_env_variables`` sees the values.
+    Workers inherit the mutated environment as a fresh import; the launching process
+    relies on the override loop reaching inherited settings.
+    """
     if args.host:
```
aidemsined, line 461: > [!WARNING]
> `--server --host <addr>` binds a network interface without enabling `SERVER_MODE`.

The `if args.host: ... elif args.server:` chain means an explicit `--host` wins and `os.environ["SERVER_MODE"]` is never written. So `--server --host 0.0.0.0` produces a network-reachable socket with `SERVER_MODE` still false, leaving every `@local_only` endpoint (decorators.py:165) and the SSH-metadata branches in `views.py` open.

This is carried over verbatim from the `main()` block this PR deletes rather than introduced here — but the PR rewrites exactly these lines, and the ticket's stated risk is `SERVER_MODE` reachability, so it seems like the moment to decide. The new CLI test only exercises `Namespace(host=None, server=True, port=None)`, so the combination is unpinned either way.

**Suggested action:** set `SERVER_MODE` whenever `args.server` is passed, independently of `args.host` (i.e. hoist it out of the `elif`), and add a case covering `--server --host` together. If the current precedence is deliberate, a comment saying so would stop the next reader from filing this again.

backend/ttnn_visualizer/app.py
```diff
@@ -470,21 +465,28 @@ def main():
...
+    """Apply CLI env mutations and return the config singleton.
 
-    if args.port:
-        config.PORT = args.port
+    Split out of ``main()`` so ``--server`` / ``--host`` / ``--port`` can be asserted
+    without binding a socket or spawning gunicorn. Relies on the override loop — no
+    hand-patches of ``HOST`` / ``SERVER_MODE`` / ``PORT``.
+    """
+    _apply_cli_env_overrides(args)
+    return cast(DefaultConfig, Config())
```
aidemsined, line 492: > [!WARNING]
> The singleton assumption this function depends on is documented but not enforced.

`Config()` is a singleton (settings.py:576-585), so if an instance already exists by the time this runs, the env writes in `_apply_cli_env_overrides` are silently ineffective and `--server` yields `SERVER_MODE=False` with no error. That is exactly what the removed hand-patch was papering over — its comment said "Config is a singleton that may have been created before we set env vars".

I traced the current ordering and it is correct: `Config()` appears only at app.py:132 (inside `create_app`, called later) and lazily inside a function in `database_migrations.py:78`, and `parse_args()` constructs nothing — so line 477 is genuinely the first construction. The concern is that the guarantee now lives only in a docstring. The test can't catch a regression either, because it sets `Config._instance = None` itself before calling in.

**Suggested action:** assert `Config._instance is None` here (or rebuild explicitly) so a future import-time `Config()` fails loudly instead of silently disabling the flag.

backend/ttnn_visualizer/settings.py
```diff
@@ -262,6 +260,44 @@ def _parse_ssh_port(env_value: Optional[str]) -> int:
...
+# Derived or constant attributes the override loop must not touch. String-typed derived
+# values (``GUNICORN_BIND``, ``SQLALCHEMY_DATABASE_URI``, ``APPLICATION_DIR``) would
+# otherwise accept an env string and diverge from their parents; ``Path`` / ``dict``
+# ones are declined by ``_coerce_env_value`` as a backstop, but skipping them here
+# keeps the loop from warning on every visit. ``APP_DATA_DIRECTORY`` /
+# ``REPORT_DATA_DIRECTORY`` are skipped so a late ``.env`` cannot update the parent
+# while leaving Path children and the DB URI on the import-time tree — the class body
+# (and ``main()``'s ``--tt_metal_home`` cascade) remain the ways to set them.
+# Constants have no env story. When adding a new constant or derived string attribute,
+# list it here — type decline only catches ``Path`` / ``dict``.
```
> The set flattens three different rationales into one frozenset, which makes it hard to edit safely.

Three distinct claims are mixed here: derived-from-other-settings (`GUNICORN_BIND`, `SQLALCHEMY_DATABASE_URI`, `LOCAL_`/`REMOTE_DATA_DIRECTORY`, `STATIC_ASSETS_DIR`) which must *never* be overridable; structural constants (`DB_VERSION`, `TEST_CONFIG_FILE`, the four `*_DIRECTORY_NAME`s); and settings nobody has made configurable *yet* (`SESSION_COOKIE_SECURE`, `SESSION_COOKIE_SAMESITE`, `DEV_SERVER_*`, `PRINT_ENV`). Those have different lifetimes, and a maintainer adding a setting can't tell from a flat set which rule applies.

`SESSION_COOKIE_SECURE = False  # For development on HTTP` (settings.py:494) is the case that will bite — it's a deployment hardening knob, not a derived value, and a TLS-fronted `SERVER_MODE` deploy currently has no way to set it.

Also worth noting the stated invariant is already violated in its own file: the comment says "Constants have no env story", but `TESTING` is a constant that is deliberately left overridable (and was reachable pre-PR). The set has no way to express "constant, but deliberately configurable".

**Suggested action:** split into two named frozensets unioned at the point of use, or at minimum add blank-line groups with a one-line comment each. Costs nothing and makes the classification reviewable.

backend/ttnn_visualizer/settings.py
```diff
@@ -462,27 +498,37 @@ class DefaultConfig(object):
...
+                    or key in _ENV_OVERRIDE_SKIP
+                    or hasattr(value, "__get__")
+                ):
+                    continue
+
+                env_value = os.getenv(_env_name_for(key))
+                if env_value is None:
+                    continue
+
+                setattr(self, key, _coerce_env_value(key, value, env_value))
```
aidemsined, line 560: > [!WARNING]
> Three of the newly-reachable settings are worth naming explicitly in the PR description.

Measured at this head, the pre-PR loop reached `['DEBUG', 'TESTING']` on `ProductionConfig` and `[]` on `DevelopmentConfig`; post-PR the reachable set is 22 settings. Most are benign, but three aren't just late strings:

- `MALWARE_SCANNER` declares `None`, so `_coerce_env_value` passes the raw string through untouched (settings.py:393) and it is `shlex.split()` and executed at file_uploads.py:190.
- `GUNICORN_APP_MODULE` is an arbitrary import target handed to the gunicorn subprocess (app.py:579).
- `MAX_CONTENT_LENGTH` declares `None` when unset, so a late unreadable value warns and keeps **no upload limit at all** (settings.py:374-379) — the outcome `_parse_max_content_length`'s hard raise exists to prevent at import.

The env-name surface isn't new — all three already read env in the class body — so this isn't a new capability. What changes is the window: values are now honoured up to `create_app`'s second `load_dotenv` (app.py:117-119) rather than only at import, which makes `backend/.env` a config surface for the whole set rather than for `FLASK_DEBUG`/`TESTING` only. Anyone who can write that file can already write `settings.py`, so this reads as acceptable — but the ticket asked for the blast radius to be enumerated in review, and these three are the ones a future reader will want named.

**Suggested action:** list the newly-reachable set in the PR description, calling out these three.

backend/ttnn_visualizer/tests/test_settings.py
```diff
@@ -476,20 +480,117 @@ def test_every_env_alias_names_a_real_setting():
...
+    declared_local = config.LOCAL_DATA_DIRECTORY
+    declared_remote = config.REMOTE_DATA_DIRECTORY
+    config.override_with_env_variables()
+
+    assert config.APPLICATION_DIR == declared_app_dir
+    assert config.SQLALCHEMY_DATABASE_URI == declared_uri
+    assert config.APP_DATA_DIRECTORY == declared_app_data
+    assert config.REPORT_DATA_DIRECTORY == declared_report_data
+    assert config.LOCAL_DATA_DIRECTORY == declared_local
+    assert config.REMOTE_DATA_DIRECTORY == declared_remote
```
aidemsined: > [!WARNING]
> These two assertions can't fail — same vacuity class you already caught for `GUNICORN_BIND` just above.

The test sets env for `APPLICATION_DIR`, `SQLALCHEMY_DATABASE_URI`, `APP_DATA_DIRECTORY` and `REPORT_DATA_DIRECTORY` (lines 537-540) but never for `LOCAL_DATA_DIRECTORY` / `REMOTE_DATA_DIRECTORY`. Both therefore take the `env_value is None: continue` branch (settings.py:526-527) and pass identically with those two names removed from `_ENV_OVERRIDE_SKIP` — the skip isn't load-bearing for either.

Setting the env wouldn't fully fix it either: both are `Path`, so `_coerce_env_value` declines them at settings.py:400-405 regardless of the skip list. The other four assertions in this test are genuinely non-vacuous (all `str`, all would take the env string), so it's specifically these two.

**Suggested action:** drop the two lines, or move them to a case where the skip is what does the work — otherwise they read as coverage that isn't there.

backend/ttnn_visualizer/settings.py
```diff
@@ -462,27 +498,37 @@ class DefaultConfig(object):
...
+
+        self.GUNICORN_BIND = f"{self.HOST}:{self.PORT}"
```
> Consider expressing this as a descriptor rather than skip-entry plus hand-recompute.

The file already has the abstraction — `_AllowedOrigins` and `_UsageRecordingEnabled` (settings.py:170-210) are read-time computed descriptors, and the loop skips anything with `__get__` for free. And this isn't a one-off: counting derived values there are nine (`GUNICORN_BIND`, `SQLALCHEMY_DATABASE_URI`, `LOCAL_`/`REMOTE_DATA_DIRECTORY`, `STATIC_ASSETS_DIR`, `APP_DATA_DIRECTORY`, `REPORT_DATA_DIRECTORY`, plus the two already-descriptor ones), now handled three different ways: read-time descriptor, frozen-by-skip, and hand-recompute.

A `_GunicornBind` descriptor would delete the skip entry, delete this line, and make the value un-stale-able for *any* later `HOST`/`PORT` mutation rather than only for the loop's. It would also remove the caveat the test had to write around at test_settings.py:534-536 ("`GUNICORN_BIND` is always rewritten… so it cannot prove the skip alone"). Non-data descriptors still let an explicit instance assignment win, so nothing that assigns it breaks.

**Suggested action:** optional for this PR, but worth deciding now rather than after a third mechanism appears.

backend/ttnn_visualizer/settings.py
```diff
@@ -462,27 +498,37 @@ class DefaultConfig(object):
...
+                # the raw environment string over one would shadow it with an unparsed
+                # value. Derived / constant attrs are listed in ``_ENV_OVERRIDE_SKIP``
+                # because a string would apply and diverge from their parents (or, for
+                # constants, because they have no env story).
+                if (
+                    key.startswith("_")
+                    or key in _ENV_OVERRIDE_SKIP
+                    or hasattr(value, "__get__")
+                ):
+                    continue
```
aidemsined, line 550: > [!WARNING]
> Skipping is silent, but declining warns — the asymmetry is backwards for a denylist.

`_keep_declared` logs whenever a value is refused, so an operator who sets an uncoercible value gets feedback. But a variable matching a skipped attribute is dropped here with no signal at all: setting `SQLALCHEMY_DATABASE_URI` or `GUNICORN_BIND` in `.env` now does nothing and says nothing. For a hand-maintained skip list that's the one place operator feedback matters most, since the failure is indistinguishable from a typo in the variable name.

**Suggested action:** a `logger.debug` (or `warning`) in the skip branch when `os.getenv(_env_name_for(key))` is non-`None` — three lines, and the natural counterpart to the coercion warnings.

backend/ttnn_visualizer/app.py
```diff
@@ -445,18 +445,13 @@ def _record_launch(config):
...
 
-def main():
-
-    run_command = sys.argv[0].split("/")
-    if run_command[-1] == "ttnn-visualizer":
-        os.environ.setdefault("FLASK_ENV", "production")
-
-    args = parse_args()
+def _apply_cli_env_overrides(args: argparse.Namespace) -> None:
+    """Write ``--host`` / ``--server`` / ``--port`` into the environment.
```
aidemsined, line 447: > [!WARNING]
> This covers three of the four CLI flags that write config, so the PR description slightly overstates the cleanup.

`--tt_metal_home` still gets the old treatment 20 lines further down (app.py:494-510): `os.environ[...]` **plus** hand-patches of `config.TT_METAL_HOME`, `config.APP_DATA_DIRECTORY` and `config.SQLALCHEMY_DATABASE_URI`. The ticket only named `HOST` / `SERVER_MODE` / `PORT`, so this isn't an AC miss — but "manual patching of config attributes removed in favour of relying solely on the override mechanism" reads as complete when there are still two mechanisms, one just smaller.

Two knock-ons worth a look: the helper's name promises more than it does, and the surviving cascade recomputes only 2 of the 5 settings derived from that root — `REPORT_DATA_DIRECTORY`, `LOCAL_DATA_DIRECTORY` and `REMOTE_DATA_DIRECTORY` are left stale, so `display_mode_info_without_db` prints stale directories in the launching process. The new `_ENV_OVERRIDE_SKIP` comment points readers at that cascade as a sanctioned way to set `APP_DATA_DIRECTORY` / `REPORT_DATA_DIRECTORY` without noting what it doesn't update.

**Suggested action:** soften the PR description, and either extend the cascade to the remaining three or note the limitation where the comment recommends it. A shared `DefaultConfig._recompute_derived()` called from both here and the override loop would collapse the three sites that currently duplicate these derivations.

aidemsined (commented):
## Unanchored findings

**`Config._instance` is assigned before the override runs, so a strict-boolean refusal leaves a poisoned singleton.** `settings.py:582-583` (outside the diff, hence no inline anchor) does `cls._instance = cls._determine_config()` *before* calling `override_with_env_variables()`, so a raise leaves a non-`None` instance behind.

This was unreachable before this PR: the loop only visited `DEBUG`/`TESTING`, neither of which is in `_STRICT_BOOLEANS`, and every other failure path goes through `_keep_declared`. Widening the reach creates the window — `SERVER_MODE` is now reachable, and `_coerce_bool` (settings.py:334-339) raises out of `_coerce_env_value` un-caught, unlike the `_ENV_PARSERS` branch (376-379) and the `int` branch (386-389), both of which are wrapped.

Reproduced at this head with `SERVER_MODE=Ture` and `FLASK_ENV=production`: the first `Config()` raises `SERVER_MODE='Ture' is not a recognised boolean`, and the second returns `ProductionConfig` with `SERVER_MODE=False`, an empty instance `__dict__`, no `GUNICORN_BIND` recompute, and `_refuse_debug_under_server_mode` never run — i.e. a typo in `.env` can leave the process on the permissive posture rather than failing closed. It also affects the suite: any test exercising the strict refusal through `Config()` poisons the singleton for every later test in the session.

Whether this is merge-blocking depends on whether anything catches the raise — if it always propagates and kills the process, the exposure is small. Worth a deliberate call given the setting involved. Fix is two lines: build into a local, override, then assign `cls._instance`.

## General observations

- No test pins "subclass declarations win". `settings.py:502-503` claims it and it is true — `reversed((Sub, DefaultConfig, object))` visits the subclass last — but it is one dropped `reversed()` from inverting, and inverting it would make `ProductionConfig.DEBUG = False` lose to `DefaultConfig.DEBUG` in exactly the misconfiguration case where it matters. An unparseable `FLASK_DEBUG` against `TestingConfig` (declared `True`) vs `ProductionConfig` (declared `False`) would distinguish the two orders.
- A key declared at several MRO levels is coerced once per level, so an unparseable `DEBUG`/`TESTING` logs the `_keep_declared` warning twice. The final `setattr` is the intended one, so this is cosmetic — but the duplicate will read as a bug to whoever sees it.
- The descriptor handling is correct, and worth recording since the PR body describes it as "explicitly skipped": the actual guard is the duck-typed `hasattr(value, "__get__")` at settings.py:521, not `_ENV_OVERRIDE_SKIP`. The loop reads `cls.__dict__.items()` (raw descriptor object), so neither `__get__` is invoked during the walk. Both are non-data descriptors, so an instance `setattr` *would* shadow them — if `_AllowedOrigins` were ever flattened to a plain list, the loop would start assigning the raw comma string and `origin in allowed_origins` would degrade from a list-membership test to a substring test. Adding both names to the skip set would make the intent explicit and testable.
- Boolean vocabulary unification checks out end to end. `_parse_env_bool` and the override path both route through `_coerce_bool` → `parse_bool`, and `usage._as_bool` → `str_to_bool` reaches the same function. No path found where `"false"`, `"0"` or `"no"` coerces to `True`. One residual asymmetry, pre-existing and documented: `"no"` means *false* at the `str_to_bool` call sites but *unrecognised, keep default* in the config layer.
- Two stale comments the cleanup pass missed: `settings.py:172-173` still describes "a `--port` override applied to the config object rather than the environment", a path this PR deletes; and `tests/test_settings.py:218-219` still asserts that `TESTING`/`DEBUG` are "the sole settings a concrete config class declares", which is the exact statement this PR falsifies.
- `AGENTS.md:201` still ends "don't put a guard in the loop expecting it to cover one", but the rewritten `CONVENTIONS.md` section it links to no longer states that rule — the detail record dropped a clause the summary still asserts.
- `SESSION_COOKIE_SECURE` / `SESSION_COOKIE_SAMESITE` are now permanently denylisted, so a hosted HTTPS deploy under `SERVER_MODE` ships session cookies without `Secure`. Follow-up issue rather than a change here.
- `to_dict` (settings.py:553-559) evaluates `getattr(self, key)` twice per key — once for the `callable(...)` filter, once for the value — so each `PRINT_ENV` dump runs `_UsageRecordingEnabled.__get__` twice and rebuilds the allowlist twice. Pre-existing and startup-only; hoisting into a local halves it.
- `tests/test_settings.py:507` imports `_config_after_cli_env` from `app.py`, putting a CLI-wiring assertion in the settings test module. There is no `test_app.py`, so there is nowhere obviously better today — flagging so the coupling is a choice rather than an accident.
- The commit split is worth preserving in the merge message. `57326f38` reverses nothing but fixes two things the suite would not have caught: adding `APP_DATA_DIRECTORY`/`REPORT_DATA_DIRECTORY` to the skip set (they were overridable in the first commit while their children were skipped), and switching `monkeypatch.delenv` to `setenv` so the raw `os.environ` writes get restored on teardown instead of leaking into the rest of the session. That second one is invisible in the monolithic diff.
- Longer-term direction, not for this PR: a declarative field helper (`PORT = env_setting("PORT", "8000")`) that parses at import *and* registers the attribute as overridable would make the allowlist fall out of the declaration instead of being maintained beside it.

Verification not run from here — `pnpm flask:test`, `pnpm flask:lint` and `pnpm flask:mypy` are still worth running before merge, `mypy` in particular against the new annotated signatures in `app.py`.

---

backend/ttnn_visualizer/settings.py
dcblundell: Agreed, and confirmed. The two `load_dotenv` calls target different files — `settings.py:25` searches from the working directory, `create_app` targets `backend/.env` explicitly — so the class-body read and the loop read genuinely can differ, which is the window you described.

Taken the recompute option rather than the skip, since `TT_METAL_HOME` is documented in `.env.sample` and freezing it would make another documented variable inert. New `DefaultConfig.recompute_derived_settings()` rebuilds `APP_DATA_DIRECTORY`, `REPORT_DATA_DIRECTORY`, `LOCAL_`/`REMOTE_DATA_DIRECTORY`, `SQLALCHEMY_DATABASE_URI` and `GUNICORN_BIND` as one group at the end of the override loop.

Covered by `test_a_late_tt_metal_home_carries_the_whole_path_tree` (whole tree including the DB URI) and `test_an_explicit_report_directory_still_wins_over_the_derived_one` (explicit variable still beats the derivation, children follow the winner).

Same method fixes the `--tt_metal_home` cascade you raised separately.

backend/ttnn_visualizer/tests/test_settings.py
dcblundell: Agreed on the fail-open direction — that is the one the current guard cannot see.

Added `test_the_settings_inventory_is_pinned`, which asserts `vars(DefaultConfig)` partitions exactly into `_OVERRIDABLE_SETTINGS` (24 names) and `_ENV_OVERRIDE_SKIP` (21), descriptors and methods excluded, and that the two sets are disjoint. A new attribute lands in neither and fails. `_REQUIRED_ENV_OVERRIDE_SKIPS` is deleted — subsumed, as you predicted.

Left the allowlist inversion itself for a follow-up. The pinned partition already forces the classification, which was the load-bearing part of the ask; inverting the storage is then a mechanical change on top.

backend/ttnn_visualizer/settings.py
dcblundell: Agreed — verified they were unreachable by accident, and this PR would have made that permanent.

Dropped both skip entries and added the class-body `os.getenv` reads. `test_a_documented_dev_server_setting_is_readable` covers both, and `test_a_dev_server_override_reaches_the_allowlist` covers the path that actually matters — a non-default Vite port reaching `_build_allowed_origins` through the `_AllowedOrigins` descriptor.

`SESSION_COOKIE_SECURE` / `SESSION_COOKIE_SAMESITE` got the explicit answer you asked for too — see the reply on the frozenset thread.

backend/ttnn_visualizer/app.py
dcblundell: Agreed, and not deliberate — `--server`'s own help text says "and enable server mode", so the current precedence contradicts the documented behaviour.

Hoisted the `SERVER_MODE` write out of the `elif`, so `--server` sets it regardless of `--host`; `--host` still wins the bind address. `test_an_explicit_host_does_not_suppress_server_mode` pins the combination.

backend/ttnn_visualizer/app.py
dcblundell: Agreed — thanks for tracing the current ordering rather than just flagging it.

Added `assert Config._instance is None` at the top of `_config_after_cli_env`, with the message naming what would be silently ignored. `test_the_cli_refuses_to_run_after_the_config_singleton_exists` sets a live instance and asserts it raises, so the guarantee no longer lives only in a docstring.

backend/ttnn_visualizer/settings.py
dcblundell: Agreed — split into `_ENV_OVERRIDE_DERIVED`, `_ENV_OVERRIDE_CONSTANTS` and `_ENV_OVERRIDE_UNCONFIGURED`, unioned into `_ENV_OVERRIDE_SKIP` at the point of use, each with its own comment. `test_the_three_skip_rationales_stay_disjoint` keeps them from overlapping.

The cookie settings sit in `_UNCONFIGURED` with a comment naming them as the first entries to revisit and what moving one costs — the explicit "no, for now" you asked for rather than an implicit one.

On the self-contradiction: you are right that the comment was wrong, though `TESTING` was never in the set — it is overridable and stays so. The misleading "constants have no env story" sentence is gone, and the third group is exactly the "constant, but deliberately configurable" gap you identified.

The derived group also gained a wrinkle worth naming: `APP_DATA_DIRECTORY` / `REPORT_DATA_DIRECTORY` are skipped by the loop but honoured by `recompute_derived_settings`, so `_RECOMPUTE_HONOURS` marks them.

backend/ttnn_visualizer/settings.py
dcblundell: Agreed, and I verified all three mechanisms: `MALWARE_SCANNER` declares `None` so the raw string passes through and reaches `shlex.split` + `subprocess.run`; `GUNICORN_APP_MODULE` is an arbitrary import target for the gunicorn subprocess; and `MAX_CONTENT_LENGTH` does downgrade the import-time hard raise to warn-and-keep-`None`, i.e. no limit, when the late value is unreadable.

Agreed too that the env-name surface is not new — only the window is — and that anyone who can write `backend/.env` can already write `settings.py`.

Not done yet: this is a PR-description change and the description has not been updated. Leaving this thread open as the reminder.

backend/ttnn_visualizer/tests/test_settings.py
dcblundell: Confirmed on both points — both assertions took the `env_value is None` branch, and setting the env would not have rescued them because `Path` is declined by type regardless of the skip.

`test_a_skipped_setting_is_not_overridden` now uses only settings the loop is the sole writer of: `APPLICATION_DIR`, `STATIC_ASSETS_DIR`, `DB_VERSION`, `SQLITE_DB_PATH`, `SESSION_COOKIE_SAMESITE`. The child-directory behaviour is asserted for real in the new cascade tests, where the parent actually changes.

Checked the new tests against the same failure mode by reverting each fix and confirming they go red.

backend/ttnn_visualizer/settings.py
dcblundell: Not taking the descriptor in this PR, but the count has moved in your direction: `recompute_derived_settings()` is now the single owner of all the derived values, called from both the override loop and `main()`'s `--tt_metal_home` handling. So it is two mechanisms (read-time descriptor, group recompute) rather than three, and `GUNICORN_BIND` is no longer special-cased — it is recomputed alongside the path tree, and the skip test no longer has to route around it.

The remaining argument for descriptors is that the group would become un-stale-able for *any* later mutation, not just the ones that call the recompute. Worth doing, but as a follow-up rather than growing this PR further — happy to raise an issue if you agree.

backend/ttnn_visualizer/settings.py
dcblundell: Agreed — the asymmetry was backwards, and an inert variable really is indistinguishable from a typo.

Added `_report_ignored_skip`, called from the skip branch. Went with `warning` rather than `debug`: it only fires when the operator actually set the variable, which makes it the direct counterpart to the coercion warnings.

One exclusion: `APP_DATA_DIRECTORY` / `REPORT_DATA_DIRECTORY` are exempt via `_RECOMPUTE_HONOURS`, because the recompute does apply them — warning there would send an operator chasing a variable that works. `test_an_ignored_variable_says_so` and `test_a_variable_the_recompute_honours_is_not_reported_as_ignored` cover both sides.

backend/ttnn_visualizer/app.py
dcblundell: Agreed on all three points, including the stale-directory knock-on — confirmed the old cascade reached `APP_DATA_DIRECTORY` and the DB URI but left `REPORT_DATA_DIRECTORY`, `LOCAL_` and `REMOTE_DATA_DIRECTORY` on the import-time values.

Taken the shared-method suggestion: the block is now the env write plus `config.recompute_derived_settings()`, so the three duplicated derivation sites collapse to one and all five children follow the new root. That also removes the caveat the `_ENV_OVERRIDE_SKIP` comment needed about what the cascade does not update.

The PR description still overstates the cleanup and has not been corrected yet.

backend/ttnn_visualizer/settings.py
dcblundell: Done now — the description has a "Reachability: what this widens" section listing all 24 reachable settings, with `MALWARE_SCANNER`, `GUNICORN_APP_MODULE` and `MAX_CONTENT_LENGTH` called out individually and the names-vs-window distinction stated. Resolving; reopen if the framing isn't what you wanted.

backend/ttnn_visualizer/app.py
dcblundell: Description corrected too — the cleanup claim is now scoped to `HOST` / `SERVER_MODE` / `PORT`, and says plainly that `--tt_metal_home` still writes `os.environ` and `config.TT_METAL_HOME` directly because it runs after `Config()`, with only the hand-rolled cascade replaced.

PR #1874, opened 2026-08-13
Refactor event logging tests

### LINKED ISSUE #1871: Usage log: exercise O_APPEND with subprocess writers in tests
Follow-up from review on #1851 (deferred from [this thread](https://github.com/tenstorrent/ttnn-visualizer/pull/1851#discussion_r3770128868)).

## Problem

`test_concurrent_writers_never_truncate_a_line` uses a `ThreadPoolExecutor` inside one process. That does not meaningfully exercise the `O_APPEND` atomicity claim in `_append_line`'s docstring — which is about separate *instances* sharing the log with no coordination. The current test would also pass if every writer funnelled through one shared buffered file object.

#1828's test plan asked for: *"concurrent appends from two writers interleave without truncating a line"* — subprocess writers match that intent.

## Proposed change

Replace (or supplement) the ThreadPool test with two `subprocess` writers that each append many lines to the same redirected usage log, then assert:

- line count equals total writes
- every line is a well-formed `app_start` record (no truncated / interleaved mid-line fragments)

Keep the existing in-process test only if it still adds value under a clearer name; otherwise remove it so the suite does not overclaim.

## Out of scope

NFS interleave behaviour (already documented as skippable malformed lines). This follow-up is about pinning the local-filesystem `O_APPEND` contract the module actually promises.

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1871.

This pull request refactors the concurrency test for usage event logging in `test_usage.py`. It replaces the previous in-process thread-based test with a more robust subprocess-based approach, ensuring that concurrent writers simulate real-world scenarios more accurately. The update introduces helper functions and a subprocess script to facilitate this change.

Concurrency and robustness improvements:

* Replaced the thread-based concurrency test with a subprocess-based test (`test_concurrent_subprocess_writers_never_truncate_a_line`), ensuring that concurrent writes to the usage log are properly tested across separate Python interpreters for better simulation of real-world usage.
* Added helper functions `_spawn_usage_writer` and `_await_usage_writer` to launch and manage subprocesses that write usage events, including an inline Python script (`_WRITER_SCRIPT`) for event recording. [[1]](diffhunk://#diff-5b2042663a71f72c1d56fe7981a739bcf2d0df07dbe17c9b1af424240c2e3c53R35-R50) [[2]](diffhunk://#diff-5b2042663a71f72c1d56fe7981a739bcf2d0df07dbe17c9b1af424240c2e3c53R85-R111)
* Introduced a timeout constant (`_SUBPROCESS_WRITER_TIMEOUT_SECONDS`) to prevent hanging tests in CI environments.

Dependency and import updates:

* Replaced the import of `ThreadPoolExecutor` with imports for `subprocess` and `sys` to support the new subprocess-based concurrency test.

### DIFF
--- a/backend/ttnn_visualizer/tests/test_usage.py
+++ b/backend/ttnn_visualizer/tests/test_usage.py
@@ -105,7 +105,16 @@ def _spawn_usage_writer(directory: Path, count: int) -> subprocess.Popen:
 
 
 def _await_usage_writer(child: subprocess.Popen) -> None:
-    _, stderr = child.communicate(timeout=_SUBPROCESS_WRITER_TIMEOUT_SECONDS)
+    try:
+        _, stderr = child.communicate(timeout=_SUBPROCESS_WRITER_TIMEOUT_SECONDS)
+    except subprocess.TimeoutExpired:
+        child.kill()
+        _, stderr = child.communicate()
+        pytest.fail(
+            f"usage writer timed out after {_SUBPROCESS_WRITER_TIMEOUT_SECONDS}s"
+            f"{f': {stderr}' if stderr else ''}"
+        )
+
     assert child.returncode == 0, stderr
 
 
--- a/backend/ttnn_visualizer/tests/test_usage.py
+++ b/backend/ttnn_visualizer/tests/test_usage.py
@@ -13,6 +13,7 @@ collector — totals never go down, and no line can be forged.
 import os
 import subprocess
 import sys
+import time
 from pathlib import Path
 from types import SimpleNamespace
 
@@ -20,8 +21,10 @@ import pytest
 from ttnn_visualizer import usage
 from ttnn_visualizer.settings import DefaultConfig
 from ttnn_visualizer.usage import (
+    _REQUIRED_FIELDS,
     COUNT_FIELD,
     RUN_ID_ENV_VAR,
+    RUN_ID_FIELD,
     USAGE_RECORDING_ENV_VAR,
     DeploymentMode,
     UsageEvent,
@@ -32,21 +35,10 @@ from ttnn_visualizer.usage import (
     record_event,
 )
 
-# Enough for two children to finish 100 appends each on a slow CI runner; a hang
-# should fail the test rather than wedge the suite.
+# One budget for the whole set of children, not one each: enough for all of them to
+# finish their appends on a slow CI runner, and a hang fails the test rather than
+# wedging the suite.
 _SUBPROCESS_WRITER_TIMEOUT_SECONDS = 30
-
-_WRITER_SCRIPT = """
-import sys
-from pathlib import Path
-
-from ttnn_visualizer import usage
-from ttnn_visualizer.usage import UsageEvent, record_event
-
-usage.USAGE_DIRECTORY = Path(sys.argv[1])
-for _ in range(int(sys.argv[2])):
-    record_event(UsageEvent.APP_START)
-"""
 
 
 @pytest.fixture
@@ -82,40 +74,67 @@ def total_events(lines):
     return sum(int(parse(line).get(COUNT_FIELD, "1")) for line in lines)
 
 
-def _spawn_usage_writer(directory: Path, count: int) -> subprocess.Popen:
-    """Append ``count`` events from a fresh interpreter into ``directory``.
+def _run_usage_writers(directory: Path, writers: int, writes_each: int) -> None:
+    """Append ``writes_each`` events into ``directory`` from ``writers`` interpreters.
 
-    ``monkeypatch`` does not cross process boundaries, so the child sets
-    ``USAGE_DIRECTORY`` itself — the same override the fixture uses in-process.
+    All-or-nothing on purpose. The children only overlap if every one is spawned
+    before any is awaited, and a separate spawn helper and await helper make
+    ``spawn(); await(); spawn(); await()`` just as natural to write — which reads
+    fine, passes, and silently proves nothing about interleaving.
+
+    Must be called from within the ``usage_directory`` fixture: the children inherit
+    a copy of ``os.environ``, and it is that fixture's ``delenv`` that keeps a
+    developer's own ``TTNN_VISUALIZER_RUN_ID`` from reaching them and collapsing
+    their run ids into one.
     """
     child_env = {
         **os.environ,
-        # A developer shell with the switch off would otherwise inherit silence;
-        # ``record_event`` then no-ops and the child still exits 0.
+        # A developer shell with the switch off would otherwise inherit silence.
         USAGE_RECORDING_ENV_VAR: "true",
     }
+    command = [
+        sys.executable,
+        "-m",
+        "ttnn_visualizer.tests.usage_writer",
+        str(directory),
+        str(writes_each),
+    ]
 
-    return subprocess.Popen(
-        [sys.executable, "-c", _WRITER_SCRIPT, str(directory), str(count)],
-        env=child_env,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        text=True,
-    )
-
-
-def _await_usage_writer(child: subprocess.Popen) -> None:
+    children: list[subprocess.Popen] = []
+    deadline = time.monotonic() + _SUBPROCESS_WRITER_TIMEOUT_SECONDS
     try:
-        _, stderr = child.communicate(timeout=_SUBPROCESS_WRITER_TIMEOUT_SECONDS)
-    except subprocess.TimeoutExpired:
-        child.kill()
-        _, stderr = child.communicate()
-        pytest.fail(
-            f"usage writer timed out after {_SUBPROCESS_WRITER_TIMEOUT_SECONDS}s"
-            f"{f': {stderr}' if stderr else ''}"
-        )
+        for _ in range(writers):
+            # Inside the ``try`` so a fork/exec failure on a loaded runner — the kind
+            # of thing that only ever happens in CI — orphans nothing.
+            children.append(
+                subprocess.Popen(
+                    command,
+                    env=child_env,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    text=True,
+                )
+            )
 
-    assert child.returncode == 0, stderr
+        for child in children:
+            try:
+                remaining = max(deadline - time.monotonic(), 0)
+                _, stderr = child.communicate(timeout=remaining)
+            except subprocess.TimeoutExpired:
+                child.kill()
+                _, stderr = child.communicate()
+                pytest.fail(
+                    f"usage writers did not finish within "
+                    f"{_SUBPROCESS_WRITER_TIMEOUT_SECONDS}s"
+                    f"{f': {stderr}' if stderr else ''}"
+                )
+
+            assert child.returncode == 0, stderr
+    finally:
+        for child in children:
+            if child.poll() is None:
+                child.kill()
+                child.communicate()
 
 
 def test_the_path_ignores_every_environment_derived_data_directory(monkeypatch):
@@ -279,30 +298,41 @@ def test_concurrent_subprocess_writers_never_truncate_a_line(usage_directory):
     # Two separate instances sharing the log — the case ``_append_line``'s
     # ``O_APPEND`` claim is about. An in-process ThreadPool would still pass if
     # every writer funnelled through one shared buffered file object.
-    writes_per_writer = 100
-    total_writes = writes_per_writer * 2
+    writers = 2
+    # The detection margin is thin: at 100 appends each a genuinely broken
+    # ``_append_line`` (one line split across two ``os.write`` calls) survives about
+    # one run in ten, because the write windows only overlap for as long as they do
+    # by accident of interpreter startup taking about as long as the loop. 500 costs
+    # ~0.02s and widens the window enough to catch it every time.
+    writes_each = 500
 
-    writers = [
-        _spawn_usage_writer(usage_directory, writes_per_writer),
-        _spawn_usage_writer(usage_directory, writes_per_writer),
-    ]
-    try:
-        for writer in writers:
-            _await_usage_writer(writer)
-    finally:
-        for writer in writers:
-            if writer.poll() is None:
-                writer.kill()
-                writer.communicate()
+    _run_usage_writers(usage_directory, writers=writers, writes_each=writes_each)
 
     lines = read_lines(usage_directory)
 
-    assert len(lines) == total_writes
+    assert len(lines) == writers * writes_each
+
+    parsed = []
     for line in lines:
-        fields = parse(line)
+        try:
+            fields = parse(line)
+        except ValueError:
+            # A line severed mid-token leaves a fragment with no ``=`` in it, which
+            # ``parse`` raises on. Quote it: the module's whole reason for existing
+            # should not report as a traceback out of a shared helper.
+            pytest.fail(f"truncated or interleaved line: {line!r}")
+
+        # Token count as well as key set. Two whole lines merged carry exactly the
+        # same four keys as one, so comparing key sets alone would not see the seam.
+        assert len(line.split(" ")) == len(_REQUIRED_FIELDS) + 1
+        assert set(fields) == {*_REQUIRED_FIELDS, RUN_ID_FIELD}
         assert fields["event"] == "app_start"
-        assert "ts" in fields
-        assert "schema_version" in fields
+
+        parsed.append(fields)
+
+    # The cross-process premise, pinned rather than merely arranged: every assertion
+    # above is equally satisfied by all 1000 lines coming from one interpreter.
+    assert len({fields[RUN_ID_FIELD] for fields in parsed}) == writers
 
 
 def test_app_start_carries_the_baseline_fields(usage_directory):
--- a/backend/ttnn_visualizer/tests/test_usage.py
+++ b/backend/ttnn_visualizer/tests/test_usage.py
@@ -11,7 +11,8 @@ collector — totals never go down, and no line can be forged.
 """
 
 import os
-from concurrent.futures import ThreadPoolExecutor
+import subprocess
+import sys
 from pathlib import Path
 from types import SimpleNamespace
 
@@ -30,6 +31,22 @@ from ttnn_visualizer.usage import (
     record_app_start,
     record_event,
 )
+
+# Enough for two children to finish 100 appends each on a slow CI runner; a hang
+# should fail the test rather than wedge the suite.
+_SUBPROCESS_WRITER_TIMEOUT_SECONDS = 30
+
+_WRITER_SCRIPT = """
+import sys
+from pathlib import Path
+
+from ttnn_visualizer import usage
+from ttnn_visualizer.usage import UsageEvent, record_event
+
+usage.USAGE_DIRECTORY = Path(sys.argv[1])
+for _ in range(int(sys.argv[2])):
+    record_event(UsageEvent.APP_START)
+"""
 
 
 @pytest.fixture
@@ -63,6 +80,33 @@ def parse(line: str):
 def total_events(lines):
     """Cumulative count the way the collector derives it: ``count``, default 1."""
     return sum(int(parse(line).get(COUNT_FIELD, "1")) for line in lines)
+
+
+def _spawn_usage_writer(directory: Path, count: int) -> subprocess.Popen:
+    """Append ``count`` events from a fresh interpreter into ``directory``.
+
+    ``monkeypatch`` does not cross process boundaries, so the child sets
+    ``USAGE_DIRECTORY`` itself — the same override the fixture uses in-process.
+    """
+    child_env = {
+        **os.environ,
+        # A developer shell with the switch off would otherwise inherit silence;
+        # ``record_event`` then no-ops and the child still exits 0.
+        USAGE_RECORDING_ENV_VAR: "true",
+    }
+
+    return subprocess.Popen(
+        [sys.executable, "-c", _WRITER_SCRIPT, str(directory), str(count)],
+        env=child_env,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+
+
+def _await_usage_writer(child: subprocess.Popen) -> None:
+    _, stderr = child.communicate(timeout=_SUBPROCESS_WRITER_TIMEOUT_SECONDS)
+    assert child.returncode == 0, stderr
 
 
 def test_the_path_ignores_every_environment_derived_data_directory(monkeypatch):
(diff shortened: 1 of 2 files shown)


backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -222,19 +275,34 @@ def test_the_run_id_is_inherited_when_it_is_safe(usage_directory, monkeypatch):
     assert parse(read_lines(usage_directory)[0])["run_id"] == "abc12345"
 
 
-def test_concurrent_writers_never_truncate_a_line(usage_directory):
-    writes = 200
+def test_concurrent_subprocess_writers_never_truncate_a_line(usage_directory):
+    # Two separate instances sharing the log — the case ``_append_line``'s
+    # ``O_APPEND`` claim is about. An in-process ThreadPool would still pass if
+    # every writer funnelled through one shared buffered file object.
+    writes_per_writer = 100
```
aidemsined: > [!CAUTION]
> 100 writes per child is under-powered — the test passes a genuinely broken implementation about 10% of the time.

I checked this by falsification rather than by reading: patching `_append_line` to emit each line as two `os.write` calls (keeping `O_APPEND`, so this is the real interleave mechanism rather than a strawman) and running this test 20 times. It caught the break in **18/20 runs**. Two runs passed a broken implementation. The malformed-line yield in a detecting run is only 1–9 lines out of 200, so the margin is thin.

Raising `writes_per_writer` to 500 gave **20/20** detection, at a cost of +0.02s (0.12s → 0.14s).

Worth noting the good news first: the test is sound. Dropping `O_APPEND` entirely fails hard, the buffered-`open(path, "a")` substitution passes for the correct reason (mode `"a"` implies `O_APPEND`), and the two children genuinely overlap — measured start skew is 0.1–5ms against a 33–60ms write window, so ~95% of each window is concurrent. This isn't the vacuity #1871 was filed about.

The residual concern is that the overlap is incidental rather than enforced. It holds because interpreter startup happens to be ~35ms for both children and the write loop is ~35ms; a slow-import regression anywhere in `usage`'s import chain could shrink that window with no test failing. Raising the write count widens it and fixes the detection rate in the same change, which is why I'd prefer it to adding a rendezvous barrier.

**Suggested action:** `writes_per_writer = 500`.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -65,6 +82,42 @@ def total_events(lines):
...
+        _, stderr = child.communicate(timeout=_SUBPROCESS_WRITER_TIMEOUT_SECONDS)
+    except subprocess.TimeoutExpired:
+        child.kill()
+        _, stderr = child.communicate()
+        pytest.fail(
+            f"usage writer timed out after {_SUBPROCESS_WRITER_TIMEOUT_SECONDS}s"
+            f"{f': {stderr}' if stderr else ''}"
+        )
+
+    assert child.returncode == 0, stderr
```
aidemsined: > [!CAUTION]
> This exit-code check can't catch the failure mode that matters — a child that writes nothing still exits 0, and its stderr is discarded here.

`record_event` swallows every exception and only logs a warning (usage.py:344-349), so no write-path problem can produce a non-zero exit code. This assertion therefore guards import and argv errors only.

Reproduced: with the target directory made unwritable, the child exited **0** while its stderr carried five copies of `Unable to record usage event app_start: [Errno 13] Permission denied: ...`. All of it was thrown away, because `stderr` is only surfaced on a non-zero exit or a timeout. The test then fails as a bare `assert 0 == 200` with no indication of cause — which is the hardest possible failure to diagnose from CI logs.

**Suggested action:** two cheap options, either or both — have `_WRITER_SCRIPT` `sys.exit(1)` unless recording is genuinely enabled (and optionally verify the log actually grew), and return the captured `stderr` from this helper so it can be attached to the count assertion's message at line 300.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -222,19 +275,34 @@ def test_the_run_id_is_inherited_when_it_is_safe(usage_directory, monkeypatch):
...
-    assert len(lines) == writes
-    assert all(
-        line.startswith("ts=") and " event=app_start " in f"{line} " for line in lines
-    )
+    assert len(lines) == total_writes
+    for line in lines:
+        fields = parse(line)
+        assert fields["event"] == "app_start"
+        assert "ts" in fields
+        assert "schema_version" in fields
```
aidemsined: > [!WARNING]
> The structural check is weaker than "well-formed" — a merged fragment can satisfy all three assertions.

`parse` splits on `=` with `maxsplit=1` and later duplicate keys overwrite earlier ones, so a merged line such as `ts=T1 event=app_stts=T2 event=app_start schema_version=1 run_id=x` passes every assertion here. In practice it's caught only because a merge leaves a paired tail fragment elsewhere that trips `parse` — i.e. by luck rather than by this check.

Two things would tighten it. An exact set comparison catches merges directly:

```python
assert set(fields) == {"ts", "event", "schema_version", "run_id"}
```

And the three field names are hand-copied from `usage._REQUIRED_FIELDS` (usage.py:109), which is what production uses to decide whether a line is a fragment (usage.py:421). Add a fourth required field there and this test keeps passing on lines the collector-side code would reject. Importing the constant is already established style in this suite — `test_settings.py` imports `_STRICT_BOOLEANS`, `_ENV_PARSERS` and `_ENV_ALIASES`.

**Suggested action:** assert the exact field set, and derive the required names from `_REQUIRED_FIELDS` rather than restating them.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -222,19 +275,34 @@ def test_the_run_id_is_inherited_when_it_is_safe(usage_directory, monkeypatch):
...
 
     lines = read_lines(usage_directory)
 
-    assert len(lines) == writes
-    assert all(
-        line.startswith("ts=") and " event=app_start " in f"{line} " for line in lines
-    )
+    assert len(lines) == total_writes
+    for line in lines:
+        fields = parse(line)
```
aidemsined: > [!WARNING]
> The one outcome this test exists to detect surfaces as a helper crash rather than an assertion failure.

A line severed mid-token yields a single-element list from `parse` (line 77), so a genuine interleave raises `ValueError: dictionary update sequence element #0 has length 1; 2 is required` from inside the shared helper — not `assert`. Pytest's traceback does show the offending line so it is diagnosable, but the least readable failure in the module being the one for the module's whole reason for existing seems worth fixing.

**Suggested action:** wrap the parse and fail with the fragment quoted, e.g.

```python
try:
    fields = parse(line)
except ValueError:
    pytest.fail(f"truncated or interleaved line: {line!r}")
```

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -222,19 +275,34 @@ def test_the_run_id_is_inherited_when_it_is_safe(usage_directory, monkeypatch):
...
+    writes_per_writer = 100
+    total_writes = writes_per_writer * 2
 
-    with ThreadPoolExecutor(max_workers=8) as executor:
-        for _ in range(writes):
-            executor.submit(record_event, UsageEvent.APP_START)
+    writers = [
+        _spawn_usage_writer(usage_directory, writes_per_writer),
+        _spawn_usage_writer(usage_directory, writes_per_writer),
+    ]
```
aidemsined: > [!WARNING]
> Both children are spawned outside the `try`, so a failing second spawn leaks the first.

If the second `Popen` raises — fork/exec failure on a loaded runner, which is exactly the kind of thing that happens in CI and nowhere else — the first child is never killed or reaped. That's the orphan case `_SUBPROCESS_WRITER_TIMEOUT_SECONDS` exists to prevent, reachable by a path the timeout doesn't cover.

Everything else in the teardown is correct, for the record: the timeout path does `kill()` + `communicate()`, and the `finally` reaps any survivor even when `pytest.fail` raises out of the first await (`Failed` derives from `BaseException`, so the `finally` still runs).

**Suggested action:** move the second spawn inside the `try`, or use `contextlib.ExitStack`. This also falls out for free if the spawn/await pair is collapsed into a single helper.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -222,19 +275,34 @@ def test_the_run_id_is_inherited_when_it_is_safe(usage_directory, monkeypatch):
...
+                writer.kill()
+                writer.communicate()
 
     lines = read_lines(usage_directory)
 
-    assert len(lines) == writes
-    assert all(
-        line.startswith("ts=") and " event=app_start " in f"{line} " for line in lines
-    )
+    assert len(lines) == total_writes
```
aidemsined: > [!WARNING]
> Nothing asserts that both writers actually contributed — 200 lines from one interpreter would satisfy every assertion in this test.

The cross-process premise is the whole point of the change, and it's currently implied by the setup rather than pinned by an assertion. Each child generates its own `run_id`, so it's one line to check:

```python
assert len({parse(line)["run_id"] for line in lines}) == 2
```

That converts "we spawned two writers" from something the test arranges into something the test verifies, and it would fail loudly if a future refactor accidentally serialised or deduplicated the writers.

**Suggested action:** add the distinct-`run_id` assertion after the count check.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -31,6 +32,22 @@
...
+import sys
+from pathlib import Path
+
+from ttnn_visualizer import usage
+from ttnn_visualizer.usage import UsageEvent, record_event
+
+usage.USAGE_DIRECTORY = Path(sys.argv[1])
+for _ in range(int(sys.argv[2])):
+    record_event(UsageEvent.APP_START)
+"""
```
aidemsined: > [!WARNING]
> These eight lines are the only Python in the PR that no quality gate ever inspects.

CI runs `black --check`, `isort --check-only` and `mypy` over `backend/ttnn_visualizer`, and the tests package sits inside that path — so every other line in this change is formatted, import-sorted and type-checked, and this string is not. A typo here is a runtime `SyntaxError` in a child process, surfaced only as a non-zero exit with a traceback buried in captured stderr. It's also not navigable: no go-to-definition, no coverage, no debugger.

The `_import_settings_with` precedent in `test_settings.py` uses `python -c` too, but its payload is a single `import` statement — nothing a linter could have an opinion about. This payload parses argv, mutates a module global and runs a loop; it's program logic, and it grows the moment anything else is needed from the child.

`backend/ttnn_visualizer/tests/` is already a package with `__init__.py`, so a sibling `usage_writer.py` (no `test_` prefix, so pytest won't collect it) runs via `[sys.executable, "-m", "ttnn_visualizer.tests.usage_writer", ...]` with no packaging or `testpaths` change.

**Suggested action:** extract to a real module. Locality is a fair counter-argument, but it's weak against three lost gates — and this is the suite's first subprocess writer, so it sets the pattern for the next one.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -65,6 +82,42 @@ def total_events(lines):
     return sum(int(parse(line).get(COUNT_FIELD, "1")) for line in lines)
 
 
+def _spawn_usage_writer(directory: Path, count: int) -> subprocess.Popen:
```
aidemsined: > [!WARNING]
> The split spawn/await pair makes the incorrect usage as natural to write as the correct one.

The pair is only correct when *every* spawn precedes *every* await. Nothing in the API expresses that — `spawn(); await(); spawn(); await()` type-checks, reads perfectly naturally, and silently degrades the test to two serialised writers, at which point it proves nothing about interleaving and still passes.

The cost also shows at the call site: the test body carries six lines of kill/reap `finally` bookkeeping inside what is otherwise a declarative assertion suite.

A single helper spawns all, joins all with the timeout, and kills stragglers in its own `finally`:

```python
def _run_usage_writers(directory, writers=2, writes_each=500): ...
```

That makes correct usage the only usage, gives the orphan fix (second spawn inside the `try`) one obvious home, and reduces the test body to roughly three lines.

**Suggested action:** collapse into one all-or-nothing helper. Keep it private and module-local; there's no second consumer.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -31,6 +32,22 @@
     record_event,
 )
 
+# Enough for two children to finish 100 appends each on a slow CI runner; a hang
+# should fail the test rather than wedge the suite.
```
aidemsined: > [!WARNING]
> The comment describes a shared budget, but the timeout is applied per child.

"Enough for two children to finish 100 appends each" reads as one budget covering both. `_await_usage_writer` passes this value to each `communicate()` call separately and the children are awaited serially, so the worst-case wall clock for a systematically wedged pair is 60s, not 30s.

Bounded either way, so this is about the comment matching the code rather than a real hang risk.

**Suggested action:** reword to say the budget is per child, or convert to a single deadline shared across both `communicate()` calls.

backend/ttnn_visualizer/tests/test_usage.py
```diff
@@ -31,6 +32,22 @@
...
+_SUBPROCESS_WRITER_TIMEOUT_SECONDS = 30
+
+_WRITER_SCRIPT = """
+import sys
+from pathlib import Path
+
+from ttnn_visualizer import usage
+from ttnn_visualizer.usage import UsageEvent, record_event
+
+usage.USAGE_DIRECTORY = Path(sys.argv[1])
```
aidemsined: > [!WARNING]
> The child re-implements part of the `usage_directory` fixture, and depends on the rest of it invisibly.

The fixture sets `USAGE_DIRECTORY`, resets `usage._run_id`, and clears both env vars. The child sets only `USAGE_DIRECTORY` and re-adds `USAGE_RECORDING_ENABLED`. It ends up correct on the run-id front only because `monkeypatch.delenv` has already mutated the parent's `os.environ` by the time `_spawn_usage_writer` copies it into `child_env` — so the helper silently requires being called from inside that fixture, and nothing in its signature or docstring says so.

The consequence is drift: add a new reset to the fixture later (a `disabled` marker, another env var) and the subprocess half of the suite silently won't have it, which is the failure mode that produced this ticket in the first place.

Also minor, on the docstring above at line 88-89 — it explains that the child sets `USAGE_DIRECTORY` itself, but the assignment it's describing lives here in the script string, twenty lines earlier. The two halves of one mechanism are documented apart from each other.

**Suggested action:** make the fixture dependency explicit — either document it as the helper's precondition, have the fixture own the spawn, or give the child the same reset list the fixture applies.

aidemsined (commented):
The headline is that this works: I tried to falsify it rather than just read it, and it holds up.

| Mutation to `_append_line` | Result |
|---|---|
| Drop `os.O_APPEND` | **Fails** — 1 line vs 200 |
| Keep `O_APPEND`, emit the line as two `os.write` calls | **Fails** in 18/20 runs, with real mid-line fragments |
| Replace the hand-rolled loop with buffered `open(path, "a")` | Passes — correctly, since `"a"` implies `O_APPEND` and one 73-byte flush is one syscall |
| Unmodified | Passes |

The two children genuinely overlap — measured start skew 0.1–5ms against a 33–60ms write window, so ~95% of each window is concurrent. The old `ThreadPoolExecutor` test is fully gone, the child writes into `tmp_path` rather than a real `~/.ttnn-visualizer/usage/`, and recording is genuinely enabled in the child. This does not reproduce the overclaim #1871 was filed about.

The two things worth acting on are that 100 writes is under-powered as a guard (18/20, not 20/20), and that a child which fails but exits 0 throws its own explanation away. Both are cheap.

## General observations

> [!NOTE]

- The short-write / `written == 0` branch in `_append_line` (usage.py:311-317) can't be reached by any concurrency test — a partial `os.write` never happens for a 73-byte record on a regular local file. Only a unit test monkeypatching `os.write` to return a short count covers it; the existing failure test only exercises the raising path. Worth a follow-up issue rather than anything here.
- This is now the second in-tests subprocess idiom with no shared home: `_import_settings_with` in `test_settings.py` uses `subprocess.run` with env-based child config, this uses `Popen` with argv, text pipes and an explicit timeout. Neither is wrong, but there are now two answers to "how do we run something in a fresh interpreter here", and `AGENTS.md` asks for shared fixture helpers over repeated ad-hoc setup.
- The `O_APPEND` contract still lives only in `_append_line`'s docstring, with nothing at the production site pointing back at this test. `CONVENTIONS.md` carries an entry explaining why the *other* out-of-process test must run in a subprocess, but never mentions `usage.py`, the usage log or `O_APPEND` at all. Without that record, a later cleanup pass could swap this back to a ThreadPool for speed and nothing would object — which is how the original overclaim arose.
- `multiprocessing` with `spawn` looks like it would remove the inline source and the argv marshalling, but it doesn't: a `spawn` target must be a module-level function anyway, and the default start method on Linux is `fork`, whose child inherits the parent's monkeypatched globals and warm imports — i.e. not a separate instance in the sense the docstring claims. The `subprocess` choice is right; one sentence in the test comment saying so would stop it being re-litigated.
- Writer count is an unnamed literal in two places (`writes_per_writer * 2` and the two-element list) which must agree by hand. Falls out for free if the spawn/await pair becomes a single helper taking `writers=2`.
- The kill/reap `finally` in the test body is now redundant with the helper's own kill/reap, which arrived in the second commit. Either drop it, or note that it covers the spawn-succeeded-but-await-never-ran case.
- `_spawn_usage_writer`'s return annotation is bare `subprocess.Popen` where `subprocess.Popen[str]` is accurate given `text=True`, and valid at runtime on the 3.10 floor. `stdout=subprocess.PIPE` is set but never inspected — correct as output suppression, just undocumented.
- If the first writer fails, the second is killed and its stderr dropped, so when both children hit the same environmental problem only one diagnostic is ever shown.
- No test-module move warranted. The repo has no integration-test split and `test_settings.py` already carries an inline subprocess helper; extracting the harness restores the top-of-file readability without inventing a directory convention.
- Cost is negligible: 0.12s for the two-child run, 0.22s for the whole invocation warm, and the 30s timeout is ~250× headroom. `sys.executable` is used so the child runs the same venv. CI is `ubuntu-latest` only with no xdist, so no Windows or parallel-worker exposure. Worth knowing that `_append_line` does a `mkdir` + `open` + `close` per line, so this test drives 200 `mkdir` syscalls and raising the write count scales slightly super-linearly (1000/child ≈ 0.29s) — irrelevant at the real event rate of ~250/day.

The commit split is worth keeping in the merge message: the second commit fixed `_await_usage_writer` so a hung child is killed and reaped with its stderr surfaced, rather than raising a bare `TimeoutExpired` with the output discarded. That's invisible in the squashed diff.

---

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Done — `writes_each = 500` in 660c13b2. Confirmed the same way you found it: with `_append_line` patched to emit each line as two `os.write` calls, the test now fails 5/5.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Done, both halves. The child is now a real module (`backend/ttnn_visualizer/tests/usage_writer.py`) that checks `is_recording_enabled()` before writing and reads the log back afterwards, exiting 1 with a stderr message on either failure. `_run_usage_writers` surfaces that via `assert child.returncode == 0, stderr`.

Verified against an unwritable directory: exit 1, with the five swallowed `Permission denied` warnings carried out in stderr rather than discarded.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Tightened — but not with the exact set comparison, because that doesn't catch the case you demonstrate.

`_format_line` emits exactly four fields for a bare `record_event(APP_START)`, so both halves of a merge carry the identical key set and the seam leaves it unchanged. Your own example, `ts=T1 event=app_stts=T2 event=app_start schema_version=1 run_id=x`, has keys `{ts, event, schema_version, run_id}` — exactly the expected set, so `set(fields) == {...}` passes it.

Token count is what catches it:

```python
assert len(line.split(" ")) == len(_REQUIRED_FIELDS) + 1
assert set(fields) == {*_REQUIRED_FIELDS, RUN_ID_FIELD}
```

The set comparison is kept alongside it and the names are derived from `_REQUIRED_FIELDS` as you suggested, noting it's a 3-tuple so `run_id` is added explicitly. With `_append_line` deliberately broken, the token-count assertion is the one that fires: `AssertionError: assert 5 == (3 + 1)`.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Done, as suggested. It's no longer the assertion that fires first for a merge — the token-count check on the thread above catches that — but it still covers a line severed with no `=` left in the fragment.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Done — both spawns are inside the `try` now, which fell out of the helper collapse in the line 85 thread as you predicted.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Done — `assert len({fields[RUN_ID_FIELD] for fields in parsed}) == writers` after the count check.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Agreed — extracted to `backend/ttnn_visualizer/tests/usage_writer.py`, run as `-m ttnn_visualizer.tests.usage_writer`. No packaging or `testpaths` change needed, and it resolves through the existing editable install exactly as the `python -c` payload did.

It's also where the exit-code checks from the line 118 thread now live, which makes the locality argument weaker still: that logic would have been a third and fourth line of unlinted string.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Done — `_run_usage_writers(directory, writers, writes_each)` spawns all, joins all against one deadline, and kills stragglers in its own `finally`. The test body is assertions only now.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Fixed by converting rather than rewording, since the helper collapse gave the shared deadline somewhere to live: one `time.monotonic()` deadline is computed up front and the remaining time passed to each `communicate()`, so 30s is now genuinely the budget for the set. Comment updated to match.

backend/ttnn_visualizer/tests/test_usage.py
dcblundell: Documented as a precondition on `_run_usage_writers`, naming the fixture and the reason: the children inherit a copy of `os.environ`, and it is that fixture's `delenv` that keeps a developer's own `TTNN_VISUALIZER_RUN_ID` from reaching them and collapsing their run ids into one. The new distinct-run-id assertion from the line 300 thread now fails loudly if that dependency is ever broken, which the drift you describe would have done silently.

The split docstring is gone too — the `USAGE_DIRECTORY` override and its rationale sit together in `usage_writer.main` rather than twenty lines apart.

PR #1875, opened 2026-08-13
Multi-select behavior for performance chart device time filters

### LINKED ISSUE #1868: Performance: shift-click union and click-again-to-clear on the histogram bucket controls
Follow-up to #1754, which #1843 closes. #1843 landed duration-bucket filtering with the chart controls, the Table tab chip, report-change reset, and unit tests. Union and clear are currently only reachable from the chip, not from the histogram itself.

## Goal

Make the bucket controls beneath the histogram columns support the full selection gesture:

- **shift-click** a control → add/remove that bucket from the selection (union), rather than replacing it
- **click the sole selected control again** → clear the filter

## Why it isn't already done

`usePrefilterPerfTableByDurationBucket` replaces the selection *and* navigates to the Table tab, so shift-click needs to suppress that navigation to be usable — otherwise the user is moved off the charts tab before they can add a second bucket. `usePrefilterPerfTableByOpCode` has the same replace-and-navigate shape, so the two should be designed together rather than one growing a special case.

`ClickAnnotationEvent` exposes `event: MouseEvent`, so `shiftKey` is available on the existing handler.

## Acceptance criteria

- [ ] Shift-clicking a bucket control adds it to the selection and stays on the Charts tab
- [ ] Shift-clicking a selected bucket control removes it from the selection
- [ ] Clicking the sole selected control again clears the filter
- [ ] Plain click keeps its current behaviour: replace the selection and move to the Table tab
- [ ] `usePrefilterPerfTableByOpCode` and `usePrefilterPerfTableByDurationBucket` share one shape for "amend the selection without navigating"
- [ ] Unit tests cover union, removal and clear; the existing assertions that a plain click *replaces* the selection stay green

## Notes

- Keyboard and assistive-technology access to the bucket controls is tracked separately on #1737; this issue is about the pointer gesture only.
- `src/hooks/usePrefilterPerfTableByDurationBucket.ts` documents replace-not-union as the current scope boundary and points here.

## Related

- #1754 — duration-bucket filtering (delivered by #1843)
- #1843 — the PR this defers from
- #1737 — keyboard-operable path for chart-driven filtering

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1868.

This pull request introduces multi-select and toggle behavior for performance chart-driven table filters, allowing users to add or remove duration buckets or op codes from the filter without navigating away from the charts tab. It also updates the UI and documentation to clarify these new interactions and adds supporting utility code and comprehensive tests.

**Enhancements to chart-driven table filtering:**

* Added multi-select and toggle support for histogram bucket and op code filters in the performance charts. Users can now shift-click to add/remove filters or click the selected filter again to clear it, all without leaving the charts tab. The filter logic now distinguishes between replacing and amending the selection, and the relevant hooks (`usePrefilterPerfTableByDurationBucket`, `usePrefilterPerfTableByOpCode`) were updated to accept an options parameter for this behavior. [[1]](diffhunk://#diff-f338f0198973527b73857e4934fb7d752adfd3dac9fcec0168b5d310aae540a0L28-R39) [[2]](diffhunk://#diff-197ea8fc2173363a90a2b0c5fa8fb437cd830e1bc8ea30b0af11d6519ece255eR7-R34) [[3]](diffhunk://#diff-ec55bb0b6550e798236ca0512103282bb3f063058547fa6fb8275758deaf1495R38-R43) [[4]](diffhunk://#diff-157c6efafae4a38393be87d8dc4e57235f676134ed65ce582fcd5755b7c46510R1-R12)

* Updated the histogram bucket annotation positioning to use a paper fraction and center anchor, ensuring that controls remain visually aligned with their columns even after incremental redraws by Plotly. This fixes a UI bug where selected controls could jump sideways.

**User interface and documentation updates:**

* Improved the chart hint message to explain the new filter interactions, including shift-click to add/remove and click-again-to-clear.

**Testing improvements:**

* Added and enhanced tests to verify the new toggle/amend filter behavior for both duration buckets and op codes, as well as the improved annotation positioning logic. Test helpers and mocks were updated to support simulating shift-click and other mouse events. [[1]](diffhunk://#diff-1aac51f1160139f717670a6b3c96cc7646bd8dcf7e25725f15da2e94b6af29aaR386-R412) [[2]](diffhunk://#diff-e4b7c4112a06675354e43c18aa3dbf6814cd2875236ea3a23f4b34b1e8e5b12dR97-R170) [[3]](diffhunk://#diff-2f67405eb82c3d1b3feeabce62a7565a283299b7ad4f352c9e46ae4ed0f6a9d5R87-R141) [[4]](diffhunk://#diff-6efc0831e4dfff1e4e4067bddc9702e88fc5c04bb3bfafc919d7d0befaa5ba9aL52-R52) [[5]](diffhunk://#diff-6efc0831e4dfff1e4e4067bddc9702e88fc5c04bb3bfafc919d7d0befaa5ba9aL61-R61) [[6]](diffhunk://#diff-1aac51f1160139f717670a6b3c96cc7646bd8dcf7e25725f15da2e94b6af29aaL279-R280) [[7]](diffhunk://#diff-1aac51f1160139f717670a6b3c96cc7646bd8dcf7e25725f15da2e94b6af29aaL304-R321) [[8]](diffhunk://#diff-2f67405eb82c3d1b3feeabce62a7565a283299b7ad4f352c9e46ae4ed0f6a9d5R37-R42) [[9]](diffhunk://#diff-e4b7c4112a06675354e43c18aa3dbf6814cd2875236ea3a23f4b34b1e8e5b12dR33-R44)

**Supporting utility code:**

* Introduced a reusable `toggleListMembership` function to handle toggling values in filter lists, used by both duration bucket and op code filters.

These changes collectively improve the usability and flexibility of performance chart filtering, making it easier for users to explore data without unnecessary navigation.


tests/usePrefilterPerfTableByOpCode.spec.tsx
```diff
@@ -78,4 +84,59 @@ describe('usePrefilterPerfTableByOpCode', () => {
...
+                    [perfSelectedTabAtom, PerfTabIds.CHARTS],
+                ]}
+            >
+                <Probe />
+            </TestProviders>,
+        );
+
+        fireEvent.click(screen.getByRole('button', { name: 'amend-matmul' }));
+
+        expect(screen.getByTestId('raw-op-filter')).toHaveTextContent('Conv2d');
```
aidemsined: > [!CAUTION]
> `toHaveTextContent` is a substring match, so this removal test cannot fail on the behaviour it names.

Starting from `['Conv2d', 'Matmul']` and amend-removing `Matmul`, the assertion checks that the rendered text *contains* `Conv2d`. A complete no-op regression leaves `Conv2d,Matmul` — which contains `Conv2d` — and passes. So does a duplicate-append regression.

The same defect is on `:71`, which matters more because that's the AC6 "plain click still replaces" assertion: from `['Conv2d','Softmax']`, clicking `Matmul` asserts `toHaveTextContent('Matmul')`, and a union regression producing `Conv2d,Softmax,Matmul` still contains `Matmul` and passes. `:85` is weak for the same reason, and `:102`'s `'Conv2d,Matmul'` would tolerate an extra appended entry.

The sibling bucket spec already does this correctly — exact `.textContent).toBe('1,10')` at `:68, :82, :95, :112, :131, :149`. The two specs currently disagree on assertion style and only one of them can actually fail.

**Suggested action:** switch this spec to exact comparisons, e.g. `expect(screen.getByTestId('raw-op-filter').textContent).toBe('Conv2d')`, and apply the same at `:71`, `:85` and `:102`. `:139`'s `toHaveTextContent('')` happens to work (jest-dom special-cases an empty expectation) but is fragile — `.textContent).toBe('')` matches the bucket spec at `:149`.

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -125,9 +134,13 @@ function PerfDurationHistogram({
                 return;
             }
 
-            prefilterPerfTableByDurationBucket(entry.bucket.minUs);
+            const { minUs } = entry.bucket;
+            const isSoleSelected = selectedBucketMinUsList.length === 1 && selectedBucketMinUsList[0] === minUs;
```
aidemsined: > [!CAUTION]
> This measures "sole selected" against the whole cross-report filter list, not against the buckets this histogram actually draws — so click-again-to-clear can silently become replace-and-navigate.

`selectedBucketMinUsList` is the global `durationBucketFilterListAtom` (line 75). The chart's buckets come from `buildDurationHistogram(rows)` (line 73) with `rows = filteredPerfData` (PerfCharts.tsx:28) — active report only, already narrowed by the op-code legend. But the Table-tab chip populates that same atom from `buildLogDecadeBuckets(combinedRows)` (usePerfReportFiltering.ts:91), which spans active **plus** comparison rows. The filter's value space is strictly larger than the chart's.

Reachable sequence: on the Table tab, select decades `[1, 1000]` from the chip where the `1000` decade exists only in a comparison report (or belongs to an op code the user then deselects in the legend). Switch to Charts — the histogram draws and highlights exactly **one** control. Plain-click it. The user sees a sole selected control and a hint promising a clear, but `length === 2`, so `shouldAmend` is false: the filter is replaced with `[1]` and they're thrown to the Table tab.

No test covers a selection containing a bucket this chart doesn't render.

**Suggested action:** scope the test to what the chart draws, e.g. compare against the intersection of `selectedBucketMinUsList` with `histogramData.buckets`. Deciding this inside the hook instead (see the note on line 141) would make the scoping natural, since the hook could compare against the bucket list it's handed.

src/hooks/usePrefilterPerfTableByDurationBucket.ts
```diff
@@ -25,10 +25,18 @@ export function usePrefilterPerfTableByDurationBucket() {
...
-        (minUs: DurationBucket['minUs']) => {
-            setFilters([minUs]);
+        (minUs: DurationBucket['minUs'], options?: PerfTablePrefilterOptions) => {
+            if (options?.amend) {
+                setFilters((current) => toggleListMembership(current, minUs));
+            } else {
+                setFilters([minUs]);
+            }
+
             setIsStackedView(false);
```
aidemsined: > [!WARNING]
> This now fires on the amend and clear paths too, which makes the gesture non-idempotent.

It sits outside both `amend` guards, so every shift-click and every click-to-clear resets the stacked view. The docstring above (`:18-20`) justifies the reset by the user landing on a table that would otherwise show neither the filter nor the tag explaining it — but `amend` exists precisely to suppress that navigation, so the justification doesn't reach this path.

It's worst on the clear case: a user in stacked view goes to Charts, shift-clicks a bucket on and back off, and returns to the Table tab with an empty filter *and* their view mode silently reverted. Nothing on screen explains it, because the subscribers to `isStackedViewAtom` are unmounted while they're on Charts (`renderActiveTabPanelOnly`, Performance.tsx:322).

The existing test at `:153-170` pins the current behaviour for the amend-with-a-result case, so that half looks deliberate — it's the clear-to-empty case that reads as unconsidered.

**Suggested action:** gate on `!options?.amend`, or on the resulting selection being non-empty. The test would need updating either way.

src/definitions/PerfDurationHistogram.ts
```diff
@@ -46,7 +46,7 @@ export const PERF_DURATION_HISTOGRAM_EMPTY_MESSAGE = 'No device ops available fo
 export const PERF_DURATION_HISTOGRAM_ACTIVE_REPORT_SUBTITLE = 'Active report only';
 
 export const PERF_DURATION_BUCKET_FILTER_HINT =
-    'Click a duration range below the chart to filter the table by op duration';
+    'Click a duration range below the chart to filter the table; shift-click to add or remove ranges, or click the selected range again to clear';
```
aidemsined: > [!WARNING]
> "click the selected range again to clear" is only true when it's the *sole* selection — with two selected, the user gets the opposite.

`isSoleSelected` requires `selectedBucketMinUsList.length === 1` (PerfDurationHistogram.tsx:138). With two ranges selected, plain-clicking one of them takes the replace path: the selection collapses to just that range and the user is navigated to the Table tab. Someone following this hint after building a multi-selection gets exactly what the hint told them wouldn't happen.

**Suggested action:** "…or click the only selected range again to clear".

tests/PerfDurationHistogram.spec.tsx
```diff
@@ -371,6 +383,33 @@ describe('PerfDurationHistogram duration bucket controls', () => {
         expect(getSelectedFlags()).toEqual([false, true]);
     });
```
aidemsined, line 408: > [!WARNING]
> The branch most likely to regress — a plain click while two or more buckets are selected — has no test.

`:378` starts from a 1-item selection of a *different* bucket, and `:404` from a 1-item selection of the *same* bucket. Nothing exercises a plain click against a multi-selection, which is the case that distinguishes replace-semantics from toggle-semantics.

Concretely: rewrite `PerfDurationHistogram.tsx:138` as the obvious "simplification" `selectedBucketMinUsList.includes(minUs)` and every test in this suite still passes — yet plain-clicking one of two selected buckets would now *remove* it rather than collapsing the selection to it, and would stay on the Charts tab instead of navigating.

Also worth noting `:378` never asserts the tab moved to `TABLE`, so the replace test doesn't guard the navigation half of AC4 either.

**Suggested action:** add a case starting from a two-bucket selection, plain-clicking one, asserting the selection collapses to that bucket and the tab moves to `TABLE`. A component-level shift-click removing the *last* remaining bucket would also be worth having — only the hook-level spec covers the clear path today.

src/functions/toggleListMembership.ts
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+/** Append `value` if absent, otherwise remove it — the MultiSelectField / chart-amend toggle. */
+export function toggleListMembership<T>(list: readonly T[], value: T): T[] {
```
aidemsined: > [!WARNING]
> The docstring names `MultiSelectField` as a consumer, but it wasn't converted — so this adds another copy of the pattern rather than collapsing the existing ones.

`MultiSelectField.tsx:34-39` still carries a character-for-character copy of these three lines and doesn't import this helper. So does `SimpleMultiselect.tsx:24` in one-line form. `PerfTableToolbar.tsx:44-54` is a near-copy with an extra idempotency branch, and `PerfChartFilter.tsx:104-106` is the same shape over `Marker` objects. Only the two new hooks call this.

That's comfortably past the ≥3 threshold that justifies a helper, so the consolidation is worth finishing — `MultiSelectField`'s `updateHandler: Dispatch<SetStateAction<T[K][]>>` takes this signature directly.

One caveat if you do: `PerfChartFilter` toggles object identities, and `includes`/`!==` here is SameValueZero, so it must **not** be converted without a value-equality variant. That's also the argument for tightening the signature:

```ts
export function toggleListMembership<T extends string | number>(list: readonly T[], value: T): T[]
```

Both current call sites pass primitives (`minUs: number`, `opCode: string`), so nothing breaks today — but the generic `<T>` advertises more than the implementation delivers, and the docstring points a future reader straight at the two generic components where it would silently misbehave.

**Suggested action:** adopt it at the primitive call sites and constrain the type parameter — or, if the conversion is out of scope, drop the `MultiSelectField` claim from the comment.

tests/PerfDurationHistogram.spec.tsx
```diff
@@ -301,12 +302,23 @@ describe('PerfDurationHistogram duration bucket controls', () => {
...
-        expect(getAnnotations().map((annotation) => annotation.x)).toEqual(traceCategories);
+
+        const annotations = getAnnotations();
+        // Anchored to the plotting area by fraction, not to the x axis: an axis-referenced x is
+        // resolved against a stale category range when react-plotly redraws a single re-styled
+        // annotation, jumping the selected control sideways (#1868). (index + 0.5) / count centres
+        // each control over its evenly-spaced column, and 'center' holds it there.
+        const bucketCount = traceCategories?.length ?? 0;
+        expect(annotations.map((annotation) => annotation.x)).toEqual(
+            traceCategories?.map((_, index) => (index + 0.5) / bucketCount),
```
aidemsined: > [!WARNING]
> This recomputes the implementation's own formula, so it's a change-detector for the expression rather than a check of the result — and it's weaker than the assertion it replaced.

`(index + 0.5) / bucketCount` here is the same expression as `PerfDurationHistogram.tsx:107`, so the two can only ever agree. The previous version compared `annotation.x` against the trace's category labels — an independent value — so this assertion lost independence in the diff.

The `xref === 'paper'` and `xanchor === 'center'` checks below it are genuine literals and do catch a revert, so the test isn't worthless. But the positioning maths itself is unpinned.

Worth being explicit about a second limitation: the bug being fixed is a Plotly rendering artifact during an incremental redraw *after a selection*, and a mocked `Plot` under jsdom can't reproduce it. This test asserts static layout props on first render and never clicks, so the actual #1868 symptom is unverified by CI either way.

**Suggested action:** hardcode `[0.25, 0.75]` for the two-bucket fixture, and add a second assertion after `clickBucket(0)` so the test at least proves the fractions survive a selection change.

src/hooks/usePrefilterPerfTableByOpCode.ts
```diff
@@ -4,21 +4,34 @@
...
 
-            setFilters([opCode]);
-            showPerfTable();
+            if (options?.amend) {
+                setFilters((current) => toggleListMembership(current, opCode));
+            } else {
+                setFilters([opCode]);
+            }
+
+            if (!options?.amend) {
```
aidemsined: > [!WARNING]
> The two hooks are now near-identical, and the one place they differ looks accidental rather than intended.

Both bodies read: guard, `if (options?.amend) setFilters(cur => toggleListMembership(cur, v)) else setFilters([v])`, then `if (!options?.amend) showPerfTable()`. The only differences are the atom, the empty-value guard, and the fact that the bucket hook also calls `setIsStackedView(false)` — which this hook has no equivalent of at all.

AC5 asks the two to "share one shape for amend without navigating". Right now that's enforced by a shared *type* and two independent implementations of the same branching, with the `!options?.amend` condition duplicated across two files and needing to stay in sync. A small shared internal — something like `useApplyPerfTablePrefilter(atom, { afterReplace })` — would make the shared shape a property of the code rather than of a type annotation.

**Suggested action:** extract the common body, and confirm the stacked-view asymmetry is deliberate. If op-code filtering should also reset the stacked view, that's a bug here; if it shouldn't, it's worth a comment saying why the two differ.

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -125,9 +134,13 @@ function PerfDurationHistogram({
...
+            const shouldAmend = Boolean(event.event?.shiftKey) || isSoleSelected;
+
+            prefilterPerfTableByDurationBucket(minUs, shouldAmend ? { amend: true } : undefined);
```
aidemsined: > [!WARNING]
> The component decides the policy and hands the hook an outcome, which inverts the ownership and overloads `amend`.

The hook owns the filter atom and already has the current list in hand inside `setFilters((current) => …)`, so it's better placed to decide replace vs toggle vs clear than the caller is. As written, the component has to read `durationBucketFilterListAtom` and re-derive the rule — it happens to have that atom for rendering, but the op-code side doesn't, so completing AC5 there would mean pulling a new atom read into a component purely to re-derive the same policy. That's one rule implemented twice, over two atoms, in two components.

It also means `amend: true` currently encodes two different inputs — "the user shift-clicked" and "a plain click landed on the sole selection" — collapsed at the call site. The name describes the first but not the second.

Passing the gesture rather than the decision (`{ shiftKey }` or `{ additive: true }`) and letting the hook translate would fix the naming and the overload together. It would also fix the scoping bug flagged on line 138 by construction, since the hook could compare against the bucket list it's given rather than the global filter.

**Suggested action:** worth considering as a follow-up rather than blocking this PR — but it's the change that makes the op-code half cheap to finish.

src/components/performance/PerfDurationHistogram.tsx
```diff
@@ -90,14 +90,23 @@ function PerfDurationHistogram({
...
-                x: bucket.label,
-                xref: 'x',
+                // Pin each control to its column by paper fraction, not an axis-referenced x.
+                // Selecting a control flips its fill, so react-plotly redraws that single annotation in
+                // place; Plotly's incremental redraw resolves an axis-referenced x against a stale
+                // category range, and the selected control jumps sideways (#1868). A paper x sidesteps
+                // the axis: the columns are evenly spaced, so (index + 0.5) / bucketCount lands on the
+                // same centre as the bar, and it holds through the redraw. 'center' keeps the box on
+                // that centre — paper's default 'auto' anchor would pull the edge columns inward.
+                x: (index + 0.5) / bucketCount,
```
aidemsined, line 114: > [!WARNING]
> The formula is correct, but it silently depends on a Plotly invariant the comment doesn't name.

`(index + 0.5) / bucketCount` is exact only because Plotly's bar cross-trace calc autoranges a category position axis to `[-0.5, N-0.5]` (`minDiff === 1`, `padded: false`), which maps category centre `i` to paper fraction `(i + 0.5) / N`. I checked this holds at N=1 as well, `xref: 'paper'` is relative to the plot area so the `l: 50` margin can't skew it, and `xaxis.fixedrange: true` (PlotConfigurations.ts:226) rules out zoom/pan drift. The fix is right, and the explanation of *why* `xanchor: 'center'` is required rather than incidental is genuinely good.

The gap is that `[-0.5, N-0.5]` is the entire load-bearing step and isn't stated. Anyone later adding an explicit `xaxis.range`, a numeric x axis, a `categoryorder`, or a non-bar overlay trace to this chart detaches every control from its column — with no visible error and no failing test, since the spec recomputes this same expression.

**Suggested action:** one sentence in the existing comment block naming the assumption. This also looks like a good first entry for the chart-interaction rules in the `AGENTS.md` / `CONVENTIONS.md` pair, which don't currently cover annotation anchoring.

aidemsined (commented):
The gesture logic is right and the in-code prose is unusually good — the comment explaining why `xanchor: 'center'` is required rather than incidental is exactly the kind of thing that saves the next person, and retiring the old "shift-click union is #1868" docstring rather than leaving it stale is the right instinct.

Two things I'd want resolved before merge: the op-code spec's assertions are substring matches that can't fail on the behaviour they name, and `isSoleSelected` is measured against the global filter list rather than the buckets this chart draws, which makes click-again-to-clear misfire in a reachable case.

Three hypotheses I chased that came back **clean**, recorded so nobody re-checks them:

- **Mock fidelity is fine.** Production reads `event.event?.shiftKey` and the mock builds `{ index, event: mouseEvent }`, which is the same shape Plotly really emits on `plotly_clickannotation` — a partial event, not an invented one. No manufactured passes.
- **No empty-array hazard.** Clearing yields `[]`, which is the atom's own default and is gated as "no filter" by `durationBucketFilterSet.size > 0` and `selectedMinUsSet.size === 0`. The user sees all rows, not zero — which is what makes the AC3 clear work at all.
- **Navigation suppression is total.** `useShowPerfTable` is the only writer of `perfSelectedTabAtom` outside the `<Tabs onChange>` binding and the only caller of `window.scrollTo`, so nothing else can move the user off Charts during an amend.

## General observations

> [!NOTE]

- Both hints now render together on the histogram. `PerfChartFrame.tsx:37` prepends `PERF_CHART_TABLE_FILTER_HINT` whenever the chart is clickable, so "Click an operation to filter the performance table" sits directly above the new shift-click hint — two different interaction contracts on one chart with nothing explaining the difference. Both are individually correct, since the bars and the controls beneath genuinely obey different rules, and leaving the op-code hint alone was the right call rather than an omission. Worth revisiting once the op-code path is settled.
- `xanchor: 'center'` gives up Plotly's edge-pull, which exists to stop edge annotations being clipped. With `r: 0` the last control's right half now reaches the container edge — fine at the fixture width, tightest at high decade counts in a narrow container.
- Latent stale-handler trap worth a one-line note: `handleAnnotationClick` now changes identity on every selection change, and react-plotly only re-binds handlers when `data`/`layout`/`config` change by reference. It's safe today only because `bucketAnnotations` has the same dependency set, so the layout reference changes in lockstep. Narrow that memo later and the handler goes stale silently, with `isSoleSelected` reading a frozen list.
- `event.event?.shiftKey` uses optional chaining on a field the Plotly types declare non-optional, and the mock always supplies it — the defensive branch is dead in both production and test.
- `toggleListMembership` has no spec of its own, despite being a new shared `src/functions/` export and the repo covering helpers at that granularity elsewhere.
- A bucket that's selected but empty in the active report renders filled yet carries `captureevents: false`, so it's a visibly-selected control that can't be clicked. Pre-existing, but the new hint now explicitly invites clicking it.
- The commit subject `Add shift-click union` describes about half of what landed. The annotation repositioning is the higher-risk change in the diff — it touches how every control is placed and rests on an unstated Plotly assumption — and a future bisect for "controls are in the wrong place" won't find it from that subject. It does belong in this PR though: before amend existed, every bucket click navigated away, so the in-place restyle was never observed. I'd push back on splitting it out.
- `toggle*` implies mutation for a function that returns a new array; the repo's prefix table has no `toggle*` row. Marginal.
- The updated hint is a ~130-character sentence carrying three rules behind a semicolon and a comma. `PerfChartFrame` already accepts `hints?: string[]` and renders one per entry, so a second constant would use the existing API for what it was built for.
- Performance is better than I expected going in: the Table panel is unmounted while on Charts (`renderActiveTabPanelOnly`), so amending does **not** re-run the row-filtering pipeline per click. Each click re-renders only the histogram and issues one `Plotly.react` layout pass over ~100 SVG rects. This only becomes visible if the legend cap or bucket count grows a lot.

Nothing was run from here — `npm run lint`, `tsc --noEmit` and the vitest specs are unverified. So is the actual #1868 symptom: no test in this diff can reproduce the Plotly redraw jump under a mocked `Plot`, so it's worth shift-clicking through several buckets in a browser before merge.

---

src/definitions/PerformanceCharts.ts
```diff
@@ -35,6 +35,12 @@ export const PERF_CHART_TABLE_FILTER_HINT = 'Click an operation to filter the pe
 
 export type OnOpCodeClick = (opCode: string) => void;
```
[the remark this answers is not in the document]
dcblundell: Leaving `OnOpCodeClick` one-arg for now. #1868 is scoped to the histogram bucket controls ("this issue is about the pointer gesture only"), and AC5 asks the two hooks to *share one shape* for amend-without-navigating, which the shared `PerfTablePrefilterOptions` gives them. Widening the chart click type with no caller that forwards a modifier would be speculative — the shape is in place for whenever an op-code chart grows the gesture.

src/hooks/usePrefilterPerfTableByOpCode.ts
```diff
@@ -4,21 +4,34 @@
...
 export function usePrefilterPerfTableByOpCode() {
     const showPerfTable = useShowPerfTable();
     const setFilters = useSetAtom(rawOpCodeFilterListAtom);
 
     return useCallback(
-        (opCode: string) => {
+        (opCode: string, options?: PerfTablePrefilterOptions) => {
             if (!opCode) {
                 return;
             }
```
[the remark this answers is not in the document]
dcblundell: Correct that no production call site passes the amend option for op codes, but that does not conflict with the acceptance criteria: #1868 covers the histogram bucket controls only, and its op-code criterion (AC5) is that the two hooks share one shape for amend-without-navigating, not that op-code amend be reachable from the charts.

The restructure in f5ade3c7 does tighten this, though. Callers now report the gesture and the values they drew, and a shared `resolvePerfTablePrefilter` owns replace/toggle/clear for both hooks — so the op-code path picks up the union behaviour for free the moment a chart forwards a modifier. Click-again-to-clear stays gated on a caller declaring `visibleValues`, so the op-code plain click still replaces as before.

tests/usePrefilterPerfTableByOpCode.spec.tsx
dcblundell: Agreed, and worse than described — `:71` was the one that mattered. Switched all five to exact `.textContent).toBe(...)`, matching the bucket spec. Confirmed the removal case now fails on a no-op regression rather than passing on the substring.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Confirmed and fixed in f5ade3c7. The value spaces do diverge exactly as described — `buildDurationHistogram(filteredPerfData)` against `buildLogDecadeBuckets(combinedRows)` at usePerfReportFiltering.ts:91.

Took the line-141 route rather than intersecting in the component: the click now passes `visibleValues: bucketMinUsList` and `resolvePerfTablePrefilter` judges the sole selection against the drawn buckets, so the scoping falls out of where the decision lives. Toggling still runs against the full selection, so clearing the drawn control leaves an undrawn bucket another view contributed alone.

Covered by `clears against the drawn values, not the rest of the cross-report filter`.

src/hooks/usePrefilterPerfTableByDurationBucket.ts
dcblundell: Agreed on the clear case. Now gated on the resulting selection being non-empty rather than on `amend`: the reset is justified by landing on a table that cannot show the filter, which says nothing about a filter that no longer exists — and the amend-with-a-result case keeps it, so the existing test stays green as it should.

Added `keeps the stacked view when amending clears the filter`.

src/definitions/PerfDurationHistogram.ts
dcblundell: Fixed — now "click the only selected range again to clear".

tests/PerfDurationHistogram.spec.tsx
dcblundell: Good catch, and the mutation reproduced. Added `collapses a multi-bucket selection onto the plain-clicked bucket` (asserts both the collapse and the move to `TABLE`), added the missing `TABLE` assertion to the replace test, and added the component-level shift-click-removes-the-last-bucket case.

Re-ran your mutation against the extracted resolver (`visibleSelection.includes(value)` in place of the sole-selection check): three tests now fail, including the component-level one.

src/functions/toggleListMembership.ts
dcblundell: Constrained to `T extends string | number` and adopted it in `SimpleMultiselect`. Dropped the `MultiSelectField` claim and replaced it with a note that primitives are deliberate, naming `PerfChartFilter` as the caller that must not be converted.

`MultiSelectField` itself I left alone: two of its six call sites have `T[K]` including `null` (`buffer_type`, `layout`), so adopting it means widening the helper past primitives and constraining `K` to primitive-valued keys, which generic indexed access does not narrow cleanly. More churn than the duplicate is worth here — happy to take it as a follow-up if you disagree.

tests/PerfDurationHistogram.spec.tsx
dcblundell: Agreed. Hardcoded `[0.25, 0.75]` for the two-bucket fixture and added a second assertion after a shift-click, so the fractions have to survive a selection change.

Kept your framing on the limitation in the comment: a mocked `Plot` under jsdom cannot reproduce the redraw artefact, only that the anchoring holds.

src/hooks/usePrefilterPerfTableByOpCode.ts
dcblundell: Extracted, and the asymmetry was real but deliberate — now documented rather than left to be guessed at.

The common body is a pure `resolvePerfTablePrefilter(current, value, options)` returning `{ selection, shouldShowPerfTable }` rather than a shared hook: it keeps the policy in one testable place without threading a jotai atom through a generic hook, and each hook is now down to its atom plus whatever it does after.

On the stacked view: op-code filtering should not reset it, because the stacked table honours the raw op code filter (`filteredStackedRows`, PerfReport.tsx:223) while having no per-op device time for the bucket filter to act on. Said so in the hook docstring.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Took this now rather than as a follow-up — you were right that it is the change that fixes line 138 by construction, and doing them separately would have meant writing the scoping rule twice.

Callers now pass `{ additive, visibleValues }` and `resolvePerfTablePrefilter` translates. That splits the two inputs `amend` was collapsing: `additive` is the shift gesture, and the sole-selection clear is derived from the values the caller says it drew. Only callers that declare `visibleValues` get click-again-to-clear, on the grounds that it is the affordance of a control rendering its own selected state — so the op-code path keeps replacing on a plain click instead of silently changing behaviour.

src/components/performance/PerfDurationHistogram.tsx
dcblundell: Added — the comment now names the `[-0.5, N-0.5]` autorange as the load-bearing step and lists what detaches the controls from their columns (an explicit `xaxis.range`, a numeric x, a non-bar overlay trace).

Leaving the `AGENTS.md` / `CONVENTIONS.md` entry out of this PR; worth its own change rather than a chart-interaction section bolted onto a bugfix.

PR #1877, opened 2026-08-13
Scope circular-buffer accounting to the device that owns the cores

### LINKED ISSUE #1844: Operation Details: per-device CB fan-out is summed, inflating per-core CB memory by the device count

aidemsined:
Closes #1844

## Summary

On a multi-device report the DeviceOp **CBs** list repeated every circular buffer once per device with no device shown, and every figure derived from per-core CB pressure was multiplied by the device count — a 4 KiB CB on an 8-device mesh reported 32 KiB.

The emitted graph was correct. A mesh op allocates each CB on every device and mesh addressing is uniform, so address, size and core range legitimately match across devices. The visualizer dropped the device dimension: `processMemoryAllocations` keyed its per-core accumulator by position alone, so all eight devices' allocations landed in the same `"0,0"` bucket and summed.

Per-core accumulation is now keyed by `(device_id, x, y)` and collapsed with a **max** across devices, so a per-core figure stays per-core. Identical CBs across devices fold into one row carrying a device count (`4 KiB x 120 cores x 8 devices`); the duplicate graph nodes are suppressed at render time rather than filtered out of the graph, because `mergeDevices` and `preprocessConnections` reference nodes by id. A repeat allocation of the same CB *on the same device* is a genuine second allocation and keeps its own row. Reports with no `device_id` accumulate into a single bucket exactly as before.

The pressure modal and the legend rows it drills into previously printed the same figure two ways (`× 120 cores` vs `x 120 cores`); both now use the legend's form. MLIR shape labels keep `×`, where it separates dimensions rather than multiplying a count.

## Legend column alignment

The longer size string made a pre-existing layout problem obvious: legend rows are independent grids sitting at different collapsible depths, and with left-anchored share-based tracks both the nesting indent (up to 80px between a `Buffer allocate` row and a CB row) and the address digit count (25px between a 7- and a 10-digit address) shifted every column downstream.

Since every row already ends at the same right edge, the address is now the only flexible track — it swallows both sources of variation — and every track to its right has a definite width. The two optional slots are placed by class rather than source order, so a row with no buffer type leaves that column empty instead of pulling its shape info left, which collapses the old 5-track / 6-track split into one template. Measured across all 34 legend rows of a multihost report, every column resolves to a single position. Alignment holds to roughly 1400px viewport width; below that the address track reaches its floor and the indent shows through again as it did before, without overflowing or clipping.

## Test plan

- [x] 9 cases in `tests/processMemoryAllocations.spec.ts` covering per-core peak parity against the single-device equivalent, collapse to one allocation per CB, `byCore` staying keyed by core position, interleaved device emission, same-device repeats, unattributed CBs, aliased CBs, and the no-`device_id` path.
- [x] 6 cases in `tests/DeviceOperationsFullRender.spec.tsx` closing the same loop at the render layer: row count, device label, a single `CBs` heading, and peak parity.
- [x] 1 case in `tests/CircularBufferPressureModal.spec.tsx` for the device count on collapsed modal rows.
- [x] Mutation-checked, since a collapse test can pass vacuously: disabling the row suppression reports 16 rows instead of 2; clearing the heading latch on skip yields three headings instead of one; reverting the per-device keying prints 48 KiB where the single-device graph prints 6 KiB.
- [x] Full suite green (1667 tests), `tsc`, `eslint`, `stylelint` clean.
- [x] Verified by hand against a multihost resnet50 report at `/operations/5`.

## Notes

The canonical surviving row is the first CB node seen in graph order, not the lowest `device_id` — device ids are not emitted sorted. Identities are slotted by how many times the device has already allocated that CB, so a genuine second allocation on one device keeps its own row and groups with the other devices' seconds rather than displacing the first (c189a9aa, reported by Copilot).

`cbFanout` is deliberately not cleared at `circular_buffer_deallocate_all`. It is read after the graph walk completes, so clearing it per-operation would leave only the last DeviceOp's records and every earlier one would render its full per-device fan-out again. Retention is bounded by graph size and dies with the memoised result.

Two related issues stay open and are deliberately out of scope. #1842 is the rank union across processes, of which this is the sibling bug *within* a rank; and the graph params carry `device_id` 0–7 while `buffers.device_id` is always 0, so the graph describes eight devices where the buffers table covers one.

image
image

## Review round 2 (@dcblundell)

Four must-fixes and nine should-fixes, all addressed in 3c71aaf5.

**Correctness.** The CB identity is now scoped to the enclosing op. `resetLiveCBs()` only ran at `circular_buffer_deallocate_all`, so on the `function_end` safety-net path a later op's second device folded onto an already-snapshotted row, making its legend read "x 2 devices" while its own modal snapshot still said 1. Probed the capture to confirm the scoping is safe: 0 of 128 CB identities in `multihost_poc_jun24_2321` span more than one enclosing `function_start`.

**Performance.** The branch had regressed the graph walk by re-keying the per-core map by device: `maxCbPerCore()` rescanned it once per node, and `getCoresInRangeList` re-expanded the same `core_range_set` per device. Both fixed (running max; module-level cache with a `CoreCoordList` readonly type so the shared arrays are safe by construction). In the renderer, the duplicate skip now precedes the allocation lookup and `renderMemoryInfo`, the lookup is a Map rather than a per-node linear scan, and `processMemoryAllocations` is memoised so it no longer reruns on every legend click.

200 ops x 8 CBs x 64 cores, `dev` and the reviewed commit rewired to an uncached expander so the cache cannot flatter them:

| devices | nodes | `dev` | @c189a9aa | now |
|---|---|---|---|---|
| 1 | 2,202 | 11 ms | 11 ms | 6 ms |
| 8 | 13,402 | 75 ms | 89 ms | 47 ms |
| 32 | 51,802 | 303 ms | 474 ms | **198 ms** |

**Tidy-up.** `getCoreCountLabel` / `getDeviceCountLabel` in `formatting.ts` replace the copy duplicated across both CB surfaces, which had already drifted again (the modal rendered `x 1 cores`). `BaseMemoryParams.device_id` is now `?: number | string`, which is what the corpus actually holds: one report omits it on all 20 CB nodes, another emits it as a string.

**Tests.** The leak above; the repeat counter resetting with the live set; devices disagreeing on size at one address; one device allocating an extra CB; and the snapshot reaching the modal through its button, including a mixed graph that a blanket device count would pass. Each new invariant was mutation-verified to fail against the previous code.

### Scope decisions

- **Buffer half of the peak** — probed all 24 local reports: 5510 L1 `buffer_allocate` nodes, max distinct `device_id` per `(address, size)` is **1**, and the multihost capture has **zero** L1 buffer nodes. `totalBuffer` is not inflated. Caveat recorded on #1844, since absence in the only mesh capture is not proof.
- **L1 plot legend** — a third CB list (`cbList` via `L1Plots`) still repeats per device. Row repetition only; the numbers are envelope-based and safe. Split to #1879 rather than bundled, as it merges with fragmentation entries and needs its own grouping design.

### DIFF
--- a/src/components/operation-details/DeviceOperationsFullRender.tsx
+++ b/src/components/operation-details/DeviceOperationsFullRender.tsx
@@ -236,7 +236,8 @@ function useDeviceOperationsFullRenderModel(args: {
     } = args;
 
     const selectedAddress = useAtomValue(selectedAddressAtom);
-    const { memoryAllocationList, peakMemoryLoad, cbPressureByOpId } = processMemoryAllocations(deviceOperations);
+    const { memoryAllocationList, peakMemoryLoad, cbPressureByOpId, cbFanout } =
+        processMemoryAllocations(deviceOperations);
 
     const formatTensor = useCallback((node: Node) => formatTensorRendering(node, details), [details]);
 
@@ -448,8 +449,15 @@ function useDeviceOperationsFullRenderModel(args: {
                         />
                     );
                 } else if (nodeType === NodeType.circular_buffer_allocate) {
+                    // Returning rather than falling through leaves
+                    // `consecutiveCBsOutput` set, so the devices we skip here
+                    // don't re-emit the "CBs" heading. #1844
+                    if (cbFanout.duplicateNodeIds.has(node.id)) {
+                        return;
+                    }
                     const cb = node.params;
                     const numCores = parseInt(cb.num_cores, 10) || 1;
+                    const deviceCount = cbFanout.deviceCountByNodeId.get(node.id) ?? 1;
 
                     const variance = cb.allocateOperationId;
                     // `globally_allocated='1'` CBs alias an existing L1
@@ -509,6 +517,7 @@ function useDeviceOperationsFullRenderModel(args: {
                                     onLegendClick={onLegendClick}
                                     colorVariance={variance}
                                     isGloballyAllocated={isGloballyAllocated}
+                                    deviceCount={deviceCount}
                                 />
                                 {memoryInfo}
                             </div>
@@ -534,6 +543,7 @@ function useDeviceOperationsFullRenderModel(args: {
             return output;
         },
         [
+            cbFanout,
             cbPressureByOpId,
             details,
             formatTensor,
--- a/src/components/operation-details/DeviceOperationsFullRender.tsx
+++ b/src/components/operation-details/DeviceOperationsFullRender.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
-import React, { Fragment, JSX, useCallback, useState } from 'react';
+import React, { Fragment, JSX, useCallback, useMemo, useState } from 'react';
 import {
     Button,
     ButtonVariant,
@@ -236,8 +236,20 @@ function useDeviceOperationsFullRenderModel(args: {
     } = args;
 
     const selectedAddress = useAtomValue(selectedAddressAtom);
-    const { memoryAllocationList, peakMemoryLoad, cbPressureByOpId, cbFanout } =
-        processMemoryAllocations(deviceOperations);
+    // Walking the graph is proportional to node count, which a mesh capture
+    // multiplies by the device count, so it cannot run on renders driven by
+    // `selectedAddress` — every legend click is one. Memoising also makes the
+    // three returned objects stable, without which the `renderNodes`
+    // `useCallback` below memoises nothing. #1844
+    const { memoryAllocationList, peakMemoryLoad, cbPressureByOpId, cbFanout } = useMemo(
+        () => processMemoryAllocations(deviceOperations),
+        [deviceOperations],
+    );
+    // A linear scan per node is quadratic over a graph with one entry per node.
+    const memoryDetailsByNodeId = useMemo(
+        () => new Map<number, AllocationDetails>(memoryAllocationList.map((data) => [data.id, data])),
+        [memoryAllocationList],
+    );
 
     const formatTensor = useCallback((node: Node) => formatTensorRendering(node, details), [details]);
 
@@ -254,7 +266,14 @@ function useDeviceOperationsFullRenderModel(args: {
 
             nodes.forEach((node, index) => {
                 const nodeType = node.node_type;
-                const memoryDetails = memoryAllocationList.find((data) => data.id === node.id);
+                // Ahead of the lookup and `renderMemoryInfo` below, both of
+                // which were being paid for rows that are then dropped — most
+                // CB nodes in a mesh capture. Returning here keeps the "CBs"
+                // heading latch untouched, same as the guard it replaces. #1844
+                if (nodeType === NodeType.circular_buffer_allocate && cbFanout.duplicateNodeIds.has(node.id)) {
+                    return;
+                }
+                const memoryDetails = memoryDetailsByNodeId.get(node.id);
                 const memoryInfo = renderMemoryInfo(memoryDetails, peakMemoryLoad);
 
                 if (nodeType === NodeType.function_start) {
@@ -449,9 +468,6 @@ function useDeviceOperationsFullRenderModel(args: {
                         />
                     );
                 } else if (nodeType === NodeType.circular_buffer_allocate) {
-                    if (cbFanout.duplicateNodeIds.has(node.id)) {
-                        return;
-                    }
                     const cb = node.params;
                     const numCores = parseInt(cb.num_cores, 10) || 1;
                     const deviceCount = cbFanout.deviceCountByNodeId.get(node.id) ?? 1;
@@ -544,7 +560,7 @@ function useDeviceOperationsFullRenderModel(args: {
             cbPressureByOpId,
             details,
             formatTensor,
-            memoryAllocationList,
+            memoryDetailsByNodeId,
             onLegendClick,
             peakMemoryLoad,
             selectedAddress,
--- a/src/components/operation-details/DeviceOperationsFullRender.tsx
+++ b/src/components/operation-details/DeviceOperationsFullRender.tsx
@@ -449,9 +449,6 @@ function useDeviceOperationsFullRenderModel(args: {
                         />
                     );
                 } else if (nodeType === NodeType.circular_buffer_allocate) {
-                    // Returning rather than falling through leaves
-                    // `consecutiveCBsOutput` set, so the devices we skip here
-                    // don't re-emit the "CBs" heading. #1844
                     if (cbFanout.duplicateNodeIds.has(node.id)) {
                         return;
                     }
--- a/src/components/operation-details/MemoryLegendElement.tsx
+++ b/src/components/operation-details/MemoryLegendElement.tsx
@@ -43,10 +43,6 @@ interface MemoryLegendElementProps {
      * reads as a tensor view rather than a fresh allocation.
      */
     isGloballyAllocated?: boolean;
-    /**
-     * Devices this row stands for, when a mesh op allocated the same CB on each
-     * of them and the duplicate rows were collapsed into this one. #1844
-     */
     deviceCount?: number;
 }
 
--- a/src/components/operation-details/MemoryLegendElement.tsx
+++ b/src/components/operation-details/MemoryLegendElement.tsx
@@ -11,7 +11,7 @@ import { DeviceOperationLayoutTypes, FragmentationEntry, MarkerType, MarkerTypeL
 import { OperationDetails } from '../../model/OperationDetails';
 import { getBufferColor, getTensorColor } from '../../functions/colorGenerator';
 import { formatMemorySize, prettyPrintAddress } from '../../functions/math';
-import { toReadableShape, toReadableType } from '../../functions/formatting';
+import { getCoreCountLabel, getDeviceCountLabel, toReadableShape, toReadableType } from '../../functions/formatting';
 import 'styles/components/MemoryLegendElement.scss';
 import { L1_SMALL_MARKER_COLOR, L1_START_MARKER_COLOR } from '../../definitions/PlotConfigurations';
 import { selectedBufferColourAtom, showHexAtom } from '../../store/app';
@@ -86,9 +86,8 @@ export const MemoryLegendElement = ({
 
     const derivedTensor = operationDetails.getTensorForAddress(chunk.address);
     const isPerCoreBuffer = bufferType !== StringBufferType.DRAM && bufferType !== StringBufferType.SYSTEM_MEMORY;
-    const numCoresLabel =
-        isPerCoreBuffer && numCores && numCores > 0 ? ` x ${numCores} ${numCores === 1 ? 'core' : 'cores'}` : '';
-    const deviceCountLabel = deviceCount > 1 ? ` x ${deviceCount} devices` : '';
+    const numCoresLabel = isPerCoreBuffer && numCores && numCores > 0 ? ` ${getCoreCountLabel(numCores)}` : '';
+    const deviceCountLabel = deviceCount > 1 ? ` ${getDeviceCountLabel(deviceCount)}` : '';
 
     const resolvedColour =
         chunk.tensorId || derivedTensor
--- a/src/components/operation-details/MemoryLegendElement.tsx
+++ b/src/components/operation-details/MemoryLegendElement.tsx
@@ -43,6 +43,11 @@ interface MemoryLegendElementProps {
      * reads as a tensor view rather than a fresh allocation.
      */
     isGloballyAllocated?: boolean;
+    /**
+     * Devices this row stands for, when a mesh op allocated the same CB on each
+     * of them and the duplicate rows were collapsed into this one. #1844
+     */
+    deviceCount?: number;
 }
 
 export const MemoryLegendElement = ({
@@ -60,6 +65,7 @@ export const MemoryLegendElement = ({
     numCores,
     userL1ZoomRange,
     isGloballyAllocated = false,
+    deviceCount = 1,
 }: MemoryLegendElementProps) => {
     const showHex = useAtomValue(showHexAtom);
     const selectedBufferColour = useAtomValue(selectedBufferColourAtom);
@@ -86,6 +92,7 @@ export const MemoryLegendElement = ({
     const isPerCoreBuffer = bufferType !== StringBufferType.DRAM && bufferType !== StringBufferType.SYSTEM_MEMORY;
     const numCoresLabel =
         isPerCoreBuffer && numCores && numCores > 0 ? ` x ${numCores} ${numCores === 1 ? 'core' : 'cores'}` : '';
+    const deviceCountLabel = deviceCount > 1 ? ` x ${deviceCount} devices` : '';
 
     const resolvedColour =
         chunk.tensorId || derivedTensor
@@ -169,6 +176,7 @@ export const MemoryLegendElement = ({
                     <>
                         {formatMemorySize(chunk.size, 2)}
                         {numCoresLabel}
+                        {deviceCountLabel}
                         {isGloballyAllocated && (
                             <Tooltip
                                 content={
--- a/src/functions/processMemoryAllocations.ts
+++ b/src/functions/processMemoryAllocations.ts
@@ -5,8 +5,12 @@
 import { DeviceOperationNode, Node, NodeType } from '../model/APIData';
 import { L1_NUM_CORES } from '../definitions/L1MemorySize';
 import { StringBufferType } from '../model/BufferType';
-import { AllocationDetails, CBAllocationSummary, CBPressureSnapshot } from '../model/MemoryAllocations';
+import { AllocationDetails, CBAllocationSummary, CBDeviceFanout, CBPressureSnapshot } from '../model/MemoryAllocations';
 import { getCoresInRangeList } from './math';
+
+// Stands in for `device_id` when the graph omits it, so a report with no device
+// dimension accumulates into one bucket exactly as it did before #1844.
+const SINGLE_DEVICE_KEY = 'single';
 
 export function processMemoryAllocations(
     graph: Node[],
@@ -15,39 +19,64 @@ export function processMemoryAllocations(
     peakMemoryLoad: number;
     memoryAllocationList: AllocationDetails[];
     cbPressureByOpId: Map<number, CBPressureSnapshot>;
+    cbFanout: CBDeviceFanout;
 } {
     let peakMemoryLoad = 0;
     const memoryAllocationList: AllocationDetails[] = [];
     const curOpList: { name: string; id: number; deviceId?: string | number }[] = [];
-    const cbBytesByCore = new Map<string, number>();
+    // Keyed `${device}|${x},${y}` because core (0,0) of device 0 is not core
+    // (0,0) of device 6; summing the mesh's identical CBs is what inflated every
+    // total by the device count. #1844
+    const cbBytesByDeviceCore = new Map<string, number>();
+    // CBs whose `core_range_set` resolved to no cores, per device.
+    const unattributedByDevice = new Map<string, number>();
     // Live CB allocations since the last `circular_buffer_deallocate_all` (or
-    // since the DeviceOp started, whichever came last). Mirrors `cbBytesByCore`
-    // so the snapshot can attribute pressure back to specific CB events.
+    // since the DeviceOp started, whichever came last), one entry per CB rather
+    // than per device. Mirrors `cbBytesByDeviceCore` so the snapshot can
+    // attribute pressure back to specific CB events.
     let liveCBs: CBAllocationSummary[] = [];
+    // Lets a later device bump an existing CB's `deviceCount` instead of adding
+    // a row. Same lifetime as `liveCBs`.
+    let liveCBByIdentity = new Map<string, { summary: CBAllocationSummary; deviceKeys: Set<string> }>();
     const cbPressureByOpId = new Map<number, CBPressureSnapshot>();
+    const cbFanout: CBDeviceFanout = {
+        deviceCountByNodeId: new Map<number, number>(),
+        duplicateNodeIds: new Set<number>(),
+    };
+
+    const resetLiveCBs = () => {
+        liveCBs = [];
+        liveCBByIdentity = new Map();
+    };
 
     const snapshotCBPressure = (opId: number) => {
-        if (liveCBs.length === 0 && cbBytesByCore.size === 0) {
+        if (liveCBs.length === 0 && cbBytesByDeviceCore.size === 0 && unattributedByDevice.size === 0) {
             return;
         }
         const byCore: Record<string, number> = {};
         let maxBytes = 0;
-        let unattributedBytes = 0;
-        for (const [k, v] of cbBytesByCore.entries()) {
-            byCore[k] = v;
-            if (k === '?') {
-                unattributedBytes = v;
-            } else if (v > maxBytes) {
-                maxBytes = v;
+        for (const [key, bytes] of cbBytesByDeviceCore.entries()) {
+            const core = key.slice(key.indexOf('|') + 1);
+            const perCore = Math.max(byCore[core] ?? 0, bytes);
+            byCore[core] = perCore;
+            if (perCore > maxBytes) {
+                maxBytes = perCore;
             }
         }
-        // Last writer wins if a DeviceOp triggers multiple snapshots. v1
-        // assumption: one `circular_buffer_deallocate_all` per DeviceOp.
+        let unattributedBytes = 0;
+        for (const bytes of unattributedByDevice.values()) {
+            unattributedBytes = Math.max(unattributedBytes, bytes);
+        }
+        if (unattributedByDevice.size > 0) {
+            byCore['?'] = unattributedBytes;
+        }
+        // Copied per entry because `deviceCount` stays mutable on the live
+        // objects; a snapshot must not change after it is taken.
         cbPressureByOpId.set(opId, {
             byCore,
             maxBytes,
             unattributedBytes,
-            allocations: liveCBs.slice(),
+            allocations: liveCBs.map((cb) => ({ ...cb })),
         });
     };
 
@@ -55,12 +84,12 @@ export function processMemoryAllocations(
 
     const maxCbPerCore = (): number => {
         let m = 0;
-        // Skip the '?' bucket — those bytes have no core attribution so
+        // Unattributed bytes are excluded — they have no core attribution so
         // they don't belong in a per-core peak. Mirrors the same exclusion
         // in snapshotCBPressure() so cbPeak and snapshot.maxBytes agree.
-        for (const [k, v] of cbBytesByCore.entries()) {
-            if (k !== '?' && v > m) {
-                m = v;
+        for (const bytes of cbBytesByDeviceCore.values()) {
+            if (bytes > m) {
+                m = bytes;
             }
         }
         return m;
@@ -98,27 +127,45 @@ export function processMemoryAllocations(
             // forward-compat with future tt-metal emit changes. #1651
             const rawGlobalFlag = node.params.globally_allocated as unknown;
             const globallyAllocated = rawGlobalFlag === '1' || rawGlobalFlag === 1;
+            const deviceKey = String(node.params.device_id ?? SINGLE_DEVICE_KEY);
             if (!globallyAllocated) {
                 if (cores.length === 0) {
-                    cbBytesByCore.set('?', (cbBytesByCore.get('?') ?? 0) + size);
+                    unattributedByDevice.set(deviceKey, (unattributedByDevice.get(deviceKey) ?? 0) + size);
                 } else {
                     for (const { x, y } of cores) {
-                        const k = `${x},${y}`;
-                        cbBytesByCore.set(k, (cbBytesByCore.get(k) ?? 0) + size);
+                        const k = `${deviceKey}|${x},${y}`;
+                        cbBytesByDeviceCore.set(k, (cbBytesByDeviceCore.get(k) ?? 0) + size);
                     }
                 }
             }
-            liveCBs.push({
-                nodeId: node.id,
-                address: parseInt(node.params.address, 10),
-                size,
-                numCores: cores.length,
-                coreRangeSet: node.params.core_range_set,
-                cores,
-                allocateOperationId: currentOp?.id,
-                allocateOperationName: currentOp?.name,
-                globallyAllocated,
-            });
+
+            const address = parseInt(node.params.address, 10);
+            const identity = `${address}|${size}|${node.params.core_range_set}|${globallyAllocated}`;
+            const existing = liveCBByIdentity.get(identity);
+            // A repeat within one device is a genuine second allocation, not mesh
+            // fan-out, so it keeps its own row.
+            if (existing && !existing.deviceKeys.has(deviceKey)) {
+                existing.deviceKeys.add(deviceKey);
+                existing.summary.deviceCount = existing.deviceKeys.size;
+                cbFanout.deviceCountByNodeId.set(existing.summary.nodeId, existing.deviceKeys.size);
+                cbFanout.duplicateNodeIds.add(node.id);
+            } else {
+                const summary: CBAllocationSummary = {
+                    nodeId: node.id,
+                    address,
+                    size,
+                    numCores: cores.length,
+                    coreRangeSet: node.params.core_range_set,
+                    cores,
+                    allocateOperationId: currentOp?.id,
+                    allocateOperationName: currentOp?.name,
+                    globallyAllocated,
+                    deviceCount: 1,
+                };
+                liveCBs.push(summary);
+                liveCBByIdentity.set(identity, { summary, deviceKeys: new Set([deviceKey]) });
+                cbFanout.deviceCountByNodeId.set(node.id, 1);
+            }
         }
 
         if (node.node_type === NodeType.circular_buffer_deallocate_all) {
@@ -129,8 +176,9 @@ export function processMemoryAllocations(
             if (currentOp) {
                 snapshotCBPressure(currentOp.id);
             }
-            cbBytesByCore.clear();
-            liveCBs = [];
+            cbBytesByDeviceCore.clear();
+            unattributedByDevice.clear();
+            resetLiveCBs();
         }
 
         if (node.node_type === NodeType.buffer_allocate && node.params.type === StringBufferType.L1) {
@@ -177,7 +225,7 @@ export function processMemoryAllocations(
         peakMemoryLoad = Math.max(peakMemoryLoad, cbPeak + totalBuffer);
     }
 
-    return { peakMemoryLoad, memoryAllocationList, cbPressureByOpId };
+    return { peakMemoryLoad, memoryAllocationList, cbPressureByOpId, cbFanout };
 }
 
 export const processInputsOutputs = (graph: Node[]): DeviceOperationNode[] => {
--- a/src/functions/processMemoryAllocations.ts
+++ b/src/functions/processMemoryAllocations.ts
@@ -83,18 +83,15 @@ export function processMemoryAllocations(
 
     let totalBuffer = 0;
 
-    const maxCbPerCore = (): number => {
-        let m = 0;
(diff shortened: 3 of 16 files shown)


src/functions/processMemoryAllocations.ts
```diff
@@ -98,27 +124,45 @@ export function processMemoryAllocations(
...
+                    numCores: cores.length,
+                    coreRangeSet: node.params.core_range_set,
+                    cores,
+                    allocateOperationId: currentOp?.id,
+                    allocateOperationName: currentOp?.name,
+                    globallyAllocated,
+                    deviceCount: 1,
+                };
+                liveCBs.push(summary);
+                liveCBByIdentity.set(identity, { summary, deviceKeys: new Set([deviceKey]) });
```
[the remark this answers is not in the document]
aidemsined: Confirmed and fixed in c189a9aa. I reproduced it before changing anything: with two devices each allocating the same CB twice, emitted device-major, the walk produced three rows counted 1, 2, 1 where two rows counted 2, 2 are correct. Your reading of the mechanism is exact — the `else` branch overwrote the slot, so device 1's allocation folded onto device 0's *second* allocation, and the "first node in graph order is canonical" claim was false.

Taken your suggestion: the key is now `identity|deviceRepeatCount`, where the count is how many times that device has already allocated that CB. Each repeat pairs with the other devices' repeats of the same ordinal instead of displacing the first allocation. A device can now only reach a given slot once, so the `deviceKeys.has()` guard that used to separate same-device repeats is redundant and gone — the set stays because it remains the device-count source of truth. The new map resets with `liveCBs` at `circular_buffer_deallocate_all`, since the grouping is only meaningful within one live window.

Worth recording that no total moves: per-core accounting is keyed independently in `cbBytesByDeviceCore` and never consulted this grouping, so the bug was confined to which row carried the count and how many rows appeared. Two regression tests were added and confirmed failing against the previous code — one on the device-major repeat pattern, one asserting the first node in graph order is the row the rest collapse onto.

src/functions/processMemoryAllocations.ts
```diff
@@ -129,8 +184,9 @@ export function processMemoryAllocations(
             if (currentOp) {
                 snapshotCBPressure(currentOp.id);
             }
-            cbBytesByCore.clear();
-            liveCBs = [];
+            cbBytesByDeviceCore.clear();
+            unattributedByDevice.clear();
+            resetLiveCBs();
```
dcblundell, line 197: **Must fix:** fan-out state survives a DeviceOp that ends without `circular_buffer_deallocate_all`

This is the only call site of `resetLiveCBs()`, so `liveCBByIdentity` and `cbRepeatsByDevice` leak across the `function_end` safety-net path (~L198-208) that the code explicitly supports. Probe on this revision: `op1{CB@0x1000 dev0}` then `op2{CB@0x1000 dev0, dev1}` with no dealloc puts op2's dev-1 node in `duplicateNodeIds` (its row is dropped at `DeviceOperationsFullRender.tsx:452`) and sets `deviceCountByNodeId{op1: 2}` — so op1's legend row reads "x 2 devices" while op1's own modal snapshot reports `deviceCount: 1`.

Reset the fan-out maps in the `function_end` branch too, or key `identity` by the enclosing op id. The existing test at `processMemoryAllocations.spec.ts:408` covers that path but without a device dimension.

src/functions/processMemoryAllocations.ts
```diff
@@ -5,62 +5,92 @@
...
     const maxCbPerCore = (): number => {
         let m = 0;
-        // Skip the '?' bucket — those bytes have no core attribution so
+        // Unattributed bytes are excluded — they have no core attribution so
         // they don't belong in a per-core peak. Mirrors the same exclusion
         // in snapshotCBPressure() so cbPeak and snapshot.maxBytes agree.
-        for (const [k, v] of cbBytesByCore.entries()) {
-            if (k !== '?' && v > m) {
-                m = v;
+        for (const bytes of cbBytesByDeviceCore.values()) {
```
dcblundell: **Must fix:** this scan runs once per graph node over a map the fix just grew by the device count

Re-keying `x,y` to `deviceKey|x,y` took the map from 64 to 2048 entries on a 32-device mesh while node count also scales with devices, so `maxCbPerCore()` at L218 went from O(N·cores) to O(N·devices·cores). Benchmarked at 200 DeviceOps x 8 CBs x 64 cores: `dev` 310ms vs this branch 413ms at 32 devices — a regression on exactly the reports this PR targets.

The map is mutated in only two places (the `set` at L138, the `clear` at L187), so a running `Math.max` maintained at the `set` and reset at the `clear` is exact and recovers essentially all of it (413ms -> 318ms).

src/components/operation-details/DeviceOperationsFullRender.tsx
```diff
@@ -448,8 +449,12 @@ function useDeviceOperationsFullRenderModel(args: {
                         />
                     );
                 } else if (nodeType === NodeType.circular_buffer_allocate) {
+                    if (cbFanout.duplicateNodeIds.has(node.id)) {
```
dcblundell: **Must fix:** the duplicate skip runs after the expensive work, not before it

By the time this `return` fires, the node has already paid `memoryAllocationList.find(...)` (L257, a linear scan of a list with roughly one entry per node) and `renderMemoryInfo()` (L258, four React elements plus three `formatMemorySize()` calls). On a 32-device mesh ~97% of CB nodes are duplicates, so ~50k nodes each do an O(N) scan and build JSX that is discarded — the rows left the DOM but the work behind them did not.

Hoist this `duplicateNodeIds.has()` guard above L257, and replace the `find` with a `Map<number, AllocationDetails>` built once outside the `forEach`.

src/components/operation-details/DeviceOperationsFullRender.tsx
```diff
@@ -236,7 +236,8 @@ function useDeviceOperationsFullRenderModel(args: {
     } = args;
 
     const selectedAddress = useAtomValue(selectedAddressAtom);
-    const { memoryAllocationList, peakMemoryLoad, cbPressureByOpId } = processMemoryAllocations(deviceOperations);
+    const { memoryAllocationList, peakMemoryLoad, cbPressureByOpId, cbFanout } =
```
dcblundell: **Must fix:** `processMemoryAllocations` is unmemoised, so the whole graph walk repeats on every render

That includes renders caused only by `selectedAddressAtom` changing — i.e. every legend click — which is now a several-hundred-millisecond main-thread block on a mesh report given the cost added by this branch. All three destructured values are also fresh objects each render, so the `useCallback` at L540 currently memoises nothing.

`useMemo(() => processMemoryAllocations(deviceOperations), [deviceOperations])` fixes the recompute and stabilises the deps together.

src/functions/processMemoryAllocations.ts
```diff
@@ -129,8 +184,9 @@ export function processMemoryAllocations(
...
                 snapshotCBPressure(currentOp.id);
             }
-            cbBytesByCore.clear();
-            liveCBs = [];
+            cbBytesByDeviceCore.clear();
+            unattributedByDevice.clear();
+            resetLiveCBs();
         }
 
         if (node.node_type === NodeType.buffer_allocate && node.params.type === StringBufferType.L1) {
```
dcblundell, line 200: **Should fix (decision):** is the buffer half of the peak heading inflated too?

`peakMemoryLoad = cbPeak + totalBuffer` (L233), but only `cbPeak` gained a device dimension — `totalBuffer` accumulates every L1 `buffer_allocate` node unconditionally. #1844's criterion 2 is stated over the whole "Peak L1 memory load per core" heading, and its root-cause section only establishes fan-out for `circular_buffer_allocate` nodes. The `buffers.device_id` aside in the issue is about the buffers *table*, not these graph nodes.

Worth checking the `multihost_poc_jun24_2321` capture for `device_id` on L1 `buffer_allocate` nodes before closing the issue. If they fan out, criterion 2 is half-done; if not, record that in #1844 so the scope is settled rather than assumed. Either way the new mesh fixtures are CB-only, so no test can currently tell.

src/model/MemoryAllocations.ts
```diff
@@ -62,3 +64,12 @@ export type CBPressureSnapshot = {
     /** Ordered list of CBs that were live when the snapshot was taken. */
     allocations: CBAllocationSummary[];
 };
+
+/**
+ * A mesh op emits the same CB once per device; only the first earns a row, so the
+ * rest are listed here for the renderer to skip. #1844
+ */
+export type CBDeviceFanout = {
+    deviceCountByNodeId: Map<number, number>;
```
dcblundell, line 73: **Should fix (decision):** a third CB list still repeats per device

(True location: `L1Plots.tsx:106-121` and `OperationDetails.ts:257-264`.) `cbList` is populated per `circular_buffer_allocate` node with no collapse, and `L1Plots` renders one `MemoryLegendElement` per entry when the "Circular buffers" toggle is on — so on the multihost report that legend still shows each CB eight times, unlabelled, in arbitrary order. Numbers are safe (`calculateCondensed` is envelope-based); this is the row-repetition half of criterion 3 only.

Either give `cbList` the same collapsing via a helper alongside this `CBDeviceFanout`, or amend #1844's criterion 3 to name the two surfaces in scope and open a follow-up for the plot legend. The second reads better to me — the plot legend has different grouping semantics and bundling it grows this PR again.

src/functions/processMemoryAllocations.ts
```diff
@@ -98,27 +128,52 @@ export function processMemoryAllocations(
...
+            const deviceKey = String(node.params.device_id ?? SINGLE_DEVICE_KEY);
             if (!globallyAllocated) {
                 if (cores.length === 0) {
-                    cbBytesByCore.set('?', (cbBytesByCore.get('?') ?? 0) + size);
+                    unattributedByDevice.set(deviceKey, (unattributedByDevice.get(deviceKey) ?? 0) + size);
                 } else {
                     for (const { x, y } of cores) {
-                        const k = `${x},${y}`;
-                        cbBytesByCore.set(k, (cbBytesByCore.get(k) ?? 0) + size);
+                        const k = `${deviceKey}|${x},${y}`;
```
dcblundell, line 134: **Should fix:** memoise `getCoresInRangeList` (called at ~L120)

The same `core_range_set` string is re-parsed and re-expanded once per device — precisely the work this fix exists to deduplicate, just not deduplicated here. It is the largest remaining cost: on top of the cached-peak fix it took the 32-device benchmark from 318ms to 166ms, roughly 2x overall.

A module-level `Map<string, CoreCoord[]>` in `math.ts` covers it, provided callers treat the returned array as immutable — worth asserting, since `CBAllocationSummary.cores` at L167 would then be shared across rows.

src/functions/processMemoryAllocations.ts
```diff
@@ -98,27 +128,52 @@ export function processMemoryAllocations(
             // forward-compat with future tt-metal emit changes. #1651
             const rawGlobalFlag = node.params.globally_allocated as unknown;
             const globallyAllocated = rawGlobalFlag === '1' || rawGlobalFlag === 1;
+            const deviceKey = String(node.params.device_id ?? SINGLE_DEVICE_KEY);
```
dcblundell, line 128: **Should fix:** the declared type and this runtime guard disagree about `device_id`

`CircularBufferAllocateParams` inherits a non-optional `device_id: number` from `BaseMemoryParams` (`APIData.ts:281`), yet this line falls back with `?? SINGLE_DEVICE_KEY` and two new tests assert the absent path. One of the two is wrong: if older captures really omit it, make it optional on the model — that is what a future reader will trust when writing `node.params.device_id.toString()`; if it is always present, the fallback and both tests are dead. `no-unnecessary-condition` is not enabled, so nothing catches this at build time.

src/components/operation-details/MemoryLegendElement.tsx
```diff
@@ -86,6 +88,7 @@ export const MemoryLegendElement = ({
     const isPerCoreBuffer = bufferType !== StringBufferType.DRAM && bufferType !== StringBufferType.SYSTEM_MEMORY;
     const numCoresLabel =
         isPerCoreBuffer && numCores && numCores > 0 ? ` x ${numCores} ${numCores === 1 ? 'core' : 'cores'}` : '';
+    const deviceCountLabel = deviceCount > 1 ? ` x ${deviceCount} devices` : '';
```
dcblundell: **Should fix:** the multiplier copy is duplicated across both CB surfaces and has already drifted again

This line and `CircularBufferPressureModal.tsx:703-704` format the same labels independently, and they are not identical at HEAD: the legend pluralises (`x 1 core`), the modal does not (`x 1 cores`); the legend gates the core label on `isPerCoreBuffer`, the modal on `numCores > 0`. Commit `61a5b34d` exists because these drifted once already, and the device label was then written independently in both files plus hard-coded in two specs.

A `getCoreCountLabel` / `getDeviceCountLabel` pair in `src/functions/formatting.ts`, imported by both components and both specs, makes the next glyph or pluralisation decision a one-line change.

src/scss/components/DeviceOperationFullRender.scss
```diff
@@ -159,21 +160,53 @@
...
+            }
+
+            // The cell is right-aligned, so an in-flow marker would shorten the
+            // text run on aliased rows only.
+            > .format-numbers:nth-child(3) {
+                position: relative;
+                padding-right: variables.$legend-marker-slot;
+            }
+
+            .globally-allocated-marker {
```
dcblundell, line 203: **Should fix:** this override leaves a stale comment behind in the owning sheet

`MemoryLegendElement.scss:106-110` still says the marker "Renders icon-only in the per-DeviceOp row to preserve the existing grid template (`15px 0.5fr 2fr 13fr 3fr`)". That template no longer applies to the per-DeviceOp row (six definite tracks now, L178-184), and the marker is no longer in flow there either — this rule absolutely positions it and sets `margin-left: 0`, overriding the `margin-left: 6px` declared four lines below that comment. Both halves of the stated rationale are now wrong for the scope they name.

`3acaeee0` trimmed the sibling comment in this file but missed that one.

tests/processMemoryAllocations.spec.ts
```diff
@@ -624,3 +626,201 @@ describe('processMemoryAllocations - globally_allocated CBs (#1651)', () => {
         expect(snap.allocations[0].globallyAllocated).toBe(false);
     });
 });
+
+describe('processMemoryAllocations - per-device CB fan-out (#1844)', () => {
```
dcblundell, line 630: **Should fix:** every fan-out case in this block uses a homogeneous mesh

No case has one device allocating an extra CB, and none has devices disagreeing on size at the same address — which are exactly where "max across devices" differs from "sum" and from "first device wins". Current behaviour is correct in both (`dev0{4096, 2048}` + `dev1{4096}` gives `peakMemoryLoad 6144` with rows `[deviceCount 2, deviceCount 1]`; `4096` vs `8192` at one address gives `maxBytes 8192`, not the `12288` sum), so these are cheap characterisation tests that lock in the accounting rule the fix exists for.

Also uncovered here: two CB batches separated by `deallocate_all` within one op, which is what exercises `cbRepeatsByDevice` being cleared.

tests/DeviceOperationsFullRender.spec.tsx
```diff
@@ -0,0 +1,160 @@
...
+    nextId = 0;
+    mockUseDevices.mockReturnValue({ data: [{ num_x_cores: 2, num_y_cores: 2, worker_l1_size: 1_000_000 }] });
+});
+
+afterEach(() => {
+    cleanup();
+    vi.clearAllMocks();
+});
+
+describe('DeviceOperationsFullRender - per-device CB fan-out (#1844)', () => {
```
dcblundell, line 105: **Should fix:** nothing connects this layer to the modal

The fan-out snapshot reaches the modal only through the "View per-core allocations" button. This spec asserts the heading exists (L126) but never clicks it, and `CircularBufferPressureModal.spec.tsx` hand-builds `deviceCount` rather than deriving it from `processMemoryAllocations` — so nothing would catch `cbPressureByOpId` being wired to the wrong DeviceOp for a fan-out graph, which is the user-visible face of the `function_end` leak flagged separately.

One test: render a mesh graph, click the heading link, assert the modal lists 2 rows each named `/x 8 devices/`. Worth adding a mixed case too (dev0 allocates both CBs, dev1 only the first) — all cases here are homogeneous, so the `?? 1` lookup at `DeviceOperationsFullRender.tsx:457` could regress to a blanket count without failing anything.

dcblundell (commented):
Conductor review (DRY / TEST / PERF / STANDARDS / SECURITY / REQUIREMENTS) against `c189a9aa`. 13 line-anchored comments below: 4 Must fix, 9 Should fix. No security findings. Two spec-vs-code decisions (the buffer term in the peak heading, and which CB surfaces criterion 3 covers) are flagged inline as decisions rather than defects.

src/components/operation-details/MemoryLegendElement.tsx
```diff
@@ -43,6 +43,7 @@ interface MemoryLegendElementProps {
      * reads as a tensor view rather than a fresh allocation.
      */
     isGloballyAllocated?: boolean;
+    deviceCount?: number;
```
dcblundell, line 46: **Should fix:** the new `deviceCount` prop has no leaf-level test

`MemoryLegendElement.spec.tsx` has a dedicated `core count label` block covering the sibling `numCoresLabel` exhaustively (unset / zero / DRAM / SYSTEM_MEMORY); `deviceCount` gets nothing but an indirect assertion through the full-render spec. Mirror the cases: omitted -> no label, `1` -> no label, `8` -> label present, and `{numCores: 2, deviceCount: 8}` -> accessible name matches `/x 2 cores x 8 devices/`.

Worth pinning intent while you are there: unlike `numCoresLabel`, `deviceCountLabel` is not gated on `isPerCoreBuffer`, so a DRAM row would render "x 8 devices" with no core context. Unreachable from today's only call site, but the prop contract is public.

src/functions/processMemoryAllocations.ts
aidemsined: Fixed in 3c71aaf5. Reproduced your probe exactly before touching anything: `op1{CB@0x1000 dev0}` then `op2{CB@0x1000 dev0, dev1}` with no dealloc put `op2`'s dev-1 node in `duplicateNodeIds` and left `deviceCountByNodeId{op1: 2}` against an op1 snapshot still reading `deviceCount: 1`. Legend and modal disagreeing on the same row is the worst kind of bug for this feature, since the whole point is making the count trustworthy.

Took the second of your two options — `identity` now carries `currentOp?.id`. I preferred it to resetting in `function_end` because that branch also fires for nested functions, so a reset there would tear down the fan-out state mid-op; scoping the key confines the collapse to one op without caring how the op is left.

That only works if every device's copy of a CB shares one enclosing `function_start`, which I checked rather than assumed. Across all 8 op/rank graphs in `multihost_poc_jun24_2321`: 128 CB nodes, and the number of CB identities spanning more than one enclosing `function_start` is **0** — all 8 devices sit under the same `BinaryNgDeviceOperation`. Had that been per-device `function_start` wrappers, op-scoping would have silently un-collapsed everything, so it was worth the probe.

Two tests added, each mutation-verified to fail alone against the previous code:
- your probe, asserting op1 stays at 1 and that `deviceCountByNodeId` agrees with the snapshot's `deviceCount`
- the repeat counter resetting with the live set, which the existing cases could not see: it needs batch one to hold a *subset* of the devices in batch two (`dev0`, dealloc, `dev0 + dev1`), otherwise a stale counter still lands both devices in the same slot and the bug hides

src/functions/processMemoryAllocations.ts
aidemsined: Fixed in 3c71aaf5, and your diagnosis and remedy were both right. `cbPeakPerCore` is now maintained at the `set` and reset at the `clear`, and `maxCbPerCore()` is gone.

I checked the exactness precondition before relying on it: `cbBytesByDeviceCore` is mutated in exactly the two places you name, the `set` is strictly accumulating (`(get ?? 0) + size`), and there is no single-CB deallocate node type at all — `NodeType` has only `circular_buffer_deallocate_all`. With no path that lowers a value, the running max cannot go stale, so this is exact rather than an approximation.

Reproduced your benchmark at the same shape (200 DeviceOps x 8 CBs x 64 cores), comparing `origin/dev`'s walker and the reviewed commit against the fix, with dev/head rewired to an uncached expander so the S3 cache could not leak into their numbers:

| devices | nodes | `dev` | branch @c189a9aa | fixed |
|---|---|---|---|---|
| 1 | 2,202 | 11 ms | 11 ms | 6 ms |
| 8 | 13,402 | 75 ms | 89 ms | 47 ms |
| 32 | 51,802 | 303 ms | 474 ms | **198 ms** |

Your 310 / 413 lands at 303 / 474 here, so the regression is confirmed a bit worse than measured. With this plus the `getCoresInRangeList` cache it is now 35% faster than `dev` rather than merely recovered.

src/components/operation-details/DeviceOperationsFullRender.tsx
aidemsined: Both done in 3c71aaf5. The guard moved to the top of the `forEach`, gated on `nodeType === circular_buffer_allocate` so it only intercepts the rows it is meant to, and the `find` is now a `Map<number, AllocationDetails>` built once in a `useMemo` — that scan was quadratic over a list with roughly one entry per node, independent of the fan-out.

One thing I verified before moving it, since the ordering was load-bearing: hoisting must not disturb the `consecutiveCBsOutput` latch. It is set at what is now L536 inside the CB branch *after* the old guard and cleared at L540 in the non-CB `else`, so a duplicate CB node reached neither before and reaches neither now. The existing "emits the CBs heading once" test covers that and stayed green, which is what I wanted from it — skipping a row must not open a fresh "CBs" section per skipped device.

src/components/operation-details/DeviceOperationsFullRender.tsx
aidemsined: Fixed in 3c71aaf5 exactly as suggested — `useMemo(() => processMemoryAllocations(deviceOperations), [deviceOperations])`.

You are right that it was doing double duty. The walk no longer repeats on renders driven only by `selectedAddress`, and the three returned values are now referentially stable, so the `renderNodes` `useCallback` memoises something for the first time. I also swapped `memoryAllocationList` out of its dep array for the new `memoryDetailsByNodeId`, which is what the callback actually reads now.

For scale on why this mattered: the walk itself measures 198 ms on a 32-device 51k-node graph after the other two fixes, and was 474 ms before them. Running that per legend click was the difference between a click and a stall.

src/functions/processMemoryAllocations.ts
aidemsined: Good catch that the criterion is stated over the whole heading while only `cbPeak` gained a device dimension. I checked the capture rather than closing the issue on an assumption, and widened it to the whole local corpus — grouping L1 `buffer_allocate` nodes by `(address, size)` and counting distinct `device_id`s per group:

| Finding | Result |
|---|---|
| L1 `buffer_allocate` nodes across all 24 reports | 5510 |
| Max distinct `device_id`s sharing one `(address, size)` | **1** |
| L1 `buffer_allocate` nodes in `multihost_poc_jun24_2321` | **0** — 14 `buffer_allocate` nodes, all DRAM, all `device_id=0` |

So nothing in the corpus fans a buffer out per device, and `totalBuffer` is not inflated. The repeated `(address, size)` groups that do show up are address reuse across allocate/deallocate cycles on one device. That reads as coherent: a mesh buffer is one logical allocation reported against device 0, where CBs are per-device program artifacts.

You were also right that the `buffers.device_id` line in the issue is about the table, not these nodes — it does not carry over.

The honest caveat is that the only multi-device capture has zero L1 buffers, so this is absence of evidence in the mesh case rather than proof. Recorded in #1844 with that caveat spelled out, including the note that if a future capture does emit L1 buffers per device, `totalBuffer` needs the same treatment and no current test would catch it, since the mesh fixtures are CB-only. Left as a scope note rather than code.

src/model/MemoryAllocations.ts
aidemsined: Confirmed at your true locations, and I agree with your second option — filed as #1879.

`cbList` is appended per `circular_buffer_allocate` node with no collapse, and `L1Plots` flattens every op's list into `memoryReportWithCB`, so with the toggle on that legend does show each CB eight times on the multihost report. Your read that the numbers are safe holds: `calculateCondensed` is envelope-based, so this is row repetition only.

What decided it for me is the point you make about grouping semantics. Entries there are sorted by address and merged with fragmentation entries rather than listed per DeviceOp, so a collapse needs its own decisions — which address survives, how a collapsed marker reads against the plot — instead of reusing `CBDeviceFanout`. That is design work, not a mechanical extension.

#1879 records the scope as the row-repetition half of criterion 3 for the surface the criterion did not name, and points at `getDeviceCountLabel()` so the wording stays consistent with the other two once it is done. Also noted on #1844 so criterion 3 is not read as fully closed by this PR.

src/functions/processMemoryAllocations.ts
aidemsined: Done in 3c71aaf5, and it is the biggest single win of the set — your framing that it is "the work this fix exists to deduplicate, just not deduplicated here" is exactly right.

Module-level `Map<string, CoreCoordList>` in `math.ts`. On your immutability caveat, I did not want to leave that to convention given `CBAllocationSummary.cores` now shares an array across rows, so it is enforced in the type system instead: `getCoresInRangeList` returns `CoreCoordList = readonly Readonly<CoreCoord>[]` (new export in `model/CoreCoord.ts`), and `CBAllocationSummary.cores` adopts it. That makes both `cores.push(...)` and `cores[0].x = 1` compile errors, so the assertion cannot rot the way a comment would. `tsc` came back clean, meaning nothing was mutating them.

I also checked the existing consumers rather than relying on the compiler alone — `CircularBufferPressureModal` only does `cb.cores.map(...)` and `for (const c of cb.cores)`, both read-only.

Measured contribution at 32 devices, 51,802 nodes, with the cached peak already in place: **318 ms → 198 ms**. Close to your 318 → 166, and the same conclusion.

src/functions/processMemoryAllocations.ts
aidemsined: You were right that one of the two had to be wrong. I resolved it by scanning the corpus rather than picking, and the answer is the **first** branch of your either/or, plus a wrinkle neither of us had:

| Report | CB nodes | with `device_id` | values |
|---|---|---|---|
| `abhinav` | 20 | **0** | — |
| `resnet50_mar06 diego 12` | 199 | 199 | **`'0'` (string)** |
| `multihost_poc_jun24_2321` | 128 | 128 | `0..7` (number) |
| 18 others | — | all | number |

So older captures really do omit it — the fallback and both tests are live, not dead — and one report emits it as a *string*, which means the declared `device_id: number` was wrong twice over.

Fixed in 3c71aaf5: `BaseMemoryParams.device_id` is now `?: number | string`, matching `DeviceOperationParams` in the same file, which already declared exactly that. The comment records both facts so the next reader does not write `node.params.device_id.toString()` on the strength of the old declaration. `String(... ?? SINGLE_DEVICE_KEY)` already normalised both forms, so no call site changed and `tsc` stayed clean — nothing was reading it as a bare number.

Worth noting the string case is not merely cosmetic: `String()` collapses `0` and `'0'` to the same key, so a capture mixing both still groups correctly.

src/components/operation-details/MemoryLegendElement.tsx
aidemsined: Done in 3c71aaf5 — `getCoreCountLabel` and `getDeviceCountLabel` now live in `src/functions/formatting.ts` and both components import them.

Your observation that they had *already* drifted again is the part that landed. The modal was rendering **`x 1 cores`**, which the shared helper fixes as a side effect since it carries the legend's `numCores === 1 ? 'core' : 'cores'`. I also gave the device label the same pluralisation for symmetry, even though today's only caller gates on `> 1`, so the helper is not a trap if someone later calls it unguarded.

The gating stayed at the call sites deliberately, since it genuinely differs — the legend gates on `isPerCoreBuffer`, the modal falls back to `unattributed` — so the helpers own the wording and nothing else. That keeps the next glyph or pluralisation decision a one-line change without pretending the two surfaces have the same rules.

On `61a5b34d` being evidence rather than a fix: agreed, and that is why this went in as a shared helper rather than a third hand-edit.

src/scss/components/DeviceOperationFullRender.scss
aidemsined: Fixed in 3c71aaf5. You are right that both halves were wrong for the scope they named: the template it quoted (`15px 0.5fr 2fr 13fr 3fr`) no longer applies to the per-DeviceOp row, and the marker is no longer in flow there — this rule positions it absolutely and zeroes the `margin-left: 6px` declared four lines under the comment claiming it.

Rewrote it in `MemoryLegendElement.scss` to state the part that is still true and scope-independent (icon-only outside the modal legend, because only the modal fits the full text; tooltip and aria-label carry the payload wherever the text is dropped), and to point at `DeviceOperationFullRender.scss` for the row that overrides the flow and margin. No stale template string, and the reader is sent to the override rather than being misled about it.

Fair hit that `3acaeee0` trimmed the sibling comment in that file and missed this one — I was reading the diff, and this comment is outside it.

tests/processMemoryAllocations.spec.ts
aidemsined: All three added in 3c71aaf5, and your predicted values were correct in both cases, so they went in as characterisation tests:

- **one device allocating an extra CB** — `dev0{4096@0x1000, 2048@0x2000}` + `dev1{4096@0x1000}` gives `peakMemoryLoad 6144` with rows `[deviceCount 2, deviceCount 1]`, and I asserted `maxBytes` too since that is where "max across devices" and "sum" visibly part company
- **devices disagreeing on size at one address** — `4096` vs `8192` gives `maxBytes 8192`, two rows of one device each, never the `12288` sum
- **two batches separated by `deallocate_all` in one op** — with the wrinkle noted below

The batch one is worth calling out, because the obvious shape of it proves nothing. With a symmetric `dev0, dev1 / dev0, dev1`, a stale `cbRepeatsByDevice` still lands both devices in the same slot and the result is identical either way — the test passes with the reset removed. It only bites when batch one holds a **subset** of batch two's devices (`dev0` / `dev0 + dev1`), where a carried-over counter puts dev1's first allocation and dev0's second in different slots and the batch reports two rows of one device instead of one of two. That is the shape I used, and I mutation-verified it: removing `cbRepeatsByDevice = new Map()` fails that test and only that test.

The same check on the other two — forcing a blanket device count across rows fails the extra-CB case, confirming they are not vacuous.

tests/DeviceOperationsFullRender.spec.tsx
aidemsined: Both added in 3c71aaf5, and your framing of the gap was accurate — asserting the heading exists while never clicking it left the only path the snapshot travels untested.

- **the wiring test**: renders a mesh graph, clicks `View per-core allocations`, asserts 2 `.cb-row` entries each matching `/x 8 devices/`
- **the mixed case**: `dev0` allocates both CBs, `dev1` only the first, so the modal must read `x 2 devices` on the first row and no device label on the second

The mixed one earns its place exactly as you predicted — I mutation-tested the blanket-count regression you were worried about by forcing every snapshot row to `Math.max(...deviceCounts)`, and it fails both that test and the extra-CB walker test. Under the homogeneous fixtures alone it passes clean.

One practical note for anyone extending these: the dialog renders through a portal, so the rows are not under the `container` returned by `render()` and have to be queried from `document`. My first attempt asserted on `container` and got zero rows with the click succeeding, which looks like a wiring failure but is not.

src/components/operation-details/MemoryLegendElement.tsx
aidemsined: Added in 3c71aaf5, mirroring the `core count label` block as suggested — a `device count label (#1844)` describe with: omitted → no label, `1` → no label, `8` → label present, and `{numCores: 2, deviceCount: 8}` → `/x 2 cores x 8 devices/`.

I also took your last point literally and pinned the asymmetry rather than just noting it: a fifth case renders `{numCores: 2, bufferType: DRAM, deviceCount: 8}` and asserts the core label is absent while the device label is still there. So the DRAM-row behaviour is now the documented contract instead of an accident, and if someone later decides `deviceCountLabel` should be gated on `isPerCoreBuffer` too, that test fails and forces the decision to be deliberate.

Agreed on the reasoning: the prop is public even though today's only call site is the CB row, and `numCoresLabel` having exhaustive coverage while its sibling had none was the inconsistency worth closing.

src/functions/processMemoryAllocations.ts
dcblundell: Verified and agreed — that is a thorough answer, and scanning the corpus rather than picking a side was the right call. Resolving.

One thing worth carrying forward before you close #1844: the finding is "no buffer fan-out observed across 24 local reports", which is not the same as a guarantee against a future tt-metal emit. `peakMemoryLoad = cbPeak + totalBuffer` would silently re-inflate if that ever changed, and the symptom (a heading that is wrong by a small integer factor) is exactly the kind that goes unnoticed for a release.

Could you record the corpus numbers in #1844 — or as a comment on the `totalBuffer` accumulation — so the next person meets the assumption rather than the symptom? Your own `device_id` investigation on the sibling thread is the argument for it: the same corpus turned out to contain a report emitting `device_id` as a string and another omitting it entirely, so it has form for surprising us.

dcblundell (approved):

PR #1881, opened 2026-08-14
Improve individual remote path test handling

### LINKED ISSUE #1856: Connection test reports one path at a time: a failing performance path discards the memory result already computed
Follow-up to review feedback on #1849 ([r3738061508](https://github.com/tenstorrent/ttnn-visualizer/pull/1849#discussion_r3738061508) and [r3738061523](https://github.com/tenstorrent/ttnn-visualizer/pull/1849#discussion_r3738061523)), deferred there because both changes push on the same shape and #1849 was already the PR moving the missing/empty distinction.

## Summary

`check_remote_path_for_reports` searches the memory path and the performance path in turn, and raises out of the middle the moment either one fails. So a user with a good memory path and a typo in their performance path is told about the typo and learns nothing about the path that works — even though the answer was computed a few lines earlier and then thrown away.

The original justification for this no longer holds. The two checks used to be separate SSH round trips, so short-circuiting saved real work. Since #1849 the root's existence is settled inside the same `find` command as the search, so reporting an outcome per path costs nothing extra.

## Current behaviour

`_count_reports_found` raises `RemoteConnectionException(FAILED)` for anything that isn't `RemoteSearchRootState.PRESENT`, so the second call abandons the first call's result:

```python
    profiler_count: Optional[int] = None
    if remote_connection.profilerPath:
        profiler_count = _count_reports_found(
            "Memory",
            find_folders_by_files(
                remote_connection, remote_connection.profilerPath, [TEST_DB_FILE]
            ),
        )
    ...
    performance_count: Optional[int] = None
    if remote_connection.performancePath:
        performance_count = _count_reports_found(
            "Performance", _find_performance_report_folders(remote_connection)
        )
```

`test_a_missing_performance_path_names_performance_not_the_profiler` pins this deliberately, so it's a decision rather than an oversight — just one worth revisiting.

The UI makes the loss visible rather than merely theoretical. `RemoteConnectionDialog.testConnectionStatus` seeds one `PROGRESS` placeholder per configured path (`MEMORY_REPORT_SEARCH_STATUS`, `PERFORMANCE_REPORT_SEARCH_STATUS`) and then replaces the whole list with whatever the server returns. When the performance path fails the response is `[SSH connection established, Performance directory does not exist…]`, so the memory row the user was watching **disappears** — it doesn't stay pending or show its result. The per-path placeholders already exist and already imply a per-path answer; the backend just can't supply one.

## Proposed fix

Give `RemoteReportCounts` a per-path outcome instead of `Optional[int]` per kind: each configured path carries either a count or a failure, and `test_remote_folder` renders a line for each. `_report_search_status` in `views.py` already owns the OK and WARNING copy for a path, so the failure copy moves in beside it rather than being raised from `sftp_operations.py`.

The two failure messages `_count_reports_found` raises today keep their wording, since they already name the same noun as the form field:

- `"{label} directory does not exist or cannot be accessed"` (`MISSING`)
- `"{label} directory could not be checked because the search did not complete"` (`UNKNOWN`)

## Fold in: a regular file at a folder path

The root check is `test -e` (`_report_search_command`), matching what the standalone check ran before #1849. That means a path pointing at a *file* passes the existence test and then reports as `"Memory path exists but no reports found"` — a folder-shaped warning for something that isn't a folder.

`test -d` alone would make the copy wrong in the other direction: the file does exist, so `"does not exist or cannot be accessed"` describes it worse than the empty-folder warning does. Doing it properly means `-d` **plus** a third message along the lines of `"{label} path is not a directory"` — which is the same per-path outcome work as above, hence bundling them so the copy is written once.

Unlikely misconfiguration compared with a missing path, so it rides along rather than justifying its own change.

## Acceptance criteria

- A connection test with a valid memory path and a missing performance path reports **both**: the memory report count (or empty-path warning) and the performance failure. The overall test still fails.
- The same holds with the kinds reversed, and when both paths fail.
- A path pointing at a regular file reports as not a directory, distinctly from missing and from empty.
- Every configured path produces exactly one status line, so no `PROGRESS` placeholder in `RemoteConnectionDialog` is ever left unresolved or silently dropped.
- Save gating is unchanged: any `FAILED` path still blocks saving the connection.
- Backend tests cover memory-ok/performance-missing, the reverse, both-failing, and the not-a-directory case, at view level as well as in `test_sftp_operations.py`.

## Out of scope

- Changing the `MISSING` / `UNKNOWN` / `PRESENT` distinction itself, or the `_MISSING_ROOT_EXIT_CODE` sentinel — both landed in #1849 and are the foundation this builds on.
- The listing flows (`get_remote_profiler_folders` and friends), which share `find_folders_by_files` but treat every non-`PRESENT` state as an empty result by design.

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1856.

This pull request refactors and improves the handling of remote report directory searches in the SFTP operations code. The main changes introduce more precise handling of edge cases (such as paths pointing to files instead of directories), improve error reporting for individual paths, and update the corresponding tests to reflect these changes. The result is a more robust and user-friendly experience when diagnosing remote connection and path issues.

**Improvements to remote path search and error handling:**

* Introduced a new state `NOT_A_DIRECTORY` in `RemoteSearchRootState` to distinguish between missing paths and paths that exist but are not directories, and updated the `_report_search_command` and `find_folders_by_files` to detect and report this case using a new exit code. [[1]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076R1009-R1017) [[2]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076R68-R71) [[3]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076L1161-R1193) [[4]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076R1303-R1308)
* Refactored the report path search logic: replaced the old `RemoteReportCounts` with `RemoteReportPathOutcome` and `RemoteReportSearchResults`, allowing each configured path's outcome (including errors) to be reported independently without failing the whole test if one path fails.
* Enhanced `_search_report_path` to capture and return path-specific errors instead of raising, except for connection-wide errors like authentication failures which are still raised.

**Test updates and improvements:**

* Updated and expanded tests to cover new behaviors, including the new `NOT_A_DIRECTORY` state, independent error reporting for each path, and ensuring that both paths' outcomes are reported even if one fails. [[1]](diffhunk://#diff-cf1b48187622774638f1cea2d004c0f133483e20a95795aa5d450309d77fddb4R297-R311) [[2]](diffhunk://#diff-cf1b48187622774638f1cea2d004c0f133483e20a95795aa5d450309d77fddb4R328-R370) [[3]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145R62-R75) [[4]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L774-R796) [[5]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L786-R818) [[6]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L1168-R1194) [[7]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145R1203-R1205) [[8]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L1193-R1267) [[9]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L1212-R1290)
* Added helper functions and fixtures in tests for more granular and readable assertions on path search outcomes.

**Exception and import consistency:**

* Added missing imports for new exceptions and constants in both implementation and test files to ensure consistency and correctness. [[1]](diffhunk://#diff-ab709cffd664d1dc76554fab2a4fc70dd7b21c1bc42be7ac4ed8fa3c56304076R23-R25) [[2]](diffhunk://#diff-cf1b48187622774638f1cea2d004c0f133483e20a95795aa5d450309d77fddb4L20-R37) [[3]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145L14-R17) [[4]](diffhunk://#diff-6f49e9dc09431fbb37a68fe92134102c30978c917810548cc6f08951f774c145R27-R32)


backend/ttnn_visualizer/views.py
```diff
@@ -1813,26 +1849,28 @@ def has_failures():
 
     # Both configured paths are checked and searched here, one SSH round trip
     # each: the search settles whether its root exists as part of the same
```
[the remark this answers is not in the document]
dcblundell: Not changing this: the behaviour predates the PR and doesn't have the effect described.

- The `except RemoteConnectionException` branch (add status, no early return) is unchanged from `dev` — this PR only adds a comment above it.
- An untrusted host key surfaces at `test_ssh_connection`, which runs first against the same host. `HostKeyVerificationException` subclasses `SSHException`, so it is caught by the generic `except SSHException` in `SSHClient.test_connection` and re-raised as a plain `RemoteConnectionException`. `has_failures()` is then true and the path search never runs. `_search_report_path` could only re-raise `HostKeyVerificationFailedException` if the host key changed between two round trips of the same request.
- Even in that race the frontend doesn't misclassify. `RemoteConnectionDialog` renders `axiosError.response.data` on error and the response body on success identically, and nothing keys off 422 for host-key trust — so 200 vs 422 changes nothing on screen, and the status that is rendered carries the host-key message, not a generic path failure.

The consistency point (auth returns early with its own status, host-key doesn't) is fair in the abstract, but it is pre-existing and currently unobservable, so it's out of scope here.

backend/ttnn_visualizer/sftp_operations.py
```diff
@@ -1041,77 +1054,111 @@ def _find_performance_report_folders(
...
-    performance_count: Optional[int] = None
+    performance: Optional[RemoteReportPathOutcome] = None
     if remote_connection.performancePath:
-        performance_count = _count_reports_found(
-            "Performance", _find_performance_report_folders(remote_connection)
+        performance = _search_report_path(
+            lambda: _find_performance_report_folders(
+                remote_connection, timeout_seconds=remaining_seconds()
+            )
         )
```
aidemsined, line 1178: > [!WARNING]
> A non-verdict error escaping the second search still discards the first path's outcome — a narrower instance of the defect #1856 is about.

`_search_report_path` deliberately doesn't catch the `SSHException` family because `remote_exception_handler` on `check_remote_path_for_reports` converts it. But that conversion happens *above* `_search_report_path`'s frame, so it escapes the whole function rather than being caught per-path. The decorator maps `NoValidConnectionsError`, `SSHException`, `FileNotFoundError` and the catch-all to a plain `RemoteConnectionException` with no custom status — so `is_connection_verdict` is `False`, the early return at views.py:1867 doesn't fire, and the request answers 200 with two lines against the three placeholders the dialog seeded.

Reachable sequence: the profiler search succeeds and returns a count, then the second `ssh` invocation exits 255 for a transient reason — connection reset, `MaxStartups`, a network blip — and the profiler outcome that was already computed and already correct is thrown away along with its status line.

It's rare, and saving is still correctly blocked because the surviving line is `FAILED`, so this isn't a blocker. But it's the same shape as the bug the ticket was filed for, and it's the one branch where AC4 doesn't hold.

**Suggested action:** convert per-search rather than per-function — apply `remote_exception_handler` to each search callable, or catch `SSHException` inside `_search_report_path` and convert there, so it sees the converted exception and can apply the `is_connection_verdict` test it already applies. A genuinely dead connection would then report the same line twice, which is honest and can't happen from a cold start anyway since `test_ssh_connection` runs first and `has_failures()` short-circuits. Worth a test asserting three lines for "profiler found, performance search dies with a 255".

backend/ttnn_visualizer/tests/test_sftp_operations.py
```diff
@@ -305,6 +335,268 @@ def test_a_timed_out_search_settles_nothing_about_its_root(self, connection):
...
+            budget = app.config["SSH_SUBPROCESS_TIMEOUT"]
+            with patch("subprocess.run", return_value=_completed("")) as run:
+                check_remote_path_for_reports(both_paths)
+
+        timeouts = [call.kwargs["timeout"] for call in run.call_args_list]
+        assert len(timeouts) == 2
+        assert timeouts[0] <= budget
+        # Whatever the first search spent is gone, so the second cannot ask for
+        # the full budget again.
+        assert timeouts[1] <= timeouts[0]
```
aidemsined: > [!WARNING]
> This test can't fail on the behaviour it names — reverting the shared budget leaves every assertion satisfied.

With `subprocess.run` mocked, remove the shared deadline so both searches request the full `SSH_SUBPROCESS_TIMEOUT` and the timeouts become `[budget, budget]`. Then:

- `timeouts[0] <= budget` ✓ (equal)
- `timeouts[1] <= timeouts[0]` ✓ (equal)
- `sum(timeouts) <= 2 * budget` ✓ (equal)

All three pass, so the test protects nothing. The assertion that would catch it — strict `<` — is the one not written, and the comment directly above it ("the second cannot ask for the full budget again") describes exactly the thing `<=` permits.

`remaining_seconds()` is strictly less than the budget on every call, so tightening is safe.

**Suggested action:** `assert timeouts[1] < timeouts[0]`.

Worth saying that the behaviour under test is a genuine win and deserves a real guard: a host where `find` hangs on both paths previously held the request ~250s (10 + 120 + 120) and now holds ~130s, which is the difference between surviving a typical 60–120s reverse-proxy read timeout and not.

tests/RemoteConnectionDialog.spec.tsx
```diff
@@ -283,6 +283,99 @@ describe('RemoteConnectionDialog connection test block', () => {
...
+        expect(screen.getByText('Testing SSH connection')).toBeInTheDocument();
+        expect(screen.getByText('Searching for memory reports')).toBeInTheDocument();
+        expect(screen.getByText('Searching for performance reports')).toBeInTheDocument();
+
+        resolveTest(PASSING_TESTS);
+
+        await waitFor(() => expect(screen.getByText('Found 3 memory reports')).toBeInTheDocument());
+        // Every placeholder is answered by a real result rather than being left pending.
+        expect(screen.queryByText('Searching for memory reports')).not.toBeInTheDocument();
+        expect(screen.queryByText('Searching for performance reports')).not.toBeInTheDocument();
```
aidemsined, line 394: > [!WARNING]
> These two assertions pass for the opposite of the reason the comment claims — the performance placeholder is gone because it was dropped, not because it was answered.

The test seeds and asserts three placeholders at `:338-340` (SSH, memory, performance), then resolves with `PASSING_TESTS`, which is **two** entries — SSH plus memory, with no performance line (`:131-134`). So `queryByText('Searching for performance reports')` is correctly absent, but only because the response replaced the list with something shorter and that row vanished.

That's precisely the failure mode AC4 exists to exclude ("no `PROGRESS` placeholder is ever left unresolved or silently dropped"), so the fixture demonstrates the bug rather than ruling it out.

The held-promise setup above it is a nice touch, by the way — the note about why a mock that resolves immediately would assert nothing about the placeholders is exactly right, and it's what makes the first three assertions meaningful.

**Suggested action:** resolve with a three-line fixture that includes a performance result, so the assertions pass because each placeholder was answered. Then add the case AC4 actually turns on as a separate test: a response with **fewer** lines than there were placeholders (which the backend can still produce — see the note on `sftp_operations.py`), asserting whatever the intended behaviour is.

aidemsined (commented):
Nothing merge-blocking here. The refactor is a real improvement on `RemoteReportCounts` — `Optional[int]` was carrying three meanings at once (not configured / a count / "we raised before you got here"), and moving every string into `views.py` while `sftp_operations.py` returns facts is the right seam. Every acceptance criterion has a test that would fail on regression, and the endpoint tests use exact list equality rather than containment, which is what makes them meaningful for AC4.

Three things worth a look, all in the same family: one branch where a partial outcome is still discarded, and two tests that pass for the wrong reason.

The "Code review round" commit is worth calling out — it made four substantive corrections rather than cosmetics. Replacing the subclass enumeration with `is_connection_verdict` is what makes host-key failures keep their 422 (on `dev`, `except AuthenticationFailedException` was enumerated by name and host-key failures fell through to 200). Collapsing the four inline copy guards into `_FAILURE_COPY_BY_ROOT_STATE` means a future enum member raises `KeyError` where the state arrives instead of falling through to the wrong warning. The shared timeout budget is a genuine behaviour win. And `find -H` fixes a symlinked root reading as empty.

## General observations

> [!NOTE]

- `is_connection_verdict` names a semantic ("this failure is about the connection") but tests a mechanism ("it carries a custom HTTP status"). On this path they coincide — the only reachable status-carrying exceptions are `AuthenticationFailedException` and `HostKeyVerificationFailedException`. But this module already holds the counterexample: sftp_operations.py:369-376 raises a bare `RemoteConnectionException` with a 422 for a *per-path* unreadable-folder error, citing the auth precedent as its justification. If that pattern ever migrates into the search path, `_search_report_path` re-raises it and discards the other path's answer. An explicit class attribute on the two subclasses keeps the no-enumeration benefit without inferring the category from the status code.
- views.py:1850-1852 — the comment says the branch keeps its own code for "rejected credentials or an untrusted host key", but an untrusted host key can't reach it as a verdict. `SSHClient.test_connection` has no `except HostKeyVerificationException`, so it falls into the generic `except SSHException` and is re-raised statusless: answered 200, with the host-key copy wrapped in a misleading "Ensure SSH key-based authentication is properly configured" suffix. Either narrow the comment, or add the branch so the initial SSH test answers 422 with the dedicated copy — which would match what the search path now does correctly.
- sftp_operations.py:1386-1394 — an unclassified nonzero `find` exit with no matches and no "permission denied" in stderr still falls through to `PRESENT`, so it renders as "path exists but no reports found". Pre-existing and out of this ticket's scope, but it's the same class of mis-description `NOT_A_DIRECTORY` was added to eliminate. Worth a follow-up issue.
- `find -H` also changes the listing flows the ticket declared out of scope: a symlinked `profilerPath` previously produced an empty listing and now produces its reports. Strictly better, and both halves are pinned by the new real-`find` tests — but it's a functional change to out-of-scope flows and isn't mentioned in the PR body.
- `RemoteReportPathOutcome` is four fields with an implicit two-field protocol: `error_message` being set means `root_state` is `UNKNOWN` for lack of information rather than as a finding, and callers must read them in that order. The docstring is explicit and `_report_search_status` honours it by checking `error_message` first, but nothing structural enforces it. Fine trade for two consumers in one module — noting it rather than asking for a change. The `# Not 'count': NamedTuple inherits tuple.count` note is the kind of comment that saves the next person ten minutes.
- A few things I checked that came back clean: exit code 88 avoids every conventional shell status (1, 2, 126, 127, 130, 255); `RemoteReportCounts`, `_raise_report_path_failure` and `_count_reports_found` have no remaining references anywhere in `backend/`, `src/` or `tests/`; the listing flows read only `.folders`, so `NOT_A_DIRECTORY` reaches them as an empty list exactly as `MISSING` already did; and save gating is genuinely unchanged, since `isConnectionSaveable` gates on status rather than count and so the `Optional[int]` → structured-outcome refactor can't invert it.
- The auth-failure tests drive from a `returncode=255` stderr string rather than injecting the converted exception, so they actually walk `raise_for_ssh_subprocess_error` → `remote_exception_handler`. That's what makes the "the decorator does the conversion" claim in the `_search_report_path` docstring verified rather than asserted — good instinct. Same for proving exit 88 against a real shell instead of a fabricated `CompletedProcess`.

Nothing was run from here, so `pnpm flask:test`, `flask:lint`, `flask:mypy` and vitest are unverified. Two specifics worth watching: mypy on the untyped `_FAILURE_COPY_BY_ROOT_STATE` lookup, and the two real-`find` symlink tests on CI's platform — GNU and BSD `find` do agree about `-H` with `-mindepth`/`-maxdepth`, but worth seeing green.

---

tests/RemoteConnectionDialog.spec.tsx
dcblundell: Agreed, and fixed. `PASSING_TESTS` is two entries against the three placeholders seeded above it, so the performance assertion passed by omission rather than by the row being answered.

Added a `PASSING_TESTS_BOTH_PATHS` fixture with a line per placeholder and switched this test to it, plus an assertion that the performance result actually renders — so the placeholder now goes because it was answered.

Added the shorter-than-seeded case as a separate test. Since the dialog replaces the whole list on response, the guarantee it can make on its own is that no `PROGRESS` row survives — nothing of its making is left spinning — and that is what the new test asserts. The "silently dropped" half of AC4 is the backend's to keep, and is fixed under your `sftp_operations.py` note, which means this shape should no longer arrive from our own server.

The sibling test below it was already consistent, for the record: it clears the performance path, so two placeholders match the two-line fixture.

backend/ttnn_visualizer/sftp_operations.py
dcblundell: Agreed, and fixed.

Confirmed the route end to end: `find_folders_by_files` and `_find_performance_report_folders` are undecorated, so a 255 with stderr matching none of the auth/host-key/connection fragments becomes `SSHException` at the final `raise` in `ssh_client.py`. That escapes `_search_report_path`, which only catches `RemoteConnectionException`, and is converted above it into one carrying no `http_status_code` — so `is_connection_verdict` is `False`, the early return doesn't fire, and the whole `else` block in `test_remote_folder` is skipped. Two lines against three placeholders, exactly as you described. Timeouts aren't affected, since `TimeoutExpired` is already handled per-path into `UNKNOWN`.

Converted per-search rather than per-function, as suggested: a `@remote_exception_handler`-decorated `_converted_search(remote_connection, search)` now sits between `_search_report_path` and each callable, so the conversion happens inside the `try` that already applies the verdict test. Verdict-carrying subclasses still re-raise, so the 422 behaviour and `test_a_connection_verdict_abandons_both_paths` are unchanged.

Two tests added: one at the search layer (the profiler count survives an `SSHException` from the second search) and one through the endpoint asserting 200 with three lines. I checked the first fails when the decorator is removed, so it does guard the behaviour rather than just passing.

Also corrected the comment above that handler in `views.py` — it claimed only connection failures reached there, which was the same mistaken assumption.

backend/ttnn_visualizer/tests/test_sftp_operations.py
dcblundell: Agreed, and fixed — all three assertions are strict now. You're right that `[budget, budget]` satisfied every one of them, so the test protected nothing against the behaviour its name claims.

`remaining_seconds()` is strictly decreasing and CI is `ubuntu-latest` only, so `time.monotonic()` has nanosecond resolution here and the strict form carries no flake risk.

Issue #1880, opened 2026-08-14
Operation graph: port perf overlay to React Flow and upgrade its visual encoding


aidemsined:
## Goal

Bring the perf overlay to the React Flow operation graph (#1809, phase 4) and, in the same pass, upgrade its visual language — the old encoding was shaped by what `vis-network` could paint on a canvas, and nodes are now DOM/CSS.

Parity baseline (what #1515 shipped, and what currently exists only in the retiring `OperationGraphComponent.tsx`):

- `Perf overlay` Switch with a status-dependent tooltip (`UNAVAILABLE` / `UNLINKED` / `READY`), disabled unless `READY`, reset when either report changes.
- Continuous colour ramp per op from log-normalised kernel duration (`aggregatePerfByOp` → `scoreOps` → `perfColorScale`).
- `PerfOverlayLegend` — gradient plus min/max bounds.
- `PerfOverlayOpMetric` — "Kernel duration" row with colour swatch in the side panel.

The data pipeline needs no changes: `src/functions/perfOverlay.ts`, `PERF_BINS` / `PERF_GRADIENT_CSS` in `src/definitions/GraphColors.ts`, and `PerfOverlayStatus` all carry over. `OperationGraphReactFlow` already declares `perfRows` / `isPerfReportLoaded`, currently unused.

## Why bundle look-and-feel with the port

| vis-network constraint | What it forced | Now possible |
|---|---|---|
| `size` is a canvas scalar that fights label layout, so nodes stayed at a fixed `size: 20` | Magnitude on a **single channel** (fill colour), even though the toggle's own tooltip promises "Colour and size nodes by per-op kernel duration" | A second channel: an inline duration bar (reusing the `OperationPerfRowBar` idiom already used in operation-list rows), border weight, or a rank badge |
| Labels are painted over the fill on the same canvas | Per-node label-colour flipping (`isDarkPerfColor` → light/dark font) and restore-styles that had to round-trip `(background, fontColor)` pairs | Move the perf colour off the label surface (coloured bar or border, neutral body) and the contrast bookkeeping disappears entirely |
| Node tooltip is a plain `title` string | Hover affordances limited to unformatted text | Blueprint tooltip with formatted duration, rank, and share of total |
| Legend was a `pointer-events: none` div | Static gradient only | Interactive legend — hover or click a band to dim ops outside it, reusing the filter fade that landed with the toolbar |
[...]

Issue #1809, opened 2026-07-28
Migrate the operation graph from vis-network to React Flow


aidemsined:
## Goal

Replace the operation graph's rendering layer (`vis-network`, `OperationGraphComponent.tsx`) with React Flow (`@xyflow/react`, already a dependency and used by the MLIR viewer). Reuse the MLIR viewer's rendering infrastructure without inheriting its domain complexity.

The data layer is unchanged: `GraphView.tsx`, `useOperationsList`, the perf-overlay pipeline (`perfOverlay.ts` / `GraphColors.ts`), and the side panel all stay. Only the render/interaction layer is rebuilt.

## Why now

- The op graph is a **flat DAG today** — no nested namespaces, regions, or Louvain sectioning — so the rendering swap itself is not a re-architecture.
- Typical graphs are **300–500 nodes** (larger ones exist but are rare), comfortably within React Flow's DOM/SVG budget.
- React Flow gives native rich HTML nodes/tooltips, a minimap, stock controls, and recomputes layout **without** the full `Network.destroy()`/recreate that vis does on every toggle.

## Collapse-aware from the start (#1583)

**Graph collapsing (#1583, “Auto-detect and collapse repeated subgraphs”, ms 0.97.0) is a hard architectural input to this migration.** Collapsing is exactly the MLIR capability a naive flat rebuild would throw away, so this migration is designed collapse-aware even though the collapse *feature* ships after core parity. If the op graph is rebuilt permanently-flat and collapsing is bolted on later, that's a second rewrite.

What #1583 needs from the renderer (its detection algorithm stays in the data layer):

| Capability | Adapt from MLIR |
|---|---|
| Parent/block node (`groupNode`) showing block name, op count, aggregate duration/memory | `mlirGroup` + `MlirGroupContext` |
| Expand / collapse state (collapsed by default; double-click / expand-one / expand-all) | MLIR namespace collapse state |
| Edge remapping — edges crossing a collapsed boundary reattach to the block node | MLIR region-edge remap |
| Buried-match filter — a filter hit inside a collapsed block surfaces (badge / auto-expand) | `resolveFilterMatches` buried-match path |

[...]

Issue #1614, opened 2026-06-03
Operation graph: stabilise vis-network DataSet across toolbar toggles (no reinit on Switch flip)


aidemsined:
## Context

Discovered while reviewing #1609 (perf overlay for #1515). Copilot flagged that toggling the new perf overlay Switch destroys + recreates the entire `vis-network` instance, resetting zoom / pan / selection. On closer inspection this is a **pre-existing** pattern in `OperationGraphComponent.tsx` — the `Hide deallocate ops` Switch has had the same behaviour for ages — but it's worth fixing once for all current and future toolbar controls instead of working around it per feature.

## Problem

The `nodes` `useMemo` in `OperationGraphComponent.tsx` lists several toolbar-driven state values as dependencies:

```ts
const nodes = useMemo(() => new DataSet<OperationNode>(...), [
    operationList,
    connectedNodeIds,
    filterOutDeallocate,   // toolbar Switch
    perfOverlayActive,     // toolbar Switch (added in #1515)
    scoreByOpId,           // changes when perf data loads
]);
```

When any of those toolbar values change, a brand-new `DataSet` is allocated. The `data` `useMemo` re-derives, which triggers the network-init `useEffect`, which calls `new Network(container, data, options)`. That:

- destroys the user's current zoom / pan / selection,
- causes a perceptible hitch on large graphs,
- discards the IO-highlight colouring applied by the last click.

For node clicks the existing code path is correct — `colorHighlightIO` calls `nodes.update(...)` in-place and the DataSet reference is unchanged. The friction is specifically on toolbar Switches.

## Affected toggles

- `filterOutDeallocate` (pre-existing).
- `perfOverlayActive` / `scoreByOpId` (added in #1515).
- Any future toolbar control that affects node appearance — e.g. the planned **metric selector** (Phase 3 in the perf overlay followups), **aggregation selector** (#1611), **multi-device filter** (Phase 7).

Fix once, every future toggle benefits.

## Proposed approach

1. Promote `nodes` from `useMemo` to a `useRef<DataSet<OperationNode>>` initialised once on mount.
2. Replace the rebuild paths with separate `useEffect`s, one per concern:
   - **Node membership** (driven by `operationList`, `connectedNodeIds`, `filterOutDeallocate`): compute the desired id set, then call `nodes.add(...)` for missing ids and `nodes.remove(...)` for ids no longer in the set. Don't tear down the whole DataSet.
[...]

Issue #1583, opened 2026-05-21
Operations graph: Auto-detect and collapse repeated subgraphs


aidemsined:
## Context

Most real models have heavy structural repetition — transformer layers, residual blocks, conv stacks. In current views (op list, op graph, perf table) every repetition is rendered as its own set of ops, so a 12-layer transformer becomes hundreds of near-identical rows / nodes. The graph view in particular becomes illegible at this scale (#1191), and finding *layer-level* signals requires manual visual pattern-matching.

## Ask

On the **operations graph only**, auto-detect strict contiguous repeated subgraphs and collapse each copy into a representative block so a 1,000-op model reads as dozens of units instead of a wall of identical ops.

## Behaviour

- Detection runs as part of graph layout (in the layout worker), on the ops that survive isolate-drop and the deallocate filter. Hiding deallocate ops can make two copies contiguous that were not before; toggling that filter folds every block again.
- Each copy becomes a `blockNode` **named from what it contains**, not by letter: the unique source-file stems of its members (`norm + attention + encoder + mlp × 24`), falling back to op names when there are no usable stems. A file covering most of the graph is dropped so the name identifies the tile and not the model; a block of a single operation type is named by that operation; and a leading prefix every stem shares is dropped, so four SentenceBERT modules read `self_attention + self_output + intermediate + output` rather than repeating `ttnn_sentencebert_` four times over. Letters survive only for disambiguation — a `· B` suffix when two patterns derive the same name, and `Block A` when a pattern yields no name at all. A derived name is strictly more informative than a letter, and the out-of-scope list below rules out only *user* rename. The node shows op count, summed Python duration, and summed memory delta. Collapsed blocks use a dashed blue frame so they do not read as ordinary ops.
- Folded by default. **Unroll** / **Fold** on the graph toolbar (all instances) and a pill chip on each block (that instance). Double-click a collapsed block to unroll it; double-click an unrolled member to fold that instance. Unrolled members are ordinary ops again — no extra grouping chrome.
- Selecting a collapsed block opens a **block details** panel (label, op range, instance *i* of *N*, aggregates, unique op types, boundary I/O). Not the first member’s operation panel: no Memory Details / Source.
- Filter: a member-name hit on a folded block counts the block as a visible match and shows `+N inside` on the node and in the filter counter. A URL / locate targeting an op inside a folded block unrolls that instance.
- Device-operation expand is opaque while a member is folded. Folding an instance drops any device-op expansion on its members.
[...]

PR #1882, opened 2026-08-14
Render the operation graph with React Flow

aidemsined:
## Summary
image

The operation graph now renders with React Flow instead of vis-network. The data layer is untouched — `GraphView`, `useOperationsList` and the side panel all stay; only the render/interaction layer is rebuilt.

The immediate win is that toggles no longer destroy and recreate the network. Filtering, selection and highlight are in-place restyles, so the layout — and any node you dragged — survives them.

Refs #1809 (phases 1–3, plus part of phase 4). See [Known limitations](#known-limitations-tracked-separately) for why this doesn't close the ticket outright.

### What renders

- New `src/components/operation-graph/` module: `OperationGraphReactFlow` (view), `OpGraphNode` / `OpGraphEdge`, `opGraphBuilder` + `opGraphLayout` (Dagre), `opGraphTypes`.
- Parallel edges bow along the chord's perpendicular. React Flow doesn't auto-curve duplicate `(source, target)` pairs the way vis did, and those are ~20% of edges on a real report — undisplaced, the twin hides under the first and loses its label. Bowing perpendicular (rather than by a fixed x offset) separates them at any orientation, including wide shallow pairs.
- Edge labels are gated on zoom. A resnet50 report carries **522** of them; below ~0.7 they draw under 8px, past reading, while still costing a text layout and repaint on every pan frame.

### Layout

- Dagre runs in a Web Worker. #1809 deferred this on the assumption of single-digit ms, but it measures 43ms on a typical 300-op report and runs into seconds on the rare large one, so it ships in v1 rather than as a later optimisation.
- If the worker dies or refuses the dispatch, layout falls back to the main thread instead of stranding the view on a stale graph. Dagre only reaches the main bundle when that actually happens.
- The spinner is held back 200ms, so the common sub-50ms build never flashes one.
- Relayout happens only when the node set or geometry changes: data/operation-range and hide-deallocate. Select, highlight, focus, locate, filter, pan and zoom are all in-place.

### Filter

- Substring/regex matcher with a 120ms debounce and sessionStorage mode persistence, replacing `SearchField` in the graph. Clearing bypasses the debounce so Escape stays instant.
- Non-matches dim rather than hide, so the matched subset keeps its position instead of the graph reflowing under the user mid-search. A query matching nothing leaves the canvas alone, because dimming every node to 18% reads as a rendering fault.
- An empty query and an uncompilable pattern both match nothing; `isRegexInvalid` is what tells the input "nothing typed" apart from "bad pattern", so a half-typed `(ttnn` shows an error and an empty field doesn't.
- The filter matches operation **names** only, while the node label also shows the id — so typing a visible id finds nothing. Split out as #1885 rather than guessed at, because widening the matched text has consequences for anchored regex queries.

### Perf overlay: deliberately not here

The overlay is **not** ported in this PR. It is scoped to #1880, which asks for more than a port: perf should get its own visual channel (a bar or a border) so it composes with the input/output highlight instead of competing with it for the node background. Painting the ramp onto the node fill — the obvious port, and what vis was forced into — makes the overlay and the I/O highlight mutually exclusive, which is exactly the compromise #1880 exists to undo. Building it here and redesigning it there would mean two passes over the same surface.

What this PR does contribute: `PerfOverlayStatus`, `PerfOverlayLegend` and `PerfOverlayOpMetric` move out of `OperationGraphComponent.tsx` into `definitions/` and `components/perf-overlay/`. That was a blocker for retiring the old component (the Styleguide imported them from it), and it is one of #1880's acceptance criteria already met. `OperationGraphReactFlow` keeps `perfRows` / `isPerfReportLoaded` as declared-but-unused props so `GraphView`'s call shape doesn't churn when the overlay lands.

Compact view is likewise absent: it was removed earlier in this branch as a control that did nothing, and that decision stands rather than being reversed by wiring it to the Dagre spacing knob.

**Edge-click navigation is deprecated, not deferred.** The vis graph focused an edge's target on click, and walked back to the source when you clicked the edge incoming to the current node. That is not carried here and is not planned — prev/next, click-to-locate and the side panel's locate buttons cover the same need. Called out so its absence reads as a decision rather than an oversight; recorded under Out of scope on #1809.

## Known limitations (tracked separately)

- **vis-network is not dropped yet.** The dead `OperationGraphComponent.tsx` and its SCSS are deleted as of `db65de8d`, but the dependency itself stays until the device-ops modal graph (`DeviceOperationsGraphComponent.tsx`) migrates — phase 5's tail.
- **Blink-on-locate and the zoom/pan help text did not carry over** from the old component. Both are phase-4 parity leftovers, filed as #1884.
- **No node cap or culling.** MLIR switches layout strategy above `DAGRE_NODE_LIMIT = 2000`; the op graph has no equivalent and `onlyRenderVisibleElements` is unused. Layout is not the risk — tight-tree measures 786ms at 4k nodes — rendering thousands of DOM nodes is.
- **The buried-match path is not carried.** #1809 asked for it up front as an inert seam for #1583, on the grounds that retrofitting is expensive. There are no group nodes yet, so it would be an abstraction with no consumer and nothing to test it against. The node-type registry, node `data` shape and edge model do leave room for a parent node. Flagging rather than quietly skipping, since the ticket's reasoning was explicit. Now recorded on #1583 as a renderer prerequisite, so it outlives this PR body.
- **Durable drag positions** stay a follow-up. Drags are ephemeral and reset on an explicit relayout, matching MLIR.
- **#1614 is superseded** — the destroy/recreate class of bug it describes no longer exists. Worth closing when this merges.

## Test plan

- [x] `pnpm test` — 1830 tests across 177 files green, including 7 new op-graph specs: `opGraphLayout` (Dagre ranking, unknown endpoints), `opGraphNodeStyles` (sizing, selection styling, the filter-dimming CSS contract), `graphFilterMatcher` (substring, regex, invalid patterns), `OpGraphEdgeLabelZoom` (the label gate at threshold), `useOpGraphLayoutWorker` (worker crash and dispatch-failure fallback, spinner delay, stale replies ignored), `opGraphLayoutWorker` (build coalescing, layout-cache hits, source-change invalidation), and `OperationGraphRebuildDiscipline` (which interactions may trigger a relayout, and node identity across a drag).
- [x] `tsc`, `eslint`, `stylelint` and `lint:spdx` clean.
- [x] Manual against a resnet50 memory report: select/highlight, prev/next, filter and match cycling, hide-deallocate, and the label gate measured at 522 → 0 → 522 across the threshold in both directions.
- [x] Rebased onto the `dev` merge at 99d08d82 and re-verified.

## Follow-ups

- #1880 — port the perf overlay and upgrade its visual encoding (phase 4's remaining half).
- #1884 — blink-on-locate and the zoom/pan help text.
- #1885 — filter should match the operation id, not just the name.
- #1887 — no node cap or culling on large reports; React Flow renders every node as DOM.
- #1888 — extract the shared seam between the op graph and the MLIR graph (filter matcher, Dagre wrapper, fitView presets).
- #1583 — collapse repeated subgraphs (phase 6; needs the group node, expand/collapse state, edge remap and the buried-match surfacing above).
- #1809 — phase 5 tail: migrate the device-ops modal graph, then drop `vis-network` / `vis-data` and the `vis` chunk in `vite.config.ts`.
- #1614 — close as superseded.


src/components/operation-graph/useOpGraphLayoutWorker.ts
```diff
@@ -0,0 +1,174 @@
...
+            if (!worker || sourceVersionRef.current === 0) {
+                return;
+            }
+            const requestId = nextRequestIdRef.current + 1;
+            nextRequestIdRef.current = requestId;
+            activeRequestIdRef.current = requestId;
+            optionsRef.current = options;
+            if (spinnerTimeoutRef.current === null) {
+                spinnerTimeoutRef.current = window.setTimeout(() => setIsBuilding(true), BUILD_SPINNER_DELAY_MS);
+            }
```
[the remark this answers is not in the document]
aidemsined: Partly acted on, and the specific failure you describe can't happen — but there was a real smell next to it.

**Why the scenario doesn't occur:** a superseded build never clears the timer. `runBuild` advances `activeRequestIdRef`, and `onmessage` drops any reply whose `requestId` isn't the active one *before* reaching `endBuild()`. So the first build finishing while the second is in flight is a no-op — its reply is discarded, the timer survives, and the second build keeps the pending spinner.

**What was actually wrong:** the timeout callback set `isBuilding` but left `spinnerTimeoutRef.current` pointing at a timer that had already fired, so "non-null" stopped meaning "a timer is pending". Harmless today, but it makes the guard read as load-bearing when it isn't. Fixed in d8cdce07 by nulling the ref as the timer fires.

**Why I didn't restart the timer per `requestId`:** that would be a regression. The delay exists so a fast build never flashes a spinner, and the clock deliberately runs from the first build of a burst rather than the latest request. Restarting it means someone toggling `Hide deallocate ops` repeatedly gets no spinner for as long as they keep toggling, however long they have actually been waiting — the 200ms window would never elapse.

I implemented your version to check, and it fails exactly one test (the new one) and no others:

```
× measures the spinner delay from the first build of a burst
 Tests  1 failed | 5 passed (6)
```

So the burst behaviour is now pinned down rather than incidental — worth having, since the restarting version looks more correct at a glance.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -0,0 +1,165 @@
...
+                    <Switch
+                        className='op-graph-toolbar-switch'
+                        checked={isPerfOverlayActive}
+                        onChange={(event: FormEvent<HTMLInputElement>) =>
+                            onPerfOverlayChange(event.currentTarget.checked)
+                        }
+                        label='Perf overlay'
+                        disabled={isDisabled || perfOverlayStatus !== PerfOverlayStatus.READY}
+                    />
+                </Tooltip>
```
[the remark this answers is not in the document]
aidemsined: Moot now — the code this points at no longer exists. The perf overlay switch was removed from this PR in 14bbf6ea; the overlay is scoped to #1880, which asks for a different visual encoding rather than a straight port, so building it here would have meant building it twice.

The underlying point is sound and worth carrying over, though, so I've noted it on #1880: a Blueprint `Tooltip` wrapping a `disabled` control won't fire, and the whole reason that switch has a three-state tooltip (`UNAVAILABLE` / `UNLINKED` / `READY`) is to explain the disabled case. A tooltip that only works when the control is *enabled* explains the one state that needs no explanation.

Worth noting the same pattern is live today in the component this PR retires (`OperationGraphComponent.tsx`, the `Perf overlay` switch), so this is a pre-existing bug being inherited rather than introduced — #1880 is the right place to fix it, since that ticket rebuilds the control anyway.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -0,0 +1,165 @@
...
+                >
+                    <Button
+                        variant={ButtonVariant.OUTLINED}
+                        disabled={isDisabled || selectedOperationId === null}
+                        onClick={() => selectedOperationId !== null && onGoToOperation(selectedOperationId)}
+                        aria-label={`Center on operation ${selectedOperationId}`}
+                    >
+                        {selectedOperationId}
+                    </Button>
+                </Tooltip>
```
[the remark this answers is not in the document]
aidemsined: Fixed in d8cdce07. Agreed — and the inconsistency with the two buttons either side of it is the giveaway that it was an oversight rather than a choice:

```tsx
aria-label={
    selectedOperationId !== null
        ? `Center on operation ${selectedOperationId}`
        : 'No operation selected'
}
```

I left the `Tooltip` `content` interpolating as-is: the tooltip is already `disabled` on the same condition, so it never renders with a null in it, and adding a second conditional would suggest a case that can't occur. The `aria-label` was the real exposure, since it's on the DOM node regardless of the disabled state.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -0,0 +1,596 @@
...
+    // mounted, so the incoming id has to be adopted rather than only seeding the
+    // initial state. Adjusting during render instead of in an effect avoids a
+    // pass where the panel still describes the operation we just left.
+    const [adoptedOperationId, setAdoptedOperationId] = useState(operationId);
+    if (operationId !== adoptedOperationId) {
+        setAdoptedOperationId(operationId);
+        if (operationId !== undefined) {
+            setSelectedOperationId(operationId);
+        }
+    }
```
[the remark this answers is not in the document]
aidemsined: Not changing this one — the render-phase update is the documented pattern here, and the effect is the anti-pattern it replaced.

React's ["You Might Not Need an Effect"](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) prescribes exactly this shape for adjusting state when a prop changes:

```js
function List({ items }) {
  const [prevItems, setPrevItems] = useState(items);
  if (items !== prevItems) {
    setPrevItems(items);
    setSelection(null);
  }
}
```

The guarantee that makes it safe: when a component calls `set` during render, React re-runs it immediately with the new state and **discards the in-progress render output without committing it**. Nothing is painted from the stale pass and no child re-renders from it.

That is precisely why the effect version is worse here. `/graphtree/:operationId` navigation keeps this component mounted, so with an effect there is a committed render where `selectedOperationId` still points at the operation we just navigated *away from* — the side panel renders that operation's name, tensors and memory-config tables, then throws them away one pass later. Visible on a slow report, and it's a wasted render of the most expensive subtree in the view.

On the specific risks you raise:

- **Render loop** — not possible. The update is guarded by `operationId !== adoptedOperationId` and the same render sets `adoptedOperationId` to `operationId`, so the guard is false on the re-run. It converges after one extra pass, and React errors out rather than looping if a guard like this is ever gotten wrong.
- **StrictMode** — fine, and this is checked in dev. The update is idempotent, which is what StrictMode's double-invocation is testing for. StrictMode warns about `setState` in an effect *cleanup* or in the render of a *different* component; a guarded self-update during one's own render is explicitly sanctioned.
- **Warnings** — none. React only warns for cross-component render-phase updates ("Cannot update a component while rendering a different component"). This one updates its own state.
- **"Keep the current selection when `operationId` becomes undefined"** — already the behaviour: the inner `if (operationId !== undefined)` guard means an undefined param adopts the new value for comparison purposes but leaves the selection alone.

The one fair criticism is discoverability, since the pattern surprises readers who haven't hit that docs page. The comment above it explains the *why*; I've left the code as-is rather than trade a real extra render for familiarity.

src/components/operation-graph/opGraphLayoutWorker.ts
```diff
@@ -0,0 +1,44 @@
...
+        operations = message.operations;
+        return;
+    }
+
+    // The view also discards mismatched replies; bailing here skips the layout.
+    if (message.sourceVersion !== sourceVersion) {
+        return;
+    }
+
+    try {
```
dcblundell, line 73: **Must fix:** worker builds are never coalesced or cancelled

The main thread discards stale replies (`useOpGraphLayoutWorker.ts:112`) but the worker still runs every `BUILD` to completion. This is reachable from the shipped UI: `FooterInfobar.tsx:45` only excludes NPE and `/operations/\d+`, so the op-range slider is live on `/graphtree`, and `RangeSlider.tsx:306` wires Blueprint's continuous `onChange` with no `onRelease` or debounce — so each pointer frame structured-clones ~500 source ops and queues a full Dagre run, and at 43ms/layout the worker falls behind realtime and keeps grinding after the drag ends.

`mlirLayoutWorker.ts` already solves this — `latestBuildByGraphId` (at most one pending build), a `newerRequestExists`/`graphVersion` re-check after layout before posting, and `postedGraphIdRef` guarding `set-graph`. A main-thread debounce helps the symptom but does not stop the queue backing up.

src/components/operation-graph/opGraphLayoutWorker.ts
```diff
@@ -0,0 +1,44 @@
...
+        const graph = buildOpGraph(operations, { hideDeallocate: message.hideDeallocate });
+        postMessage({
+            type: OpGraphWorkerMessageType.BUILT,
+            sourceVersion: message.sourceVersion,
```
dcblundell: **Should fix:** no layout result cache, so every hide-deallocate flip re-runs Dagre

The build-option space here is two values, so toggling the switch off and back on re-lays out a graph the worker built moments earlier — 43ms typical, seconds on a large report, spinner included. `mlirLayoutWorker.ts` keeps an LRU keyed on `cacheKey` for exactly this, and #1809's "Lift: `touchLruCache`" is currently unmet and not covered by #1888.

A two-entry map keyed on `${sourceVersion}:${hideDeallocate}`, cleared on `SET_GRAPH`, covers it.

dcblundell (commented):
Conductor review (DRY / TEST / PERF / STANDARDS / SECURITY / REQUIREMENTS). Must-fix and should-fix findings are left inline below. Security is clean — no injection surface, worker messaging is properly versioned, sessionStorage is bounded and validated. Consider/nit items and the spec-vs-code decisions on #1809 (buried-match path, scale slider) are not posted here.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -0,0 +1,484 @@
...
+            if (isSelected) {
+                styled = { ...styled, className: SELECTED_NODE_CLASS };
+            } else if (highlight) {
+                const relation = highlight.relationByNodeId.get(node.id);
+                if (relation) {
+                    styled = { ...styled, className: NODE_CLASS_BY_RELATION[relation] };
+                }
+            }
+            if (matchedIds && !isSelected && !matchedIds.has(node.id)) {
+                styled = { ...styled, style: { ...styled.style, opacity: FILTER_DIM_OPACITY } };
```
dcblundell: **Must fix:** filter dimming re-creates every node/edge object, defeating React Flow's equality check

`adoptUserNodes` reuses an internal node only when `userNode === internals.userNode`, and `EdgeWrapper` compares with `Object.is` — so object identity, not style identity, is what forces a re-render. With a filter active the early return at `:369`/`:391` is bypassed and every non-match gets a fresh spread. Because the memo deps include `nodes`, and `useNodesState` returns a new array per pointer frame, **dragging one node with a filter applied re-renders ~500 `NodeWrapper`s per frame**; every post-debounce keystroke does the same to nodes and ~522 edges.

Hoisting a shared `DIMMED_STYLE` constant does not fix this — the spread still yields a new object. Move dim state out of the element objects into a container class (`.op-graph-filtering .react-flow__node { opacity: .18 }` plus an un-dim class on the usually-small matched set); the SCSS hooks already exist at `OperationGraphReactFlow.scss:43-53,82-88`, and it also makes the manual `opacity` hack at `OpGraphEdge.tsx:60` unnecessary. Same treatment fixes the `{ ...styled, className }` churn at `:376`/`:380`.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -0,0 +1,484 @@
...
+                onPaneClick={handlePaneClick}
+                minZoom={MIN_ZOOM}
+                maxZoom={MAX_ZOOM}
+                nodesConnectable={false}
+                selectNodesOnDrag={false}
+                proOptions={{ hideAttribution: true }}
+            >
+                <Background />
+                <Controls />
+                {!isPanelOpen ? <MiniMap pannable /> : null}
```
dcblundell: **Must fix:** the MiniMap is effectively never visible

`isPanelOpen = selectedOperationId !== null && !isBuilding` (`:419`), and `onBuilt` (`:163-172`) always resolves a selection — falling back to `graph.nodes[0]` when nothing is selected. So after every build a node is selected and the MiniMap unmounts; on the default `/graphtree` route the user only ever sees it by clicking empty canvas. #1809 phase 2 asks for "stock Controls/MiniMap/Background".

The panel is a full-height right column (`OperationGraphReactFlow.scss:112-117`), so `position="bottom-left"` satisfies both the criterion and the "panel covers the corner" rationale in the comment. Separately, folding `isBuilding` into the same boolean mounts a ~500-rect MiniMap over the old graph on exactly the builds already slow enough to trip the spinner, then unmounts it — worth gating the panel on `isBuilding` and the minimap on selection alone.

src/components/operation-graph/opGraphBuilder.ts
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+    return candidates;
+}
+
+export function buildOpGraph(
+    operations: OpGraphSourceOperation[],
+    { hideDeallocate }: OpGraphBuildOptions,
+): OpGraphBuiltGraph {
+    const candidates = collectCandidateEdges(operations);
```
dcblundell, line 41: **Must fix:** no direct spec for the builder

This module decides node membership, deallocate filtering and parallel-edge indexing, and it runs in **both** the worker and the main-thread fallback — so a defect here is invisible in every environment. It is reached today only incidentally by `tests/useOpGraphLayoutWorker.spec.tsx:109-110` asserting a 2-node/1-edge count. It is a pure function with no DOM and no mocking cost, and is currently the worst-covered file in the PR.

Untested invariants the code itself calls out as load-bearing: connectivity is computed from `candidates` *before* the deallocate filter (`:44-49`), so hiding a deallocate must not drop its neighbour nor pull in an always-isolated op; and `parallelCountByPair` (`:72-78`) must yield unique ids with a monotonic `parallelIndex` per `(source, target)`.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -0,0 +1,484 @@
...
+
+    const { runBuild, isBuilding } = useOpGraphLayoutWorker(sourceOperations, onBuilt);
+
+    // The only input that changes the node set or the layout geometry, and so the
+    // only one that may trigger a rebuild.
+    const buildOptions = useMemo<OpGraphBuildOptions>(() => ({ hideDeallocate }), [hideDeallocate]);
+
+    // `sourceOperations` isn't read here — it's the signal that the worker holds a
+    // new report and the current build is stale.
+    useEffect(() => {
```
dcblundell, line 201: **Must fix:** relayout-trigger discipline is unguarded by any test

The whole design rests on `buildOptions` memoising on `hideDeallocate` alone and this effect depending only on `[runBuild, sourceOperations, buildOptions]`. The code is correct today — adding `filterQuery` or `selectedOperationId` to that list is the single most likely future regression, and it would compile, lint, and pass every existing test while silently relayouting on each keystroke.

Nothing anywhere counts `runBuild` invocations. The observable contract worth pinning: it increments on `operationList` identity change and on hide-deallocate toggle, and must not increment on node click, pane click, filter typing, mode toggle, or match stepping.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -0,0 +1,484 @@
...
+            // An edge between two matches stays lit so the matched subset is
+            // traceable; a selection edge outranks the filter either way.
+            if (matchedIds && !relation && !(matchedIds.has(edge.source) && matchedIds.has(edge.target))) {
+                styled = { ...styled, style: { ...styled.style, opacity: FILTER_DIM_OPACITY } };
+            }
+            return styled;
+        });
+    }, [edges, highlight, matchedIds]);
+
+    const handleNodeClick = useCallback(
```
dcblundell, line 478: **Should fix:** keyboard node selection is inert

React Flow v12 makes nodes focusable and handles Enter/Space in its own `onKeyDown`, but that path calls the internal `handleNodeClick` store action — it never invokes the user-supplied `onNodeClick` (`@xyflow/react/dist/esm/index.mjs:2278` vs `:2282`). So a keyboard user can tab to a node and press Enter: React Flow applies its own `.selected` style while `selectedOperationId`, the I/O highlight, the toolbar centre button and `OpGraphInfoPanel` all stay on the previous node.

Either drive selection through `onSelectionChange` / `node.selected`, or set `nodesFocusable={false}` so the node stops claiming focus it cannot service.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -0,0 +1,484 @@
...
+
+const NODE_TYPES = { [OpGraphNodeType.OP]: OpGraphNode };
+const EDGE_TYPES = { [OpGraphEdgeType.OP]: OpGraphEdge };
+
+const EDGE_MARKER = { type: MarkerType.ArrowClosed, width: 18, height: 18 } as const;
+
+// A large report only fits at extreme zoom-out; 3 caps zoom-in as vis did.
+const MAX_ZOOM = 3;
+const MIN_ZOOM = 0.02;
+const FOCUS_ZOOM = 1;
```
dcblundell, line 52: **Should fix:** every jump snaps the viewport to zoom 1

`focusOperation` (`:188-200`) calls `setCenter(..., { zoom: FOCUS_ZOOM })` with `FOCUS_ZOOM = 1` hard-coded, and that one function backs match cycling, prev/next op, the panel's locate buttons and the post-build recentre. A user zoomed out to survey the graph loses that view on every "next match".

This is a regression against both references: vis passed the current `scale` through (`OperationGraphComponent.tsx:481-484`), and MLIR uses `fitView({ nodes: [...], ...MLIR_FIT_VIEW_OPTIONS.localJump })`. #1809 asks for "prev/next cycling with `fitView`" and for lifting the `MlirFitView` presets — switching to `localJump` discharges both, and its `padding: 0.3, duration: 200` already match the 200ms re-derived at `:53`.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -0,0 +1,484 @@
...
+                onNextMatch={goToNextMatch}
+                selectedOperationId={selectedOperationId}
+                previousOperationId={previousOperationId}
+                nextOperationId={nextOperationId}
+                onGoToOperation={selectOperation}
+                hideDeallocate={hideDeallocate}
+                onHideDeallocateChange={setHideDeallocate}
+                isDisabled={isBuilding}
+            />
+            <ReactFlow<OpGraphFlowNode, OpGraphFlowEdge>
```
dcblundell, line 519: **Should fix:** edge-click traversal was dropped with no record anywhere

The #1809 spike comment listed it as a live parity item (`OperationGraphComponent.tsx:640-648` — clicking an edge focuses the *other* end, deliberately skipping I/O re-highlight so you can walk back). There is no `onEdgeClick` on this element, and unlike blink-locate and the help text it appears in neither the PR body's Known limitations nor any follow-up issue.

It is the only spike-flagged parity item that is neither implemented nor tracked. Either add the handler (a few lines, given `edgesBySource`/`edgesByTarget` already exist) or fold it into #1884 so the drop is a decision rather than an omission.

src/components/operation-graph/useOpGraphLayoutWorker.ts
```diff
@@ -0,0 +1,182 @@
...
+        },
+        [endBuild],
+    );
+
+    useEffect(() => {
+        const worker = new Worker(new URL('./opGraphLayoutWorker.ts', import.meta.url), { type: 'module' });
+        workerRef.current = worker;
+        // A crash emits no terminal reply, so the layout has to be redone here or
+        // the view keeps whatever graph it had with no way back.
+        worker.onerror = () => {
```
dcblundell, line 96: **Should fix:** a dead worker strands the spinner permanently

The main-thread fallback is wired only to `worker.onerror`, which fires once when the worker script fails to load or throws at top level. After that, `runBuild` keeps calling `worker.postMessage` (`:166`) on a dead worker — that does not throw, so the `catch` fallback at `:172` never runs, no reply ever arrives, `endBuild` is never called, and the view sits behind the spinner with no recovery.

Latch a `workerFailedRef` in `onerror` and route subsequent builds straight to `buildOnMainThread`, or add a per-request watchdog. The existing `FakeWorker.crash()` helper covers the first failure but not a build issued after one. Related: the worker's own `ERROR` reply path (`opGraphLayoutWorker.ts:37-43`) — the most likely real failure, since it is what a throwing `buildOpGraph` produces — is also untested.

src/components/OperationGraphComponent.tsx
```diff
@@ -22,13 +22,16 @@ import { StackTraceLanguage } from '../definitions/StackTrace';
 import MemoryConfigRow from './MemoryConfigRow';
 import { ShardSpec } from '../model/MemoryConfig';
 import { BufferTypeLabel } from '../model/BufferType';
-import { formatDuration, toReadableShape, toReadableType } from '../functions/formatting';
+import { toReadableShape, toReadableType } from '../functions/formatting';
 import SearchField from './SearchField';
 import MemoryTag from './MemoryTag';
 import { GRAPH_COLORS } from '../definitions/GraphColors';
+import { NodeRelation } from '../definitions/NodeRelation';
```
dcblundell: **Should fix:** delete this component rather than editing it

Nothing imports it after `GraphView.tsx:11` switched over, yet this PR edits it to re-import the extracted pieces — so ~1150 unreachable lines plus `OperationGraphComponent.scss` ship, and `groupTensorsByConnectedOp` (`:912-947`) becomes a live second copy of `getConnectedOpGroups` (`OpGraphInfoPanel.tsx:31-70`): same algorithm, same external-input/unconsumed-output keys, differing only in `opId` vs `operationId`. `ConnectedOpHeader` (`:948-975`) likewise duplicates the locate-button block at `OpGraphInfoPanel.tsx:141-158`. Both copies typecheck and lint, so nothing forces them to move together.

The stated blocker (the Styleguide importing the perf-overlay pieces) is resolved by this same diff, and `vis-network` stays for `DeviceOperationsGraphComponent.tsx` regardless — nothing else is gated on keeping the file. Taking up the PR body's offer.

src/components/operation-graph/OpGraphFilter.tsx
```diff
@@ -0,0 +1,157 @@
...
+    matchCount: number;
+    // 0-based cursor within `matchCount`, or null when no match is focused.
+    currentMatchIndex: number | null;
+    onPrev: () => void;
+    onNext: () => void;
+    isDisabled: boolean;
+}
+
+// Input, counter and prev/next only; the fade itself lives in the view.
+const OpGraphFilterInner = forwardRef<OpGraphFilterHandle, OpGraphFilterProps>(
```
dcblundell: **Should fix:** verbatim copy of `MlirOpFilter`, and #1888 does not cover it

`:30-157` matches `MlirOpFilter.tsx:38-166` down to `spellCheck={false} autoComplete='off'` — identical `useImperativeHandle`, `handleKeyDown`, counter-text ladder including the stale-cursor guard, and `toggleMode`. The only deltas are the class prefix, `hiddenMatchCount` (MLIR only) and `isDisabled` (op-graph only). `OpGraphFilter.scss` is byte-identical to `MlirOpFilter.scss` apart from six selector prefixes, and the filter *state harness* in `OperationGraphReactFlow.tsx:99-105, 229-251, 255-267, 307-326` duplicates `MLIRViewReactFlow.tsx` too — including `FILTER_DIM_OPACITY = 0.18` and `FILTER_DEBOUNCE_MS = 120` as separate module-private literals in two files.

#1888's stated scope is "filter matcher, Dagre wrapper, fitView presets", so this copy is deferred to nothing. Either widen #1888 explicitly, or unify now — which also gives the op-graph filter the 20 cases in `tests/MlirOpFilter.spec.tsx` that it currently lacks entirely.

src/definitions/PerfOverlayStatus.ts
```diff
@@ -0,0 +1,18 @@
...
+// Availability ("is a perf report loaded at all") and linkage ("do its ops
+// overlap this graph") are separate so the tooltip can tell the user which of
+// the two is missing.
+export enum PerfOverlayStatus {
+    UNAVAILABLE,
+    UNLINKED,
+    READY,
+}
+
+export const PERF_OVERLAY_TOOLTIP: Record<PerfOverlayStatus, string> = {
```
dcblundell, line 14: **Should fix:** two comments went stale when this map moved here

Anchored here because neither cited line falls inside a diff hunk. `BufferSummaryPlotControls.tsx:33` still reads "Mirrors the `PERF_OVERLAY_TOOLTIP` map in `OperationGraphComponent`" — the map now lives in this file, and that pointer dangles entirely once the dead component is deleted. Repoint it at `src/definitions/PerfOverlayStatus.ts`.

Separately, `GraphView.tsx:72-77` describes a tooltip distinguishing "no perf report loaded" (UNAVAILABLE) from "loaded but doesn't match this graph" (UNLINKED), but `OperationGraphReactFlow` declares `perfRows`/`isPerfReportLoaded` as unused placeholders and renders no such tooltip. Retarget it at #1880 like the props comment does, or drop it until the overlay lands.

tests/useOpGraphLayoutWorker.spec.tsx
```diff
@@ -0,0 +1,197 @@
...
+        });
+        act(() => result.current.runBuild(BUILD_OPTIONS));
+        act(() => {
+            vi.advanceTimersByTime(SPINNER_DELAY_MS / 2);
+        });
+
+        expect(result.current.isBuilding).toBe(true);
+    });
+
+    it('ignores a built reply that belongs to a superseded request', () => {
```
dcblundell, line 188: **Should fix:** missing `cleanup()` leaves a live spinner timer after teardown

`vitest.config.ts` does not set `globals: true`, so RTL auto-cleanup is off and nothing unmounts these hooks — all six stay mounted for the file's lifetime. That matches `tests/useMlirLayoutWorker.spec.tsx`, but this hook is the first with a real-timer 200ms spinner, which changes the calculus.

This last test runs on real timers, calls `runBuild` (arming the spinner at `useOpGraphLayoutWorker.ts:160`), then emits a reply with a superseded `requestId` — which returns early at `:112` *without* calling `endBuild()`. The timer is left armed on a hook that is never unmounted, racing the file teardown: sometimes clean, sometimes a `setIsBuilding` outside `act()`. `afterEach(cleanup)` resolves it — the hook's unmount effect (`:96-103`) clears the timer.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Confirmed and fixed in 0c4b79c0.

I ran the branch against a real report before touching anything, because "never visible" sounded like it might be overstated. It isn't, quite: `.react-flow__minimap` is 0 elements on a fresh `/graphtree` load and 1 after a click on empty canvas. So it *is* reachable — but only from a state the user has no reason to enter, with nothing to suggest the click. Your mechanism is exact: `onBuilt` resolves a selection on every build, so `isPanelOpen` is true in every state the user lands in.

I dropped the gating entirely rather than moving it to selection alone. Once the minimap is docked opposite the panel there is nothing left for the two to trade places over, and an unconditional mount is what phase 2's "stock Controls/MiniMap/Background" actually asks for. That also disposes of the mount-during-build churn you flagged in the second paragraph. `isPanelOpen` keeps its `isBuilding` term — it now only gates the panel, which is the case it was right for.

One wrinkle worth recording for the next person: `bottom-left` is already the `<Controls>` corner, and React Flow's own `.react-flow__panel.left { left: 0 }` matches an override at equal specificity while landing later in the cascade, so setting `left` silently does nothing. The offset is on `margin-left` instead. Measured in the browser afterwards: minimap at x=83, 15px clear of the controls, panel starts at x=940.

src/components/operation-graph/opGraphBuilder.ts
aidemsined: Fixed in 0c4b79c0 — `tests/opGraphBuilder.spec.ts`, 11 cases.

Both invariants you named are pinned directly rather than incidentally:

- **Connectivity before the filter.** Three cases: an op whose only neighbour is a hidden deallocate stays in the graph, an always-isolated op is not pulled in when a deallocate is hidden, and the edges into a hidden deallocate are dropped rather than left dangling. The first is the one that would silently regress if the filter ever moved above the `connected` set.
- **`parallelCountByPair`.** Ids and indices asserted exactly (`1-2-0`, `1-2-1`, `1-2-2`), plus a cycle case proving each ordered pair counts separately, plus a uniqueness check across a graph with several parallel pairs.

Also covered: node membership and label/`filterString` shape, case-insensitive deallocate matching across both name forms, positions coming back from Dagre rather than the seed origin, and an empty report not throwing.

src/components/operation-graph/useOpGraphLayoutWorker.ts
aidemsined: Fixed in 0c4b79c0. Went with the latch rather than a watchdog: `isWorkerDeadRef` is set in `onerror` and cleared on unmount, and `runBuild` routes straight to `buildOnMainThread` while it's set. A watchdog would have had to pick a timeout that a genuinely slow Dagre run on a large report wouldn't trip, which is the same number the spinner delay already declines to guess at.

Covered by `routes builds issued after a crash straight to the main thread`, which asserts the second build produces a graph *and* that nothing further was posted to the dead worker. I checked it fails against the unpatched hook rather than passing vacuously.

The worker's own `ERROR` reply path is now exercised too, though for a different reason — it turned out to be the only way to get a superseded fallback in flight without killing the worker, which the neighbouring spinner test needed.

tests/useOpGraphLayoutWorker.spec.tsx
aidemsined: Fixed in 0c4b79c0 — `afterEach(cleanup)`, with the reason in a comment so nobody deletes it as redundant with the `useMlirLayoutWorker` spec that doesn't have it.

Your read on why this file differs is the useful part: the real-timer spinner is what turns a shared convention into a race, and that distinction is worth keeping visible at the call site.

src/definitions/PerfOverlayStatus.ts
aidemsined: Both fixed in 0c4b79c0.

`BufferSummaryPlotControls` now points at `definitions/PerfOverlayStatus`, which is where the map lives and, as you say, survives the dead component being deleted.

`GraphView` I retargeted at #1880 rather than dropping, matching what the props comment already does — the distinction it describes is real and the flag is still computed, it just has no reader until the overlay lands. Deleting the comment would leave `isPerfReportLoaded` looking like an oversight instead of a deliberate placeholder.

src/components/operation-graph/useOpGraphLayoutWorker.ts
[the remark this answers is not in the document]
aidemsined: Following up on my own reply above: I claimed a superseded build can never clear the timer. That holds for worker replies, but not in general — there was a second path I missed, and it produces exactly the symptom described here.

`buildOnMainThread` ran `endBuild()` from an unconditional `finally`. The two stale-request checks inside it are early `return`s, so a fallback that had been superseded still fell through the `finally` and cleared the timer the *active* build was relying on. Reachable whenever a build starts while an earlier fallback's dynamic import is still pending — via a worker `ERROR` reply or a dispatch throw, neither of which latches anything.

Fixed in 0c4b79c0 by guarding that call on `requestId === activeRequestIdRef.current`, so `endBuild` is now reached only by the request that owns the spinner. Pinned by `leaves the spinner armed when a superseded fallback finishes`, which fails against the unpatched hook.

The conclusion in my earlier reply stands — restarting the timer per `requestId` would still regress the burst behaviour, and that test still fails if you try it. The right fix was to make "a superseded request never touches the spinner" true everywhere, rather than to change what the timer measures.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: This one has since been recorded — #1809's Known gaps now carries it explicitly:

> Edge-click navigation (dropped — **deprecated**). The vis graph focused an edge's target on click, and walked back to the source when you clicked the edge incoming to the current node. Not carried to React Flow and not planned: prev/next, click-to-locate and the side panel's locate buttons cover the same need.

So it's a decision rather than an omission now, which was the substance of the ask. It went onto #1809 rather than #1884 because #1884 is for parity gaps we intend to close, and this one we don't — filing it there would have implied a commitment to implement it.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Agreed on the diagnosis, but I don't think the prescription follows, so I've left `FOCUS_ZOOM` alone for now.

`fitView` derives zoom from the target's bounds and clamps it — `@xyflow/system`'s `getViewportForBounds`:

```js
const xZoom = (width - p.x) / bounds.width;
const yZoom = (height - p.y) / bounds.height;
const zoom = Math.min(xZoom, yZoom);
const clampedZoom = clamp(zoom, minZoom, maxZoom);
```

For one op node (~200×40) in a ~1200×680 viewport at `padding: 0.3`, that's `min(4.2, 11.9)` → clamped to this flow's `maxZoom`, which is 3. So the user you describe — zoomed out surveying the graph — would land on 3 instead of 1. Same view lost, further to travel back. MLIR gets away with `localJump` because its `maxZoom` is 1.5 and its nodes are large enough that fitting one is a sane framing; ours aren't.

The deeper problem is that the complaint and the AC point in opposite directions. #1809 asks for "prev/next cycling with `fitView`", but `fitView` sets zoom by construction — it cannot preserve one. The only thing that actually fixes "a user zoomed out loses that view on every next match" is passing the current zoom through, which is precisely what vis did at `OperationGraphComponent.tsx:481-484`. That's a deviation from the written AC, not an implementation of it.

Three options, and I'd rather not pick unilaterally:

1. `setCenter(x, y, { zoom: getZoom() })` — vis parity, fixes the complaint, contradicts the AC's wording.
2. `fitView({ ...localJump, maxZoom: 1 })` — satisfies the AC literally and caps the damage at today's behaviour, but still discards the user's zoom.
3. Leave `FOCUS_ZOOM = 1` and amend the AC.

Worth noting the initial post-build recentre rides the same function, and that one *does* want a defined zoom rather than whatever the viewport happens to hold — so whichever way this goes, the first focus and subsequent jumps may need to stop sharing a code path.

The `MlirFitView` lift is a separate matter and still outstanding either way; I'd rather do it as a rename to something graph-neutral than import an `MLIR_`-prefixed constant into the op graph.

aidemsined:
Pushed `a684e8a2`. This closes the seven remaining threads from the Conductor review; the earlier seven landed in `0c4b79c0` / `d8cdce07`.

| Finding | Commit |
|---|---|
| filter dimming re-creates every node/edge object | `22adeafb` |
| worker builds are never coalesced or cancelled | `22adeafb` |
| no layout result cache | `22adeafb` |
| relayout-trigger discipline unguarded by any test | `22adeafb` |
| keyboard node selection is inert | `db65de8d` |
| delete `OperationGraphComponent` rather than editing it | `db65de8d` |
| `OpGraphFilter` is a verbatim copy of `MlirOpFilter` | `a684e8a2` |

**Dimming** now sets one class on the container and exempts matches, so node and edge objects keep identity through a filter change — `adoptUserNodes`' `userNode === internals.userNode` check and `EdgeWrapper`'s `Object.is` both hold, and the styled memos no longer spread every non-match. `OpGraphEdge` loses the inline `opacity` on its label `<text>`; the label now fades with its group instead of carrying its own value.

**Coalescing** went into the worker rather than the hook, since that is where the request queues up. `pendingBuild` holds only the newest request and a `setTimeout(..., 0)` hop drains it. The task hop is the whole mechanism and is deliberate over a microtask: message events are tasks, so yielding lets the rest of a burst arrive and overwrite `pendingBuild` before Dagre starts, where a microtask would run first and build the oldest request of the burst, then the next, then the next. Nothing can arrive mid-build — the worker is single-threaded and `buildOpGraph` never yields — so there is no post-build re-check.

**The cache** is an LRU of 4 keyed on `sourceVersion` + `hideDeallocate`, cleared on `SET_GRAPH`. It uses the `touchLruCache` lift, moved to `src/functions/touchLruCache.ts` and now shared with `mlirLayoutWorker`, so #1809's "Lift: `touchLruCache`" is met.

**Keyboard selection** was the one genuinely user-facing bug in the set. React Flow's own `onKeyDown` set `.selected` without ever calling `onNodeClick`, so Enter on a focused node drew the highlight and left the panel empty. `onNodesChange` is now intercepted: `select` changes drive `selectedOperationId`, and app selection mirrors back onto `node.selected`, so toolbar- and URL-driven selection agree as well. The same interception drops `remove` changes, which also answers Copilot's suppressed note about Backspace deleting nodes from this read-only graph.

**The filter** is now one `GraphOpFilter` with a shared `GraphFilterMode` and `buildGraphFilterMatcher`; `hiddenMatchCount` and `isDisabled` become optional props. `GraphFilterMode` keeps the `'substring'` / `'regex'` string values so persisted `sessionStorage` state migrates without a version bump. That is the filter-matcher slice of #1888 — the Dagre wrapper and fitView presets are still open there.

### Verification

`pnpm test` 1830 tests across 177 files green; `tsc`, `eslint`, `stylelint` and `lint:spdx` clean.

New `tests/OperationGraphRebuildDiscipline.spec.tsx` pins the discipline in both directions — filter typing, mode changes, selection and match-stepping must not relayout; hide-deallocate and a report change must — and asserts non-matching nodes keep identity across a drag. New `tests/opGraphLayoutWorker.spec.ts` covers coalescing and cache hits against the worker directly. Manually on a resnet50 memory report, a 6-toggle burst of hide-deallocate produced 6 `BUILD` messages and 2 `BUILT` replies, and keyboard Enter on a focused node now fills the panel.

### Body updates to follow

Two spots in the PR body are stale and I'll amend them: the Known-limitations bullet offering to delete the dead component (now done — note `vis-network` itself still stays until `DeviceOperationsGraphComponent` migrates), and the test-plan reference to `opGraphFilterMatcher`, now `graphFilterMatcher`.

One follow-up not taken: the worker is created per view mount, so its layout cache dies on remount. Sharing it across mounts needs module-scoped state outside the worker with explicit invalidation, which seems worth doing only if remount-relayout turns into a real complaint.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in `22adeafb`.

Dimming moved off the element objects entirely. The container carries a single `op-graph-filtering` class and the SCSS dims `.react-flow__node` / `.react-flow__edge` beneath it, with matches exempted by `op-graph-node-match` / `op-graph-edge-match`. So a filter change now rewrites only the `className` of the nodes that changed match state, and every other node and edge object keeps its reference — `adoptUserNodes`' `userNode === internals.userNode` and `EdgeWrapper`'s `Object.is` both hold.

`OpGraphEdge` also loses the inline `opacity` it was putting on the label `<text>`; the label now inherits the fade from its group, which is what makes the edge case work without touching the edge object.

Covered two ways: `tests/opGraphNodeStyles.spec.ts` pins the CSS contract (dim rule on the container, exemption for matches), and `tests/OperationGraphRebuildDiscipline.spec.tsx` asserts a non-matching node keeps object identity across a filter change plus a simulated drag.

src/components/operation-graph/opGraphLayoutWorker.ts
aidemsined: Fixed in `22adeafb`.

Coalescing went into the worker rather than the hook, since that is where the queue actually builds up. `pendingBuild` holds only the newest request, and the first message of a burst schedules a `setTimeout(drainPendingBuild, 0)`.

The task hop is the whole mechanism and is deliberate over a microtask: message events are tasks, so yielding lets every request already queued behind this one arrive and overwrite `pendingBuild` before Dagre starts. A microtask would drain first and build the oldest request of the burst, then the next, then the next — exactly the behaviour being reported.

Nothing can arrive mid-build — the worker is single-threaded and `buildOpGraph` never yields — so the request read in the drain is provably the one that gets answered, and there is no post-build re-check.

`SET_GRAPH` clears `pendingBuild` outright: a build queued against the previous source is moot, and the view reissues one for the new source as part of the same change.

Covered by `tests/opGraphLayoutWorker.spec.ts`. Measured in the browser on a resnet50 report, a 6-toggle burst of hide-deallocate produced 6 `BUILD` messages and 2 `BUILT` replies.

src/components/operation-graph/opGraphLayoutWorker.ts
aidemsined: Fixed in `22adeafb`.

The worker now keeps `layoutCache`, an LRU of 4 keyed on `sourceVersion` + `hideDeallocate`, checked in `drainPendingBuild` before Dagre runs and populated after. Flipping hide-deallocate off and back on is served from cache instead of re-laying out.

`SET_GRAPH` clears it, so a new report never serves a stale layout — the `sourceVersion` component of the key makes that belt-and-braces, but clearing also keeps the cache from holding the previous report's graphs alive.

This uses the `touchLruCache` lift #1809 asked for: it moved out of `mlirLayoutWorkerCache.ts` to `src/functions/touchLruCache.ts` and both workers now share it, so that item is met rather than left open. `mlirLayoutWorkerCache.ts` keeps only `CACHE_LIMIT_PER_GRAPH`.

Cache hits and source-change invalidation are covered in `tests/opGraphLayoutWorker.spec.ts`; the existing `tests/mlirLayoutWorkerCache.spec.ts` still covers the helper's eviction semantics from its new location.

One limit worth naming: the worker is constructed per view mount, so the cache dies on remount. Sharing it across mounts needs state outside the worker with its own invalidation, which seems worth doing only if remount-relayout becomes a real complaint.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in `22adeafb`.

Added `tests/OperationGraphRebuildDiscipline.spec.tsx`, which mocks `useOpGraphLayoutWorker` and counts `runBuild` calls, so the assertion sits directly on the trigger rather than on a downstream side effect.

It pins the contract in both directions. Must not relayout: typing in the filter, switching substring/regex, selecting a node, and stepping through matches. Must relayout: toggling hide-deallocate, and a report change. Mount itself is asserted at exactly one build.

So adding `filterQuery` or `selectedOperationId` to that dependency list — the regression you called out, which would otherwise compile, lint and pass everything else — now fails with a count mismatch.

The same file also carries the dimming invariant from the other thread: a non-matching node keeps object identity across a filter change plus a simulated drag.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in `db65de8d`. This was the one finding in the set with a user-visible symptom, and it reproduced exactly as described — tab to a node, press Enter, React Flow paints its own `.selected` while the panel stays empty.

Rather than add a parallel key handler, `onNodesChange` is now intercepted. `select` changes are resolved to the node's operation id and drive `setSelectedOperationId`, so React Flow's internal keyboard path lands in the same place a click does; the remaining changes pass through to `useNodesState`. Selection also mirrors back onto `node.selected`, so toolbar- and URL-driven selection agree with React Flow's own styling instead of leaving two nodes looking selected.

Unselect clears app state, so Escape and a pane click behave consistently.

The same interception drops `remove` changes, which also answers Copilot's suppressed note about Backspace deleting nodes out of this read-only graph.

Covered in `tests/OperationGraphRebuildDiscipline.spec.tsx`: adopting a React Flow `select` event, mirroring app selection onto the flag, and clearing on unselect — plus the assertion that none of it triggers a relayout.

src/components/OperationGraphComponent.tsx
aidemsined: Done in `db65de8d` — `OperationGraphComponent.tsx` and `OperationGraphComponent.scss` are deleted rather than edited.

That removes the second copy of the connected-op grouping you flagged, so `getConnectedOpGroups` in `OpGraphInfoPanel.tsx` is now the only implementation.

To be precise about what this does not do: `vis-network` itself stays as a dependency until `DeviceOperationsGraphComponent.tsx` migrates, which is phase 5's tail on #1809. So this drops the unreachable lines and the dead stylesheet, not the library or the `vis` chunk in `vite.config.ts`. I've updated the PR body's Known-limitations bullet accordingly, since it previously left this as an open offer.

Two referencing comments were repointed in the same commit: the `CONVENTIONS.md` examples now cite `src/definitions/NodeRelation.ts` and `DeviceOperationsGraphComponent.tsx`, and the `CircularBufferPressureModal.scss` note no longer points at the deleted file. The `BufferSummaryPlotControls.tsx:33` pointer you raised separately is handled on that thread.

src/components/operation-graph/OpGraphFilter.tsx
aidemsined: Fixed in `a684e8a2` — the two are one component now rather than a copy each.

- `src/components/GraphOpFilter.tsx` replaces both `MlirOpFilter` and `OpGraphFilter`. `hiddenMatchCount` and `isDisabled` were the only real deltas, and they become optional props, so neither view carries the other's branch.
- `src/definitions/GraphFilterMode.ts` replaces `MlirFilterMode` and `OpGraphFilterMode`. It keeps the `'substring'` / `'regex'` string values so persisted `sessionStorage` state migrates without a version bump and both views share one storage vocabulary.
- `src/functions/graphFilterMatcher.ts` replaces `buildFilterMatcher` and `buildOpGraphFilterMatcher`, including the `isRegexInvalid` distinction between "nothing typed" and "bad pattern".
- `GraphOpFilter.scss` replaces both stylesheets, class prefix `graph-op-filter`.

Tests consolidated the same way: `tests/graphFilterMatcher.spec.ts` absorbs the matcher cases from both old specs, `tests/MlirOpFilter.spec.tsx` becomes `tests/GraphOpFilter.spec.tsx`, and `tests/mlirFilter.spec.ts` keeps only `resolveFilterMatches`.

Net 240 insertions against 506 deletions. This is the filter-matcher slice of #1888 — the Dagre wrapper and fitView presets are still open there.

Issue #1812, opened 2026-07-29
Report link status depends on transient performance-tab view filters


dcblundell:
Split out of #1810, which shares the same code path.

## Problem

Memory<->performance link status is decided by `useGetDeviceOperationListPerf` in `src/hooks/useAPI.tsx`, which matches the memory report's device-operation sequence against the performance report's rows strictly by index. The perf side comes from `usePerformanceReport`, whose query key includes the perf-tab preference atoms: `hideHostOpsAtom`, `mergeDevicesAtom`, `tracingModeAtom` and the signpost range.

So a user changing how they are *viewing* the performance report changes whether the two reports are considered to be from the same run. The clearest case is **Merge devices**: turning it off on a 2-device report returns roughly twice as many rows, interleaved per device, against an unchanged memory-side list, and the positional match fails. The link badge flips to "Failed to link", and with it the perf-table Op column, L1-pressure column, tensor-drawer gating, graph perf overlay, top-N annotations and memory/perf range synchronisation all drop out.

Signpost range selection and **Hide host ops** narrow or widen the row set the same way; the match tolerates trailing extra rows but not leading or interleaved ones.

## Why it matters

"Do these two reports come from the same run?" is a property of the reports, not of the current view. Users reasonably read the link badge as a statement about their data, and a view toggle silently invalidating it looks like a bug in the reports.

## Possible directions

- Resolve link status from an unfiltered performance-report fetch (merged devices, host ops included, no signpost range) held separately from the rows the perf tab renders, so the view filters cannot move it.
- Or make the match tolerant of the filtered shapes rather than strictly positional.
- #1800 (a shared run id across memory and performance reports) removes the need for sequence matching altogether and would make this moot.

## Related

- #1810 - the near-term correctness fix for the multi-device collapse heuristic (PR #1811). That fix is orthogonal: it makes the memory-side list correct, but the perf-side coupling described here remains.
- #1800 - tie memory and performance reports together with an explicit ID.

dcblundell:
Shipped in [v0.99.0](https://github.com/tenstorrent/ttnn-visualizer/releases/tag/v0.99.0) via #1883 (`useLinkedPerformanceReport`). Link status no longer tracks the performance tab’s view filters.

#1883 left a documented `tracingMode` carve-out; closing that properly still waits on #1800.

Issue #1800, opened 2026-07-24
Tie Memory and Performance reports together


dcblundell:
We should add an ID which links memory and performance reports together, so we can avoid the guesswork we currently employ which is hit and miss.

aidemsined:
The shared run id this issue wants may already exist on the wire, and a tt-metal branch is about to put it in the graph report too.

`ops_perf_results.GLOBAL CALL COUNT` encodes both halves of the join. Verified against `test_ttnn_moe_aug26_2217` (32-chip Blackhole, 808 rows): `gcc = (runtime_id << 10) | device_id` holds on **all 800** device-op rows, with 25 distinct runtime ids each appearing on exactly 32 devices. That matches `EncodePerDeviceProgramID` in tt-metal, so the low 10 bits are the device id and the rest is the per-op runtime id.

The graph-report side is what has been missing, and `smountenay-tt/subdevice_reports` adds it: an `operation_executions` table carrying `global_call_count` per `(operation_id, device_id)` alongside `runtime_id`, `program_id` and `command_queue_id`. Tracked as #1936. That would give an exact per-(operation, device) join rather than the positional alignment `matchDeviceOperationsToPerf` does today.

Worth noting what that would retire: the TODO in `deviceOperationMatching.ts` points here, and both heuristics layered on it become unnecessary — the per-device collapse fallback from #1810 and the signpost filter from #1943 (PR #1945). Neither is load-bearing once the two reports share a key.

Two caveats on the branch as written, since they decide whether this actually helps:

- Tracking is gated on `active_manager_id != default_manager_id`, so a run without a non-default subdevice manager emits no `program_execution` nodes and the table stays empty — no key on exactly the ordinary reports this issue is about. Needs widening to be useful here.
- It is hooked in `enqueue_mesh_workload` only, so the single-device path is not covered.

Raised both on #1936.

PR #1883, opened 2026-08-14
Decouple report linking from perf viewing

dcblundell:
Refs https://github.com/tenstorrent/ttnn-visualizer/issues/1812 — deliberately not `Fixes`. The `tracingMode` carve-out described below is a documented residual of that issue, so merging this should not close it.

Whether two reports come from the same run is a property of the reports, not of how the user happens to be viewing them. Today it is decided by a query whose key carries the performance tab's view filters, so toggling **Merge devices** — or hiding host ops, or picking a signpost range — flips the link badge to "Failed to link" and takes the perf-table Op column, the L1-pressure column, tensor-drawer gating, the graph perf overlay, top-N annotations and memory/perf range synchronisation with it.

This PR resolves link status from its own report fetch, held still against the view filters, and centralises how both fetches are keyed so they cannot drift apart.

**Query key centralisation**

- New `performanceReportQueryKey.ts` defines `PerformanceReportParams` and builds the keys for both the single and comparison report queries from one shared set of filter segments, so the two consumers cannot key on different things.
- `usePerformanceReport`, `useLinkedPerformanceReport` and `usePerformanceComparisonReport` all build their keys through it.
- `fetchPerformanceReport` takes the params as one object rather than seven positional arguments. `hideHostOps`, `mergeDevices` and `tracingMode` are adjacent booleans, so a positional signature let two of them be swapped with no type error while the key was still built from the correct object — a request cached under a key that misdescribes it.

**Filtered vs. link-pinned reports**

- `useLinkedPerformanceReport` fetches the report in the one shape the memory report can be matched against: devices merged, host ops hidden, whole run. `useProxyPerformanceReport`, `useReportLinkMatch`, `useTopNAnnotationAvailability` and `GraphView` all read it instead of the filtered view.
- `groupBy` is pinned too. It only shapes `stacked_report` and cannot affect `report` at all, but both come back from one endpoint so it is part of the query key — following the tab's grouping control would blank the link report on every switch, which is the failure this exists to prevent.
- `tracingMode` is deliberately **not** pinned, and so is a knowing carve-out from #1812, which names it alongside the three filters that are. It only reorders rows, and on a trace-captured run the traced order is the one that lines up with the memory report; pinning it would leave those reports permanently unlinkable, a worse failure than the one it would fix. Resolving against both orders and keeping whichever aligns would close the gap properly, but that is worth the second fetch only once #1800 lands.

Note that #1812's "Possible directions" bullet suggests holding the report with *host ops included*. The code hides them instead: the match is positional from index 0 and tolerates only trailing extras, while the memory-side list is device operations only, so any leading or interleaved host-op row breaks it. Hiding host ops is also the tab's own default, which is what lets the two queries share a cache entry. The issue text should be amended rather than the code.

**Range slider**

Op↔perf ids are now resolved against the link-pinned report while the performance slider still spans the rows the tab is displaying, so a mapped id can land outside the slider's own track. `clampSelectionToRange` keeps the handles inside it — but widens a selection that misses the track entirely to the whole track instead of clamping, since clamping would pin both handles to the same edge and `Performance.tsx` would read that as a legitimate one-row selection rather than falling back to the full report.

**Performance**

The op↔perf match is memoised across call sites rather than per hook invocation. `useMemo` caches per invocation, so the O(rows) match and the device-operation flatMap ran once for every consumer — and `OperationListPerfData` renders one instance per virtualised row, each previously filtering the whole match. Consumers now share one result, and the per-row lookup goes through a map keyed on op id.

**Follow-up**

Once the tab moves off its defaults the two query keys diverge and a second `perf-results/report` request goes out. Report generation is uncached server-side and CPU-bound on a single gevent worker, so that is worth removing at the backend rather than by unpinning these filters — tracked in #1886.

**Testing**

- `performanceReportQueryKey.spec.ts` — link-resolution and perf-tab queries share a cache entry at the tab's defaults and diverge when filters differ; signpost identity; comparison key shape and separation.
- `perfRangeSelection.spec.ts` — clamp when the selection overlaps the track, widen when it misses.
- `useLinkedPerformanceReport.spec.tsx` — the link fetch holds still across view-filter changes, and the match survives them.


src/functions/performanceReportQueryKey.ts
```diff
@@ -0,0 +1,68 @@
...
+export const LINKED_PERFORMANCE_REPORT_FILTERS = {
+    startSignpost: null,
+    endSignpost: null,
+    hideHostOps: true,
+    mergeDevices: true,
+    groupBy: StackedGroupBy.OP,
+} as const;
+
+const getSignpostKey = (label: string, signpost: Signpost | null) =>
+    `${label}:${signpost ? `${signpost.id}${signpost.op_code}` : null}`;
```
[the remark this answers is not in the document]
dcblundell: Good catch — fixed in 66fdd61b by delimiting the two fields. Worth noting the practical risk was narrow: signposts are deduplicated by `op_code` server-side, so `op_code` alone already identifies a signpost within a report and `id` is redundant defence. A collision would have needed a digit-leading signpost label plus exact arithmetic alignment. Still, the delimiter costs nothing and the key is client-only React Query state, so no migration concern.

src/functions/performanceReportQueryKey.ts
```diff
@@ -0,0 +1,136 @@
...
+interface PerformanceReportQueryKeyInput {
+    instanceId: string;
+    params: PerformanceReportParams;
+}
+
+export const getPerformanceReportQueryKey = ({
+    name,
+    instanceId,
+    params,
+}: PerformanceReportQueryKeyInput & { name: string | null }) => [
```
aidemsined, line 132: > [!CAUTION]
> `instanceId` doesn't fix the collision the commit message describes, and that collision is reachable today.

`4a0289d9` justifies adding `instanceId` with "a local copy and a remote-synced copy of the same timestamped run … collided on a single entry". But within one browser session `getOrCreateInstanceId()` returns one value for *both* selections, so this key is still identical for the local and the remote report — the name segment is a bare basename (via `getReportId`'s `pathBasename`), so two reports of the same timestamped run genuinely coincide on every segment.

The reachable sequence, which I traced:

1. Select the remote report — `RemoteSyncConfigurator.tsx:392` calls `clearReportCaches`, the report is fetched and stored under key K.
2. Select the local report of the same basename — `handleSelectPerformance` (`LocalFolderSelector.tsx:298-309`) calls only `updateInstance` and `setActivePerformanceReport`. It clears **nothing**.
3. Key K is still populated at `staleTime: Infinity`, so the local selection renders the remote report's rows.

`handleSelectProfiler` and the two `clearActive` callbacks have the same gap — the only handlers calling `clearReportCaches` are the ones at `:207` and `:245`.

**Suggested action:** add `performanceReportLocationAtom` (or the active report's path) to the key — `AGENTS.md:164` already mandates location as part of report identity — or route the select/clear paths through `clearReportCaches`. Keeping `instanceId` for the NPE-precedent defence is fine either way; the commit rationale just needs correcting so the next reader doesn't think this case is covered.

src/hooks/useAPI.tsx
```diff
@@ -852,61 +909,38 @@ export const useGetDeviceOperationsListByOp = () => {
...
+ * @description One entry per matched device operation, pairing its op id with
+ * the perf row it matched. An operation that matched no perf row still gets an
+ * entry, with `perfId` left `undefined` — consumers filter or coerce it
+ * themselves (`RangeSlider` leans on `Number(undefined)` being falsy).
+ *
+ * The perf ids come from the link-pinned report (`useLinkedPerformanceReport`),
+ * not from the rows the performance tab is displaying. A row id survives host-op
+ * and signpost filtering, so consumers can still join against those views; ids
+ * do not survive `mergeDevices: false`, where an operation's per-device rows
+ * each carry their own id and only the merged representative joins.
```
aidemsined, line 942: > [!CAUTION]
> Worth verifying against a real 2-device report before merge — the subset assumption here may not hold, and if it doesn't the Op column points at the wrong operation.

`enrichRowData` (`enrichPerfRowData.ts:36-45`) builds `opIdByPerfId` from `useOpToPerfIdFiltered()` — now the **merged** report's ids — and looks up displayed rows by `row.id`, which under `mergeDevices: false` come from the **unmerged** report. If merged ids are a strict subset of unmerged ids, the result is merely incomplete (only the merged representative joins), which is what the comment here assumes.

The PR's own fixture suggests they're not a subset but a *different id space*: in `useLinkedPerformanceReport.spec.tsx`, merged ids `{2,3}` map Matmul→2 and Softmax→3, while unmerged id `3` is Matmul-on-device-1. If tt-perf-report assigns ids that way in reality, an unmerged row joins to a merged entry describing a different operation — and the Op column, the L1-pressure column and the tensor-drawer target all silently point at the wrong op.

What makes this worth blocking on rather than noting: before this PR, `opIdsMap` was `[]` under a divergent filter and the column was *hidden*. Now the column is present and confident. A wrong value is worse than an absent one here.

`csv_queries.py:658-660` says ids are source-CSV row numbers, which reads as consistent with the fixture rather than with the subset assumption. I couldn't settle it — tt-perf-report isn't vendored, so I can't check its assignment under `merge_devices=false`.

**Suggested action:** confirm against a real 2-device report. If the id spaces can disagree, drop `op` for rows whose id isn't in the merged map, or suppress the Op/L1/drawer affordances while `mergeDevices` is false. Either way a spec covering the unmerged join would pin it — there isn't one today.

src/functions/performanceReportQueryKey.ts
```diff
@@ -0,0 +1,136 @@
...
+ * filter, plus the one view control it still follows.
+ *
+ * Exported so the hook and its tests compose the pinned set the same way. A test
+ * that spread `LINKED_PERFORMANCE_REPORT_FILTERS` itself would assert against its
+ * own composition, and would keep passing if the hook started pinning
+ * `tracingMode` too. The `PerformanceReportParams` return type also makes a new
+ * filter added to that interface a compile error until it is either pinned above
+ * or forwarded here deliberately.
+ */
+export const getLinkedPerformanceReportParams = (tracingMode: boolean): PerformanceReportParams => ({
```
aidemsined, line 59: > [!WARNING]
> `Fixes #1812` will auto-close an issue that this code says stays open.

`tracingMode` is forwarded here rather than pinned, so the link query still varies with it — and #1812 names tracing mode explicitly among the filters that must not move link status. The code agrees: `useAPI.tsx:1264` says "#1812 stays open on that residual until the both-orders resolution lands", and both new specs carry comments describing the gap as unresolved.

The persistence guard in `ReportLinkStatus.tsx:62` suppresses only the localStorage write, so a verdict reached with tracing on doesn't outlive the toggle. The badge itself and all six downstream features still flip when tracing mode changes.

That's a defensible scope decision — the carve-out is documented, tested and deliberate. The problem is purely that `Fixes` closes the issue on merge, so the tracked residual disappears with it.

**Suggested action:** change to `Refs #1812`, or split the tracing-mode residual into a follow-up issue and `Fixes` that one.

src/hooks/useAPI.tsx
```diff
@@ -1160,36 +1194,36 @@ export const usePerfMetas = (reportNames: string[] | null | undefined): (MetaDat
...
-            name,
-            `startSignpost:${startSignpost ? `${startSignpost.id}${startSignpost.op_code}` : null}`,
-            `endSignpost:${endSignpost ? `${endSignpost.id}${endSignpost.op_code}` : null}`,
-            `hideHostOps:${hideHostOps ? 'true' : 'false'}`,
-            `mergeDevices:${mergeDevices ? 'true' : 'false'}`,
-            `tracingMode:${tracingMode ? 'true' : 'false'}`,
-            `groupBy:${groupBy}`,
-        ],
+        queryFn: () => (name !== null ? fetchPerformanceReport(name, params) : Promise.resolve(EMPTY_PERF_RETURN)),
+        queryKey: getPerformanceReportQueryKey({ name, instanceId: getOrCreateInstanceId(), params }),
```
> This runs in the render body of a hook that's reached once per virtualised row, and it touches `sessionStorage` synchronously each time.

`getOrCreateInstanceId` (`src/libs/axiosInstance.ts:18-38`) constructs a `URLSearchParams` from `window.location.search` and does a synchronous `sessionStorage.getItem` on every call. `OperationListPerfData` renders once per row of the virtualiser (`OperationList.tsx:538`) and reaches here through `useGetDeviceOperationListPerfByOpId → useGetDeviceOperationListPerf → useLinkedPerformanceReport`.

With a ~40-row window that's ~40 sessionStorage reads plus 40 `URLSearchParams` allocations per render pass, and the window re-renders on every scroll tick. New in `4a0289d9` — before that, this hook didn't read the instance id.

Slightly ironic given the same commit's `memoiseLatest` work is specifically about removing per-row cost from this path.

**Suggested action:** cache the id at module scope, or wrap it so the storage read happens once per session. `:500`, `:660`, `:684`, `:710` and `:1293` call it the same way, though those aren't on the per-row path.

AGENTS.md
```diff
@@ -162,6 +162,8 @@ Applies on touch to views that draw data-proportional visuals (NPE chip cluster
...
+- **Anything deciding whether two reports describe the same run reads `useLinkedPerformanceReport()`, never `usePerformanceReport()`.** The latter keys on the performance tab's view filters, so a view toggle would move the answer; the former pins them. Build both keys through `performanceReportQueryKey.ts` rather than assembling a key inline. See #1812.
+- **Derived values shared by many hooks go through `memoiseLatest`, not `useMemo`.** `useMemo` caches per invocation, so an O(rows) derivation read by one component per virtualised row runs once per row. Only use it when every argument comes from a source all consumers share (a query result, not a per-caller `useMemo`) — a cache of one degrades to no cache when two live callers pass different arguments. Results are shared, so treat them as immutable, and pair any `queryClient.clear()` with the matching cache reset (`clearReportCaches`).
```
aidemsined: > [!WARNING]
> This states the rule more broadly than `CONVENTIONS.md` does, and the broad form is already contradicted by the file it points at.

As written — "Derived values shared by many hooks go through `memoiseLatest`, not `useMemo`" — it reads as an unconditional repo-wide preference. `CONVENTIONS.md:523` is correctly narrower: shared by several hooks **and** by one component instance per virtualised row. The row fanout is what makes the trade pay.

Two hooks in `useAPI.tsx` already violate the broad form: `usePerformanceRange` (`:947-959`, two live consumers — `RangeSlider.tsx:63` and `Performance.tsx:71`) and `useL1PressureByOperation` (`:1132-1157`, two live consumers — `Performance.tsx:73` and `useTopNAnnotations.tsx:65`). Converting either would hit the two-live-callers degradation `memoiseLatest`'s own doc comment warns about, where alternating callers thrash a cache of one.

So a reader following the AGENTS.md line literally would make the codebase worse, and the dual record is meant to keep the summary and the detail in step.

**Suggested action:** lead with the row-fanout condition rather than "shared by many hooks", so the one-liner and the expanded rule agree.

src/components/RangeSlider.tsx
```diff
@@ -111,14 +113,14 @@ function Range() {
...
             const updatedMin =
                 Number(matchMin) ||
-                (selectedOperationRange[0] < opIdsMap[0].opId ? perfMin! : selectedPerformanceRange[0]);
+                (selectedOperationRange[0] < opIdsMap[0].opId ? rangeMin : selectedPerformanceRange[0]);
             const updatedMax =
                 Number(matchMax) ||
                 (selectedOperationRange[1] > opIdsMap[opIdsMap.length - 1].opId
-                    ? perfMax!
+                    ? rangeMax
                     : selectedPerformanceRange[1]);
```
aidemsined, line 122: > [!WARNING]
> This effect got the clamp; its perf→op sibling below didn't, and it has the same id-space problem.

The effect at `:127-158` maps a selection in the *filtered* perf id space back to op ids through `opIdsMap`, which is now the *pinned* id space. Under `mergeDevices: false` the nearest-perfId `reduce` at `:136` and `:144` settles on whichever merged representative is numerically closest, which can be an unrelated operation — the same two-id-space issue as the enrich join, reached from the other direction.

Before this PR the path was effectively inert, because the match returned `[]` under a divergent filter and there was nothing to map. Pinning the match brings it to life.

Also worth a look while you're in there: `opMin!` / `opMax!` at `:153` and `:157` are still non-null asserted, and the guard at `:128` checks `perfRange` but never `operationRange` — the same latent unwrap you just fixed one block above.

**Suggested action:** clamp the result the way this effect does, or add a comment explaining why the reverse direction doesn't need it.

tests/Performance.spec.tsx
```diff
@@ -264,6 +264,33 @@ describe('Performance route', () => {
...
+    // resolve against the link-pinned report while this table spans the rows the tab
+    // is showing, so a mapped selection can land entirely outside the report — here
+    // it must fall back to the full span rather than render an empty table.
+    it('falls back to the whole report when the selected range misses it entirely', () => {
+        (usePerformanceReport as Mock).mockReturnValue({
+            data: { report: [DRAM_PERF_ROW], stacked_report: [], signposts: [] },
+            isLoading: false,
+            error: null,
+        });
+        (usePerformanceRange as Mock).mockReturnValue([1, 1]);
```
aidemsined: > [!WARNING]
> A zero-width range makes this test unable to distinguish the branch it claims to pin.

With `usePerformanceRange` mocked to `[1, 1]`, `clampSelectionToRange(40, 50, 1, 1)` returns `[1,1]` on the **widen** branch and `[1,1]` on the **clamp** branch — and the pre-PR `selectionMissesReport ? perfRange : selectedRange` also returns `[1,1]`. All three produce the single row the assertion expects, so the comment above about "the other half of `clampSelectionToRange`'s widen contract" isn't something this test can observe.

**Suggested action:** give the range a width so the two branches diverge — e.g. two fixture rows with ids `1` and `9`, `usePerformanceRange` returning `[1, 9]`, selection still `[40, 50]`, asserting 2 rows. Clamping would then yield `[9,9]` and one row, and the test would earn its comment.

Worth noting the sibling specs in this PR mostly get this right — `perfRangeSelection.spec.ts` pins the clamp and widen branches separately with ranges where they actually differ, which is why this one stands out.

aidemsined (commented):
The core decoupling is right, and the key module is the best-built thing in the diff. Typing `getFilterKeySegments` as `Record<keyof PerformanceReportParams, string>` turns "the two consumers cannot key on different things" from a claim into a compile error — a filter added to the interface can't reach the request without also reaching the key. Same trick on `getLinkedPerformanceReportParams`'s return type for the pinned set. Three of the four filters #1812 names are genuinely pinned to literals rather than snapshotted at first render, which is the distinction that matters.

I checked every current consumer of `usePerformanceReport` against the new AGENTS.md rule — `RangeSlider.tsx:56`, `Performance.tsx:68`, `useAPI.tsx:947` — and none of them decides whether two reports describe the same run. **Zero existing violations**, so the rule lands clean.

Two things I'd want settled before merge, both flagged inline, plus a scope-of-closure point about the `Fixes` keyword.

**A note on breadth:** at 25 files and ~1900 diff lines this is at the edge of what one pass covers with confidence. I read the three new modules, the full after-state of the changed regions of `useAPI.tsx`, and every consumer of the changed hooks end to end. Two files got less than they deserve: `PerfTable.tsx` (I traced `opIdsMap` and the tensor-drawer gate but didn't audit the sort/column machinery around them) and `tests/Performance.spec.tsx` beyond the one added case. I also **could not** verify tt-perf-report's row-id assignment under `merge_devices=false` — it's an external dependency, not vendored — which is exactly what the second inline finding turns on.

## General observations

> [!NOTE]

- `isLinkResolutionCanonical` is a persistence-policy predicate living in a module named for query keys, and `ReportLinkStatus.tsx:13` now imports from that module for a reason unrelated to keys. The module is really "the pinned-filter policy for link resolution" — either rename it or move the predicate next to `reportLinks.ts`.
- `EMPTY_PERF_RETURN` (`useAPI.tsx:97`) is a shared mutable object handed out both as resolved query data and as a `memoiseLatest` argument, under a convention that says results are shared and must be treated as immutable. `Object.freeze` or `as const` costs nothing.
- The tracing-mode carve-out is now explained in five places — `performanceReportQueryKey.ts:64-72`, `useAPI.tsx:1252-1264`, `ReportLinkStatus.tsx:57-61`, `CONVENTIONS.md:517` — plus two test headers. Keep one canonical copy and point the rest at it; five will drift when the both-orders resolution lands.
- `useOpToPerfIdFiltered` no longer filters anything — every matched operation gets an entry, `perfId` possibly undefined. The new doc comment says so explicitly, which is the right call given the rename cost, but the name is now actively misleading to anyone reading a call site.
- `useGetDeviceOperationListPerfByOpId` calls a hook inside an argument expression (`useAPI.tsx:928-929`). Legal and unconditional so no lint complaint, but it reads as a plain function call and hides that it subscribes to three queries.
- No test covers `clearReportCaches`, which is the only thing standing between a report switch and a stale module-level match, with four call sites depending on it. Populate a cache, call it, assert recompute — four lines.
- `memoiseLatest.spec.ts`'s "recomputes after a reset, and stops referencing what it held" only asserts the recompute count; the release half is untested and genuinely hard to test without a `FinalizationRegistry`. Worth trimming the name to what it checks.
- `performanceReportQueryKey.ts:32-37` reaches into capacity planning ("CPU-bound on a single worker", "#1886") in a module header. Good context, wrong home — that belongs in the issue.
- On the second query's cost: at tab defaults every pinned value equals its atom default, so both hooks produce an identical key and React Query serves one entry from one request. The divergence only starts when the user touches a filter — and the stacked-grouping control is the cheapest way to trip it while being the one a user has least reason to think is expensive. `RangeSlider` is mounted app-wide, so once diverged both queries stay live at `staleTime: Infinity` on every route. The module comment already says this and points at #1886; I agree with the trade.

Two things worth calling out as done well. `f7adeedf` fixed a real bug in this PR's own new code — `memoiseLatest` was writing `cache.args` *before* calling `compute`, so a throw paired the new arguments with the previous result and poisoned that argument set permanently; the spec covering it would fail against the pre-fix version. And `4a0289d9` caught that `useReportLinkMatch.spec.tsx` mocked `useAPI` wholesale, so renaming the mocked hook kept it green while checking nothing — which means one of this PR's headline behaviours was untested from `fae7a871` through `f7adeedf`. Both are the kind of finding that only comes from re-reading your own work adversarially.

Nothing was run from here: `npm run lint`, `npx tsc --noEmit` and the eight affected specs are all unverified. The `Record<keyof …>` guard and the `PerformanceReportParams` return type are compile-time claims that only `tsc` can actually confirm, so that one's worth seeing green specifically.

---

src/functions/performanceReportQueryKey.ts
dcblundell: Agreed, and confirmed reachable — fixed in 20c088ab.

You are right that `instanceId` cannot separate these: it comes from `sessionStorage`, so both selections in one tab carry the same value, and `name` is a bare basename via `getReportId` → `pathBasename`. And `handleSelectPerformance` really does clear nothing — the only `clearReportCaches` callers are `LocalFolderSelector:207`/`:245` and `RemoteSyncConfigurator:382`/`:392`.

I went with location rather than path, because location is what the server actually discriminates on: `_apply_requested_performance_name` (`views.py:1186`) joins `?name=` onto `Path(instance.performance_path).parent`, and that parent is the uploads folder for a local report and the synced folder for a remote one. `path` would also work within a session but is unstable across one — per `Reports.ts:17-20` it is the remote path while a report is freshly selected and the synced folder name after a reload, so it would fragment the cache without buying anything extra.

`performanceReportLocationAtom` is now a key segment on both the single-report and comparison keys, with a case in `performanceReportQueryKey.spec.ts` pinning each. The `instanceId` doc comment now says what it does and does not cover, so the next reader does not repeat the commit message's mistake.

One small correction for the record: AGENTS.md:164 mandates the report's *path* as part of report identity, not its location. Same conclusion either way.

src/hooks/useAPI.tsx
dcblundell: Good question to raise, but I can settle it: the id spaces are not divergent, and the subset assumption holds.

`tt-perf-report` is not vendored, but it is a pinned dependency (`pyproject.toml:24` → `tt-perf-report==1.2.8`) and installs into `.venv`, so the source is readable. In `perf_report.py`:

- `:2164` — `df["ORIGINAL_ROW"] = df.index + 2`, stamped from the source CSV's row index **before** any merging.
- `:2183` — `df = merge_device_rows(df)`, which picks the max-duration constituent row (or the first block's, for collectives) and keeps that row's dict.
- `:2192` — `op_data["ID"] = Cell(row["ORIGINAL_ROW"])`, in the loop common to both the merged and unmerged paths.

Merging never renumbers, so a merged row's id is always one of its constituent unmerged rows' ids. Merged ids are therefore a strict subset of unmerged ids, and the enrich join degrades exactly as `useOpToPerfIdFiltered`'s doc comment describes: the representative row joins, the other devices' rows get `op === undefined`. Incomplete, never wrong — so the Op column, L1 column and drawer target stay correct under `mergeDevices: false`.

The fixture is what made this look worse than it is, and you were right to flag it. It numbered merged rows `2, 3` against unmerged `2, 3, 4, 5`, which the real tool cannot produce and which does create the collision you traced (unmerged `3` = Matmul on device 1 vs merged `3` = Softmax). 20c088ab derives `mergedRows` from `unmergedRows` instead of renumbering, with a comment recording why, so the fixture now models the real id relationship.

Not resolving the underlying ask entirely: a spec covering the unmerged join genuinely does not exist. I would rather raise that separately than grow this PR further.

src/functions/performanceReportQueryKey.ts
dcblundell: Agreed — the residual is real and deliberate, so it should not be auto-closed. Changed the PR body to `Refs`, with a line saying why it is not `Fixes`.

src/hooks/useAPI.tsx
dcblundell: Agreed — fixed in 20c088ab. Confirmed the path: `OperationList:538` renders `OperationListPerfData` per virtual row → `useGetDeviceOperationListPerfByOpId` → `useGetDeviceOperationListPerf` → `useLinkedPerformanceReport` → `usePerformanceReportQuery`, whose `queryKey` is rebuilt in the render body.

Cached at module scope rather than wrapped per call site, so `:500`, `:660`, `:684`, `:710` and `:1293` benefit too. Safe to cache: nothing in the app rewrites `?instanceId=` mid-session, and `SocketProvider:17` already binds the id at module load, so it was effectively fixed per page load already.

AGENTS.md
dcblundell: Agreed on both counts — the one-liner was broader than CONVENTIONS.md:523, and the broad form is contradicted by `usePerformanceRange` (2 live consumers) and `useL1PressureByOperation` (2 live consumers), which correctly stay on `useMemo`.

20c088ab leads with the row-fanout condition and names those two as the counter-example, so the summary and the expanded rule now agree.

src/components/RangeSlider.tsx
dcblundell: Split verdict — the guard is fixed in 20c088ab, the clamp is deliberate and now documented rather than added.

**On the clamp.** I do not think the reverse effect needs one. The forward effect maps *into* the perf track, whose bounds come from `usePerformanceRange` (the **view** report) while the ids come from the link-pinned one, so a mapped id can genuinely land outside a signpost-narrowed view — hence the clamp. The reverse effect maps into the operation track, and both sides come from the same source: every `opId` in `opIdsMap` is an id from the operations list, and `useOperationListRange` derives the track's ends from that same list, with the fallbacks being those ends. So the result is inside the track by construction. I have added a comment saying exactly that, since the asymmetry is not self-evident.

The id-space half of this also relaxes given the finding on the enrich thread: merged ids are a strict subset of unmerged ids (`perf_report.py` stamps `ORIGINAL_ROW` before `merge_device_rows` and merging keeps a constituent row), so the nearest-perfId `reduce` settles on a real representative rather than a renumbered stranger. It can still pick a neighbouring operation when the exact id is filtered out, but that is inherent to a nearest-neighbour fallback and predates this PR — and an approximate range endpoint is a much softer failure than a wrong Op column.

**On the guard — you are right, and it was worse than the two lines you spotted.** `selectedOperationRange` is a global atom that outlives a report switch, so `operationRange` can be null while a selection is still set. 20c088ab adds `operationRange` to the `:128` guard and destructures the ends from it, dropping both assertions. The same latent unwrap existed in the two `InputGroup` handlers further down (`opMin!` / `opMax!` inside a block guarded only on `selectedOperationRange`); those now fall back to the current handle instead.

tests/Performance.spec.tsx
dcblundell: Agreed — fixed in 20c088ab. Confirmed: with the range mocked `[1, 1]`, `clampSelectionToRange(40, 50, 1, 1)` returns `[1, 1]` on both branches, and the pre-PR expression returned `[1, 1]` too, so the assertion could not distinguish any of the three.

Taken your suggested shape almost exactly: two rows with ids `1` and `9`, `usePerformanceRange` returning `[1, 9]`, selection still `[40, 50]`, asserting 2 rows. Widening keeps both; clamping to the far edge would keep one. Added a comment recording why the range must have width, so it does not quietly collapse back to a single row later.

PR #1889, opened 2026-08-17
Convert navigation to collapsible side menu

dcblundell:
Replaces the horizontal top navbar with a vertical, collapsible side rail.

Screenshot 2026-08-21 at 09 24 32

Screenshot 2026-08-21 at 09 24 19

## Navigation

- `MainNavigation` is replaced by `SideNavigation`: a vertical rail beside `main` that collapses to an icon-only strip. Collapse state is persisted (`navigationCollapsed` in `localStorage`) and read on init, so a collapsed rail paints collapsed on first render instead of expanding and reflowing the page beside it.
- Nine hand-written tooltip/button blocks collapse into one descriptor table (`src/definitions/NavigationItems.ts`) plus one render loop. Disabled reasons, active-path matching and per-tab identity colours each now have a single home.
- `RouteRequirements` — the `ProtectedRoute` redirect guard — is derived from those same descriptors, so "which report does this route need" is stated once rather than in two independent tables that drift silently in both directions.
- Collapsing is **manual only**. The old navbar's responsive auto-collapse (1180px) and type shrink (1370px) are intentionally not carried over; the trade-off is tracked in #1930.

## Server-mode banner

The hosted-deployment install banner moves out of the navigation into its own `ServerModeBanner` component, so restyling the navigation can't take the hosted build's only signpost with it. Its pointer listener now only reaches React when the reveal threshold is actually crossed, rather than on every `mousemove`.

## Behaviour fixes riding along

These are not restyling, and are worth reviewing on their own terms:

- **Topology reachability.** Cluster availability was mirrored into an atom written only by the footer's range slider, which unmounts when there is no active report — so a report *with* cluster data left Topology enabled for reports without it. It now reads the report-scoped query directly, and that query is gated on an active report, so it no longer fires a guaranteed 404 on every report-less page load.
- **Modal re-entry.** Activating Topology while the cluster overlay was already open recorded the overlay's own route as its background, leaving the overlay with itself underneath. The previous guard was `pointer-events: none`, which keyboard activation went straight past; the check now lives in the navigation handler.
- **Cluster overlay.** Switched from `absolute` to `fixed` with `overscroll-behavior: contain`, so it no longer slides off-screen on long pages or chain-scrolls the page beneath it.
- **Blueprint stacking.** `.bp6-button-group` / `.bp6-control-group` get `isolation: isolate`. Blueprint raises intent-coloured children to z-index 10–16 without a stacking context, so those values escaped into the root one and painted over full-screen overlays. **Blast radius is every button group in the app** — flagging it explicitly.

## Layout follow-on

`$app-navbar-height` is gone. MLIR and graph view heights, and the home page's sizing, are retuned for the narrower `main`. The performance report's width floor moves from `.perf-report` onto `.perf-table`, so a narrow viewport scrolls the table inside its own container rather than scrolling the whole document — and the table's unvirtualised rows don't re-run auto layout every time the rail is toggled.

## Conventions

- The navigation mixin partial moves from `src/scss/definitions/` to `src/scss/mixins/`, matching the split CONVENTIONS.md draws between the two.
- `useGetClusterDescription` gains explicit `useQuery` generics and `staleTime: Infinity` (#1920). Its `initialData: null` sentinel is dropped — under `staleTime: Infinity` it would have counted as fresh forever and the descriptor would never have been fetched at all.

## Test coverage

New: click-through navigation (including the modal background-location state and the re-entry guard), all four `isActivePath` branches, MLIR's disabled branch under a stubbed non-dev build, collapse persistence and first-paint restore, the cluster query's `enabled` gate, and the derived `RouteRequirements` table. `TestProviders` gains an `initialEntries` pass-through so specs can place the router.


src/scss/components/SideNavigation.scss
```diff
@@ -0,0 +1,161 @@
...
+                flex: 0 0 27px;
+                overflow: hidden;
+            }
+        }
+
+        // Hidden rather than unrendered so the button keeps its accessible name. Takes the
+        // inline badge with it, which has nowhere to go beside a bare icon.
+        .bp6-button-text {
+            display: none;
+        }
```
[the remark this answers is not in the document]
aidemsined: Checked this against the installed Blueprint rather than assuming, and I think the finding is incorrect — the badge is a **child** of `.bp6-button-text`, not a sibling, so `display: none` already takes it.

`node_modules/@blueprintjs/core/package.json` is `6.6.1`, matching the `~6.6.1` pin. The render path in `lib/cjs/components/button/buttons.js:87` is:

```js
hasTextContent && jsxs(Text, {
  className: cx(Classes.BUTTON_TEXT, textClassName),
  tagName: "span",
  children: [text, children]     // <- children go INSIDE the span
})
```

`SideNavigation.tsx:85` passes `{item.badge ? <small>{item.badge}</small> : null}` as `children` of `<Button>`, so it lands in that `children` slot — inside the span, as a sibling of the *text node*, not of the span. Hiding `.bp6-button-text` removes the whole subtree.

Worth saying where the confusion came from, because it's ours: the comment at `SideNavigation.scss:81` states that Blueprint "renders it as the label's sibling". That comment is wrong, and it's the natural thing to read before reaching line 140. I'd fix the comment — it's the part that will mislead the next person.

Also bounding the impact either way: only one item carries a badge (`NavigationItems.ts:104`, MLIR), so even on the stated premise this would have been confined to a single row.

src/scss/components/SideNavigation.scss
```diff
@@ -0,0 +1,161 @@
...
+    flex: 0 0 auto;
+    display: flex;
+    flex-direction: column;
+    gap: 15px;
+    height: 100vh;
+    width: variables.$side-nav-width-expanded;
+    padding: 15px 8px;
+
+    // The fixed footer info bar crosses the foot of the rail, so the last item clears it.
+    padding-bottom: variables.$app-status-bar-height + 15px;
```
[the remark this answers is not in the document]
aidemsined: Confirming this one — it holds up, and I think it's the most substantive open item on the PR. Adding the stacking and scroll analysis, which is what makes it a blocker rather than a cosmetic overlap.

The arithmetic: the rail reserves `$app-status-bar-height + 15px` = **49px** (`SideNavigation.scss:21`), and `_variables.scss:65` defines `$app-status-bar-height: 34px` as the footer's *closed* `min-height`, with its own comment noting it grows when opened. When `sliderIsOpen`, the footer gains the `gap` plus the whole `RangeSlider` block — `.inputs` as a two-row flex column, `.slider-container` and `.slider` each with their own margins — on top of the 24px `.current-data` row. That's comfortably 100px+ against 49px reserved.

Two things turn the overlap into "unclickable" rather than "clipped":

- **Stacking.** `.app-footer` is `position: fixed; bottom: 0; width: 100%` with an explicit `z-index: $z-index-2-interactive` (= 1). `.side-navigation` is `position: sticky` and declares **no** `z-index`, so it's `auto`. The footer paints over the rail unconditionally, and it spans the full viewport width so it genuinely crosses the rail's foot rather than stopping at `main`.
- **No escape.** The rail is `height: 100vh` with `overflow: hidden` (`:16`, `:39`), so the covered items can't be scrolled into view.

The consequence is that while the range slider is open — the normal working state on Performance and Operations — the collapse toggle and the last one or two items are covered. That's worst in the collapsed state, where the toggle is the only way back to expanded.

On the fix: I don't think a static SCSS constant can work here, since the expanded height is content-driven. A CSS custom property set on `.app-footer` that both it and the rail's `padding-bottom` read would keep the two in step without either having to know the other's contents.

src/components/SideNavigation.tsx
```diff
@@ -0,0 +1,124 @@
...
+            <div className='side-navigation-items'>
+                {items.map((item) => {
+                    const tooltipContent = getTooltipContent(item, isCollapsed);
+
+                    return (
+                        <Tooltip
+                            key={item.route}
+                            content={tooltipContent}
+                            position={Position.RIGHT}
+                            disabled={!tooltipContent}
```
aidemsined, line 64: > [!CAUTION]
> This `aria-label` is the *only* thing giving the collapsed rail accessible names — but the comment explaining it says the opposite, and the test that claims to pin it can't.

Three things are out of step here.

**The mechanism.** `SideNavigation.scss:132` says the label is "Hidden rather than unrendered so the button keeps its accessible name". `display: none` removes the subtree from the accessibility tree entirely, so it does *not* preserve a name. The collapsed rail has names purely because of this `aria-label`. The behaviour is correct; the stated reason is inverted, which is exactly the shape of comment that gets the `aria-label` deleted as redundant in a later cleanup.

**The test can't catch it.** `tests/sideNavigation.spec.tsx:171-180` ("keeps every item reachable by name while collapsed") queries `getByRole('button', {name: /reports/i})`. `vitest.config.ts` doesn't enable `test.css`, so no stylesheet applies in jsdom — the label text nodes are present in both states and satisfy the accessible-name computation on their own. The test passes with `aria-label` removed. Its own comment at `:174-175` claims the labels "have to survive as `aria-label` rather than as text nodes", which is the thing it doesn't establish.

So today, deleting this prop leaves nine unlabelled icon buttons in the collapsed rail and the suite stays green.

**Suggested action:** fix the comment at `SideNavigation.scss:132` to name `aria-label` as the source, and change the assertion to `expect(button).toHaveAttribute('aria-label', 'Reports')` so it tests the mechanism rather than the rendered text.

src/components/SideNavigation.tsx
```diff
@@ -0,0 +1,124 @@
...
+                            fill
+                        >
+                            <Button
+                                text={item.label}
+                                aria-label={item.label}
+                                onClick={() => handleNavigate(item)}
+                                active={item.isActive}
+                                icon={item.icon}
+                                disabled={item.isDisabled}
```
aidemsined, line 79: > [!WARNING]
> The active route is conveyed by colour and a CSS class alone — no `aria-current`.

I checked what Blueprint emits for `active` in 6.6.1 rather than assuming: `buttons.js` sets only `aria-disabled`, never `aria-pressed` or `aria-current`. So a screen-reader user gets no signal about which view they're on, and neither does anyone who can't distinguish the item colours.

This is parity with the deleted nav rather than a regression — the old `MainNavigation` didn't have it either. Raising it because a nav rewrite with an open accessibility thread is the natural moment, and it's a two-word change:

```tsx
aria-current={item.isActive ? 'page' : undefined}
```

Related, same area: the toggle at `:96-115` carries `aria-expanded` but no `aria-controls`, so the expanded/collapsed state isn't tied to the thing it controls.

**Suggested action:** add `aria-current` on the nav buttons; `aria-controls` on the toggle if you want the pair complete.

src/store/app.ts
```diff
@@ -50,6 +50,15 @@ export const showHexAtom = atomWithStorage('showHex', false); // Used in Buffers
 export const showMemoryRegionsAtom = atomWithStorage('showMemoryRegions', true); // Used in Buffers and Operation Details
 export const renderMemoryLayoutAtom = atomWithStorage('renderMemoryLayout', false); // Used in Buffers and Operation Details
 
+// App shell / navigation
+// Persisted because a rail width is a stable preference rather than session state.
+// `getOnInit` because this one drives the app shell's layout: without it a user who
+// collapsed the rail renders it expanded first and the page beside it reflows on every
+// load.
+export const isNavigationCollapsedAtom = atomWithStorage('navigationCollapsed', false, undefined, {
```
aidemsined, line 58: > [!WARNING]
> The persistence here is entirely untested — the suite passes identically with `atomWithStorage` swapped for a plain `atom`.

`getOnInit: true` is used nowhere else in the repo, and the comment above gives a specific reason for it (avoiding a first-paint reflow for a user who collapsed the rail). That makes it the most interesting decision in this file, and nothing pins it.

Every collapse test either starts from a cleared store (`tests/sideNavigation.spec.tsx:42`) or hydrates the atom directly, so none of them exercises the round trip. `:110 "starts expanded"` sits in the collapsing describe block but asserts the default value, not persistence.

For what it's worth I think the choice is right: persisting matches the sibling UI-preference atoms above it (`showHex`, `showMemoryRegions`, `renderMemoryLayout`) rather than the report-scoped ones below, and a nav preference should outlive a report switch. It just isn't defended.

**Suggested action:** one test that sets collapsed, remounts, and asserts the rail comes back collapsed — plus an assertion that the `navigationCollapsed` key is written. That protects both the storage choice and the `getOnInit` flag.

src/components/ServerModeBanner.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+// How close to the top of the viewport the pointer has to come before the banner
+// reveals itself.
+const REVEAL_THRESHOLD_PX = 80;
+
+/**
+ * Points hosted visitors at the installable build. Lives beside the navigation rather
+ * than inside it so that restyling the navigation can't take the hosted deployment's
+ * only signpost with it.
+ */
```
aidemsined, line 22: > [!WARNING]
> This docblock states a placement invariant that no test verifies — and the spec that used to check it was deleted in this PR.

The comment says the banner "Lives beside the navigation rather than inside it so that restyling the navigation can't take the hosted deployment's only signpost with it". That's the component's stated reason for existing, and it's a claim about where `Layout` mounts it — not something the unit spec can see.

`tests/Layout.spec.tsx:26-27` mocks both `SideNavigation` and `ServerModeBanner` to `null`, so there's now no test that `Layout` renders either one, let alone that the banner sits outside `.app-shell`.

Worth knowing this used to be covered: `1c700cbb` replaced `tests/layout.spec.tsx` (79 lines) with the new banner unit spec. The deleted file asserted three things through `Layout` — that the rail mounts inside `.app-shell`, that the banner renders under `SERVER_MODE`, and that it doesn't on a local install. The new spec covers the banner's *behaviour* well (the listener-attach and listener-detach assertions at `:42` and `:85` are the real ones), but not its placement.

**Suggested action:** restore the two `Layout`-level assertions from the deleted spec alongside the unit spec. The invariant this docblock describes is the one thing worth protecting, and it's the one thing now uncovered.

src/scss/mixins/_navigation.scss
```diff
@@ -0,0 +1,106 @@
...
+            cursor: default;
+            pointer-events: none;
+        }
+    }
+
+    small {
+        font-size: 11px;
+        position: absolute;
+        border-radius: 4px;
+        padding-left: 4px;
```
aidemsined, line 101: > [!WARNING]
> The positional declarations in this `small` block are dead — the mixin's only consumer overrides all three at equal specificity.

`nav-button-styles` has exactly one consumer, `SideNavigation.scss:82-88`, which immediately sets `position: relative; top: -5px; right: -3px` on the same selector later in the cascade. So `position: absolute`, `right: 0` and `top: 0` here never apply. Only `font-size`, `border-radius`, `padding-*` and `background-color` survive.

Worth contrasting with the sibling: `button-colours` genuinely has two consumers (this file and `OperationDetailsNavigation.scss:19`) and earns its place in a shared definitions file. `nav-button-styles` doesn't yet — it was promoted to shared on the strength of one caller, and the dead positioning is the cost of that.

**Suggested action:** drop the positional declarations from the mixin, or move the whole `small` block down into `SideNavigation.scss` so the shared file only holds what's actually shared.

src/components/SideNavigation.tsx
```diff
@@ -0,0 +1,124 @@
...
+    'https://docs.tenstorrent.com/tt-tm-assets/Logo/Standard%20Lockup/svg/tt_logo_color-orange-whitetext.svg';
+const LOGO_WIDTH = 150;
+
+// Collapsed, the label is the only thing identifying an icon, so it has to reach the
+// tooltip; expanded, the label is already on screen and only a blocked item has
+// something left to say.
+function getTooltipContent(item: ResolvedNavigationItem, isCollapsed: boolean): string {
+    if (item.disabledReason) {
+        return item.disabledReason;
+    }
```
aidemsined: > [!WARNING]
> When an item is both collapsed and disabled, the tooltip loses the one thing identifying which item it belongs to.

In that branch the function returns `disabledReason` alone. On a collapsed icon-only button that produces a tooltip reading *"Upload or select an active memory report to enable this feature"* with nothing naming the view — and four of the nine items share that exact string, so hovering three of them in sequence gives identical text.

The expanded case is fine, because the label is on screen next to it.

The function itself is good — pure, four lines, and the comment genuinely explains why the two cases differ. It's just that the collapsed-and-disabled combination drops the wrong half.

**Suggested action:** compose both when collapsed, e.g. `` `${item.label} — ${item.disabledReason}` ``.

src/scss/_blueprintjs.scss
```diff
@@ -103,6 +103,17 @@ $pt-intent-text-colors: (
...
 
+// Blueprint raises an intent-coloured button or input inside a group to z-index 10-16 so
+// its border paints over the neighbour it shares an edge with. Those values only ever need
+// to beat siblings in the same group, but Blueprint sets them without a stacking context,
+// so they escape into the root one and land on top of our overlay scale ($z-index-3-overlay
+// is 10) -- an active segmented control then paints over a full-screen overlay covering the
+// page. Isolating the group keeps the border ordering intact and keeps the numbers local.
+.bp6-button-group,
+.bp6-control-group {
+    isolation: isolate;
```
aidemsined, line 120: > [!WARNING]
> App-wide Blueprint behaviour arriving in a commit titled "Handle topology scrolling".

The comment here is genuinely good — it names Blueprint's z-index 10–16 range, explains those values only need to beat siblings, and identifies `$z-index-3-overlay: 10` as the collision. No complaint about the reasoning.

The concern is scope and discoverability. `isolation: isolate` on every `.bp6-button-group` and `.bp6-control-group` in the app creates a new stacking context wherever those classes appear, and it landed in `933d7105` alongside the ClusterView fix — so nobody reviewing "topology scrolling" would expect to be reviewing a global override.

I bounded the blast radius and **nothing is broken today**: of the five `usePortal={false}` popovers in the codebase (`InitialMessage.tsx:23`, `PerfTableToolbar.tsx:104`, `PerfHeuristicFlags.tsx:45`, `perfFunctions.tsx:133` and `:218`), none currently sits inside a button or control group. It's a latent trap, not a live bug — but the next inline-portal popover placed inside a group will be trapped under its siblings with no obvious cause.

**Suggested action:** either add the constraint to the comment ("no inline-portal popover may live inside a group"), or scope the isolation to the groups that actually collide with the overlay rather than all of them.

aidemsined (commented):
The three-way split — `NavigationItems.ts` as data, `useMainNavigationItems.ts` as resolution, `SideNavigation.tsx` as presentation — is a clear improvement on what it replaces. The old `MainNavigation` repeated a nine-line tooltip/button block eight times with the gates inlined per item; adding a nav item is now one object literal, and `NAV_DISABLED_REASON` being a `Record<NavRequirement, string>` makes a new requirement without a reason a type error.

I checked the rewrite for behavioural parity rather than reading the line counts, and it holds up: all nine items keep their routes, icons and colour classes; every gate is reproduced, including the MLIR dev bypass (`!import.meta.env.DEV && !hasActiveMlir` → `import.meta.env.DEV || !!activeMlirJson`, logically identical); all four disabled-reason strings carried over verbatim; the modal-background navigation for Topology survives. `isActivePath` is actually *safer* than the base — the old `location.state?.background?.pathname.includes(path)` would throw on an undefined pathname, the new `!!backgroundPath &&` guard can't. The only loss is the decorative separator icon between Reports and Operations, which has no rail equivalent.

Two things I want to single out as done well. `useAPI.tsx:553` adding `enabled: !!activeProfilerReport?.path` to `useGetClusterDescription` belongs in this PR precisely because the PR is what moved that query to app scope — without it every cold start would fire a guaranteed 404. And `tests/sideNavigation.spec.tsx:81` pins the actual regression this rewrite fixes: the old `hasClusterDescriptionAtom` was written only by `RangeSlider`, which unmounts with no active report, so a report with cluster data left Topology enabled for a subsequent report without it.

The layout ripple is mostly clean. `$app-navbar-height` had exactly two consumers at base and both were updated; `$bp-nav-large`/`$bp-nav-medium` died with the file that owned them; `.app-header` and `.nav-container` are fully gone; the `NEW_MENU` flag left zero residue across `.env.sample`, `settings.py`, `_build_spa_client_config`, `ServerConfig.ts` and `getServerConfig.ts`. One hard-coded literal escaped, below.

## Unanchored findings

**`src/scss/components/PerfChartsIndex.scss:32` still reserves the deleted navbar band.** `.perf-charts-index-menu { max-height: calc(100vh - 80px) }` subtracts 80px of vertical chrome that no longer exists, so that dropdown is permanently ~80px shorter than it should be. It survived the sweep because it's a raw `80px` rather than a `$variable` — the same shape as the `top: 80px` in `ClusterView.scss` that `933d7105` did catch. Cosmetic rather than broken, but it's the one remaining piece of the ripple. Either drop the subtraction or express it against `$app-status-bar-height` so the next chrome change finds it.

**`CONVENTIONS.md:245` still cites `MainNavigation.scss`** as the canonical example of a PascalCase component stylesheet, in the PR that deletes that file. The PR touched `CONVENTIONS.md` to rename the `hasClusterDescriptionAtom` example, so this was just missed two hundred lines earlier. The line already lists other examples, so it can lose this one.

## General observations

> [!NOTE]
> Minor items and follow-ups — not merge-blocking.

- **No linked issue, and the body contradicts the title.** The body says *"New horizontal menu"*; the title says *"collapsible side menu"*; the code builds a `flex-direction: column` rail. The title is right. A reviewer arriving cold can't tell whether the horizontal wording describes a scrapped approach, a follow-up, or this change — worth a one-line fix to the body.
- `$mlir-footer-clearance: $app-status-bar-height + 6px` reproduces the old literal 40px via an unexplained `+ 6px`, while the comment says the clearance *is* the footer info bar (34px). The net box math is unchanged from base, so this isn't a regression — but the comment asserts a rationale the arithmetic doesn't support. Either say what the 6px is or drop it.
- Three changes in `PerfReport.scss` are unrelated to navigation: `cursor: help` on the heuristic flag, and `.table-footer td` padding replaced with `line-height: 1` (which silently changes the footer row height). The `min-width: 1240px` removal in the same commit *is* rail-driven — the rail takes 180px, so a 1240px floor would force a horizontal scrollbar on laptops — and should stay.
- `items` in `useMainNavigationItems.ts:67` is rebuilt every render while `handleNavigate` beside it is `useCallback`-wrapped. Nine items make this irrelevant to performance; the asymmetry just reads as unfinished.
- Three tests assert `getByAltText('tenstorrent')` has a `src` containing `tt_logo_color` — a component constant that can't distinguish collapsed from expanded and would pass with the collapse feature deleted. The adjacent class assertion at `:147` is what carries the weight. `:188` uses `getButtonWithText`, which throws when absent, so the `toBeInTheDocument()` after it can never fail; the sibling at `:199` correctly uses `queryByRole`.
- `vitest.config.ts` sets neither `restoreMocks` nor `clearMocks`, so the `addEventListener`/`removeEventListener` spies in `tests/ServerModeBanner.spec.tsx:43` and `:87` stay installed for the rest of the file. Nothing breaks — both retain the real implementation — but it's an unbalanced pair in a spec about unbalanced listener pairs.
- Accessibility follow-up beyond this PR: `OperationDetailsNavigation.tsx:78` renders a second `<nav>` with no label, so on an operation-details page a screen reader now lists "Main navigation" and one bare "navigation". Strictly better than the two bare entries before, and labelling the second is out of scope here.
- Also out of scope but worth recording: the shell's `<h1>` went with the old `<header>`, and `SideNavigation.tsx:56` replaces it with a plain `<span className='visualizer-title'>`. Home, Operations, Tensors, Buffers and Graph now have no `<h1>` at all — NPE, MLIR, Buffers-by-operation and Performance define their own.

Not run from here: `npm run lint`, `tsc --noEmit`, and the suite. Two specifically worth watching — `tests/serverMode.spec.tsx` and `tests/localFolderSelector.spec.tsx` both mock `useGetClusterDescription` and sit downstream of the `enabled` change. And the lint run matters for the `{ isOpen: _isOpen, ... }` destructure at `SideNavigation.tsx:96`, which depends on an unused-args ignore pattern I couldn't confirm by reading. Visually, the combination worth checking is the range slider open at 1280px in the collapsed state — see the thread on `SideNavigation.scss:21`.

src/components/SideNavigation.tsx
dcblundell: Confirmed and fixed. I checked the claim by deleting the `aria-label` prop and running the suite: the test passed, so nine unlabelled icon buttons would indeed have shipped green.

- The SCSS comment now states that `display: none` removes the label from the accessibility tree too, and that `aria-label` is therefore the sole source of names — worded so the prop does not read as redundant.
- The assertion is now on the attribute itself (`toHaveAttribute("aria-label", "Reports")`), split into its own test from the enabled/disabled one. Re-ran the deletion: it now fails.

src/components/SideNavigation.tsx
dcblundell: Agreed, and verified independently — Blueprint 6.6.1 emits only `aria-disabled` from its button modules, no `aria-pressed` or `aria-current`. Confirmed the old `MainNavigation` had none either, so this is a pre-existing gap rather than a regression, and a fair moment to close it.

Added `aria-current={item.isActive ? "page" : undefined}`, plus `aria-controls` on the toggle pointing at the items container (which gained an id), so `aria-expanded` names what it expands. Both are covered, and I checked the `aria-current` test fails when the prop is removed.

src/store/app.ts
dcblundell: This one is already addressed on the branch — I think the comment was written against `f728cb53`, before the review round landed.

`tests/sideNavigation.spec.tsx` now has a `SideNavigation collapse persistence` block doing what you describe: it seeds `navigationCollapsed` in `localStorage`, asserts the rail paints collapsed on first render (which is what pins `getOnInit`), and separately asserts the key is written on toggle. Both halves assert the exact stored value, so the write and read formats are pinned against each other.

Agreed on the reasoning for keeping `atomWithStorage` — it matches the sibling UI-preference atoms rather than the report-scoped ones. Resolving as covered; reopen if you think the round trip wants asserting in a single test instead of two.

src/components/ServerModeBanner.tsx
dcblundell: The gap is real and I have fixed it — but the supporting history in this comment is not correct, flagging so it does not get cited later.

I read `tests/layout.spec.tsx` at `1c700cbb^` in full. It contained only the two usage-recording tests (`starts usage recording on mount`, `balances every start with a teardown`). Grepping it for `app-shell`, `SERVER_MODE`, `banner` or `navigation` returns nothing but the `MainNavigation` mock line. It was a case-rename to `Layout.spec.tsx` plus a mock swap, not a replacement — the 79-line delete and 3-line add in the stat are the two sides of that rename. So the placement assertions were never there to restore; they had to be written fresh.

Done in `tests/Layout.spec.tsx`: the child stubs are now identifiable markers instead of `null`, and three tests assert the rail and `main` sit inside `.app-shell` while the banner and the fixed footer sit outside it. Verified by moving `<ServerModeBanner />` inside the shell — the banner test fails, as it should.

src/scss/mixins/_navigation.scss
dcblundell: Agreed — this came up in our own review pass too. Confirmed `nav-button-styles` has exactly one consumer and that it overrides `position`, `top` and `right` at equal specificity later in the cascade, so those three declarations never applied.

Dropped them from the mixin, which now carries appearance only (`font-size`, `border-radius`, `padding`, `background-color`), with a note that placement belongs to the consumer because it depends on the shape of the button the badge hangs off. The rail keeps its `position: relative` nudge in `SideNavigation.scss`, and I reworded the comment there so it no longer contrasts with an absolute default that has gone.

src/components/SideNavigation.tsx
dcblundell: Good catch — this was the most user-visible item in the batch. Confirmed four items share the `PROFILER_REPORT` string, so a collapsed rail gave four identical tooltips naming nothing.

Taken your suggestion: the collapsed-and-blocked branch now composes `` `${item.label} — ${item.disabledReason}` ``. Expanded stays as the reason alone, since the label is already on screen beside it.

The function moved to `src/functions/getNavigationTooltip.ts` so it can be tested directly — Blueprint mounts the tooltip only on hover and jsdom does not drive that machinery (I tried both `mouseEnter` and `mouseOver` on the popover target; neither opens it). Exporting it from the component tripped `react-refresh/only-export-components`, hence the separate module. All four branches are now covered in `tests/getNavigationTooltip.spec.ts`.

src/scss/_blueprintjs.scss
dcblundell: Verified your bounding: all five `usePortal={false}` sites are where you say, and none of those files references `ButtonGroup` or `ControlGroup`, so nothing is trapped today. Agreed it is latent rather than live.

Taken the first of your two suggestions — the comment now states the constraint the rule buys: an inline (`usePortal={false}`) popover must not live inside a button or control group, because the stacking context traps it under the group siblings, and that failure has no obvious symptom at the call site.

Declining the narrowing: scoping to "the groups that actually collide with the overlay" means maintaining a list that goes stale the moment someone adds a group, and the failure mode of a missing entry is the original bug returning silently. The blanket rule with a documented constraint seems the better trade.

On discoverability — fair point that it landed under `Handle topology scrolling`. The PR description has been rewritten and now calls this out under its own heading with the app-wide blast radius flagged explicitly.

tests/ServerModeBanner.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+        expect(screen.queryByTestId(TEST_IDS.SERVER_MODE_BANNER)).not.toBeInTheDocument();
+    });
+
+    it('attaches no pointer listener in a local deployment', () => {
+        const addEventListener = vi.spyOn(window, 'addEventListener');
+
+        render(<ServerModeBanner />);
+
+        expect(addEventListener).not.toHaveBeenCalledWith('mousemove', expect.any(Function));
```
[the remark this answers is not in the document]
dcblundell: Correct, and this was a regression I introduced in the previous round when I added `{ passive: true }` to the registration. Verified by forcing `SERVER_MODE: true` inside that test: it still passed, so the assertion had gone vacuous exactly as you describe.

Now asserts on the recorded event types (`spy.mock.calls.map(([type]) => type)` must not contain `mousemove`), and I added the positive counterpart in server mode so the negative assertion cannot go vacuous again unnoticed. Re-ran the same mutation: it now fails.

src/components/SideNavigation.tsx
```diff
@@ -0,0 +1,123 @@
...
+                            position={Position.RIGHT}
+                            disabled={!tooltipContent}
+                            fill
+                        >
+                            <Button
+                                text={item.label}
+                                aria-label={item.label}
+                                // Blueprint emits only `aria-disabled` for `active`, so
+                                // without this the current view is conveyed by colour alone.
+                                aria-current={item.isActive ? 'page' : undefined}
```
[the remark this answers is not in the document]
dcblundell: Confirmed — I counted `[aria-current="page"]` in a modal-open render and got 2 (Operations and Topology). Taken your suggested split.

`ResolvedNavigationItem` now carries `isCurrentPage` alongside `isActive`: `isCurrentPage` matches `location.pathname` only, `isActive` keeps the background match. `aria-current` is keyed off the former, the identity colours off the latter, so the dual visual state is unchanged while exactly one item is announced as current.

At most one item can match `isCurrentPage` — no route in `ROUTES` is a substring of another and `HOME` is excluded from the nested match — and there is now a test asserting the count is 1, that it is Topology, and that both items still carry `bp6-active`.

src/components/ServerModeBanner.tsx
```diff
@@ -0,0 +1,84 @@
...
+        return null;
+    }
+
+    return (
+        <div
+            className='server-mode-banner'
+            data-testid={TEST_IDS.SERVER_MODE_BANNER}
+            style={{
+                transform: isRevealed ? 'translateY(0)' : 'translateY(-100%)',
+            }}
```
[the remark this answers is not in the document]
dcblundell: Agreed. The banner is translated off-screen rather than removed, so the links stayed tabbable, and the only reveal path was `mousemove`.

Fixed with the keyboard equivalent rather than `inert`, which would have removed the reveal path entirely: the reveal moved from an inline `style` to a `revealed` class, so both transform values live in the stylesheet and a `:focus-within` rule can reveal the banner too. The inline style was the blocker — it outranks any stylesheet rule, so `:focus-within` could not have worked while it was there. Tabbing into either link now brings the banner into view rather than moving focus somewhere invisible.

Worth flagging that the `:focus-within` half is not covered by a test: `vitest.config.ts` does not enable `test.css`, so no stylesheet applies in jsdom and CSS-driven state is unobservable there. The spec asserts the class contract instead, and I noted the limitation in the test.

src/scss/components/SideNavigation.scss
```diff
@@ -0,0 +1,169 @@
...
+            @include navigation.nav-button-styles;
+
+            // Reduce horizontal padding so button icons stay in the same position when the rail collapses.
+            padding-left: 12px;
+            padding-right: 12px;
+
+            // A full-width rail button would strand a corner superscript at the far edge,
+            // so the badge sits inline after the label and is nudged up into superscript
+            // position. Blueprint already renders it as the label's sibling; the shared
+            // mixin supplies its appearance and leaves placement here.
```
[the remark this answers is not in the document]
dcblundell: You are right on the structure. I dumped the rendered markup to check rather than reasoning from the source, and Blueprint emits `<span class="bp6-button-text">MLIR<small>beta</small></span>` — the badge is inside that span, after the label text.

Comment corrected to say so, and to draw the connection you point at: that containment is exactly why the collapsed `display: none` on `.bp6-button-text` takes the badge with it.

Issue #1610, opened 2026-06-03
Operation graph: perf overlay — discoverability (hover tooltips, linked count, unmapped style)


aidemsined:
## Context

Follows up on #1515 (Operation graph: perf overlay) which shipped a Switch + continuous color ramp + per-op metric in the property panel. This issue cleans up the rough edges so the overlay reveals its data without forcing a click on every node.

## Scope

1. **Hover tooltip on graph nodes.** When the overlay is active, hovering a node surfaces its kernel duration and rank (e.g. `Kernel duration · 12.5 ms  (#7 of 198)`). `vis-network` already supports `title` on nodes — derive from `deviceTimeByOpId` and the `scoreOps` output.
2. **"N of M ops linked" indicator.** Surface the partial-join count near the perf overlay Switch in the toolbar — currently the user has no way to tell whether 5% or 95% of graph ops have perf data. Pull `linkedOpCount` (already computed in `OperationGraphComponent`) and `operationList.length`.
3. **Muted style for unmapped ops.** When overlay is active but a node has no perf row, render it in a muted grey instead of the default `GRAPH_COLORS.normal` (which currently looks identical to a `t=0` colored node). Makes "no data" visually distinct from "low duration".

## Acceptance criteria

- [ ] Hovering any node with perf data shows duration + rank in a `vis-network` tooltip; hovering an unmapped node shows "no perf data".
- [ ] Toolbar shows e.g. `Perf overlay (180/302)` next to the Switch when status is `READY`; hidden when status is `UNAVAILABLE` / `UNLINKED`.
- [ ] Nodes without perf data render in a distinct muted style, both at zoom-in and zoom-out, when overlay is on.
- [ ] Styleguide example covers the muted style.

## Out of scope

- Metric selector (separate phase).
- Aggregation method selector (#TBD-Phase4).

## Notes

- Smallest-impact follow-up to #1515 — touches only `OperationGraphComponent.tsx`, `perfOverlay.ts`, and SCSS.
- Builds straight on top of existing infrastructure; no backend changes, no atom changes.

PR #1890, opened 2026-08-17
Op graph perf overlay: encode kernel duration on a bar that composes with selection and highlight

aidemsined:
image

## Summary

Brings the perf overlay to the React Flow operation graph and changes how it encodes duration. The previous version recoloured the whole node, which put perf in direct competition with the input/output highlight (same channel) and left selection as the only survivor. Here per-op kernel duration goes on an inset bar drawn as a `::after` pseudo-element, sized and coloured from two CSS custom properties, so all three signals are readable at once. Also adds the overlay's discoverability affordances: a hover chip with duration, rank and share of total, an N-of-M linked count beside the toolbar switch, and explicit "no perf data" copy wherever an op isn't in the report.

Refs #1880
Refs #1610

### Frontend

- **`operation-graph/opGraphPerfOverlay.ts`** (new) — the graph-scoped overlay model layered on the shared `functions/perfOverlay.ts` primitives: status derivation, ramp anchoring, ranking, the per-node style patch, and the hover label. Everything is scoped to the ops the canvas is currently showing rather than the whole report, so an operation-range selection or hidden deallocates re-anchor the ramp instead of pinning it to a maximum the user can't see. The same intersection drives the linked count and the hover's rank.

- **Bar geometry** (`OperationGraphReactFlow.scss`) — the bar is absolutely positioned so it never becomes a flex item. That keeps the node's measured box identical whether the overlay is on or off, which matters because Dagre laid the graph out against those sizes: an in-flow bar would make every toggle a relayout. Its size divides by the live viewport zoom, because a size fixed in graph units is scaled down with the graph and went sub-pixel at exactly the zoom a large report fits in — the encoding disappearing where a whole-graph scan needs it most. The height is capped at the node, so at overview zoom the bar grows into a full-node fill rather than vanishing.

- **Unmapped ops** set neither custom property, so the CSS fallback paints nothing while a mapped but coolest op still keeps a visible stub. That difference is what separates "no perf data" from "fastest op" — the two were indistinguishable under the old fill encoding.

- **`OperationGraphReactFlow.tsx`** — overlay state, the styling pass that writes the custom properties alongside (not instead of) the selection class and the filter dim, the hover chip, the legend, and the panel wiring. Overlay intent is scoped to the report it was enabled for: without that reset, a report leaving and re-entering a linkable state switched the encoding back on by itself. Zoom reaches the stylesheet through a React Flow store subscription that writes the custom property directly, since routing a value that changes on every wheel frame through React state would re-render the whole graph to move one bar. The published value is quantised to proportional buckets, so a wheel gesture invalidates style a handful of times instead of once per frame.

- **`OpGraphToolbar.tsx`** — the overlay switch with a tooltip that stays reachable in all three status states (Blueprint drops pointer events on a disabled control, so the explanation for *why* it's disabled has to survive), plus the linked count.

- **`OpGraphInfoPanel.tsx`** — the selected op's kernel duration and a swatch matching its bar, reusing the existing `PerfOverlayOpMetric`.

- **`Styleguide.tsx`** — the legend, the op metric in each state, and the node bar across the score ramp including the unmapped no-bar case (#1880 criterion 7, #1610 criterion 4).

## Also in this PR: the graph viewport height

Called out separately so it isn't approved implicitly — no criterion in #1880 or #1610 asks for it, and it is user-visible across the whole Graph view.

`.operation-graph-react-flow` moves from `height: 80vh` to `calc(100vh - navbar - main's top padding - status bar)`, with a negative bottom margin cancelling the page's trailing gutter. The overlay's legend docks to the bottom of the graph beside the zoom controls and minimap, and at `80vh` that band floated above the page's trailing padding rather than sitting at the visible canvas edge, which is what the new chrome has to line up against.

The chrome numbers this depends on are now shared variables — `$main-padding`, `$main-padding-bottom`, `$data-padding-bottom`, `$reactflow-panel-origin-left` — rather than private copies of values owned by `_base.scss`, `_common.scss` and the MLIR stylesheet. `$minimap-width` is now authored onto `.react-flow__minimap` instead of assuming React Flow's default. Compiled CSS is byte-identical to the literals it replaces.

## Known limitations

- #1610's third item asked for a muted grey *fill* on unmapped ops. That's superseded here: a grey fill would re-enter the same channel competition this PR exists to remove, so "no data" is carried by the absent bar plus the hover and panel copy instead.
- At overview zoom a hot op's bar covers its node, so perf outranks the input/output highlight there. Deliberate — the highlight isn't legible at that scale anyway.
- The hover chip is pointer-only and deliberately carries no ARIA role; the panel's Kernel duration row is the keyboard/AT route to the same figure.
- `MLIRViewReactFlow.scss` scopes its panel offsets under a bare `.react-flow`, so they leak app-wide and this branch out-specifies them to dock its own chrome. Root cause tracked in #1899.

## Test plan

- [ ] `pnpm test` — 1927 pass. Coverage: overlay status/scoping/ranking and the style-patch contract (`opGraphPerfOverlay.spec.ts`), the shared rank comparator's tie-break (`perfOverlay.spec.ts`), the bar's geometry, zoom floor and containing block (`opGraphPerfBarStyles.spec.ts`, regexes built from the exported property names), the toolbar switch including the mid-build disable (`OpGraphPerfOverlaySwitch.spec.tsx`), the panel metric (`OpGraphPanelPerfMetric.spec.tsx`), the report-scoped reset, zoom publishing, panel wiring, hover path, legend and toggle-off node identity (`OperationGraphRebuildDiscipline.spec.tsx`), and the three `isPerfReportLoaded` states at the route (`GraphViewPerfReportLoaded.spec.tsx`).
- [ ] `pnpm run lint:ts`, `pnpm lint`, `pnpm run lint:scss`, `pnpm lint:spdx` clean.
- [ ] Manual: load a profiler report with a matching performance report, enable the overlay, and confirm the bars read against selection and the input/output highlight simultaneously; that an op absent from the perf report shows no bar while the coolest mapped op shows a stub; that zooming out to fit the whole graph keeps both distinguishable; that toggling the overlay doesn't move the graph; and that switching to a different report leaves the overlay off.
- [ ] Manual: with only a profiler report loaded, confirm the switch is disabled and its tooltip still explains why.
- [ ] Manual: `/styleguide` renders the bar ramp and the unmapped no-bar node.


src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -115,7 +138,30 @@ const OperationGraphInner = ({ operationList, operationId }: OperationGraphReact
...
+        performance: activePerformanceReport,
+    });
+    if (
+        overlayReportScope.profiler !== activeProfilerReport ||
+        overlayReportScope.performance !== activePerformanceReport
+    ) {
+        setOverlayReportScope({ profiler: activeProfilerReport, performance: activePerformanceReport });
+        setIsPerfOverlayEnabled(false);
+        setPerfHover(null);
+    }
```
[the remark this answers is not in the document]
aidemsined: Keeping the render-phase adjustment. This is React's [documented pattern for adjusting state when props change](https://react.dev/reference/react/useState#storing-information-from-previous-renders), not the anti-pattern the warning targets. The two conditions it requires both hold: the update is guarded by a comparison, so the second pass finds the scope already equal and stops, and it sets state on the same component during its own render — React discards that render and re-runs before anything paints. The app mounts under `StrictMode` (`src/main.tsx:50`), and the double invocation is a no-op here for the same reason the loop terminates.

The reason it isn't just style: `isPerfOverlayActive` feeds `styledNodes`, so an effect lets one commit through between the report change and the reset — a frame of the new graph coloured by the previous report's ramp. `useLayoutEffect` narrows that to a pre-paint layout pass rather than removing it, and both cost an extra render pass that the render-phase update avoids.

The same file already resolves the identical problem this way for `adoptedOperationId` (landed in #1882), where the effect version left the details panel describing the operation we had just navigated away from.

The other half of your read was right, though: comparing `ReportFolder` objects by reference was wrong. 28c55f13 keys the scope on `report?.path`, so a refetch that rebuilds an equivalent object no longer reads as a report swap, with a regression test in `tests/opGraphPerfOverlayReset.spec.tsx`.

tests/opGraphPerfOverlayReset.spec.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const [isEnabled, setIsEnabled] = useState(false);
+    const activeProfilerReport = useAtomValue(activeProfilerReportAtom);
+    const activePerformanceReport = useAtomValue(activePerformanceReportAtom);
+    const [scope, setScope] = useState({ profiler: activeProfilerReport, performance: activePerformanceReport });
+    if (scope.profiler !== activeProfilerReport || scope.performance !== activePerformanceReport) {
+        setScope({ profiler: activeProfilerReport, performance: activePerformanceReport });
+        setIsEnabled(false);
+    }
+    return [isEnabled, setIsEnabled] as const;
+};
```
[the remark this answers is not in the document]
aidemsined: The hook mirrors the component deliberately, so the render-phase update stays for the reasons in the [thread on `OperationGraphReactFlow.tsx`](https://github.com/tenstorrent/ttnn-visualizer/pull/1890#discussion_r3809879677) — a test that reset in an effect would stop characterising the code it exists to pin.

Two parts of this were fair and are fixed in 28c55f13:

- The hook now keys on `report?.path`, matching the component, plus a test for the case that motivated it: the same report arriving as a rebuilt object must not drop the overlay.
- The comment saying nothing is written during render was simply wrong — it described an earlier draft. It now says the mirror includes the render-phase adjustment, and the sentence about the DOM is scoped to what it was actually about: intent is read through the rendered button rather than captured into module scope, so no assertion depends on when the hook runs.

aidemsined:
Brought the branch up to date with `dev` and pushed three commits on top.

**Merge (539f0f4f)** — `dev` renamed `OpGraphFilter` to `GraphOpFilter` and replaced per-node opacity dimming with a container class, so the perf bar now inherits the dim as part of the node group instead of needing its own rule. Also added a `useStoreApi` stub to the React Flow test harness, which the overlay's zoom subscription needs.

**Layout (f0a9ac80, 460fd9f5)** — the bottom-left chrome sat well above the status bar and the legend overlapped the minimap. The container was `height: 80vh`, so on a tall window it ended hundreds of pixels short of the viewport. It now fills the band between the navbar and the status bar, and a negative bottom margin drops the trailing page padding that would otherwise only add a scrollbar. The legend sits beside the minimap, and an override neutralises a bare-`.react-flow` rule in `MLIRViewReactFlow.scss` that lifts bottom-left panels app-wide. Result is a consistent 8px gap with the controls, minimap and legend on one baseline.

**Review feedback (28c55f13)**

| Finding | Outcome |
|---|---|
| Report scope compared `ReportFolder` by reference | Fixed — keyed on `report?.path`, with a regression test for the same report arriving as a rebuilt object |
| `height: calc(100vh - 92px - ...)` hardcoded the navbar height | Fixed — `$mlir-navbar-height` generalised to `$app-navbar-height`, and the status bar promoted to `$app-status-bar-height`, sourced from `.app-footer`'s own closed `min-height` rather than a literal |
| Move the overlay reset from render into an effect | Declined — rationale in the [inline thread](https://github.com/tenstorrent/ttnn-visualizer/pull/1890#discussion_r3809879677); an effect paints a frame of the new graph under the old ramp |

Also trimmed the comments this PR added by about 40 lines, keeping the intent the code can't state and dropping the sentences that restated it.

Full suite green (183 files, 1911 tests), and the layout re-verified in the browser at two window heights.

aidemsined:
Working through the four suppressed findings from the latest Copilot pass. Three were right and are fixed in d0b2fce2; the fourth is a real gap I'd rather not fold into this PR.

**Node identity under drag — fixed.** This was the serious one. `styledNodes` rebuilt every linked node on each render, so a drag frame handed React Flow a few hundred fresh objects instead of the one the pointer moved. The filter path was already guarded by `keeps non-matching nodes byte-for-byte across a drag frame`, and the overlay walked straight past it because it patches every scored node rather than the handful a filter marks. Styled nodes are now cached against their source node, class list and style patch, all stable between score changes. The new test fails on the old code and passes on the new, and a companion test proves the cache still lets a changed ramp through.

**`isPerfReportLoaded` from row count — fixed.** Now `perfReport?.report !== undefined`, so a report that parsed to zero rows reads as `UNLINKED` rather than telling the user to load the report they already have.

**Reset test duplicated the implementation — fixed, and the criticism was demonstrably correct.** I disabled the production reset and ran both: the standalone mirror stayed green at 5/5 while the ported real-component tests went red 3/5. All five cases now run against `OperationGraphReactFlow` through the existing React Flow and worker mocks, and `tests/opGraphPerfOverlayReset.spec.tsx` is deleted.

**Styleguide coverage — not in this PR.** Correct that #1880 asks for it and that the PR description lists it as outstanding. It needs bar, low-duration stub and unmapped-node entries alongside the existing legend and metric, which is a self-contained piece of work with no bearing on the overlay's behaviour. I'd rather land it separately than grow this diff.

Full suite green at 1913 tests.

dcblundell (commented):
Combined six-lens review (maintainability, tests, performance, conventions, security, requirements conformance against #1880 and #1610). Must-fix and should-fix findings are left inline below.

SECURITY found nothing: the change is pure frontend derivation with no trust boundary, and the custom-property values are internally synthesised and clamped rather than report-derived. Requirements conformance: 9 of 12 criteria implemented, 1 unimplemented (styleguide entries, in both issues), 1 deliberately superseded, 1 partial on naming.

Three spec-vs-code divergences look like the **issue** needs amending rather than the code — #1610's "muted grey fill" for unmapped ops (superseded by the no-bar encoding, as #1880's own comment anticipated), #1880's "never competes with the I/O highlight" versus the deliberate overview-zoom override, and criterion 8's `--graph-*` namespace (two of the three new properties are not colours). Also worth fixing: the PR's test plan names `opGraphPerfOverlayReset.spec.tsx`, which does not exist — that coverage landed in `OperationGraphRebuildDiscipline.spec.tsx:449` instead.

src/routes/GraphView.tsx
```diff
@@ -63,8 +63,12 @@ const GraphView = () => {
...
     // (UNAVAILABLE) from "loaded but doesn't match this graph" (UNLINKED).
     // `perfOverlayRows` collapses both into "empty"; this flag preserves the
-    // distinction. Unread until the overlay lands, alongside `perfRows`. #1880
-    const isPerfReportLoaded = Boolean(perfReport?.report?.length);
+    // distinction, which is what the toggle's disabled tooltip reads. Keyed on
+    // the selection rather than the row count, so a report that parsed to zero
+    // rows still reads as loaded instead of telling the user to load the report
+    // they already have. The query resolves an empty report when nothing is
+    // selected, so its data alone cannot answer this. #1880
+    const isPerfReportLoaded = activeReportFolderName !== null && perfReport !== undefined;
```
dcblundell, line 76: **Must fix:** `isPerfReportLoaded` has no test at any level

Reverting this to the pre-fix `perfReport !== undefined` leaves the entire suite green (verified by mutation). This line decides UNAVAILABLE vs UNLINKED — which tooltip the user reads, and whether the switch can ever enable — and it was rewritten twice on this branch (d0b2fce2, then 5360c8ee), i.e. a bug fixed twice with no regression test. `opGraphPerfOverlay.spec.ts` cannot cover it: the helper takes the boolean as an input, so neither half of the conjunction is exercised. Suggest a route-level spec asserting three states: name null + query resolved → false; name set + report in flight → false; name set + `report: []` → true.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -546,12 +719,32 @@ const OperationGraphInner = ({ operationList, operationId }: OperationGraphReact
...
+                </div>
+            ) : null}
             {isPanelOpen ? (
                 <OpGraphInfoPanel
                     operationId={selectedOperationId}
                     operationList={operationList}
                     operationNamesById={operationNamesById}
                     onLocateOperation={focusOperation}
+                    isPerfOverlayActive={isPerfOverlayActive}
+                    perfDeviceTimeNs={selectedPerfAggregate?.deviceTimeNs}
```
dcblundell, line 781: **Must fix:** panel perf props are wired here but tested only in the leaf

Hard-coding `perfDeviceTimeNs={undefined}` / `perfColor={undefined}` keeps every spec green — the panel would read "No perf data" for every op in the report and nothing would fail. `OpGraphPanelPerfMetric.spec.tsx` feeds the props directly, so it cannot see a broken parent. This matters doubly because the panel is the only keyboard/AT-reachable route to the metric. Add a case to `OperationGraphRebuildDiscipline.spec.tsx`: select a node via `emitNodeChanges`, assert the formatted duration and the swatch colour, then select an op absent from `PERF_ROWS` and assert the no-data label.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -425,8 +490,47 @@ const OperationGraphInner = ({ operationList, operationId }: OperationGraphReact
...
+    );
+
+    // Derived, so perf data arriving or going away can't leave a stored flag
+    // disagreeing with what the encoding can show.
+    const isPerfOverlayActive = isPerfOverlayEnabled && perfOverlay.status === PerfOverlayStatus.READY;
+
+    // Zoom goes straight to the DOM: it changes every wheel frame, and through
+    // state it would re-render the whole graph per frame to move one bar.
+    // `OpGraphEdge` can afford `useStore` only because it selects a boolean.
+    useEffect(() => {
```
dcblundell, line 509: **Must fix:** the zoom-publishing effect is entirely uncovered

Removing both the initial `writeZoom(...)` and the `flowStore.subscribe(...)` return keeps all overlay specs green, so the whole #1610 zoom compensation can become dead code without a failure. `opGraphPerfBarStyles.spec.ts` asserts the CSS *reads* `--op-graph-perf-zoom`; nothing asserts anything ever *writes* it. Blocked today by the stub at `OperationGraphRebuildDiscipline.spec.tsx:52` — `subscribe: () => () => {}` never invokes its callback. Give it a subscriber list and a `setZoom` helper and this unblocks in a few lines: initial write, one write per distinct zoom, and unsubscribe on deactivate.

src/scss/components/OperationGraphReactFlow.scss
```diff
@@ -52,6 +65,57 @@
         &.op-graph-node-output {
             background-color: var(--graph-output-node);
         }
+
+        // Magnitude goes on an inset bar: the fill belongs to the I/O highlight
+        // and the border to selection. Absolute so it cannot resize the node and
+        // turn every overlay toggle into a relayout. #1880
+        &::after {
```
dcblundell, line 84: **Must fix:** styleguide criterion is unimplemented and the deferral is untracked

#1880 criterion 7 and #1610 criterion 4 both ask for styleguide entries covering the new encoding, but `src/routes/Styleguide.tsx` is untouched by this branch — no example of the bar, and none of the unmapped no-bar state. The bar is the one part of this PR a reader cannot see without a live graph and a linked report, which is what the criterion exists to prevent. The PR lists this as a known limitation, but there is no TODO and no follow-up issue, so the deferral disappears on merge. Either add a static host with a few mock `.react-flow__node-opNode` elements at fixed `--op-graph-perf-scale` values plus one with neither property set, or open a tracking issue and link it. (Anchored here; the real gap is `src/routes/Styleguide.tsx:914`.)

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -425,8 +490,47 @@ const OperationGraphInner = ({ operationList, operationId }: OperationGraphReact
...
+        let lastZoom: number | null = null;
+        const writeZoom = (zoom: number) => {
+            if (zoom !== lastZoom) {
+                lastZoom = zoom;
+                container.style.setProperty(PERF_BAR_ZOOM_VAR, String(zoom));
+            }
+        };
+        writeZoom(flowStore.getState().transform[2]);
+        // No selector support in the vanilla store, so every change calls back
+        // and the guard above keeps it to one write per zoom change.
```
dcblundell: **Should fix:** quantise the zoom before writing it

`--op-graph-perf-zoom` is inherited and consumed by `min-width` and `height` of `.op-graph-node::after`, so every write invalidates style for the whole node subtree and dirties one pseudo-box per node. `onlyRenderVisibleElements` is not set, so N is the full report, and the `zoom !== lastZoom` guard does not help — d3-zoom emits a new scale nearly every frame. The jank lands at exactly the overview zooms this compensation exists for. One line fixes it: bucket with `Math.round(zoom * 10) / 10` before comparing, which keeps the bar floor within ~5% (imperceptible) and cuts writes from ~60/s to a handful per gesture. Same trick as `OpGraphEdge.tsx:22`, which selects a threshold crossing rather than live zoom for this reason.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -486,12 +613,49 @@ const OperationGraphInner = ({ operationList, operationId }: OperationGraphReact
         setSelectedOperationId(null);
     }, []);
 
+    // Box read on enter, not per mousemove: crossing a node boundary is orders
+    // of magnitude rarer than moving, and the read forces a layout flush.
+    const handleNodeMouseEnter = useCallback((event: ReactMouseEvent, node: OpGraphFlowNode) => {
```
dcblundell, line 649: **Should fix:** `getBoundingClientRect()` per mouse-enter forces a whole-graph layout

The comment correctly justifies this against per-mousemove, but sweeping a dense graph still crosses node boundaries many times a second, and each fire follows a commit that moved the tooltip — so layout is dirty and this read flushes it synchronously for the entire graph DOM, the one operation whose cost scales with node count. Cache the rect in a ref and refresh it from a `ResizeObserver` on the container plus window resize/scroll; the container's offset only changes on resize, scroll, and panel open/close.

src/components/operation-graph/opGraphPerfOverlay.ts
```diff
@@ -0,0 +1,144 @@
...
+// custom property leaves the node geometry, and so the Dagre layout, alone. #1880
+export const PERF_BAR_SCALE_VAR = '--op-graph-perf-scale';
+export const PERF_BAR_COLOR_VAR = '--op-graph-perf-color';
+
+// Sizes inside the scaled viewport are graph units, not screen pixels. The SCSS
+// divides by the live zoom to hold an on-screen floor, without which the
+// encoding washes out at exactly the zoom a whole report fits in. #1610
+export const PERF_BAR_ZOOM_VAR = '--op-graph-perf-zoom';
+
+export const NO_PERF_DATA_LABEL = 'No perf data';
```
dcblundell: **Should fix:** two sources of truth for one user-visible string

`PerfOverlayOpMetric.tsx:25` still hardcodes the same `'No perf data'` literal, so the hover chip and the panel row can drift. `OpGraphPanelPerfMetric.spec.tsx:73` asserts the component's output *via this constant*, which reads as coupled but is not — renaming the constant fails a test that points at the wrong file, and changing the component's literal fails it with no hint that the hover label is the other half. Per AGENTS.md a constant shared across modules belongs in `src/definitions/`, naturally beside `PERF_OVERLAY_TOOLTIP` in `PerfOverlayStatus.ts`; move it there and have the panel consume it.

tests/opGraphPerfBarStyles.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+});
+
+describe('perf overlay bar zoom floor', () => {
+    it('divides both dimensions by the live zoom so the bar holds its on-screen size', () => {
+        // Fixed graph units go sub-pixel at the zoom a large report fits in,
+        // losing the encoding exactly where a whole-graph scan needs it. #1610
+        const body = ruleBody(PERF_BAR);
+
+        expect(body).toMatch(/height:.*calc\(\s*\d+px\s*\/\s*var\(--op-graph-perf-zoom,\s*1\)\s*\)/);
```
dcblundell: **Should fix:** the TS↔SCSS custom-property contract is unpinned

The property names are re-typed as literals here even though `opGraphPerfOverlay.ts:21-27` exports `PERF_BAR_ZOOM_VAR` / `PERF_BAR_SCALE_VAR` / `PERF_BAR_COLOR_VAR` as the source of truth. This is the one seam the spec exists to protect: rename a constant and the bar silently stops painting while every assertion here still passes, because the regexes match SCSS text the JS no longer writes. Build the regexes from the imported constants, as `opGraphPerfOverlay.spec.ts` already does.

src/components/operation-graph/opGraphPerfOverlay.ts
```diff
@@ -0,0 +1,144 @@
...
+            status: PerfOverlayStatus.UNLINKED,
+            totalOpCount: graphOperationIds.length,
+        };
+    }
+
+    const { scoreByOpId, minNs, maxNs } = scoreOps(aggregatesByOpId);
+
+    const rankByOpId = new Map<number, number>();
+    let totalNs = 0;
+    const slowestFirst = Array.from(aggregatesByOpId.values()).sort((a, b) => b.deviceTimeNs - a.deviceTimeNs);
```
dcblundell: **Should fix:** rank sort has no tie-break, and diverges from the repo's other ranked surface

`pickTopN` in `src/functions/topNAnnotations.ts:46-51` sorts the same values but breaks ties on `opId` ascending, and its doc comment says why: determinism across renders. Without it, equal-duration ops here are ranked by perf-report row order, so the hover's `#N of M` can disagree with the Buffer Summary top-N over the same data and shuffle between report reloads. Smallest fix is the same `opId` ascending tie-break; better, lift the comparator into `src/functions/perfOverlay.ts` so both ranked surfaces share one ordering rule.

src/scss/components/OperationGraphReactFlow.scss
```diff
@@ -52,6 +65,57 @@
...
+            content: '';
+            position: absolute;
```
dcblundell, line 86: **Should fix:** the bar's containing block comes from a vendored rule nothing pins

`.react-flow__node-opNode` declares no `position`, so this resolves against `@xyflow/react`'s own `.react-flow__node { position: absolute }`. The suite asserts the bar's declarations in detail but nothing pins the invariant they all depend on: if that vendor rule ever changed, every bar would anchor to `.operation-graph-react-flow` (which *is* relative) and pile up in the canvas corner with a green suite. Add `position: relative` to the node rule and assert it alongside the geometry tests.

src/scss/components/OperationGraphReactFlow.scss
```diff
@@ -9,10 +9,23 @@
...
+
 .operation-graph-react-flow {
     position: relative;
     width: 100%;
-    height: 80vh;
+
+    // Fills the band between the navbar and the status bar, less `main`'s 15px
+    // top padding. The negative margin drops the 50px of trailing page padding,
+    // which here would only add a scrollbar.
+    height: calc(100vh - #{variables.$app-navbar-height} - 15px - #{variables.$app-status-bar-height});
```
dcblundell: **Should fix:** unrequested scope — the graph viewport height contract changed

`height: 80vh` becomes `calc(100vh - …)` with `margin-bottom: -50px`, arriving via f0a9ac80 / 460fd9f5. No criterion in #1880 or #1610 asks for this, and it is a user-visible change for the whole Graph view (different height on every viewport, plus a negative margin that eats the page's trailing padding). Either split it into its own PR, or call it out in the body as a deliberate second change — reviewers of a perf overlay should not implicitly approve a layout change.

src/scss/components/OperationGraphReactFlow.scss
```diff
@@ -9,10 +9,23 @@
 @use '../definitions/colours' as *;
 @use '../definitions/variables' as variables;
 
+// Bottom-left chrome — zoom controls, minimap, perf legend — on one baseline.
+// The origin and margin are React Flow's own.
+$chrome-dock-bottom: 8px;
+$panel-origin-left: 12px;
+$panel-margin: 15px;
+$minimap-margin-left: 56px;
+$minimap-width: 200px;
```
dcblundell, line 19: **Should fix:** promote the remaining magic layout numbers

This branch does the right thing promoting `$app-navbar-height` / `$app-status-bar-height`, then leaves four chrome numbers as private copies of values owned elsewhere: the `15px` at :27 is `main`'s padding (`_base.scss:58`), the `-50px` at :28 is `_base.scss:59`'s 30px plus `.data-padding`'s 20px (`_common.scss:222`) — a silent breakage waiting on any edit to either — `$panel-origin-left` duplicates `MLIRViewReactFlow.scss:34`, and `$minimap-width` here is React Flow's *default*: nothing in the repo sets it, so the legend's `left` depends on an undeclared upstream value and drifts on a library bump. For the minimap specifically, either set `width: $minimap-width` on `.react-flow__minimap` so this variable is authoritative, or lay the bottom-left chrome out with flex so the legend follows the minimap's real box.

src/scss/components/OperationGraphReactFlow.scss
```diff
@@ -107,6 +171,13 @@
         }
     }
 
+    // Outranks MLIRViewReactFlow.scss's bare-`.react-flow` rule, which lifts
+    // bottom-left panels 16px app-wide. Specificity, not stylesheet order.
+    .react-flow .react-flow__panel.bottom.left {
```
dcblundell, line 191: **Should fix:** this wins a specificity race against another view's stylesheet

`MLIRViewReactFlow.scss` nests `.react-flow__panel.bottom.left` under a bare top-level `.react-flow`, so it applies app-wide — that unscoped block is the root cause, and out-specifying it from a second component sheet entrenches the leak rather than fixing it. Scoping the MLIR block under `.mlir-page` (or an app-owned wrapper) removes the need for this rule entirely. If that is out of scope here, the comment should say the cause is an unscoped stylesheet rather than only describing the mechanism.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -546,12 +719,32 @@ const OperationGraphInner = ({ operationList, operationId }: OperationGraphReact
...
+                    <PerfOverlayLegend
+                        minNs={perfOverlay.minNs}
+                        maxNs={perfOverlay.maxNs}
+                    />
+                </div>
+            ) : null}
+            {perfHoverLabel !== null && perfHover !== null ? (
+                <div
+                    className='op-graph-perf-hover'
+                    style={{ left: perfHover.x, top: perfHover.y }}
```
dcblundell, line 768: **Should fix:** `role='tooltip'` with nothing referencing it via `aria-describedby`

An ARIA role without the relationship it implies gives assistive tech no path to the content, and the trigger is pointer-only — `onNodeMouseEnter` with no focus equivalent. That leaves the panel's Kernel duration row as the only AT-reachable route to perf data, which is the wiring proven untested elsewhere in this review. Either wire `aria-describedby` from the node plus a focus trigger, or drop the role and rely on the panel.

src/components/operation-graph/opGraphPerfOverlay.ts
```diff
@@ -0,0 +1,144 @@
...
+    overlay: OpGraphPerfOverlay,
+    isActive: boolean,
+): Map<string, CSSProperties> | null => {
+    if (!isActive) {
+        return null;
+    }
+    const styleByNodeId = new Map<string, CSSProperties>();
+    for (const [opId, score] of overlay.scoreByOpId) {
+        // `CSSProperties` has no index signature for custom properties.
+        styleByNodeId.set(String(opId), {
```
dcblundell, line 149: **Should fix:** `isDarkPerfColor` is now dead in production

With the colour written to a custom property here rather than onto the label surface, `src/functions/perfOverlay.ts:220` has no production caller — only `tests/perfOverlay.spec.ts` imports it. #1880's first comment predicted exactly this ("It disappears entirely once the colour moves off the label surface"), so deleting it is the evidence the encoding change worked. Remove the helper and its spec block. Worth doing in the same pass: the `PERF_BINS` comment at `GraphColors.ts:22-27` ("Colour-only encoding for now: vis-network's `size` option is ignored…") is now false on both counts.

tests/OperationGraphRebuildDiscipline.spec.tsx
```diff
@@ -365,3 +396,121 @@ describe('OperationGraphReactFlow filter dimming', () => {
         expect(nodeById(after, '5')).toBe(nodeById(before, '5'));
     });
 });
+
+describe('OperationGraphReactFlow perf overlay identity', () => {
```
dcblundell, line 461: **Should fix:** remaining coverage gaps in the new overlay surfaces

Each of these mutations survives the full suite: the legend mount and its `minNs`/`maxNs` (replacing the condition with `false` stays green, and `PerfOverlayLegend` has no spec anywhere in `tests/`); the whole hover path including its `isPerfOverlayActive` gating (attaching `onNodeMouseEnter` unconditionally stays green); both `setPerfHover(null)` guards, whose comments describe the bug they prevent; and the ordinary toggle-*off* path, which no test exercises — including whether node identity returns to the raw builder objects, the property this file exists to defend. Separately, `OpGraphPerfOverlaySwitch.spec.tsx:37` hardcodes `isDisabled={false}`, so the mid-build disable reason is never asserted — one extra row on the existing `it.each`.

src/routes/GraphView.tsx
aidemsined: Fixed in 276ae945 — route-level spec at `tests/GraphViewPerfReportLoaded.spec.tsx`, with the three states you named:

- no report selected + query resolved to `{ report: [] }` → false
- report selected + still in flight → false
- report selected + `{ report: [] }` → true

The flag never reaches the DOM, so the spec stubs `OperationGraphReactFlow` and asserts on the props the route actually handed it. Reverting the line to `perfReport !== undefined` fails the first case; dropping the other half fails the third.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in 276ae945. Two cases in `OperationGraphRebuildDiscipline.spec.tsx` under `perf overlay panel wiring`:

- `hands the selected op its duration and the colour the graph drew it with` — selects through `emitNodeChanges`, then asserts the formatted duration and the swatch colour, read back off the styled node so a divergence between the graph's ramp and the panel's swatch fails too
- `reads no data for an op the perf report never mentioned`

Hard-coding either prop back to `undefined` now fails both.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in 276ae945, and you were right that the stub was the whole obstacle. `subscribe` keeps a subscriber list now and the harness exposes a `setZoom` helper, which made five cases cheap (`perf bar zoom compensation`):

- `publishes the current zoom when the overlay turns on` — fails if the initial `writeZoom(...)` goes
- `republishes when a gesture crosses a bucket`
- `coalesces sub-bucket frames into a single write`
- `never publishes a zero, which would invalidate the bar geometry`
- `unsubscribes when the overlay is switched off` — fails if the `subscribe` return is dropped

src/scss/components/OperationGraphReactFlow.scss
aidemsined: Implemented rather than tracked. `Styleguide.tsx` has a `Node bar — the score ramp on the graph` section in 276ae945: four mock `.react-flow__node-opNode` elements at fixed `--op-graph-perf-scale` / `--op-graph-perf-color` values, plus a fifth that sets neither, so the unmapped no-bar state sits beside the ramp instead of being described in prose.

Two things the preview needed, both noted in comments: the page imports `OperationGraphReactFlow.scss` (the bar is a pseudo-element of the graph's node rule, and nothing else on the styleguide pulls that sheet in), and the host compounds `.operation-graph-react-flow` with a styleguide class to neutralise the view's viewport-filling height and trailing negative margin. No follow-up issue needed as a result.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in 276ae945, with one deviation from the suggested `Math.round(zoom * 10) / 10`: this view sets `minZoom={0.02}` (`OperationGraphReactFlow.tsx:72`), and absolute tenths round the entire overview range to `0`. `calc(2px / 0)` is invalid, so the declaration drops and the bar disappears at exactly the zooms the compensation exists for.

`getQuantisedPerfZoom` buckets proportionally instead — `Math.round(log(zoom) / log(1.1))` — which is uniform across the whole 0.02–3 range, holds the bar floor within ~5% of true, and returns `1` (the SCSS's own fallback) for non-finite or non-positive input. `never publishes a zero, which would invalidate the bar geometry` pins the degenerate case, `coalesces sub-bucket frames into a single write` pins the reduction in writes.

Thanks for the `OpGraphEdge.tsx:22` pointer — same reasoning, different quantiser because that one only needs a single threshold.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in 276ae945, as suggested: the box is read into `containerBoundsRef` by an effect gated on `isPerfOverlayActive`, refreshed by a `ResizeObserver` on the container plus `resize` and capture-phase `scroll` — capture because an ancestor scrolling moves the box without scrolling the window.

`handleNodeMouseEnter` keeps `?? containerRef.current?.getBoundingClientRect()` as a fallback for environments without `ResizeObserver`, matching how `ChipCongestionCanvas` degrades.

src/components/operation-graph/opGraphPerfOverlay.ts
aidemsined: Fixed in 276ae945 — `NO_PERF_DATA_LABEL` now lives in `src/definitions/PerfOverlayStatus.ts` beside `PERF_OVERLAY_TOOLTIP`, `PerfOverlayOpMetric.tsx` consumes it instead of its own literal, and `opGraphPerfOverlay.ts` re-imports rather than re-declares. The coupling `OpGraphPanelPerfMetric.spec.tsx` implied is real now, so changing either surface's copy is no longer possible without changing both.

tests/opGraphPerfBarStyles.spec.ts
aidemsined: Fixed in 276ae945 — the regexes are built from the imported `PERF_BAR_ZOOM_VAR` / `PERF_BAR_SCALE_VAR` / `PERF_BAR_COLOR_VAR` through a `varRef(name, fallback)` helper, so a rename fails here instead of passing against SCSS text the TS no longer writes. The same file also picked up the containing-block assertion from the [`position: relative` thread](https://github.com/tenstorrent/ttnn-visualizer/pull/1890#discussion_r3815042445).

src/components/operation-graph/opGraphPerfOverlay.ts
aidemsined: Took the second option in 276ae945: `getRankComparator(valueOf)` lives in `src/functions/perfOverlay.ts` and is now the only ordering rule for both ranked surfaces — `buildOpGraphPerfOverlay`'s `slowestFirst` and `pickTopN` in `topNAnnotations.ts`, whose local `compareByOpIdAsc` is deleted.

Covered by a `getRankComparator` block in `perfOverlay.spec.ts`, including `ranks equal values identically regardless of input order`, which is the property perf-report row order was breaking.

src/scss/components/OperationGraphReactFlow.scss
aidemsined: Fixed in 276ae945 — `.react-flow__node-opNode` declares `position: relative`, with a comment naming the vendor rule it used to lean on, and `opGraphPerfBarStyles.spec.ts` asserts it (`anchors the bar to the node it annotates`). Removing the declaration now fails a test rather than surfacing as a corner pile-up in a live graph.

src/scss/components/OperationGraphReactFlow.scss
aidemsined: Fair — it is a second change and it was not called out. The PR body names it explicitly now: what changes (`80vh` → the band between navbar and status bar, less `main`'s top padding), why the negative margin is there, and that no criterion in #1880 or #1610 asks for it, so it gets reviewed on its own merits rather than implicitly.

276ae945 also removes the part that made it fragile. The literals were private copies of `_base.scss`'s padding and `.data-padding`'s gutter; they are the shared `$main-padding` / `$main-padding-bottom` / `$data-padding-bottom` now, so an edit to either propagates instead of silently desyncing.

Happy to lift both hunks into their own PR if you would rather not carry the layout change here — say the word.

src/scss/components/OperationGraphReactFlow.scss
aidemsined: Fixed in 276ae945. All four come from `_variables.scss` now: `$main-padding`, `$main-padding-bottom` and `$data-padding-bottom`, which `_base.scss` and `_common.scss` consume as well so the copies are gone in both directions, and `$reactflow-panel-origin-left`, which `MLIRViewReactFlow.scss` also uses instead of its own `12px`.

For the minimap I took the first option: `.react-flow__minimap` declares `width: $minimap-width`, so the legend's offset depends on a width we own and a library bump cannot move it. Verified the compiled CSS is byte-identical before and after the promotion.

src/scss/components/OperationGraphReactFlow.scss
aidemsined: Took the second option in 276ae945 — the comment names the cause now (`MLIRViewReactFlow.scss` nests its panel offsets under a bare top-level `.react-flow`, so they leak into every graph in the app) and says out-specifying it is containment rather than the fix.

#1899 tracks scoping that block under an app-owned wrapper, which removes this rule entirely. Keeping it out of this PR because it changes MLIR view layout and wants its own verification pass.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Dropped the role in 276ae945 — the chip is `aria-hidden='true'` with a comment saying why: the trigger is pointer-only, and the panel's Kernel duration row carries the same figure on a keyboard-reachable path, so advertising a relationship no `aria-describedby` provides was worse than staying decorative.

That leans on the panel wiring, which is no longer the untested half — see the two cases in the [panel props thread](https://github.com/tenstorrent/ttnn-visualizer/pull/1890#discussion_r3815042384). A focusable node plus `aria-describedby` is the better end state, but it needs keyboard navigation for the graph, which this PR does not add.

src/components/operation-graph/opGraphPerfOverlay.ts
aidemsined: Deleted in 276ae945 — the helper, `DARK_BG_LUMINANCE_THRESHOLD`, the three BT.601 weights and the spec block.

The `PERF_BINS` comment is rewritten too: it says the ramp is one of two dimensions on an inset bar whose width carries the same score, instead of the two claims about vis-network's `size` option that stopped being true when the encoding moved.

tests/OperationGraphRebuildDiscipline.spec.tsx
aidemsined: All fixed in 276ae945, in `OperationGraphRebuildDiscipline.spec.tsx` under `perf overlay chrome`:

- `mounts the legend with the linked ops bounds` — `PerfOverlayLegend`'s first coverage anywhere, asserting `minNs`/`maxNs` rather than just the mount
- `attaches the hover handler only while the overlay is on`
- `shows the hovered op its duration and rank`, `clears the hover when the pointer leaves`, `drops a hover the overlay left behind when it is switched off` — the two `setPerfHover(null)` guards, one case each
- `returns the nodes to the builder objects when the overlay goes off` — node identity on the toggle-off path, the property this file exists to defend

`OpGraphPerfOverlaySwitch.spec.tsx` takes `isDisabled` as a parameter now, with the extra `it.each` row: `READY` + disabled (`a graph still being laid out`), so the mid-build reason is asserted instead of assumed.

src/scss/components/OperationGraphReactFlow.scss
aidemsined: Correction to my reply above: `position: relative` was the wrong value, and it broke the graph. React Flow positions nodes out of flow and offsets them from the pane origin, so taking them back into document flow left the nodes stacking in DOM order while the edge SVG kept drawing at the store's coordinates — every arrow detached from its node. Confirmed in the browser: nodes computed to `relative` with a 172px flow-stacked stride against the layout's real 126px.

ef3763f7 uses `absolute` instead, which establishes the containing block just as well and is the value the vendor rule was already supplying — declaring it was only ever meant to stop depending on someone else for it, not to change it.

Worth noting for this thread's original point: the assertion you asked for did not catch this, because `relative` and `absolute` both satisfy "has a containing block" and jsdom computes no layout, so every geometry test around it passed against a graph that rendered wrong. It now pins the value, with a second case naming `relative` as the specific regression.

PR #1892, opened 2026-08-18
Validate logging content

### LINKED ISSUE #1829: Usage signals: validated local ingest endpoint for frontend events
Sub-issue of #1819. Depends on the event log writer.

Caveat 18 settles that recording happens frontend-side: React Query caching, prefetching and retries inflate backend API counts, and the interactions we most care about — chart views, table toggles, filters, playback — never reach the API at all. That means the frontend needs somewhere local to post to, which is the one piece of network surface the "no network calls" framing in #1819 does not cover. It is also the most security-sensitive part of this work, which is why it is split out for its own review.

## Why this needs careful validation

Nothing in a local install is authenticated. `ALLOWED_ORIGINS` decides which pages may call us at all, and it is the only gate (see the trust boundaries section of `AGENTS.md`). An endpoint that trusts its request body therefore lets any permitted page write arbitrary lines into the exact file we are asking IT to parse — log injection into a privacy-reviewed artefact whose bounded contents are the entire promise being made. Client-side enum validation protects nothing here; the enums have to be enforced in the handler.

## Scope

- `POST /api/usage` in [`views.py`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/backend/ttnn_visualizer/views.py), decorator order `@api.route → @local_only` per AGENTS.md.
- **No `@with_instance`.** The log is machine-scoped, not report-scoped, so this route takes no `instanceId` query param — a deliberate exception to the report-backed route convention, worth a comment saying why.
- `@local_only` returns 403 under `SERVER_MODE`, which is exactly the hosted-instance behaviour caveat 2 asks for. The recording-disabled check in the writer is then belt and braces rather than the only guard.
- Accept a small batch of events per request, since the client buffers. Cap the batch size and the request size.
- **Reject rather than coerce.** Unknown event name, unknown detail key, out-of-enum detail value, a value containing a newline or `=` (which would let one event forge extra log lines), or a batch over the cap — all rejected. Nothing free-form reaches the file.
- The server supplies `ts` and `schema_version`; the client does not get to set them.
- **`app_start` is server-owned and cannot be posted by a client.** It is a Tier 1 event, so a literal reading of the out-of-scope note below would permit it — but the server records launches itself in `main()`, and a client able to forge one could distort the `deployment_mode` population that every other figure is read against (caveat 4 already warns that launches overcount). The handler refuses it alongside unknown event names. The client-postable set is therefore exactly `report_loaded`, `report_load_failed`, `view_opened`, `view_engaged`.
- Errors through `response_bad_request` / `response_unprocessable_entity` from `exceptions.py`. No hand-rolled `jsonify`.
- **The request-size cap is the one exception: it answers 413.** Setting `request.max_content_length` per request (Flask ≥ 3.1) lets Werkzeug refuse the body *before it is read*, which a manual `request.content_length` check cannot do for a chunked body — and `MAX_CONTENT_LENGTH` defaults to no limit at all, so nothing is inherited. The refusal therefore arrives as `RequestEntityTooLarge`. The app's `HTTPException` handler still renders it as `{"error": …}`, so the response shape is unchanged; only the status differs.
- **Require the JSON content type; do not pass `force=True` to `get_json`.** This is part of the threat model rather than pedantry: an `application/json` POST is not a CORS-simple request, so a hostile origin cannot make one without a preflight `ALLOWED_ORIGINS` refuses, whereas a `text/plain` body would sail straight through the only gate this deployment has. It follows that the client's beacon flush has to send a typed `Blob` — see #1830.

## Tests

`backend/ttnn_visualizer/tests/test_usage_endpoint.py`, using the shared `client` fixture:

- 403 under `SERVER_MODE`
- rejection of an unknown event name, an unknown detail key, an out-of-enum detail value, an embedded newline, and an oversized batch. For the out-of-enum case use a value that could never become valid (`not_a_view`) — **not** a deferred-but-planned one such as a view awaiting its page, which would make a future enum addition fail a test that has nothing to do with it. If a deferral is worth pinning, give it its own named test whose failure means exactly what it says.
- an accepted batch appends exactly N well-formed lines
- a rejected batch appends nothing at all — partial acceptance of a mixed batch must not leave half the events written

## Out of scope

The frontend client. Which events exist beyond the Tier 1 and Tier 2 set.

## Verification

`pnpm flask:test`, `pnpm flask:lint`, `pnpm flask:mypy`.

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1829.

This pull request introduces a comprehensive suite of tests for the usage logging and ingestion endpoint, and refactors test utilities for better reuse. It also expands the usage event schema with new event types and enums, enforces stricter validation on client-supplied usage events, and improves documentation around event timestamps. The changes ensure robust validation, atomicity of event batch writes, and security against malformed or malicious client data.

**Testing and Validation Enhancements:**

* Added `test_usage_endpoint.py`, a new test module covering the `/api/usage` endpoint, including validation of event batches, boundary conditions (batch size, body size), and security checks to prevent partial writes or leaking client data in error responses.
* Added comprehensive tests in `test_usage.py` to guarantee every client-postable event has a validation rule, all detail fields are backed by enums, no field collisions exist between client and server, and that batch writes are atomic (all-or-nothing).
* Refactored test utilities (`usage_directory`, `read_lines`, `parse`, `total_events`) into `conftest.py` for reuse across test modules, and removed their local definitions from `test_usage.py`. [[1]](diffhunk://#diff-5c0e6008ed590e8dc8f4b4ae5d59c46d2e1948bba1b6c2cbe6dcbd937782a736R15-R24) [[2]](diffhunk://#diff-5c0e6008ed590e8dc8f4b4ae5d59c46d2e1948bba1b6c2cbe6dcbd937782a736R59-R92) [[3]](diffhunk://#diff-5b2042663a71f72c1d56fe7981a739bcf2d0df07dbe17c9b1af424240c2e3c53L44-R50)

**Usage Event Schema Expansion:**

* Added new enums and event types to `usage.py`: `UsageEvent` now includes `REPORT_LOADED`, `REPORT_LOAD_FAILED`, `VIEW_OPENED`, and `VIEW_ENGAGED`. Introduced `ReportKind`, `ReportSource`, `ReportLoadFailureReason`, and `UsageView` enums, and defined which fields are valid for each event type via `CLIENT_EVENT_DETAIL_FIELDS` and `_DETAIL_FIELD_ENUMS`. [[1]](diffhunk://#diff-bec298e1068fbb1826513e0899e09be594dfe4de9929b621bc0a79156578fa61R122-R125) [[2]](diffhunk://#diff-bec298e1068fbb1826513e0899e09be594dfe4de9929b621bc0a79156578fa61R141-R234)

**Documentation and Code Quality:**

* Improved documentation in `usage.py` to clarify that event timestamps (`ts`) are assigned on the server at batch receipt, not by the client, and that all values must come from closed enums to prevent injection or data leakage.

**Type and Import Improvements:**

* Updated type hints for clarity and correctness, e.g., using `Mapping` and `Type` for enum mappings in `usage.py`.

These changes significantly strengthen the reliability, security, and maintainability of usage event collection and validation.


backend/ttnn_visualizer/tests/usage_writer.py
```diff
@@ -34,8 +46,13 @@ def main(argv: list[str]) -> int:
...
 
-    for _ in range(count):
-        record_event(UsageEvent.APP_START)
+    if batch_size > 1:
+        batch = [(UsageEvent.VIEW_OPENED, {"view": UsageView.OPERATIONS})] * batch_size
+        for _ in range(count // batch_size):
+            record_events(batch)
+    else:
+        for _ in range(count):
+            record_event(UsageEvent.APP_START)
```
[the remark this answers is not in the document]
dcblundell: Fixed — the writer now finishes with a short final batch, so it writes exactly `count` events for any `batch_size`, including `batch_size > count`.

One correction to the report: it wasn't silent. The post-write `len(written.splitlines()) < count` check caught the shortfall and exited non-zero. I verified by reverting the fix, and the test fails with `wrote fewer than 10 lines`. The real cost was a *confusing* failure rather than a hidden one — a test about cross-process interleaving failing because of arithmetic in the helper, which is a bad place to send the next person debugging a flake.

Also took the parenthetical suggestion, from the other end: the concurrency test now runs `writes_each=10, batch_size=3`, so the remainder path is genuinely exercised rather than the fix being theoretical.

backend/ttnn_visualizer/usage.py
```diff
@@ -314,32 +472,97 @@ def _append_line(line: str) -> None:
...
+
+    if _bytes_since_size_check < LOG_SIZE_CHECK_INTERVAL_BYTES:
+        return False
+
+    _bytes_since_size_check = 0
+
     try:
-        if server_mode is None:
-            server_mode = _server_mode_from_app_context() or False
+        return get_usage_log_path().stat().st_size > MAX_LOG_BYTES
```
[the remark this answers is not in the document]
dcblundell: Good catch on the overclaim, but `>=` would introduce a worse problem, so I've fixed the docstring instead and kept the comparison.

`compact_if_needed` skips when `st_size <= MAX_LOG_BYTES` (`usage.py:802`) — it compacts only when the log is *over* the cap. `_log_is_full`'s `>` matches that deliberately, and the two have to agree. With `>=` here, a log sitting exactly on the cap would refuse every append while launch compaction still skipped it as not-yet-over, leaving it stuck refusing indefinitely. That is strictly worse than one batch of overshoot.

The boundary also isn't where the precision is lost. The check is amortised over `LOG_SIZE_CHECK_INTERVAL_BYTES`, so the log can already be up to 256 KB past the cap before `_log_is_full` runs at all — `>=` would not have made "stops growing at `MAX_LOG_BYTES`" true either.

So the defect was my wording, not the operator. The module docstring now reads "stops growing shortly past `MAX_LOG_BYTES`" and names the amortisation as the bound, and `_log_is_full` carries a comment explaining why it is `>` and not `>=`, so this doesn't get raised again.

backend/ttnn_visualizer/usage.py
```diff
@@ -84,11 +97,27 @@
...
+# amortise instead, at ~2,400 events between checks. Once a check finds the log over the
+# cap that verdict is cached rather than re-derived (see :func:`_is_log_full`), so the
+# overshoot is this interval plus one batch — not this interval per refused batch.
+LOG_SIZE_CHECK_INTERVAL_BYTES = 256 * 1024
+
+# The largest batch one request may carry, which is also the largest number of lines one
+# ``os.write`` has to hold. It lives here rather than beside the ingest route's byte cap
+# because what it bounds is ``_append_line``'s atomicity, not merely an HTTP body: a
+# second batch caller would otherwise get no cap at all, and raising it in ``views.py``
+# would silently relax a guarantee documented here.
```
aidemsined: > [!WARNING]
> This comment claims a guarantee the module doesn't make — the cap is only enforced in `views.py`.

The reasoning for placing the constant here is that what it bounds is `_append_line`'s atomicity rather than merely an HTTP body, and that "a second batch caller would otherwise get no cap at all". But neither `_write_events` (:752) nor `record_events` (:884) checks it — the only enforcement is `views.py:2619`. So a second batch caller today gets exactly the no-cap situation the comment says the placement prevents.

It matters because `_append_line`'s docstring at :612 defers its single-`os.write` obligation to this bound. `record_events([event] * 5000)` builds a ~550 KB buffer and hands it to a write loop that will iterate, forfeiting the `O_APPEND` atomicity the whole no-lock design rests on. Nothing does that today, but the module is written as though it were guarded.

**Suggested action:** add the check in `_write_events` (returning `False`, consistent with its other refusals), or rewrite this comment to say it's a route-level cap that `usage.py` merely names, so the two docstrings stop implying otherwise.

backend/ttnn_visualizer/tests/test_usage_endpoint.py
```diff
@@ -0,0 +1,607 @@
...
+
+ACCEPTED_EVENTS = {
+    UsageEvent.REPORT_LOADED: {"kind": "profiler", "source": "upload"},
+    UsageEvent.REPORT_LOAD_FAILED: {
+        "kind": "performance",
+        "reason_class": "parse_error",
+    },
+    UsageEvent.VIEW_OPENED: {"view": "operations"},
+    UsageEvent.VIEW_ENGAGED: {"view": "performance"},
+}
```
aidemsined, line 223: > [!WARNING]
> `ACCEPTED_EVENTS` restates the client-postable set by hand, so a fifth event would silently get no round-trip coverage.

`test_every_client_postable_event_round_trips` parametrises over `list(ACCEPTED_EVENTS)`, so this dict is what decides which events are exercised end to end. Add an entry to `CLIENT_EVENT_DETAIL_FIELDS` and nothing here fails — the new event simply isn't tested, and the suite stays green while coverage silently narrows.

The two meta-tests in `test_usage.py` do derive correctly from the handler's tables, so this isn't a systemic problem — it's this one fixture. `CLIENT_EVENT_DETAIL_FIELDS` is already imported at `:26`, so one line closes it:

```python
assert set(ACCEPTED_EVENTS) == set(CLIENT_EVENT_DETAIL_FIELDS)
```

Related, and worth a thought rather than a change: `test_every_client_postable_event_has_a_validation_rule` pins the *rule* (`set(CLIENT_EVENT_DETAIL_FIELDS) == set(UsageEvent) - {APP_START}`) rather than the literal set #1829 names. That's the more useful invariant to hold — but it means widening the client-postable set to a fifth event passes silently everywhere, and AC7's "exactly `report_loaded`, `report_load_failed`, `view_opened`, `view_engaged`" isn't pinned anywhere. Pinning the four names once would make widening a deliberate, visible edit.

**Suggested action:** add the set-equality assertion here.

backend/ttnn_visualizer/usage.py
```diff
@@ -414,38 +642,274 @@ def _append_line(line: str) -> None:
             if written == 0:
                 raise OSError("usage log write returned 0 bytes")
             offset += written
```
aidemsined, line 644: > [!WARNING]
> A short write followed by `OSError` leaves a torn line on disk while the caller is told nothing was written.

If `os.write` returns a short count and the next iteration raises, the bytes already written stay in the file, but `record_events` returns `False`. The comment just above at :636-637 acknowledges exactly this shape — "Leaving a partial line would look identical to an NFS interleave" — so the residue is understood. The problem is that the docstrings at :756 (`_write_events`) and :888 (`record_events`) both state all-or-nothing unconditionally, and the endpoint's atomicity tests reinforce that reading.

The gap is narrow: `test_a_failed_write_does_not_fail_the_request` monkeypatches `_append_line` wholesale, so this loop is never exercised under failure. A genuine short write followed by a failure isn't deterministically reproducible, which is why I'd treat this as a documentation fix rather than asking for a test.

**Suggested action:** qualify the two docstrings — something like "all-or-nothing except under a short write that then fails, which is indistinguishable from an interleave and is quarantined by `_summarise`". That keeps the guarantee honest without pretending to a stronger one.

backend/ttnn_visualizer/views.py
```diff
@@ -2542,3 +2563,80 @@ def get_latest_version():
...
+    ``@local_only`` is the control that matters. Nothing here is authenticated and
+    ``ALLOWED_ORIGINS`` is the only other gate, so the handler validates its body
+    against a closed schema rather than trusting it: a permitted page must not be able
+    to write arbitrary lines into the file we are asking IT to parse.
+    """
+    # Before anything reads the stream. Werkzeug enforces this both against a declared
+    # `Content-Length` and while reading a stream the server has terminated, which a
+    # manual `request.content_length` check would miss for chunked bodies. The resulting
+    # `RequestEntityTooLarge` renders as a 413 through the app's `HTTPException` handler.
+    request.max_content_length = MAX_USAGE_REQUEST_BYTES
```
aidemsined, line 2597: > [!WARNING]
> This line carries an invisible Flask 3.1 dependency.

Assigning `request.max_content_length` per request is a Flask 3.1 addition — on 3.0 the attribute is read-only and the assignment raises `AttributeError`, so every request to this route would 500 rather than degrading to an uncapped body.

`pyproject.toml:15` pins `Flask>=3.1.0,<4.0`, so this is correct today and I'm not suggesting a code change. But the coupling is only expressed in the pin, and the failure mode if someone relaxes it is loud-but-confusing (a 500 on the ingest route, nothing pointing at Flask) rather than a quiet loss of the cap.

**Suggested action:** one clause in the existing comment naming the 3.1 requirement, so the pin and this line stay connected.

aidemsined (commented):
I went at this as a security review first, and tried to defeat the validator rather than confirm it exists. **I could not get a log-injection payload through `/api/usage`**, and the reason is structural rather than lucky: `validate_client_event` doesn't sanitise the client's string and pass it on — `usage.py:875` does `resolved[field] = _DETAIL_FIELD_ENUMS[field](value)`, and `_detail_value` then writes `member.value`. The object written is never the object received.

That kills the whole validate-then-mutate class, and it also defeats the cleverer variants — a `str` subclass with a lying `__eq__`, a value differing only in Unicode normalisation form, anything that survives a membership test but changes on the way to the file. Newline, `\r`, `=`, space, NUL and U+2028/U+2029 all fail at the enum long before `_format_line` sees them. All nine acceptance criteria in #1829 are met.

One finding I can't anchor, because the line is untouched by this diff:

## Unanchored findings

**`_SAFE_VALUE_PATTERN` accepts a trailing newline** — `usage.py:154` is `re.compile(r"^[A-Za-z0-9._:+-]+$")`, used with `.match()` at `:539`. Python's `$` matches *before* a trailing newline, so `_is_safe_value("forged\n")` returns `True`. Confirmed directly:

```
'abc'     -> True
'abc\n'   -> True     <-- the hole
'abc\r'   -> False
'abc\n\n' -> False
'abc\nx'  -> False
```

This is **not client-reachable** — the enum re-materialisation above stops every posted value well short of here. The reachable vectors are `get_run_id()` (`usage.py:487-489`) via `TTNN_VISUALIZER_RUN_ID=$'forged\n'`, and `get_application_version()` via a version string ending in a newline. The impact is one truncated event plus one orphan line; `_summarise` correctly quarantines the orphan as unparseable, but the line count the collector reads is then wrong. A *full* forged event isn't achievable, because injecting `ts=` after the newline needs an `=`, which the character class does reject, and `$` permits only one newline and only at the very end.

What makes it worth fixing anyway is that this is the exact function whose entire stated job is preventing this, and `get_run_id`'s own comment says "an unchecked one carrying a newline would forge log lines" immediately before calling the guard that misses it. `test_every_enum_value_is_safe_to_write_unquoted` inherits the weakness too. The fix is one token — `re.fullmatch`, or `\Z` in place of `$` — and adding `"trailing\n"` to the parametrisation at `test_usage.py:430` gives it a test that fails today.

The line predates this PR, so this is scope-adjacent rather than a regression. But this PR is what makes the function load-bearing.

## General observations

> [!NOTE]

- The top-level payload isn't closed the way the event envelope is. `views.py:2614` does `payload.get("events")` and ignores everything else, while `usage.py:830` closes the envelope with the reasoning that an ignored key "lets a client believe it is sending a field that is being dropped". Same argument, one level up, not applied. Nothing is written either way, so this is consistency rather than security — and note `test_oversized_body_is_refused_before_it_is_parsed` currently relies on a stray `padding` key being tolerated, so closing it needs that test moved to padding inside a valid structure.
- `test_a_full_batch_of_the_largest_event_is_still_one_write` overclaims slightly: the `data.startswith(b"ts=")` filter can't observe a continuation chunk, so the docstring's "exercises where the short-write loop would begin to iterate" isn't what the assertion sees. The per-event-loop claim it also makes is valid and is the one worth keeping.
- The body cap has only its rejecting boundary pinned — a body at exactly `MAX_USAGE_REQUEST_BYTES` should be accepted. The batch cap has both sides (50 accepted, 51 refused), which is what makes it convincing.
- `usage.py:299-302` says `CLIENT_EVENT_DETAIL_FIELDS` is "Exported so the frontend enums and the docs page can be derived from one source", but nothing derives from it yet — no schema route, no codegen, no usage page under `docs/`. Fine as forward intent; worth phrasing as intent so a later reader doesn't assume the derivation already exists.
- No test asserts a disallowed `Origin` is actually refused. Framework-owned, so low value, but the `application/json` requirement at `views.py:2603-2608` reasons about preflight and it'd be cheap to pin.
- Ordering of the guards is right and worth noting: the batch-size check runs before the validation loop (a 51-event batch costs one `len()`, not 51 validations), and the recording-disabled check runs before `get_json`, so a disabled install pays no JSON parse. `test_a_disabled_install_answers_before_it_validates` pins the second one.

A few things I checked that came back clean, so nobody re-treads them: `app_start` can't be forged by case variation, whitespace or alias (they all fail at `UsageEvent(name)`, since str-Enum value lookup is exact); client-supplied `ts`/`schema_version` are **rejected** rather than silently dropped, at both the envelope and detail levels, so there's no contradiction with reject-not-coerce; the regex character class is ASCII-only under `re`; and `_record_launch` is the only other write path, with no frontend client for this route existing yet.

Not run from here: `pnpm flask:test`, `flask:lint`, `flask:mypy`. Two worth watching — the 413 assertion is framework-behaviour-dependent, and mypy on the `Dict[str, Enum]` → `Sequence[Tuple[UsageEvent, Mapping[str, Any]]]` flow is a variance question a machine should answer.

Separately: nice to see `_WRITER_SCRIPT` landed as a real module. It picks up lint, type-checking and formatting now.

---

backend/ttnn_visualizer/usage.py
dcblundell: Agreed, and took the code fix rather than the comment fix — the guarantee at `_append_line`:612 is worth keeping true rather than downgrading to "a route-level cap `usage.py` merely names".

`_write_events` now refuses a batch over `MAX_USAGE_BATCH_EVENTS`, returning `False` alongside its other refusals, so the bound binds every batch caller instead of the one that happens to check first. The constant's comment now says where enforcement lives rather than implying placement alone achieves it. Pinned by `test_a_batch_over_the_cap_is_refused_by_the_writer`.

No behavioural change to the app: the route still rejects with a 400 before the writer sees the batch.

Fixed in 43568a62.

backend/ttnn_visualizer/tests/test_usage_endpoint.py
dcblundell: Agreed — added the set-equality assertion, as its own test with a docstring explaining why the fixture needs pinning at all.

On the related point, which I agree is the more interesting one: left as-is deliberately, so it stays visible rather than half-addressed. `test_every_client_postable_event_has_a_validation_rule` still pins the rule rather than AC7's four literal names, so widening the client-postable set to a fifth event does still pass everywhere. Worth a deliberate decision rather than folding into a review-comment pass.

Fixed in 43568a62.

backend/ttnn_visualizer/usage.py
dcblundell: Agreed, and agreed on documentation over a test — the failure needs a short write *followed by* a raise, which isn't deterministically reproducible, and the residue is genuinely indistinguishable from the interleave `_summarise` already quarantines.

Both docstrings now qualify the guarantee: `_write_events` cross-references `_append_line`, and `record_events` spells out that the bytes already written stay on disk while it returns `False`, costing a skipped line rather than a misread one.

Fixed in 43568a62.

backend/ttnn_visualizer/views.py
dcblundell: Agreed — correct today, and the coupling was only in the pin. Added a clause to the existing comment naming the Flask >= 3.1 requirement and the failure mode if the pin is relaxed (a 500 on this route, not a quietly uncapped body), so the two stay connected.

Fixed in 43568a62.

Issue #1833, opened 2026-08-04
Usage signals: record view_engaged (the primitive all activity is derived from)


dcblundell:
Sub-issue of #1819. Depends on the frontend recording client. Serves Q1, Q2 and Q4.

#1819 is explicit that if only one thing ships carefully it should be this one. Every figure about reach and repeat use is derived from `view_engaged` at query time, because neither alternative works: `app_start` overcounts badly since the browser auto-launches on every start (`LAUNCH_BROWSER_ON_START`, `settings.py:150`), and `report_loaded` misses days of work in a long-lived tab. This is the only event that means "a human did something deliberate".

## The decision this issue has to make

**What the threshold is** — time open, interaction count, or both. Worth one round of discussion rather than picking a number, because of an asymmetry: the *definition of "active"* stays revisable at query time, but the *threshold itself is baked in at record time* and cannot be applied retroactively. Getting it wrong means either inflated activity we cannot deflate later, or interactions we never recorded.

Record the answer in the docs page so it is recoverable when someone reads a panel in six months.

## Requirements

- At most one event per view visit. Re-engaging the same view without leaving it must not emit a second event, or the counter becomes an interaction count wearing an activity label.
- Requires a genuine interaction — not a render, not a route change, not a data fetch settling.
- The auto-launched browser landing on `/` must not produce one (caveat 4). This is the failure mode that would make the reach numbers look like good news.
- **Must not add work to hot paths.** No per-`mousemove`, per-scrub or per-row bookkeeping. The NPE timeline and the performance table are called out by name in the canvas and rendering rules in `AGENTS.md`; detecting engagement must not become the thing that makes a scrub janky. Prefer a coarse listener on a container with a cheap boolean latch over anything per-datum.

## Tests

- a view opened and abandoned below the threshold emits nothing
- a view crossing the threshold with an interaction emits exactly one event
- the boot landing on `/` emits nothing
- re-entering the same view after leaving emits a second event; interacting repeatedly within one visit does not
- interaction detection does not run per pointer event

## Out of scope

The Prometheus recording rule that pins the definition of "active" — that lives with the dashboard sub-issue. The two are related and should be reviewed together, since the recording rule is what stops a panel being written against `view_opened` and silently reintroducing the auto-launch inflation.

Issue #1832, opened 2026-08-04
Usage signals: record view_opened and reconcile the view enum with the real routes


dcblundell:
Sub-issue of #1819. Depends on the frontend recording client. Serves Q4, coarsely.

## Scope

One mapping from route to `UsageView`, derived from `ROUTES` ([`src/definitions/Routes.ts:5-16`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/definitions/Routes.ts#L5-L16)) rather than open-coded, and emitted from a single place — a `Layout`-level effect or a route wrapper — rather than from each page component, so a new page cannot silently go uncounted.

## The enum in #1819 does not match the routes

The ten proposed views need reconciling against what actually exists before this is implemented:

- **`topology` is a modal route, not an unimplemented one.** `ROUTES.CLUSTER` has `element: null` in [`routeObjectList.tsx:63-66`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/routes/routeObjectList.tsx#L63-L66) only because [`Layout`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/components/Layout.tsx#L66) renders `<ClusterRenderer />` itself as an overlay over the background route, keyed on `location.state.background`. `MainNavigation` opens it with `navigate(ROUTES.CLUSTER, { state: { background: location } })` and Escape closes it with `navigate(-1)`. So it is a real, countable surface — and it is the one a mapping derived from the route *elements* will silently drop, producing a permanent zero that reads as "nobody opens topology". **Derive the mapping from `ROUTES`, which does contain `CLUSTER`, not from `routeObjectList`.** It is currently the only modal route, so this is one special case rather than a category.

  Two things follow for wiring. Because the overlay renders over a background route, `location.pathname` is `/cluster` while the background page is still mounted — an effect keyed on pathname will fire, but one keyed on the rendered page component will not. And because it is opened only by a nav-button click and dismissed with `navigate(-1)`, the open/close pair is a genuine deliberate action, which makes it a strong `view_engaged` candidate (#1833) and means the back-navigation on close must not be counted as opening the background view a second time.
- **`ROUTES.STYLEGUIDE` exists but is absent from the ten.** Recommend leaving it out: it is a development surface and counting it would pollute reach.
- **`reports` is the index route (`/`)**, not a named `ROUTES` entry.
[...]

Issue #1831, opened 2026-08-04
Usage signals: record report_loaded and report_load_failed


dcblundell:
Sub-issue of #1819. Depends on the frontend recording client. Serves Q3 and the onboarding half of Q5.

#1819 calls `report_loaded` the single highest-value event in Tier 1, because every feature counter is meaningless without it as a denominator (caveat 6). Getting the wiring sites right matters more here than anywhere else.

## Wiring sites

Reports reach the app by four routes, and the atoms are set in more than one place per route. Line numbers are from `dev` at the time of writing and will drift.

**Remote SSH sync** — `source=remote_sync`, in [`RemoteSyncConfigurator.tsx`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/components/report-selection/RemoteSyncConfigurator.tsx): `applyProfilerReportSelection` / `applyPerformanceReportSelection` (356–372), `mountAndActivateFolder` after a successful `POST /api/remote/use` (436–466), and post-sync activation (around 517).

**Upload and local tree** — in [`LocalFolderSelector.tsx`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/components/report-selection/LocalFolderSelector.tsx): upload handlers (183, 221) give `source=upload`; the local folder picker (246) gives `source=local_tt_metal`.

**Consider a single choke point.** [`useActivatingReport.ts`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/hooks/useActivatingReport.ts) (18–28) already wraps select and mount for both profiler and performance reports, so it may be the one seam rather than five call sites — provided `source` can be threaded through it without contorting the signature.

**Boot restore must not count as a load.** [`useRestoreInstance.tsx`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/hooks/useRestoreInstance.tsx) (80–95) sets the active report from the backend instance on every launch, and `LAUNCH_BROWSER_ON_START` defaults to true (`settings.py:150` — note #1819 cites `settings.py:62`, which is stale). Counting it would reproduce the `app_start` overcount of caveat 4 inside the one event used as a denominator, which is worse than losing the signal. Either exclude it or record it as a distinguishable restore.

[...]

Issue #1819, opened 2026-07-30
Usage signals: local-only recording, out-of-band collection, Prometheus/Grafana dashboard


ayerofieiev-tt:
## Problem

We have no idea how much ttnn-visualizer is used, by how many people, or which parts of it matter. Every prioritisation call today — is the Performance table worth more investment than NPE? does anyone use report comparison? is MLIR worth pushing past beta? — is made on anecdote and whoever spoke up most recently in Discord.

We want a small, boring set of usage signals that answer a handful of specific questions, and nothing beyond that.

## Approach and hard constraints

The tool **does not report anything anywhere**. That is a fixed design constraint, not a default to be revisited:

1. **The tool records usage data to a local file only.** No network calls, no endpoints, no opt-in "share telemetry" prompt, no background upload. If the local file is never read by anything, nothing leaves the machine, ever.
2. **Collection is out-of-band.** IT/DevInfra enables reading that local file on managed machines, through the same mechanism they already use for machine-level metrics. The tool is not involved and has no knowledge of whether collection is happening.
3. **Export to Prometheus, surfaced in Grafana.**

The consequence worth stating explicitly: this is a *deliberately partial* measurement. We are trading coverage for the guarantee that the tool itself never phones home. See Caveats.

## Questions the dashboard has to answer

Writing these down first, because otherwise we will ship forty counters and learn nothing:

- **Q1 — Reach.** How many distinct machines run the tool in a given week, and is that trending up or down?
- **Q2 — Depth.** Of those, how many use it more than once? (One-time launches and repeat users are very different signals.)
- **Q3 — Data mix.** Which report types are actually loaded — profiler/memory, performance, NPE, MLIR — and how are they getting in (upload, remote SSH sync, local `TT_METAL_HOME`)?
- **Q4 — Feature value.** Which views and which features inside them are used, and which are effectively dead code?
- **Q5 — Health.** Where do people fail or bounce — failed report loads, failed remote connections, unusably slow views?

Anything proposed below that doesn't serve one of these five should be cut.

## Proposed initial event set

The tool records each of these as a **timestamped event line in a local append-only log** — `<event, timestamp, details>`, the ordinary analytics shape. The IT-side collector aggregates those events into Prometheus counters at collection time. See "Local format" below for why the split, and for the guardrails on `details`.

[...]

PR #1898, opened 2026-08-19
Add the frontend usage recording client

### LINKED ISSUE #1830: Usage signals: frontend recording client (enums, buffered sender, config gating)
Sub-issue of #1819. Depends on the ingest endpoint. No events are wired here — this is the mechanism the event sub-issues plug into.

There is no analytics, telemetry or instrumentation code anywhere in `src/` today, so this establishes the pattern rather than following one.

## Scope

**Enums** — `src/definitions/UsageEvent.ts`

String-valued enums with `SCREAMING_SNAKE_CASE` members, per the naming rules in `AGENTS.md`: `UsageEvent`, `ReportKind`, `ReportSource`, `ReportLoadFailureReason`, `UsageView`. String-valued because these cross a serialisation boundary. Every call site uses the enum member, never the literal.

These duplicate the backend enums, which is unavoidable given validation has to happen server-side. Decide how the two are kept honest — a test asserting parity against a shared fixture is probably enough — and say so in the module, because a silent divergence means events the client emits and the server rejects.

**Endpoint** — add `USAGE = '/api/usage'` to [`src/definitions/Endpoints.ts`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/definitions/Endpoints.ts). Never inline the URL.

**Sender** — `src/functions/recordUsage.ts`

- Buffer in memory and flush off the render path. No per-frame, per-row or per-hover work: caveat 12, and the canvas and rendering performance rules in `AGENTS.md` exist because scrub and hover responsiveness on large reports cost real effort to get right.
- Flush on idle, on a timer, and on `visibilitychange` / `pagehide` — the last via `sendBeacon`, since this is a long-lived SPA whose tabs stay open for days and a buffer that only flushes on unload will lose events.
- **The `pagehide` / `visibilitychange` flush must beacon a typed `Blob`**: `navigator.sendBeacon(url, new Blob([body], { type: 'application/json' }))`. A bare-string beacon is sent as `Content-Type: text/plain`, which the ingest endpoint refuses by design (see #1829). Same-origin, so no preflight is involved and it simply works; a cross-origin beacon with that type *is* preflighted, and `ALLOWED_ORIGINS` refuses it, which is the posture we want. Getting this wrong fails silently in both directions — the endpoint 400s and this client swallows post failures by design — so the loss surfaces only as a thin dashboard panel weeks later. `test_plain_text_body_is_refused` pins the server half; nothing can pin the client half until this issue lands.
- Failures are silent. A usage post must never raise a toast, block a user action, or retry aggressively.

**Gating** — never post under `SERVER_MODE`, and never post when recording is switched off. Both come from [`getServerConfig()`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/src/functions/getServerConfig.ts), which needs a new flag threaded through `_build_spa_client_config` ([`app.py:66-75`](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/backend/ttnn_visualizer/app.py#L66-L75)) and `ServerConfig` (`src/definitions/ServerConfig.ts`), with `test_spa_client_config.py` extended to cover it. Checking client-side is a courtesy, not a control — the endpoint already refuses — but it avoids a pointless request on every interaction in the hosted deployment.

## Tests

`tests/recordUsage.spec.ts`, mocking `src/libs/axiosInstance` in the existing style:

- events batch rather than posting one request each
- each flush trigger fires exactly once
- no request at all when disabled or in server mode
- a rejected or failed post is swallowed and does not resurface

## Out of scope

Wiring any actual event. Tier 3 and Tier 4 events.

## Verification

`pnpm test`, `pnpm lint`, `pnpm lint:ts`, `pnpm lint:spdx`.

dcblundell:
Closes #1830. Sub-issue of #1819, and consumes the ingest endpoint from #1892.

Adds the buffered sender that the Tier 1 and Tier 2 event issues (#1831, #1832, #1833) plug into, plus the enums and config gating it needs.

**No event is wired.** The only production reference added is `initUsageRecording()` in `Layout.tsx`, which starts the flush lifecycle and records nothing on its own. Keeping the transport and the call sites apart is the point of splitting this issue out — but it does mean the diff touches `Layout.tsx`, which can read like scope creep at a glance.

## What changed

**`src/functions/recordUsage.ts`** — the sender. Events are buffered in memory and posted off the render path, so recording one costs a predicate and an array push. The views this will instrument include the NPE timeline and the performance table, where per-frame work is exactly what the canvas rules in `AGENTS.md` exist to prevent (caveat 12 on #1819).

Flushes on idle, on a 30s timer, and on `visibilitychange` / `pagehide` via `sendBeacon`. A long-lived SPA keeps tabs open for days, so a buffer that only flushed on unload would lose most of what it held.

**`src/definitions/UsageEvent.ts`** — `UsageEvent`, `ReportKind`, `ReportSource`, `ReportLoadFailureReason`, `UsageView`, and the `UsageEventPayload` union.

**`src/functions/isUsageRecordingEnabled.ts`** — the gate, kept out of `getServerConfig.ts` for the reason `isDirectReportMode.ts` documents: that module is mocked wholesale by a dozen specs.

**Config plumbing** — `USAGE_RECORDING_ACTIVE` published by `_build_spa_client_config`, typed in `ServerConfig`, read through `isFlagEnabled`, and `USAGE = '/api/usage'` in `Endpoints`.

## Things a reviewer should know

**The `Blob` type on the beacon is load-bearing, not defensive.** `/api/usage` requires `application/json` so the request is non-simple and a hostile origin needs a preflight `ALLOWED_ORIGINS` refuses. A bare-string beacon goes as `text/plain` and is rejected — verified against the real endpoint (400).

**`pagehide` drops a refused batch rather than falling back to a post.** The document is being discarded there, so an async post would not reliably run; a fallback would read as a safety net while being close to dead code. `visibilitychange` does fall back, because a hidden tab usually survives.

**Failed batches are never re-buffered.** The endpoint answers 204 whether it wrote or recording is off, so there is nothing to act on. Re-queueing would grow the buffer without bound against a dead endpoint and resubmit a malformed batch forever.

**The published flag is recomputed, not read from `app.config`.** `create_app` runs `from_object` before applying `settings_override`, so a snapshot of `USAGE_RECORDING_ACTIVE` resolves against the pre-override `SERVER_MODE` and reads true for a hosted app. `test_the_published_usage_state_follows_the_posture_not_the_config_snapshot` fails if someone simplifies this back to a config lookup.

**Parity lives in pytest, not vitest.** `usage.py` owns the vocabulary and `UsageEvent.ts` is a copy; nothing at runtime would report them diverging, because the endpoint answers 422 and this client swallows it by design. `test_usage_frontend_parity.py` is the only thing that notices. It is in pytest because Python can introspect `CLIENT_EVENT_DETAIL_FIELDS` directly, whereas a vitest spec would need a Python toolchain the frontend suite does not have. Mutation-checked: drifting an enum value or dropping a required detail key each fail it.

**`isFlagEnabled` is extracted here** from `isServerModeEnabled`, which now delegates to it — a feature flag should not read as a security posture. #1889 introduces the same helper, so whichever lands second resolves by keeping either copy.

**Also removes the retired `USAGE_RECORDING_ENABLED`** and the warning it drove. Worth stating plainly: that was the live variable in v0.98.0, so an operator who opted out with it and upgrades now starts recording again with no signal. Removing the warning closes that migration window by decision rather than by expiry.

## Verification

`pnpm test` (1939 tests / 184 files), `pnpm lint`, `pnpm lint:ts`, `pnpm lint:spdx`, `pnpm flask:test` (1033 tests), `pnpm flask:lint`, `pnpm flask:mypy` — all clean.

Also checked end to end against the real endpoint, which no axios-mocking spec can reach: the exact body this client builds returns 204 and writes two well-formed logfmt lines to `events.log`; a flattened payload returns 422; a `text/plain` body returns 400.


src/components/Layout.tsx
```diff
@@ -28,6 +30,11 @@ function Layout() {
     const location = useLocation();
     const state = location.state as { background?: Location };
 
+    // Starts the usage flush lifecycle; it records nothing on its own. Here rather than at
+    // module scope so importing the sender has no side effect, and so the listeners are
+    // owned the way every other listener in this app is.
+    useEffect(() => initUsageRecording(), []);
```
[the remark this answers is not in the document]
dcblundell: Checked this rather than taking it at face value, and I don't think it applies here.

`react-hooks/exhaustive-deps` only tracks values declared inside the component — props, state, locals. `initUsageRecording` is a module-scope import, which the rule treats as stable and never asks for. The rule is genuinely on (`plugin:react-hooks/recommended`, `eslint.config.cjs:57`), and `eslint src/components/Layout.tsx --max-warnings 0` is clean, so this is a passing check rather than an absence of coverage.

Adding the dep would be harmless at runtime but would imply the import can change identity, which it can't — and `AGENTS.md` asks that a warning's validity be assessed before acting on it. There's no warning to satisfy here, so I've left it as is.

Happy to reconsider if you're seeing it flagged in a context I haven't.

backend/ttnn_visualizer/app.py
```diff
@@ -81,6 +81,13 @@ def _build_spa_client_config(app: Flask) -> dict:
         "TT_METAL_HOME": app.config["TT_METAL_HOME"],
         "REPORT_DATA_DIRECTORY": str(app.config["REPORT_DATA_DIRECTORY"]),
         "USERNAME": _get_client_username(server_mode),
+        # Recomputed rather than read from ``app.config``: ``from_object`` resolves the
+        # ``_UsageRecordingEnabled`` descriptor before ``settings_override`` is applied,
+        # so the snapshot can claim recording is on for an app that is in server mode.
```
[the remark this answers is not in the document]
dcblundell: Good catch — fixed in 969d59d8.

The descriptor was renamed to `_UsageRecordingActive` in #1892. This branch was cut from that PR's head before the rename landed, so the merge carried my older wording through, leaving the comment pointing at a symbol that no longer exists in exactly the place a reader goes to understand the recompute. `AGENTS.md` treats stale comments as bugs, so this counts.

I grepped for other survivors of the same merge and `_UsageRecordingEnabled` now appears nowhere in `backend/` or `src/`.

src/components/Layout.tsx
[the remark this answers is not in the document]
dcblundell: Leaving as is.

backend/ttnn_visualizer/usage.py
```diff
@@ -457,8 +429,6 @@ def is_recording_enabled(server_mode: Any = False) -> bool:
     The file half of the off switch exists because an environment variable is
     per-shell, and so easy to set in one terminal and lose in the next.
```
aidemsined, line 430: > [!CAUTION]
> An operator who opted out of usage recording in v0.98.0 will silently start recording again, and this PR deletes the warning that would have told them — before it ever shipped.

I verified the chain rather than inferring it:

- `git show v0.98.0:.env.sample` line 21 is `# USAGE_RECORDING_ENABLED=true`, under the comment "Off here, or by creating `~/.ttnn-visualizer/usage/disabled`". So `USAGE_RECORDING_ENABLED=false` was a **released, documented** opt-out.
- `origin/dev` still carries `_RETIRED_RECORDING_ENV_VAR = "USAGE_RECORDING_ENABLED"` (usage.py:96), `_warn_about_the_retired_env_var` (:398), and the call from `is_recording_enabled` (:460).
- At this head, all three are gone — zero survivors across `backend/`, `src/`, `.env.sample` or `docs/`.

Since `origin/dev` hasn't been released, the deprecation warning is being removed *before any user sees it*. The net effect on the next release: someone who deliberately turned recording off gets it back on, with no log line, no banner, and nothing left in the tree that explains why.

The polarity inversion makes this sharper than a rename. The opt-out is now `USAGE_RECORDING_DISABLED=true` (docstring above, :423), so a stale `USAGE_RECORDING_ENABLED=false` isn't merely unrecognised — it can't accidentally still work, and its plain reading is the *opposite* of what it now does.

I take the commit message's word that this was a decision rather than an oversight. But a commit message isn't discoverable by the operator it affects, and this is the one setting in the file where the user's expressed intent is a privacy preference.

**Suggested action:** keep `_warn_about_the_retired_env_var` for one release cycle, or land the migration note where an upgrading operator will actually meet it — release notes, or beside the current opt-out in `.env.sample:18-25`. Separately: this retirement is outside #1830's scope in a PR titled "Add the frontend usage recording client", and would have read more clearly as its own change.

src/functions/recordUsage.ts
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * Failures are silent and unretried. The endpoint answers 204 whether it wrote, whether
+ * recording is switched off locally, and whether the batch was dropped, so there is
+ * nothing here a caller could act on — and instrumentation must never destabilise the
+ * application it measures.
+ */
+
+// Mirrors MAX_USAGE_BATCH_EVENTS in backend/ttnn_visualizer/usage.py, which bounds the
+// atomicity of a single append rather than merely an HTTP body. test_usage_frontend_parity
+// pins the two equal so a batch this client builds can never be refused wholesale.
```
aidemsined, line 28: > [!WARNING]
> The batch cap is pinned; the endpoint's *byte* cap isn't known to the client at all.

Credit where it's due — the count cap is handled exactly right. `MAX_BUFFERED_EVENTS = 50` equals `MAX_USAGE_BATCH_EVENTS` in `usage.py:112`, it's pinned by `test_usage_frontend_parity.py:148-158`, and because `takeBatch` splices at 50 and `recordUsage` flushes on reaching it, a request can never carry more than 50. The endpoint's 400-on-oversized-batch path is unreachable from this client.

The gap is the other cap. `MAX_USAGE_REQUEST_BYTES = 16 * 1024` (`views.py:161`, enforced at `:2597` as a 413) appears nowhere in `src/`. Today there's real headroom: 50 copies of the largest branch of `UsageEventPayload` — `report_load_failed` with `cluster_descriptor`/`unsupported_version`, ~107 bytes — is about 5.4 KB, a third of the cap. So there's no live bug.

But the invariant is load-bearing and undefended. The first event detail that carries free text — a report name, an error string, a file path — pushes 50 events past 16 KB, the endpoint answers 413, and this client swallows post failures by design. That's verbatim the failure #1830 warns about: "surfaces only as a thin dashboard panel weeks later." Nothing would fail; nothing would log.

**Suggested action:** extend `test_usage_frontend_parity.py` with a case that builds `MAX_USAGE_BATCH_EVENTS` copies of the largest event the client's union can express and asserts the serialised envelope is under `MAX_USAGE_REQUEST_BYTES`. That module already reads both `UsageEvent.ts` and `views.py`, so it's the natural home, and it turns the headroom into something that fails loudly when consumed. At minimum, name the byte cap and the current headroom in this comment so whoever adds the next detail field knows the constraint exists.

src/functions/recordUsage.ts
```diff
@@ -0,0 +1,238 @@
...
+    // cap here. Dropping rather than growing is the deliberate answer if it ever is.
+    if (buffer.length >= MAX_BUFFERED_EVENTS) {
+        return;
+    }
+
+    buffer.push(payload);
+
+    if (buffer.length >= MAX_BUFFERED_EVENTS) {
+        flushUsage();
+        return;
```
aidemsined: > [!WARNING]
> The cap path posts synchronously and bypasses the coalescing floor every other path respects, so there's no floor on request rate.

`MIN_BATCH_WINDOW_MS` governs the idle path and the 30s ceiling, but reaching 50 buffered events fires `postEvents` inline, in the same tick as the event that filled the buffer. No rate limit applies.

With the events #1830 actually plans this is unreachable — `report_loaded` and `view_opened` are navigation-scoped, so 50 of them is 50 navigations. The concern is `VIEW_ENGAGED`, which is exactly the shape that ends up wired to interaction, and this module's own docblock at `:15-18` names the NPE timeline and the performance table as the surfaces it will instrument. A `VIEW_ENGAGED` on scroll or hover over a 5,000-row performance table becomes 5,000 events → 100 POSTs issued as fast as the scroll produces them, each with an `instanceId` interceptor pass and a JSON serialisation, on the main thread, mid-scroll.

That's the cliff the canvas/rendering rules in `AGENTS.md` exist to prevent, and it would land on whoever wires the Tier-2 event rather than here — where it's cheapest to prevent.

**Suggested action:** track the last flush time and, on the cap path, defer to a `setTimeout` rather than posting inline — or drop events when the cap is hit inside the coalescing window, which the module is already comfortable doing at `:180-184`. Worth a burst test either way: 5,000 events in a tight loop, asserting the request count stays bounded.

src/functions/recordUsage.ts
```diff
@@ -0,0 +1,238 @@
...
+        navigator.sendBeacon(getUsageEndpointUrl(), new Blob([body], { type: 'application/json' }));
+
+    if (sent) {
+        return;
+    }
+
+    // On pagehide the document is being discarded and an async post is not guaranteed to
+    // run, so falling back there would read as a safety net while being close to dead
+    // code. A hidden tab usually survives, so it is worth trying.
+    if (allowFallback) {
```
aidemsined, line 175: > [!WARNING]
> This comment's correctness depends on a browser event-ordering assumption that no test pins.

To be clear up front: **the claim in the PR body checks out.** Per the HTML unload algorithm, `pagehide` fires *before* the visibility state flips to `hidden`, so on a genuine tab close `handlePageHide` drains first and `handleVisibilityChange` finds an empty buffer. The drop-wins behaviour is real, and the reasoning for not falling back here — the document is being discarded, so an async post wouldn't reliably run — is sound.

The hardening request is that if any browser ever fires `visibilitychange` first, `postEvents` gets issued during document discard, which is precisely the thing this comment says it avoids. And the composite sequence isn't covered: `spec:232-244` tests hide-then-close with a **successful** beacon, and `:284-303` tests refusal on each trigger **independently**. The interesting path — hidden, beacon refused, fallback post issued, then `pagehide` — is untested.

**Suggested action:** add a test dispatching `pagehide` then `visibilitychange` with `sendBeacon` stubbed to return `false`, asserting exactly one drop and zero posts. And note in this comment that the drop-wins behaviour depends on `pagehide` preceding `visibilitychange`, so the assumption is visible to whoever next touches the ordering.

tests/recordUsage.spec.ts
```diff
@@ -0,0 +1,522 @@
...
+
+        expect(sendBeacon).toHaveBeenCalledTimes(1);
+        expect(post).not.toHaveBeenCalled();
+
+        const [url, blob] = sendBeacon.mock.calls[0];
+
+        expect(url).toBe('/api/usage');
+        // The type is load-bearing: it makes the request non-simple, and a bare-string
+        // beacon would go as text/plain and be refused by the route.
+        expect((blob as Blob).type).toBe('application/json');
```
aidemsined, line 270: > [!WARNING]
> This pins the beacon's `type` but never reads its bytes, so a correctly-typed `Blob` with the wrong body shape passes.

The `type` assertion is genuinely valuable and does the job #1830 asked for — a bare string has no `.type`, so the assertion gets `undefined` and fails. That's the pin the ticket said nothing could provide until this issue landed, and it works.

What's missing is the shape. The axios path has a body assertion at `:126-128`, and the spec's own comment there explains why: a flattened payload passes a call-count check while the endpoint's closed-envelope check (`usage.py:830`) rejects it outright. That reasoning applies identically to the beacon path, which is the path that runs on every real tab close — and it has no equivalent assertion. A regression that beacons `{ event, details }` at the top level instead of `{ events: [...] }` would ship green.

`Blob.text()` works in jsdom, so it's about three lines:

```ts
const body = JSON.parse(await (blob as Blob).text());
expect(body).toEqual({ events: [VIEW_OPENED] });
```

**Suggested action:** assert the parsed beacon body against the expected envelope, matching what the axios path already does.

aidemsined (commented):
The transport is well built and the reasoning is written down where the next person will find it. The thing I went looking hardest for — whether the client can post a batch the endpoint refuses — came back clean, and for the right reason rather than by luck: `MAX_BUFFERED_EVENTS` equals `MAX_USAGE_BATCH_EVENTS`, the parity test pins them together, and `takeBatch` splices at 50 before any `await`, so #1892's 400-on-oversized-batch is unreachable from here. The byte cap is the half nothing knows about, flagged inline.

The one item I'd want resolved before merge isn't in the ticket's scope at all: retiring `USAGE_RECORDING_ENABLED` removes a released, `.env.sample`-documented opt-out's deprecation warning before that warning ever reaches a release, and the polarity inverted on the way through.

Some things I checked that came back clean, so nobody re-treads them:

- **AC4 holds for the right reason.** `recordUsage` is a memoised predicate, an integer compare, a push, and an early-returning `scheduleFlush` — no allocation, no `JSON.stringify`, no `new Date()`, no `getServerConfig()`. And the `getServerConfig` concern is closed twice over: `isUsageRecordingEnabled` memoises into a module-level `boolean | null`, and even the uncached path only reads `window.TTNN_VISUALIZER_CONFIG` — it's `axiosInstance.ts` that touches `sessionStorage`, separately cached.
- **There is no standing interval.** The "30s timer" is a one-shot `setTimeout` armed inside `scheduleFlush` and cleared on every flush, so an idle tab with an empty buffer has zero pending timers and zero wakeups. Strictly better than an interval, and it removes the double-interval hazard a remount would otherwise create.
- **Buffer growth is hard-bounded at 50**, even against a dead endpoint, because failed batches are dropped rather than re-buffered. Worst-case resident memory is a couple of KB.
- **The gating is genuinely server-side.** `app.py:90` recomputes `is_recording_enabled(server_mode)`, which returns `False` under `SERVER_MODE`, so the client check really is the courtesy its own comment claims. `test_spa_client_config.py:60-76` is the case that matters — it flips `SERVER_MODE` in the dict and asserts the *built* config follows, so reverting to a plain config lookup fails it.
- **The parity test is better than the ticket asked for.** It compares values in both directions (set equality), and `test_every_client_facing_enum_is_paired` enumerates every `Enum` in `usage.py` via `vars()` so a new one must be paired or explicitly declared server-only — closing the hand-maintained-list hole one level up. It also fails *loudly* on a parse miss rather than degrading to a vacuous pass, which is the specific way a regex-over-TypeScript test usually rots.
- **No vacuous tests found** in either new file, which is worth saying for 522 + 213 lines of them. The one I suspected — "flushes once on the idle fallback and does not flush again on the interval" — is sound, since the 2s window fires first and `clearScheduledFlush` cancels the 30s ceiling.

## General observations

> [!NOTE]
> Minor items — not merge-blocking.

- `String(Endpoints.USAGE)` at `recordUsage.ts:54` is redundant; `Endpoints` is a string enum so the member is already a `string`. Reads as a guard against something the type system forbids.
- `recordUsage.ts:180-184` is unreachable — `recordUsage` flushes the moment the buffer hits the cap and `takeBatch` splices before any `await`, so the buffer can't be *at* cap on entry. The comment admits it, which makes it honest dead code rather than a bug, and it costs one integer compare. Worth keeping; noting it so nobody spends time working out how to reach it.
- `flushUsage` is exported with no production consumer — the only external callers are in the spec. The docblock at `:206-216` names `initUsageRecording` as the entry point but says nothing about `flushUsage`, so nothing stops someone calling it from a component and defeating the batching this module exists to provide. One line would fix that.
- `initUsageRecording` is safe under StrictMode double-invocation, because `addEventListener` de-duplicates identical `(type, listener, capture)` triples and both handlers are stable module-scope references — and `Layout.spec.tsx:63-78` correctly pins the *balance* rather than the count, which is the assertion that catches a leaked listener. What isn't guarded is two concurrent owners, where the first teardown removes listeners the second still needs. `Layout` renders once so it's theoretical; a module-level `let started = false` returning a no-op teardown would close it for free.
- Events carry no client timestamp, so an event buffered before a tab is hidden and beaconed hours later is timestamped at receipt. Fine for reach and DAU, lossy for session reconstruction — a server-side decision from #1892, not this PR's, raised only so it's on the record.
- `test_the_usage_recording_state_is_published_under_both_postures` asserts key *presence*, not value. That matches its stated purpose (absent must only ever mean "nothing inlined"), so no change wanted — flagging so it isn't misread as a value assertion.

Worth noting how good `0cc8593e` is, since it's the kind of commit that usually doesn't happen. It fixed four defects that shared one cause — the specs passed with or without the behaviour — including a dev diagnostic that could never fire (axios resolves only on 2xx, so the 422 went to a bare `.catch`), and `requestIdleCallback`'s `timeout` being an upper bound on delay rather than a lower one, which meant each event got its own request on a quiet tab. The `BOOLEAN_TRUE_VALUES` rename in `12dcf3cd` is the same instinct: renaming a constant once it governs a second setting is exactly how you avoid a constant silently controlling something it isn't named for.

Nothing was run from here. `npm run lint` is the one I'd most want green, specifically whether `plugin:compat/recommended` accepts the positive `typeof requestIdleCallback === 'function'` guard *and* the `cancelIdleCallback` call inside it — that's a different global the plugin may check independently under the repo's browserslist. Then `tsc --noEmit`, the three vitest specs, and `pnpm flask:test` over the four touched backend test modules.

---

src/functions/recordUsage.ts
dcblundell: Fixed in 2d0897c8 — you're right, and this was the most valuable of the five.

The cap path no longer posts inline. Overflow is dropped and the scheduled flush drains within `MIN_BATCH_WINDOW_MS`, so the request rate is bounded however fast a caller records rather than scaling with the gesture. Dropping is consistent with what the module already does at the cap, and losing overflow beats issuing 100 posts mid-scroll.

Added the burst test you asked for: 5,000 events in a tight loop, asserting one request carrying 50 events. Mutation-checked — restoring the inline post fails two tests.

src/functions/recordUsage.ts
dcblundell: Fixed in 2d0897c8. Confirmed your figures independently: 50 copies of the largest branch serialise to 5,412 bytes against a 16,384 cap, and `MAX_USAGE_REQUEST_BYTES` appears nowhere in `src/`.

Took the parity-test option rather than only the comment, since a comment doesn't fail. It builds a full batch of the widest event the client's union can express — from the enums, so widening any of them is what moves the number — and asserts it fits. Mutation-checked by shrinking the cap to 4 KB.

The failure message names the consequence directly, because a 413 is swallowed like any other failure and would surface as events quietly going missing.

src/functions/recordUsage.ts
dcblundell: Fixed in 2d0897c8, both halves.

The ordering assumption is now stated where the behaviour is: drop-wins depends on `pagehide` firing before the visibility state flips, and if a browser reversed that the fallback would fire on the discard path this is written to avoid.

Added the composite test — refused beacon on `pagehide`, then hidden — asserting one drop and zero posts. You were right that each trigger was only covered alone. Mutation-checked by switching `handlePageHide` to allow the fallback.

tests/recordUsage.spec.ts
dcblundell: Fixed in 2d0897c8. Agreed the reasoning from the axios path applies identically here, and more so — this is the path every real tab close takes.

One correction worth recording: `Blob.text()` isn't available in this jsdom. Its `Blob` exposes only `slice`, `size` and `type`, and Node's `Response` won't accept it across realms (it stringifies to `[object Blob]`). Read it through `FileReader` instead, with a comment saying why so the next person doesn't retry the shorter form.

Mutation-checked by beaconing a flattened envelope.

backend/ttnn_visualizer/usage.py
dcblundell: Thanks for verifying the chain rather than inferring it — and the point that `origin/dev` is unreleased, so the warning would be deleted before any operator ever sees it, is a fact neither the author nor I had brought out. It's the sharpest framing of this so far.

The decision stands: `USAGE_RECORDING_ENABLED` stays removed, and the migration note is deliberately not being added either, since it would put the retired name back in the tree. Recording that plainly so it's discoverable here rather than only in a commit message: the deprecation window is being closed by decision, not by expiry, and the consequence you describe is accepted.

On scope, you're right that it doesn't belong in a PR titled "Add the frontend usage recording client" and would have read better as its own change. Noted for next time.

Leaving this open rather than resolving it — you raised it, so it's yours to close if you're satisfied, and I'd rather not tidy away feedback I didn't act on.

backend/ttnn_visualizer/usage.py
```diff
@@ -88,13 +88,6 @@
 # operator who wants no usage data sets this, and everyone else sets nothing.
 USAGE_DISABLED_ENV_VAR = "USAGE_RECORDING_DISABLED"
```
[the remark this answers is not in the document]
dcblundell: Same answer as the thread @aidemsined opened on this — the two overlap.

This is a decision rather than an oversight: `USAGE_RECORDING_ENABLED` stays removed, and the migration note is deliberately omitted too, because it would reintroduce the retired name. The consequence you describe is real and accepted — an operator who opted out on v0.98.0 and upgrades will start recording again with no signal.

Worth noting your suggestion to keep honouring the old opt-out for a deprecation window is the stronger of the two you offer, since it protects the user rather than merely informing them. It was considered and declined.

PR #1900, opened 2026-08-19
Op graph critical path: trace the longest cumulative-duration path without overriding the highlights already on screen

### LINKED ISSUE #1613: Operation graph: critical-path highlighting (longest cumulative duration through the DAG)
## Context

Follows up on #1515. The perf overlay answers "which ops are hot" but not "which _path_ is hot". Two ops at 10 ms each that run sequentially cost more wall-clock than one 15 ms op on a side branch — but the current overlay can't surface that.

## Scope

1. Compute the longest cumulative-time path through the operation graph treated as a DAG. Op duration is the only cost signal; weights come from the perf overlay's `aggregatesByOpId` (`deviceTimeNs`), which is already scoped to the ops the canvas shows.
2. New toolbar toggle: **"Highlight critical path"** (disabled unless perf overlay status is `READY`). It gates on the same `PerfOverlayStatus` the bars do, but works independently of them — the path can be traced with the bars off.
3. When enabled:
   - Edges along the critical path render in a high-contrast accent (`--graph-critical-path-edge`); off-path edges dim to 0.35 via a container-level class, so only the path itself carries per-element classes.
   - Nodes on the path get an outline rather than a colour override, leaving both the perf ramp and the input/output fill visible.
   - Side annotation: "Critical path: 81 ops · 2.4 ms · 35.0% of total device time", in the bottom-left band above the perf legend.
4. Hold the toggle in a `criticalPathScopeAtom` that carries the reports it was switched on for.

## Decisions

- **Zero-cost ops stay traversable.** An op with no perf row costs 0 and the path runs through it, so a gap in the report shortens the total rather than severing the path. Deallocate ops behave the same way when "Hide deallocate ops" is off.
- **Ties resolve on cost, then chain length, then op id.** Length second so an equal-cost tie reads as the fuller sequence instead of a prefix of it; op id last so the pick can't move with build order. Determinism rests entirely on those comparators, so the topological order is left unsorted and the pass stays O(V+E).
- **The toggle is in-memory and carries its report scope, not a bare flag.** The highlight only means anything on a `READY` overlay, so nothing is persisted. Storing the scope makes a stale intent inert the moment the report changes — the switch reads off in the same render, with no frame showing a path computed from the previous report's weights and no render-phase write to a shared atom. The view then clears the stale intent, so swapping back to the original report finds the switch off, matching the perf overlay.
- **The I/O highlight outranks the path accent.** Edges touching the selected node keep their input/output colour; the node outlines keep the path traceable across that gap. Those same edges are also exempt from off-path dimming, so clicking a node to inspect its inputs and outputs never fades half of them.
- **Filter dimming outranks off-path dimming.** While a search is active the filter's dim applies instead, since it is the narrower intent.
- **The path outline clears the selection ring.** Outline paints above box-shadow, so the outline sits at `outline-offset: 4px`, outside the 3px selection ring — a node that is both selected and on the path shows both.
- **The share is of summed device time, not wall clock.** Denominator is the overlay's total across linked ops, labelled "of total device time" and carried to one decimal to match the per-op hover; whole-percent rounding prints 0% for a short path on a large graph.
- **Cycles are a property of the report, not a malformed input.** Edges follow each tensor's consumer list, which nothing constrains to run forwards. Kahn's algorithm covers the acyclic portion, the result carries a `hasCycle` flag, the annotation reads "Critical path (partial)" so the understated total isn't mistaken for a confident one, and a console warning names the reason for whoever is debugging the report.

## Acceptance criteria

- [ ] Longest-path computation is O(V+E): Kahn topological order plus one longest-path relaxation pass over the ops the build kept. Connectivity and the deallocate filter are already applied upstream in `opGraphBuilder`, so there is no second pass to scope the subgraph.
- [ ] Deallocate ops are off the path by default, through the existing "Hide deallocate ops" default rather than a second name filter.
- [ ] Toggle integrates cleanly with the existing IO highlight on click (critical path doesn't fight the input/output colour of the focused node, and doesn't dim its edges).
- [ ] Total path duration + op count + share of total device time shown on canvas, suppressed entirely when no path exists.
- [ ] Unit tests cover: linear chain, branching graph, equal-cost ties including order-independence of both sources and edges, zero-cost pass-through, parallel edge pairs, partial and whole-graph cycles, empty graph.
- [ ] Toolbar tests cover the new switch: disabled unless `READY`, independent of the overlay switch, and the `CRITICAL_PATH_TOOLTIP` mapping in every state.
- [ ] Report-scope tests cover intent invalidation on profiler and performance swaps, no revival on swap back, and survival of a rebuilt-but-equivalent `ReportFolder`.
- [ ] The two SCSS precedence claims are asserted against the stylesheet, not just stated in comments: accent rule ordered before the I/O rules, and the outline offset larger than the selection ring's width.

## Open questions

- Should we allow the user to click an op and see "the critical sub-path that includes this op"? — punt to follow-up.

## Out of scope

- Comparison mode ("what changed on the critical path between two reports") — separate ticket.
- Per-tensor edge weights (currently edges are unweighted; node duration is the only cost signal).
- Styleguide entry for the edge accent — dropped deliberately: the annotation only reads correctly against real report data, and the accent colour is already confined to `_base.scss` with both consumers using `var()`.

## Notes

- Builds on #1515. Independent of the metric-selector and hot-ops phases.
- Probably the highest-value triage feature in the backlog after the basic overlay — "what's on the critical path" is the question that drives most perf optimisation conversations.

aidemsined:
image

## Summary

The perf overlay answers "which ops are hot" but not "which *path* is hot". Two 10 ms ops running in sequence cost more wall-clock than one 15 ms op on a side branch, and a per-node colour ramp cannot show that — the question that drives most perf conversations is which chain the time actually flows through.

This adds a **Highlight critical path** toggle that treats the operation graph as a DAG, weights each op by the device time the perf overlay already aggregates, and highlights the longest cumulative-duration path. The highlight is additive by construction: it outlines nodes rather than recolouring them, and it yields to both the input/output highlight and search dimming, so turning it on never costs you information that was already on screen.

Stacked on #1890, so the base here is `feature/1880-op-graph-perf-overlay` and this diff is the critical path work alone. GitHub retargets it to `dev` once #1890 merges.

Closes #1613

### Frontend

- **`opGraphCriticalPath.ts`** (new) — Kahn topological order plus a single longest-path relaxation, O(V+E). Connectivity and the deallocate filter are already applied by `opGraphBuilder`, so there is no second pass to scope the subgraph. Ties resolve on cost, then chain length, then op id: length second so an equal-cost tie reads as the fuller sequence rather than a prefix of it, op id last so the pick cannot move with build or edge order. Determinism rests entirely on those comparators, which is what lets the topological order stay unsorted.
- **Cycles are treated as a property of the report, not malformed input.** Edges follow each tensor's consumer list, and nothing constrains that to run forwards. Kahn covers the acyclic portion, the result carries `hasCycle`, and the annotation says "partial" so an understated total is not mistaken for a confident one.
- **`CriticalPathAnnotation.tsx`** (new) — `Critical path: 81 ops · 2.4 ms · 35.0% of total device time`, in the bottom-left band above the perf legend, suppressed entirely when no path exists. The share is of summed device time rather than wall clock, labelled as such, and carried to one decimal because whole percents print `0%` for a short path on a large graph.
- **`ReportScope.ts`** (new) + **`store/app.ts`** — `criticalPathScopeAtom` holds the report pair the toggle was switched on for instead of a bare boolean, keyed on report *paths* so a refetch that rebuilds an equivalent `ReportFolder` is not read as a report swap. A stale intent is inert in the same render it goes stale, so no frame can show a path computed from the previous report's weights, and the view then clears it — swapping back finds the switch off, matching the perf overlay.
- **`OpGraphToolbar.tsx`** — a second switch, gated on the same `PerfOverlayStatus` as the bars (both read per-op durations) but independent of them, so the path can be traced with the bars off.
- **`OperationGraphReactFlow.tsx`** — computes the path when the toggle is on and carries classes only for the path itself; the ~500 off-path elements are dimmed from the container so React Flow's identity diffing is untouched for the non-path majority.
- **SCSS** — the accent rule is ordered ahead of the I/O rules so edges touching the selected node keep their input/output colour, and those edges are exempt from the off-path dim so clicking a node to inspect its inputs never fades half of them. Search dimming wins while a filter is active, being the narrower intent. Nodes take an outline rather than a fill, offset clear of the selection ring, so a node that is selected, hot and on the path shows all three.

## Known limitations (deliberate, per the issue)

- No styleguide entry for the edge accent: the annotation only reads correctly against real report data, and the accent colour is confined to `_base.scss` with both consumers reading it through `var()`.
- Edges are unweighted — per-op device time is the only cost signal.
- Comparison mode ("what changed on the critical path between two reports") is a separate ticket.

## Test plan

- [x] `pnpm test` — 2012 tests green, including 13 algorithm cases (linear chain, branching, equal-cost ties with order-independence of both sources and edges, zero-cost pass-through, parallel edge pairs, partial and whole-graph cycles, empty graph)
- [x] Toolbar switch coverage: disabled unless `READY`, disabled mid-rebuild, independent of the overlay switch, and the full `CRITICAL_PATH_TOOLTIP` mapping
- [x] Report-scope coverage: invalidation on profiler and performance swaps, no revival on swap back, and survival of a rebuilt-but-equivalent `ReportFolder`
- [x] The two SCSS precedence claims asserted against the stylesheet rather than stated in comments: accent ordered before the I/O rules, and the outline offset larger than the selection ring's width
- [x] `tsc`, `eslint`, `stylelint`, `prettier` clean
- [ ] Manual: load a matched profiler + performance report pair, toggle the path on, then confirm it composes — select a node on the path (outline plus selection ring, its own edges keep I/O colour), search while the path is on (filter dim wins), and toggle the overlay bars independently

## Follow-ups

- "Show the critical sub-path through this op" on click — the issue's open question, punted rather than tracked; happy to open an issue if it is worth queueing.


src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -552,6 +607,9 @@ const OperationGraphInner = ({
             if (matchedIds && (isSelected || matchedIds.has(node.id))) {
                 classNames.push(MATCHED_NODE_CLASS);
             }
+            if (criticalPathNodeIds?.has(node.id)) {
+                classNames.push(CRITICAL_PATH_NODE_CLASS);
```
dcblundell, line 618: **Must fix:** the rendered half of the feature has no behavioural test

Nothing asserts that `criticalPath.nodeIds`/`edgeIds` reach React Flow as `op-graph-node-critical-path`/`op-graph-edge-critical-path` (here and line 658), that the container gains `op-graph-critical-path` (line 754), or that `CriticalPathAnnotation` renders at all. Verified by mutation: deleting all three class pushes **and** the annotation render leaves the full suite (2013 tests) green — so AC3 and AC4 of #1613 have zero coverage, and the SCSS precedence tests in `opGraphNodeStyles.spec.ts` currently guard a cascade nobody proves is reached.

The base branch established the pattern for exactly this (`filter dimming › marks the matched set`, `perf overlay identity › keeps untouched nodes byte-for-byte`). Note `OPERATION_LIST` is a straight chain, so every node lands on the path — an on-path-only assertion would pass against an unconditionally applied class. A branching fixture is needed.

tests/opGraphCriticalPath.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const edgeId = (source: number, target: number, parallelIndex = 0) => `${source}-${target}-${parallelIndex}`;
+
+const edges = (...pairs: [source: number, target: number][]) =>
+    pairs.map(([source, target]) => ({ id: edgeId(source, target), source: String(source), target: String(target) }));
+
+const weights = (...pairs: [opId: number, deviceTimeNs: number][]) => new Map<number, number>(pairs);
+
+describe('findCriticalPath', () => {
+    it('walks a linear chain and sums every op on it', () => {
```
dcblundell, line 21: **Must fix:** `nodeIds` is never read by any of the 12 cases

The spec asserts `opIds`, `edgeIds`, `totalNs` and `hasCycle`, but never `nodeIds` — yet `hasCriticalPath = criticalPath.nodeIds.size > 0` is what gates the node outlines, the container dim and the annotation. `findCriticalPath` could return an empty `nodeIds` alongside correct `opIds` and all 12 cases would still pass, silently disabling the whole feature.

Assert it on at least the linear-chain and equal-cost-tie cases. Worth adding an isolated-nodes case too (no edges, weights `1→10, 2→50, 3→20` → `opIds [2]`, `edgeIds.size 0`): that is the single-node branch the `hasCriticalPath` nulling exists for, and which the comment at `OperationGraphReactFlow.tsx:579` calls out but nothing exercises.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -141,6 +150,7 @@ const OperationGraphInner = ({
     const [nodeIndex, setNodeIndex] = useState<OpGraphNodeIndexEntry[]>([]);
     const [hideDeallocate, setHideDeallocate] = useState(true);
     const [isPerfOverlayEnabled, setIsPerfOverlayEnabled] = useState(false);
+    const [criticalPathScope, setCriticalPathScope] = useAtom(criticalPathScopeAtom);
```
dcblundell, line 158: **Should fix:** the path highlight survives an unmount, the overlay doesn't

`isPerfOverlayEnabled` is `useState` (line 152); `criticalPathScope` is a module-level atom. Leave `/graphtree` and return with the same reports and the path reads **on** while the bars read **off**. The comment at line 209 says this matches "the overlay's behaviour" — true for swap-back, false for remount.

The atom also has exactly one consumer, which AGENTS.md → State management points away from, and it leaks across tests (hence the `beforeEach` reset). Either make it local `useState` and extract one `useReportScopedIntent(reportScope)` both switches consume, or state in the atom's comment that cross-mount stickiness is deliberate and apply it to both switches.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -525,6 +548,38 @@ const OperationGraphInner = ({
...
+    const criticalPath = useMemo(() => {
+        if (!isCriticalPathActive) {
+            return EMPTY_CRITICAL_PATH;
+        }
+        const deviceTimeNsByOpId = new Map<number, number>();
+        for (const [opId, aggregate] of perfOverlay.aggregatesByOpId) {
+            deviceTimeNsByOpId.set(opId, aggregate.deviceTimeNs);
+        }
+        return findCriticalPath(nodeIndex, edges, deviceTimeNsByOpId);
+    }, [isCriticalPathActive, perfOverlay, nodeIndex, edges]);
```
dcblundell: **Should fix:** clicking any edge pays a full O(V+E) recompute

`edges` comes from `useEdgesState` with `onEdgesChange` wired stock (line 799) and `elementsSelectable` left at its default `true`, so selecting an edge produces a fresh array via `applyEdgeChanges` — triggering the whole traversal plus a re-map of all E edges in `styledEdges`. The critical path cannot change as a result of an edge selection.

Key the memo on something that only moves when the graph is rebuilt (a build counter set in `onBuilt`, or the built edge array held in a ref beside `nodeIndex`) rather than on the React Flow edge state array.

src/components/operation-graph/opGraphCriticalPath.ts
```diff
@@ -0,0 +1,175 @@
...
+        for (const edge of outgoingByNodeId.get(nodeId) ?? []) {
+            const candidateCost = nodeCost + weightOf(edge.target);
+            const candidateOpCount = nodeOpCount + 1;
+            const currentCost = cost.get(edge.target) ?? 0;
+            const currentOpCount = opCount.get(edge.target) ?? 1;
+            const currentPredecessor = predecessorByNodeId.get(edge.target);
+            // Cost, then chain length, then op id. Length second so an equal-cost
+            // tie reads as the fuller sequence rather than a prefix of it; op id
+            // last so the pick can't move with build order.
+            const isBetter =
```
dcblundell: **Should fix:** the tie-break is written twice in two different shapes

The relaxation guard here compares values directly with a `currentPredecessor !== undefined` clause; the end-node scan at ~line 142 compares precomputed deltas. Determinism — which this spec explicitly holds ("picks the same branch whichever order the sources arrive in") — depends on the two agreeing, and nothing enforces that.

Extract one total comparator and call it from both sites, e.g. `isPreferred(costDelta, opCountDelta, opIdDelta)`. At the relaxation site pass `currentPredecessor === undefined ? 0 : byOpId(...)` — behaviour-identical, since an equal-cost *and* equal-length tie can only arise once the target has a predecessor.

tests/opGraphNodeStyles.spec.ts
```diff
@@ -68,3 +68,62 @@ describe('op graph filter dimming', () => {
...
 });
+
+describe('op graph critical path styling', () => {
+    const OFF_PATH_DIM_SELECTOR = '&.op-graph-critical-path:not(.op-graph-filtering)';
+
+    it('accents the path before the I/O rules so a focused node keeps its own colours', () => {
+        // Both match at the same specificity, so the later rule wins. Order is the
+        // only thing keeping the selected node's edges their input/output colour
+        // while the rest of the path reads magenta. #1613
+        const accent = STYLESHEET.indexOf('.op-graph-edge-critical-path .react-flow__edge-path');
```
dcblundell, line 79: **Should fix:** the ordering assertion covers `-input` only, not `-output`

AC8 of #1613 asks that the accent be ordered before "the I/O rules" (plural). This compares the accent's index against `.op-graph-edge-input` alone, so moving the `.op-graph-edge-output` rule above the accent would silently break output-edge precedence with the suite green. One extra `indexOf` closes it.

tests/opGraphNodeStyles.spec.ts
```diff
@@ -68,3 +68,62 @@ describe('op graph filter dimming', () => {
...
+        // filter's 0.18 and quietly weaken search dimming whenever the path is on.
+        expect(STYLESHEET).toContain(OFF_PATH_DIM_SELECTOR);
+    });
+
+    it('exempts the selected node′s own edges from the off-path dim', () => {
+        // Clicking a node to read its inputs and outputs would otherwise fade
+        // whichever of them sit off the path. Read out of the `:not()` arguments
+        // rather than matched as text: a selector list and chained `:not()`s exclude
+        // the same set, so asserting one spelling pins the authoring style instead.
+        const exclusions = Array.from(ruleBody(OFF_PATH_DIM_SELECTOR).matchAll(/:not\(([^)]*)\)/g))
```
dcblundell, line 103: **Should fix:** the exclusion set doesn't assert the path's own member

The block checks `.op-graph-edge-input` and `.op-graph-edge-output` are in the `:not()` list, but not `.op-graph-edge-critical-path`. Drop that member from the selector and the path edges themselves dim to 0.35 — the feature's central visual — with every test still passing.

Given `151f4e40` deliberately hardened this test to read the set rather than a spelling, the missing member looks like an oversight. One more `expect(exclusions).toContain(...)` in the same block.

src/components/operation-graph/CriticalPathAnnotation.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface CriticalPathAnnotationProps {
+    opCount: number;
+    totalNs: number;
+    /** Device time across every linked op in the graph, the share's denominator. */
+    measuredNs: number;
+    /** A cycle held part of the graph back, so the path and total understate. */
+    isPartial?: boolean;
+}
+
+const CriticalPathAnnotation = ({ opCount, totalNs, measuredNs, isPartial = false }: CriticalPathAnnotationProps) => {
```
dcblundell: **Should fix:** re-renders every drag frame, constructing an `Intl.NumberFormat` each time

Not `memo`'d, and `handleNodesChange` forwards position changes into `setNodes`, so this renders on every node-drag frame — each one calling `formatDuration` → `formatSize` → `new Intl.NumberFormat` (`src/functions/math.ts:28`), which is among the more expensive things to do per frame.

All four props are primitives, so wrapping the export in `memo` bails out cleanly on every drag frame. `PerfOverlayLegend` has the same shape, so this doubles an existing per-frame cost rather than introducing it — worth fixing in both.

src/scss/components/CriticalPathAnnotation.scss
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+@use '../definitions/colours' as *;
+
+// Appearance only; the graph view that shows it owns the placement.
+.critical-path-annotation {
```
dcblundell, line 8: **Should fix:** duplicates the perf legend's chip chrome value for value

`background-color: rgba($tt-grey-2, 0.85)`, `border-radius: 4px`, `font-size: 11px`, `line-height: 1.4` and `pointer-events: none` all repeat `PerfOverlay.scss:9-19`. The two now sit stacked in the same `.op-graph-bottom-band`, so any drift is visible side by side in a single screenshot.

Extract a shared mixin (e.g. `graph-chrome-chip`) and `@include` it from both, keeping only the genuine differences (row vs column, padding). Separately: this sheet has one consumer, so per CONVENTIONS.md it should be named `CriticalPathAnnotation.scss` — `PerfOverlay.scss` is the two-consumer exception.

src/components/operation-graph/opGraphCriticalPath.ts
```diff
@@ -0,0 +1,175 @@
...
+    nodes: readonly CriticalPathNode[],
+    edges: readonly CriticalPathEdge[],
+    deviceTimeNsByOpId: ReadonlyMap<number, number>,
+): CriticalPath => {
+    if (nodes.length === 0) {
+        return EMPTY_CRITICAL_PATH;
+    }
+
+    const opIdByNodeId = new Map<string, number>();
+    const inDegree = new Map<string, number>();
```
dcblundell: **Should fix:** `inDegree`/`cost`/`opCount` skip the `valueByKey` naming pattern

AGENTS.md → Naming mandates naming `Map` accumulators after the relation they encode, and this function already does it for `opIdByNodeId`, `outgoingByNodeId`, `predecessorByNodeId` and `predecessorEdgeByNodeId` — so the file is internally inconsistent.

`opCount` is the worst of the three: here it means "ops on the best chain ending at this node", while the `opCount` prop on `CriticalPathAnnotation` means ops on the whole path. Rename to `inDegreeByNodeId` / `costByNodeId` / `opCountByNodeId`.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -159,6 +163,21 @@ const OpGraphToolbar = memo(
                         disabled={isDisabled || perfOverlayStatus !== PerfOverlayStatus.READY}
```
dcblundell: **Should fix:** the two switch blocks are structurally identical, including the READY gate

Same `Tooltip` placement, same `op-graph-toolbar-switch` class, same inline `(event) => cb(event.currentTarget.checked)`, and — the part that matters — the same duplicated `isDisabled || perfOverlayStatus !== PerfOverlayStatus.READY`. That predicate is the feature's contract ("both read per-op durations"), so having it in two places invites exactly the divergence the shared `CRITICAL_PATH_TOOLTIP`/`PERF_OVERLAY_TOOLTIP` keying was meant to prevent.

Extract a file-local `PerfGatedSwitch` taking `{ tooltip, label, checked, onChange }` and compute the gate once. Keep the "bound to the Switch, not a wrapper" comment on the extracted component — it is the reason for the shape.

src/components/operation-graph/CriticalPathAnnotation.tsx
```diff
@@ -0,0 +1,42 @@
...
+    // Device time is the only measured quantity, so the share is of summed op
+    // time, not wall clock — said plainly, because "of measured time" reads as
+    // wall clock. One decimal matches the per-op hover, where rounding to whole
+    // percent prints 0% for a short path on a large graph.
+    const share = measuredNs > 0 ? `${((totalNs / measuredNs) * 100).toFixed(1)}% of total device time` : null;
```
dcblundell: **Decision needed (spec vs code):** "of total device time" vs "Kernel duration"

This is the exact string #1613 specifies, so the code follows the spec. But the same `deviceTimeNs` measure is called **"Kernel duration"** in the legend rendered directly below it in the same band (`PerfOverlayLegend`), in `PerfOverlayOpMetric`, and in `PERF_OVERLAY_TOOLTIP[READY]` — two labels for one measure, adjacent on screen.

My suggestion is to unify on "kernel duration" and amend the issue, since the adjacency is what makes it read as drift — but following the issue verbatim is defensible and this is your call, not a defect.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -750,14 +824,22 @@ const OperationGraphInner = ({
...
             </ReactFlow>
-            {isPerfOverlayActive && !isBuilding ? (
-                <div className='op-graph-perf-legend'>
+            <div className='op-graph-bottom-band'>
+                {isCriticalPathActive && hasCriticalPath && !isBuilding ? (
+                    <CriticalPathAnnotation
+                        opCount={criticalPath.opIds.length}
+                        totalNs={criticalPath.totalNs}
+                        measuredNs={perfOverlay.totalNs}
+                        isPartial={criticalPath.hasCycle}
```
dcblundell, line 840: **Decision needed (spec vs code):** "(partial)" never renders for the case it was written for

#1613's Decisions say the annotation reads "Critical path (partial)" so an understated total isn't mistaken for a confident one. AC4 says the annotation is "suppressed entirely when no path exists". For a **whole-graph** cycle the code takes suppression — `hasCriticalPath` is false because `nodeIds` is empty — so `(partial)` never shows in exactly the case the cycle decision was written for, and only the `console.warn` fires.

The code's choice looks right to me; the issue is the thing to amend (scope "(partial)" to *partial* cycles). Flagging so the divergence is a decision rather than a silent one.

tests/opGraphCriticalPath.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+        );
+        const backwards = findCriticalPath(
+            nodes(4, 3, 2, 1),
+            edges([1, 2], [2, 4], [1, 3], [3, 4]),
+            weights([2, 10], [3, 10]),
+        );
+
+        expect(backwards.opIds).toEqual(forwards.opIds);
+        expect(backwards.totalNs).toBe(forwards.totalNs);
+    });
```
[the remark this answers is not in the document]
aidemsined: Correct, and thanks — the fixture couldn't fail. Node 1 was the only zero-indegree source in both runs, so reversing the node list left Kahn's seed identical and the assertion held for reasons unrelated to its name.

Replaced with two equal-cost sources converging on one sink (`edges([1, 3], [2, 3])`, ops 1 and 2 both at 10ns), run as `nodes(1, 2, 3)` against `nodes(3, 2, 1)` so the seed order genuinely reverses. The sink's predecessor is then settled by the op-id tie-break alone.

Verified falsifying: dropping the op-id clause from the comparator makes the reversed run answer `[2, 3]` against the forward run's `[1, 3]`.

c865f294

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Agreed, and confirmed before fixing: nothing in `tests/` referenced `op-graph-node-critical-path`, `op-graph-edge-critical-path`, `op-graph-critical-path` or the annotation, so the three class pushes and the annotation render were all deletable with the suite green.

Added `OperationGraphReactFlow critical path rendering` with the branching fixture your note called for — `BRANCHING_OPERATION_LIST` runs two routes from op 1 to op 4, and `BRANCHING_PERF_ROWS` makes 1 → 3 → 4 cost 120ns against 25ns for 1 → 2 → 4, so op 2 and its two edges are off-path and an unconditionally applied class fails:

- ops on the path marked, op 2 not
- the two traversed edges marked, the two skipped ones not
- the container gains the class only once the switch is on
- the annotation renders, reading `3 ops`, the path's own total, and its share
- switching off takes all of it back

Re-ran your mutation against the new block: deleting the two class pushes, the container class and the annotation render now fails 5 tests where it previously failed 0.

c865f294

tests/opGraphCriticalPath.spec.ts
aidemsined: Agreed — `nodeIds` appeared zero times in the spec, and it is what `hasCriticalPath` gates the outlines, the container dim and the annotation on.

Asserted on the linear-chain and equal-cost-tie cases (the second also pins that the losing branch is *absent* from the set, not merely absent from `opIds`), and added the isolated-nodes case you described — `nodes(1, 2, 3)` with no edges and weights 10/50/20 gives `opIds [2]`, `nodeIds {'2'}`, `edgeIds.size 0`. That is the single-node branch the comment at `OperationGraphReactFlow.tsx` referred to and nothing exercised.

c865f294

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: The remount asymmetry is real and worth naming — I'm keeping the atom, and I've documented it rather than reversing it. Reasoning, since this is the one place I'm not taking the suggestion:

#1613 asks for the toggle to be held in a `criticalPathEnabledAtom` by name, and an earlier review round proposed exactly this swap to `useState`; it was declined then and the scope-carrying atom was chosen deliberately, which is what removed the reset effect and the stale frame. Reversing that inside this PR would undo a decision made twice, and the *other* half of your alternative — "apply it to both switches" — is a behaviour change to the overlay from #1880, which is still in review. Which lifetime is right for a perf-derived toggle is a product call, not a detail of the critical-path change.

So: the atom's comment in `src/store/app.ts` now states that surviving a trip away from `/graphtree` is deliberate, that the report it names is what invalidates it rather than the view's lifetime, and that the overlay deliberately differs for now. Opened #1903 to decide one lifetime for both, with the three options and what each costs.

On AGENTS.md: the State management section governs where atoms live and how they're consumed — this one is declared in `store/app.ts` and read through `useAtom`, so it complies. I don't read it as ruling out a single-consumer atom, though I agree that's the weakest justification for one.

c865f294

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Valid — `onEdgesChange` is wired stock and edges are selectable (`elementsSelectable` is left at its default and the rendered edge groups carry `selectable`), so `applyEdgeChanges` hands back a fresh array on every edge click and the traversal ran again for a change that cannot move the path.

The built edge set now lives in its own `builtEdges` state written in `onBuilt` beside `setNodeIndex`, and the memo keys on that instead of React Flow's edge state.

One deviation from your suggestion worth flagging: I tried the ref first, and `react-hooks/refs` correctly objected — reading `.current` inside the memo is a render-phase ref read. State is the better shape here anyway, since it changes exactly when a build lands, and it needs no comment explaining why a ref read is safe.

Pinned by `does not retraverse the graph when React Flow replaces the edge array`, which spies on `findCriticalPath` and replaces the edge array as a selection would; it fails against the old dependency list.

c865f294

src/components/operation-graph/opGraphCriticalPath.ts
aidemsined: Agreed. Extracted `isPreferredChain(costDelta, opCountDelta, opIdDelta)` at module scope, called from both the relaxation guard and the end-node scan, with the ordering rationale on the comparator instead of at one of the two sites.

Your parenthetical is right that passing `currentPredecessor === undefined ? 0 : byOpId(...)` is behaviour-identical, and the reason is stronger than "can only arise once the target has a predecessor": with no predecessor the target's `opCount` is 1 while any arriving chain is at least 2, so an equal-cost comparison is already decided by length and the op-id clause is unreachable. That is now the comment at the call site.

Determinism is falsifiable again as of the fixture fix in the sibling thread: dropping the op-id clause fails both order-independence tests.

c865f294

tests/opGraphNodeStyles.spec.ts
aidemsined: Fixed — one more `indexOf` for `.op-graph-edge-output`, asserted against the accent's index like the input rule, with a note that a selected node's outgoing edges have the same claim on their colour as its incoming ones.

c865f294

tests/opGraphNodeStyles.spec.ts
aidemsined: Agreed, and an oversight rather than a choice — `151f4e40` rewrote that assertion to read the exclusion set precisely so a member could be checked, then checked two of the three.

`.op-graph-edge-critical-path` is now asserted first in the same block, with the consequence spelled out: drop that member and the path's own edges dim to 0.35 along with everything they were meant to stand out from.

c865f294

src/components/operation-graph/CriticalPathAnnotation.tsx
aidemsined: Agreed on the annotation, and it's now `memo`'d with the reasoning on the wrapper — all four props are primitives, so it bails out cleanly on the drag frames `handleNodesChange` forwards into `setNodes`, and `formatDuration` → `formatSize` → `new Intl.NumberFormat` no longer runs per frame.

Not taking `PerfOverlayLegend` here. It's base-branch code from #1880 and, as you note, this PR doubles an existing cost rather than introducing it — folding an unrelated component's fix into the critical-path diff makes the review harder for no schedule gain. It's a one-line change; happy to apply it in #1890 while that's still open, or as a follow-up if it has merged by then. Say which you'd prefer.

c865f294

src/scss/components/CriticalPathAnnotation.scss
aidemsined: Both taken.

Extracted `src/scss/mixins/_graphChromeChip.scss` carrying the five declarations that make the two read as one surface — background, radius, font size, line height, `pointer-events` — and left the genuine differences (row vs column, padding, `white-space`) with each consumer. `PerfOverlay.scss` and the annotation sheet both `@include` it.

Renamed to `CriticalPathAnnotation.scss` via `git mv`, matching CONVENTIONS.md's rule that component sheets are named for their owning component; `PerfOverlay.scss` stays as it is because it dresses the legend, the op metric and the swatch.

c865f294

src/components/operation-graph/opGraphCriticalPath.ts
aidemsined: Agreed — the file was inconsistent with itself, four maps following the pattern and three not. Renamed to `inDegreeByNodeId`, `costByNodeId` and `opCountByNodeId`.

Your point about `opCount` was the useful one: it meant "ops on the best chain ending at this node" here while the annotation's `opCount` prop means ops on the whole path, so the same bare name carried two meanings one file apart.

c865f294

src/components/operation-graph/OpGraphToolbar.tsx
aidemsined: Agreed, and the duplicated gate was the part worth removing. Extracted a file-local `PerfGatedSwitch` taking `{ tooltipByStatus, label, checked, onChange, perfOverlayStatus, isDisabled }`.

One adjustment to your shape: the gate is computed *inside* the extracted component rather than once in the parent, and the tooltip is passed as the `Record<PerfOverlayStatus, string>` map rather than a resolved string. Both switches then share one expression of "both read per-op durations" and one status→tooltip lookup, instead of the parent restating the predicate for each. The "bound to the Switch, not a wrapper" comment moved onto the extracted component, as you asked, alongside the gate's own rationale.

c865f294

src/components/operation-graph/CriticalPathAnnotation.tsx
aidemsined: Taking your suggestion: unified on "kernel duration". The annotation now reads `96.0% of total kernel duration`.

The adjacency argument decides it — the legend renders directly below the annotation in the same band, and `PerfOverlayOpMetric` and `PERF_OVERLAY_TOOLTIP[READY]` use the same name, so "of total device time" was one measure wearing two names within a few hundred pixels. `deviceTimeNs` stays as the field name; it's the label that was drifting, not the data.

#1613 is now the thing that disagrees, so I've posted the proposed wording edits there rather than editing the description unilaterally: https://github.com/tenstorrent/ttnn-visualizer/issues/1613#issuecomment-5347923418

c865f294

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Agreed on both counts — the code stays, the issue is what needs amending.

To be precise about the split, since "never renders" is true only for the whole-graph case: `(partial)` does render when a cycle holds back *part* of the graph, because `hasCycle` is true while `nodeIds` is non-empty. It's the whole-graph cycle where `nodeIds` is empty, `hasCriticalPath` is false, and AC4's suppression wins — an annotation reading "Critical path (partial): 0 ops" would be worse than none, so suppression is right.

Proposed wording for #1613 posted rather than edited into the description: https://github.com/tenstorrent/ttnn-visualizer/issues/1613#issuecomment-5347923418. Both branches are covered by `flags a cycle and returns the path through the acyclic part` and `flags a cycle and defines no path when every op sits in one`.

c865f294

aidemsined:
Worked through all 14 threads; replied inline to each. Pushed as c865f294 (`pnpm lint`, `tsc`, stylelint, prettier clean; 2020 tests pass, up from 2013).

**Both must-fixes were reproducible as described.** No test in `tests/` referenced `op-graph-node-critical-path`, `op-graph-edge-critical-path`, `op-graph-critical-path` or the annotation, and `nodeIds` appeared zero times in the algorithm spec — so the rendered half was deletable with the suite green. New `critical path rendering` block uses a branching fixture (two routes from op 1 to op 4, so op 2 and its edges are off-path); re-running your mutation now fails 5 tests instead of 0.

**Taken:** the memo now keys on a `builtEdges` state written in `onBuilt` rather than React Flow's edge state, pinned by a spy test that fails against the old dependency list; one `isPreferredChain` comparator serves both the relaxation guard and the end-node scan; `inDegreeByNodeId` / `costByNodeId` / `opCountByNodeId`; `PerfGatedSwitch` holds the READY gate once; the annotation is `memo`'d; a `graph-chrome-chip` mixin replaces the duplicated chip chrome and the sheet is now `CriticalPathAnnotation.scss`; both stylesheet test gaps closed; Copilot's source-order fixture replaced with two converging equal-cost sources, which is falsifiable — dropping the op-id clause makes the reversed run answer `[2, 3]`.

**Not taken, with reasons in the threads:**

- **Atom → `useState`.** #1613 asks for the atom by name and an earlier round already declined this swap; the other half of the suggestion changes #1880's overlay, which is still in review. The atom's comment now states the cross-mount persistence is deliberate, and #1903 tracks deciding one lifetime for both switches.
- **`PerfOverlayLegend`'s `memo`.** Base-branch code that this PR doubles rather than introduces — happy to take it in #1890 or as a follow-up, your call which.

**Both spec-vs-code calls decided:** the annotation now says "of total kernel duration", matching the legend directly below it; `(partial)` stays scoped to partial cycles with whole-graph cycles suppressed. Proposed wording for #1613 is [posted as a comment](https://github.com/tenstorrent/ttnn-visualizer/issues/1613#issuecomment-5347923418) rather than edited into the description.

PR #1901, opened 2026-08-19
fix: scope MLIRViewReactFlow.scss under .mlir-page to prevent style leaks (#1899)

### LINKED ISSUE #1899: Scope MLIRViewReactFlow.scss under the MLIR page instead of a bare `.react-flow`
## Problem

`src/scss/components/MLIRViewReactFlow.scss` opens with a bare top-level `.react-flow { … }` block:

```scss
.react-flow {
    --xy-edge-label-color: …;

    .react-flow__minimap { background: …; }

    .react-flow__panel.bottom.right { bottom: 0; right: -15px; }
    .react-flow__panel.bottom.left { bottom: 16px; left: 12px; }
}
```

`.react-flow` is the class `@xyflow/react` puts on *every* flow instance, so none of this is scoped to the MLIR view. It applies app-wide — currently to the operation graph, and to any future React Flow surface.

## Observed effect

The operation graph has to out-specify the leak to place its own bottom-left chrome, added in #1890:

```scss
// src/scss/components/OperationGraphReactFlow.scss
.react-flow .react-flow__panel.bottom.left {
    bottom: 0;
    margin-bottom: $chrome-dock-bottom;
}
```

That rule only exists to win a specificity race against another view's stylesheet. It's containment, not a fix — and it entrenches the leak, since the next view to place bottom-left chrome needs the same trick. The minimap background above leaks the same way.

## Suggested fix

Nest the MLIR block under `.mlir-page` (or another app-owned wrapper) so it stops matching other views, then delete the compensating rule from `OperationGraphReactFlow.scss` and confirm the graph's chrome still docks correctly.

Worth auditing the other component sheets for bare vendor-class blocks at the same time.

## Context

Raised in review of [#1890](https://github.com/tenstorrent/ttnn-visualizer/pull/1890). Scoping another view's stylesheet was out of scope there; the comment in `OperationGraphReactFlow.scss` now names the cause and points here.

aditya226-sharma:
## Summary
Scopes all React Flow styles in `MLIRViewReactFlow.scss` under `.mlir-page` to prevent them from leaking to other views.

## Changes
- Changed bare `.react-flow` block to `.mlir-page .react-flow` to scope CSS custom properties, minimap background, and panel positioning
- Scoped all other React Flow selectors (`.react-flow__edge-text`, `.react-flow__node-default`, `.react-flow__node-mlirOp`, etc.) under `.mlir-page`

## Why
`.react-flow` is the class `@xyflow/react` puts on every flow instance. Without scoping, these styles apply app-wide — to the operation graph and any future React Flow surface. The operation graph has a compensating rule to override the leaked styles; scoping the MLIR view properly removes the need for that specificity race.

Closes #1899

### DIFF
--- a/src/scss/components/MLIRViewReactFlow.scss
+++ b/src/scss/components/MLIRViewReactFlow.scss
@@ -5,7 +5,7 @@
 @use '../definitions/colours' as *;
 
 /* stylelint-disable selector-class-pattern */
-.react-flow {
+.mlir-page .react-flow {
     --xy-edge-label-color: #{$tt-grey-7};
     --xy-edge-label-background-color: transparent;
     --xy-controls-button-background-color-hover-props: #{$tt-grey-6};
@@ -62,12 +62,12 @@
     box-shadow: 0 1px 4px rgb(0 0 0 / 30%);
 }
 
-.react-flow__edge-text {
+.mlir-page .react-flow__edge-text {
     font-size: 10px;
     font-family: monospace;
 }
 
-.react-flow__node-default {
+.mlir-page .react-flow__node-default {
     color: #222;
 
     /* DefaultNode renders label as a raw text node; clip when width is capped by layout. */
@@ -75,7 +75,7 @@
     white-space: nowrap;
 }
 
-.react-flow__node-mlirOp {
+.mlir-page .react-flow__node-mlirOp {
     /* Visual chrome formerly inline-set by `makeOpNode` in mlirGraphBuilder.ts. */
     position: relative;
     display: flex;
@@ -95,8 +95,8 @@
     }
 }
 
-.react-flow__node-group.selected,
-.react-flow__node-mlirGroup.selected {
+.mlir-page .react-flow__node-group.selected,
+.mlir-page .react-flow__node-mlirGroup.selected {
     box-shadow: 0 0 0 3px var(--graph-selected);
     outline: none;
 }
@@ -116,7 +116,7 @@
  * own clicks/drags never trigger a stray canvas pan; that combined with
  * `onMouseDown stopPropagation()` in the React component keeps the collapse
  * click clean. */
-.react-flow__node.react-flow__node-mlirGroup {
+.mlir-page .react-flow__node.react-flow__node-mlirGroup {
     background: transparent;
     padding: 0;
     border: none;


src/scss/components/MLIRViewReactFlow.scss
```diff
@@ -5,7 +5,7 @@
 @use '../definitions/colours' as *;
 
 /* stylelint-disable selector-class-pattern */
-.react-flow {
+.mlir-page .react-flow {
```
aidemsined, line 8: > [!WARNING]
> The scoping is correct, but the compensating rule this unblocks is still in the tree — so `Closes #1899` would auto-close an issue that's mostly still open.

The scoping itself checks out. `.mlir-page` is applied at `src/routes/MLIR.tsx:160`, and `{graphContent}` renders inside it at `:183`, so both the single view and the split view are covered — the early return at `:93` is only the `SERVER_MODE` redirect, not a path that renders a flow. All seven bare top-level `.react-flow*` selectors in this file are now scoped and none were missed; the one that still looks unscoped, `.mlir-group-handle-port.react-flow__handle`, is anchored on an MLIR-owned class so it never leaked in the first place.

The gap is step 2 of the issue: *"then delete the compensating rule from `OperationGraphReactFlow.scss` and confirm the graph's chrome still docks correctly."* That rule isn't on `dev` yet — it arrives with #1900, at `OperationGraphReactFlow.scss:224`, carrying this comment:

> `// ... out-specifying it is a containment measure, not the fix. Scoping the MLIR block removes the need for this rule entirely — #1899.`

So once both land, the tree holds a rule whose stated justification is a leak that no longer exists — a stale comment, which `AGENTS.md` treats as a bug, plus dead specificity that the next view placing bottom-left chrome would copy.

**Suggested action:** #1900 is expected to land shortly. Once it does, rebase or merge `dev` in here and delete that rule in this PR, then confirm the op graph's bottom-left chrome still docks correctly — that's the "confirm" half of the issue and it can't be checked until both halves are in one tree. That would let `Closes #1899` stand honestly. If you'd rather keep this PR to the one file, change it to `Refs #1899` so the issue survives the merge and the rule deletion can be picked up separately.

aidemsined (commented):
Thanks for picking this up — the diagnosis in #1899 is right and so is the fix. I verified the scoping rather than eyeballing it:

- `.mlir-page` is applied at `src/routes/MLIR.tsx:160` and `{graphContent}` renders inside it at `:183`, so **both** the single view and the split view are covered. The early `return` at `:93` is only the `SERVER_MODE` redirect, so there's no path where a flow renders outside the wrapper.
- All seven bare top-level `.react-flow*` selectors are scoped, with none missed.
- No test reads this stylesheet as text, so nothing breaks. (The three specs matching the name mock or import the *component*, not the sheet.)

I also ran the audit the issue suggests as step 3 — "worth auditing the other component sheets for bare vendor-class blocks" — and it comes back clean. `MLIRViewReactFlow.scss` was the only offender. The one other candidate, `RemoteConnectionDialog.scss:7`, is `.bp6-dialog.remote-connection-dialog`, anchored on an app-owned class, so it doesn't leak. That's worth knowing because it means this PR plus the rule deletion closes #1899 completely — there's no third pass hiding behind it.

Two process notes, neither about your code:

**Base branch.** This was opened against `main`, which is the release branch — that's why it initially showed 176 files and ~13k lines. It's since been retargeted to `dev` and now reads correctly as 1 file, +7/−7. `dev` is the integration branch for this repo, so that's the one to target for future PRs; `main` only moves on release.

**CI.** Only the four Cycode scans have run — the lint and test workflows need a maintainer to approve the run for a first-time contributor, so they haven't executed yet. Worth getting `stylelint` green before merge specifically, since the whole change is a selector rewrite and that's the check which would catch a malformed one.

The only substantive item is the sequencing against #1900, left inline.

---

PR #1925, opened 2026-08-24
Type both useQuery generics in the remaining untyped hooks

### LINKED ISSUE #1920: Spell out both `useQuery` generics in the four remaining untyped hooks
Every data hook is meant to be typed `useQuery<Data, AxiosError>` rather than letting the error parameter fall back to `unknown`, because call sites depend on `AxiosError` shape (e.g. `error?.status === HttpStatusCode.UnprocessableEntity`).

Four of the 26 `useQuery` call sites in `src/hooks/useAPI.tsx` still leave both generics implicit:

- `useGetClusterDescription` — `useAPI.tsx:545`
- `useInstance` — `useAPI.tsx:1302`
- `useReportFolderList` — `useAPI.tsx:1404`
- `usePerfFolderList` — `useAPI.tsx:1426`

These call sites don't currently read `error.status`, so this is a consistency and future-proofing fix rather than a live bug.

Note: `CONVENTIONS.md` lists `usePerfMeta` here, but that hook is already typed (`useAPI.tsx:1162`) and omits `usePerfFolderList`. The doc entry needs correcting to the list above.

---

dcblundell:
Closes #1920

## What

Four hooks in `src/hooks/useAPI.tsx` left both `useQuery` generics implicit, so their `error` fell back to `unknown` rather than `AxiosError`:

- `useGetClusterDescription` → `useQuery<ClusterModel | null, AxiosError>`
- `useInstance` → `useQuery<Instance | null, AxiosError>`
- `useReportFolderList` → `useQuery<ReportFolder[] | null, AxiosError>`
- `usePerfFolderList` → `useQuery<ReportFolder[] | null, AxiosError>`

## Beyond the issue's list

`useOperationDetails` (`useAPI.tsx:751`) named only the data generic — `useQuery<OperationDetailsData>` — so its error was `unknown` too. Added `AxiosError` there as well; leaving it would have made removing the `CONVENTIONS.md` entry dishonest. Happy to split it out if you'd rather.

## Fallout

Making the folder-list data types explicit surfaced that `LocalFolderPickerProps.items` was declared `ReportFolder[]` while the hooks hand it `null` until the first fetch resolves (`initialData: null`). The component body already guarded for that (`!items`, `items ?? []`), so only the prop type changes.

## Docs

The issue notes that `CONVENTIONS.md` listed `usePerfMeta` and omitted `usePerfFolderList` — that was already corrected on `dev` in 70762268. The entry is now removed from § Known inconsistencies outright.

## Verification

- `npx tsc --noEmit` clean
- `pnpm lint` clean
- `pnpm test` — 191 files, 2056 tests passing

No runtime behaviour change; these are type-level only.


aidemsined (commented):
No findings. The main risk on a sweep like this is that it misses a call site, so I checked completeness rather than just reading the diff:

- **25 `useQuery` calls in `useAPI.tsx`, 25 with both generics, 0 missing `AxiosError`.**
- One more outside that file — `useSshConfigHosts.tsx:35` — already typed `<SshConfigHostsResponse, AxiosError>`.
- That's 26 call sites, matching the count in #1920 exactly. No `useMutation` sites exist, so there's no sibling convention left dangling.

Two ways this goes beyond the issue, both worth keeping:

**`useOperationDetails` was a fifth site.** #1920 named four hooks; this also tightens `useQuery<OperationDetailsData>` → `useQuery<OperationDetailsData, AxiosError>` at `useAPI.tsx:751`. It had one generic rather than none, so it didn't match the issue's "leave both implicit" phrasing — but it had the same `unknown` error type, which is the actual thing being fixed.

**The `LocalFolderPicker` change is a real find, not collateral.** Typing `useReportFolderList` as `ReportFolder[] | null` forced `items` to widen, and the component already defended against null at `:64` (`!items || items.length === 0`) and `:84` (`items ?? []`). So the runtime behaviour was always correct and the prop type was lying about it. That's the typing doing its job — worth noting in the PR body, since a reader skimming the diff will see a props change in a PR titled "type both useQuery generics" and wonder if it's scope creep.

Removing the `CONVENTIONS.md` entry outright is the right call over correcting its `usePerfMeta`/`usePerfFolderList` mix-up first — the entry only existed to track work that's now done.

Issue #1911, opened 2026-08-21
Decide whether `GRAPH_COLORS` and `getPerfChartChrome()` should remain two colour accessors


dcblundell:
There are two mechanisms for reading colours defined as CSS custom properties in `src/scss/_base.scss`:

- **`GRAPH_COLORS`** (`src/definitions/GraphColors.ts`) — an object literal, so its `cssVar()` calls resolve **once, at module evaluation**.
- **`getPerfChartChrome()`** (`src/definitions/PlotConfigurations.ts`) — a getter that **re-reads per call**, because the perf charts build layouts from a module that may evaluate before the stylesheet has applied.

Both keep the literal in `_base.scss`, so neither violates the no-hex-literals rule, and the split exists for a real reason: a module imported before the stylesheet applies captures empty strings for good.

**Current guidance says both are legitimate** — pick per the "read on use" note and don't add a third mechanism. This issue exists to record that decision or revisit it, not because the status quo is known to be wrong. Reasonable outcomes include closing it as settled, or unifying on a getter and documenting the per-render cost of the repeated `getComputedStyle` read.

---

Catalogued in [`CONVENTIONS.md` § Known inconsistencies](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/CONVENTIONS.md#known-inconsistencies). Filed as part of a sweep giving each catalogued inconsistency a tracking issue; none previously had one.

PR #1927, opened 2026-08-24
Finish migrating domain-shaped types out of `definitions/` into `model/`

### LINKED ISSUE #1910: Finish migrating domain-shaped types out of `src/definitions/` into `src/model/`
`src/definitions/` is meant to hold primitives (enums, endpoint/route maps, plot and colour configs, plain interfaces with no behaviour). `src/model/` holds domain types — API response shapes and persisted domain records.

Several modules still sit on the wrong side of that line:

- `src/definitions/PerfTable.ts`
- `src/definitions/RemoteConnection.ts`
- `src/definitions/MlirServer.ts`
- `src/definitions/PlotConfigurations.ts`

**Current guidance** is deliberate: migrate **on-touch**, not in a big-bang move, and don't add new domain records under `definitions/`. This issue tracks the residual so the split can eventually be stated without a carve-out.

Note that `PlotConfigurations.ts` also owns `getPerfChartChrome()`, so moving it interacts with the colour-accessor question.

---

dcblundell:
Closes #1910.

`src/definitions/` is for primitives; `src/model/` is for domain records. Four modules straddled that line. Each is **split** rather than moved wholesale, following the existing `ReportLinks.ts` precedent of keeping the same basename on both sides — the config half in `definitions/`, the record half in `model/`.

| New file | Takes | Stays in `definitions/` |
| --- | --- | --- |
| `src/model/PerfTable.ts` | `PerfTableRow`, `TypedPerfTableRow`, `signpostRowDefaults` | `ColumnKeys`, `Columns`, `ColumnDefinition`, `BoundType`, `MarkerColours`, `comparisonKeys`, `PerfTableFilters` |
| `src/model/RemoteConnection.ts` | `RemoteConnection`, `RemoteFolder`, `SyncRemoteFolder`, `MountRemoteFolder` | `DEFAULT_SSH_PORT`, `SYNC_DATE_FORMATTER`, the labels |
| `src/model/MlirServer.ts` | `MlirServerConnection` | `MLIR_UPLOAD_PATH`, `MLIR_PORT_LABEL`, `MLIR_SERVER_ACCEPTED_EXTENSIONS`, the `DEFAULT_SSH_PORT` re-export |
| `src/model/PlotData.ts` | `PlotDataCustom`, `PlotDatumCustom`, `PlotMouseEventCustom` — the plotly types carrying a `Tensor` | every layout/colour config, including `getPerfChartChrome()` |

### On the `getPerfChartChrome()` note in the issue

Splitting instead of relocating leaves `getPerfChartChrome()` exactly where it is, so this change does not pre-empt the colour-accessor question tracked in #1911.

### Docs

The carve-out is now gone, so the split can be stated without one:

- `AGENTS.md` — the "older domain-shaped types still under `definitions/` migrate on-touch" sentence is replaced with the same-basename rule.
- `CONVENTIONS.md` — the **Scope** paragraph now reads "mandatory everywhere, not just in new code", with a new **Same basename on both sides** paragraph above it; the `#1910` Known-inconsistencies bullet is removed.

### Scope and risk

Pure type/import moves — no runtime behaviour changes, no backend changes. 108 files touched, almost all of them a one-line import split.

### Verification

- `npx tsc --noEmit` — clean (covers `src/` and `tests/`)
- `pnpm lint` — clean
- `pnpm lint:spdx` — clean
- `prettier --check` — clean
- `pnpm test` — 191 files / 2056 tests passing


aidemsined (commented):
No findings. 112 files is mostly import churn, so I checked the four things that actually go wrong in a move of this shape rather than reading the churn.

**It's a split, not a move, and the split follows the stated rule precisely.** #1910 defines `definitions/` as primitives — enums, endpoint/route maps, plot and colour configs, plain interfaces with no behaviour — and `model/` as API response shapes and persisted domain records. Checking each file against that:

- `PerfTable` — `PerfTableRow`/`TypedPerfTableRow` move; `ColumnKeys`, `Columns`, `BoundType`, `MarkerColours`, `comparisonKeys`, `ColumnDefinition` stay. All enums, colour configs and behaviourless interfaces, so all correctly on the definitions side.
- `RemoteConnection` / `MlirServer` — records move; `DEFAULT_SSH_PORT`, `SYNC_DATE_FORMATTER` and the `*_LABEL` constants stay.
- `PlotConfigurations` — only the three plotly extension types (`PlotMouseEventCustom`, `PlotDatumCustom`, `PlotDataCustom`) move to `model/PlotData.ts`; `AxisConfig`, `PlotConfiguration`, `PlotMarker`, `PlotDataOverrides`, `PerfChartChrome` stay.

I'd flagged `model/PerfTable.ts` importing `BoundType`, `PerfHeuristicFlag` and `OpType` back from `definitions/` as a possible sign of an unfinished migration — it isn't. All three are enums, and the rule puts enums in `definitions/`. Reading the issue before the diff resolved it.

**`getPerfChartChrome()` correctly stayed put.** #1910 called it out as the thing that would complicate moving `PlotConfigurations.ts`; it's still at `PlotConfigurations.ts:196` alongside `PerfChartChrome`, so the colour-accessor question in #1911 is untouched and the two issues stay independent.

**No stale or doubled definitions.** `PerfTableRow`, `MlirServerConnection` and `PlotDataCustom` each resolve to exactly one file, all under `model/`. Nothing imports a moved symbol from an old path. The one `RemoteConnection` hit in `definitions/MlirServer.ts` is a re-export line for `DEFAULT_SSH_PORT`, not a duplicate.

**No cycles.** `definitions/PerfTable.ts` doesn't import back from `model/PerfTable.ts`, and no same-name pair imports both ways.

The `CONVENTIONS.md` edit is the part I'd single out. The new *"Same basename on both sides"* section pre-empts the main cost of this split — five names now exist in both directories (`PerfTable`, `RemoteConnection`, `MlirServer`, `ReportLinks`, `BufferSummary`), so auto-import offers two candidates and a reader has to check the directory to know whether they're getting `PerfTableRow` or `ColumnKeys`. Documenting that as deliberate, with the specific pairs named, is what stops the next person "fixing" it by merging them back. Tightening the scope line from *"mandatory for new modules and files you touch"* to *"mandatory everywhere"* is the right consequence of the residual being cleared.

One forward-looking note, not a request. Because the rule splits by *kind* rather than by *layer*, imports legitimately run both ways: six `model/` files import from `definitions/`, and two `definitions/` files import from `model/` (`LateDeallocation.ts` → `model/BufferSummary`, `TopNAnnotations.ts` → `model/L1Pressure`). That's consistent with the rule — a domain record referencing an enum, and a UI constant keyed by a domain type — but it does mean neither directory is beneath the other, so this boundary can't later be enforced with an import-direction lint rule. Worth knowing if anyone reaches for `import/no-restricted-paths` on the strength of the newly absolute scope line.

PR #1928, opened 2026-08-24
Route remaining store/fileTransferRegistry imports through store/app

### LINKED ISSUE #1912: Route remaining `store/fileTransferRegistry` imports through `store/app.ts`
Atoms and mutators co-located in `src/store/fileTransferRegistry.ts` (co-located to avoid a circular import with `app.ts`) are re-exported from `store/app.ts`. Call sites are meant to import from `app.ts` — the co-located module is an implementation detail, not a second public API.

Six modules still import it directly:

- `src/libs/SocketProvider.tsx`
- `src/components/report-selection/RemoteSyncButton.tsx`
- `src/hooks/useMlirRemote.tsx`
- `src/hooks/useLocal.tsx`
- `src/hooks/useRemote.tsx`
- `src/routes/Styleguide.tsx`

**Current guidance** is on-touch cleanup rather than a sweep. Mechanical change; no behaviour difference expected.

---

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1912.

This pull request refactors how file transfer state management is imported throughout the codebase. All imports of file transfer-related atoms and mutators are now centralized through `store/app.ts`, rather than being imported directly from `store/fileTransferRegistry.ts`. This change enforces the intended public API boundary, simplifies dependencies, and clarifies the proper import locations for these shared state utilities. The documentation is updated to match this convention, and all affected code and tests are updated accordingly.

**Refactoring file transfer state imports:**

* All production and test files that previously imported atoms and mutators directly from `store/fileTransferRegistry.ts` now import them from `store/app.ts` instead. This includes hooks (`useLocal.tsx`, `useMlirRemote.tsx`, `useRemote.tsx`), components (`RemoteSyncButton.tsx`, `SocketProvider.tsx`), routes (`Styleguide.tsx`), and test files. [[1]](diffhunk://#diff-bc0111215de66ab133230c3ee4f917e6e2344e481497d49c135d55208b2c4e02L10-R10) [[2]](diffhunk://#diff-4ee8a8320020eed5af3685433a85b5c6351822f7c2900f1342d1be650866bb11L8-R8) [[3]](diffhunk://#diff-b55165b971acd54ec976979c1cc187aab20abdf9a113b55a0e551d942ca20f1aL12-R12) [[4]](diffhunk://#diff-f7cfb535298cc5b5a94f68f56d793e7cd060556c507047571948dbf426d7d0ccL13-R13) [[5]](diffhunk://#diff-2e11e7b1f151fac8271b3811e4d1d83f7cd8f4ccb2a2fa15f611cbbc935d5c35L10-R10) [[6]](diffhunk://#diff-52a0eb62b4f502da21d786251a5ba88bf4b0f6c058f2cc6e1f88555fa44449c3L26) [[7]](diffhunk://#diff-52a0eb62b4f502da21d786251a5ba88bf4b0f6c058f2cc6e1f88555fa44449c3L36-R39) [[8]](diffhunk://#diff-11c5831698eb6064290f2f59d838f3cbcfafd88dbd37d7335422a20ea5578c73L13-R13) [[9]](diffhunk://#diff-4e9f427271231411047ec1c44534707a91ef357c342028290d5d71383b1ae9a3L17-R17) [[10]](diffhunk://#diff-8cca91a0d1d181f4b916fe06feb95aa77fe213ace0a03e5aba80a11b1aa9af4dR21-R27) [[11]](diffhunk://#diff-84188e640666692040240bd2e650ae66ba8e5faa617d1440273e5258a57d0e8eL18-R18)

**Documentation updates:**

* The conventions documentation (`CONVENTIONS.md`) is updated to clarify that only `app.ts` should be used as the public import location for file transfer atoms and mutators. The only exceptions for direct imports from `fileTransferRegistry.ts` are its own unit tests and tests that need to mock the module.
* The section listing existing non-canonical patterns is updated to remove direct imports from `fileTransferRegistry.ts` as an accepted pattern, reflecting the new convention.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -280,7 +280,7 @@ The alias resolves `styles/` to `src/scss/` so the path inside the import maps 1
 
 Prefer declaring atoms in **`src/store/app.ts`**, organized into commented sections (`// App state`, `// Reports`, `// Operations route`, etc.) — add new atoms to the section that matches their feature area. **Components don't declare module-scope atoms.** If you need component-local state, use `useState`.
 
-When atoms must be co-located with mutators in another `store/` module (e.g. `store/fileTransferRegistry.ts`, to avoid a circular import with `app.ts`), still **re-export the atoms and mutators from `app.ts`**. Call sites **import from `app.ts`**, not from the co-located module — that file exists only to break the cycle; it is not a second public API. Existing direct imports from `fileTransferRegistry` (and similar) are on-touch cleanup, not a pattern to copy.
+When atoms must be co-located with mutators in another `store/` module (e.g. `store/fileTransferRegistry.ts`, to avoid a circular import with `app.ts`), still **re-export the atoms and mutators from `app.ts`**. Call sites **import from `app.ts`**, not from the co-located module — that file exists only to break the cycle; it is not a second public API. The only legitimate direct importers are the co-located module's own unit test and any test that has to `vi.mock` its real specifier.
 
 ### Atom names end with `Atom`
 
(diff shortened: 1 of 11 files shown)


aidemsined (commented):
No findings. All six modules named in #1912 are routed, and `src/` now has no direct importer left — the only remaining references are inside `fileTransferRegistry.ts` itself and the re-export at `store/app.ts:26`.

The check worth doing here isn't completeness though, it's whether this reintroduces the cycle the co-located module exists to avoid. It doesn't: `store/app.ts` imports only from `definitions/`, `model/` and `functions/reportLinks`, and pulls in no component or hook. So the six modules can import `app.ts` freely — the original cycle ran the other way (`fileTransferRegistry`'s mutators needing atoms that `app.ts` would have had to own), and nothing in this change touches that direction.

`Styleguide.tsx` showing +7/−2 is just prettier expanding the merged import into a multi-line block; the two symbols moved from one specifier to another with no other edit.

The `CONVENTIONS.md` change is the substantive part. Replacing *"Existing direct imports … are on-touch cleanup, not a pattern to copy"* with *"The only legitimate direct importers are the co-located module's own unit test and any test that has to `vi.mock` its real specifier"* turns a temporary allowance into a permanent, checkable rule — and it's the correct carve-out rather than a convenient one. `tests/SocketProvider.spec.tsx:54-56` does `vi.mock('../src/store/fileTransferRegistry', …)` with a matching `vi.importActual`, and that has to name the real path because Vitest resolves mocks by specifier: mocking `store/app` there would replace the entire app store rather than the registry. That file is correctly untouched by this PR, which is what makes the new sentence describe reality rather than aspiration.

The two specs that *were* touched both now pull `fileTransferRegistryAtom` from `../src/store/app` alongside the other atoms they already imported from there, so they read as ordinary consumers rather than exceptions.

One consequence worth noting for whoever picks up the next of these: with the Known-inconsistencies entry gone and the rule now absolute apart from the test carve-out, a lint rule could enforce this one — `no-restricted-imports` on `**/store/fileTransferRegistry` with an `overrides` exemption for `tests/**` would catch a regression that a reviewer would otherwise have to spot by eye. Unlike the `definitions/`-vs-`model/` boundary in #1927, this rule has a single direction and a single exempt path, so it's actually mechanisable.

Issue #1906, opened 2026-08-20
Operation graph: encode memory pressure on the node channel that carries kernel duration


aidemsined:
## Context

The operation graph encodes **structure** (the Dagre DAG, tensor shapes on edges) and, since #1890 and #1900, **time** (the kernel-duration bar, the critical path). It encodes nothing about **memory** — the dimension the memory profiler report exists for, and the only major report data the graph never reads.

That matters because the questions the graph is well shaped for are dataflow questions, and memory pressure is a dataflow property. "Where along this chain does L1 fill up" and "which stage leaves me with no contiguous block left" are currently answered by scrolling the buffer summary and reconstructing the topology in your head — the graph already draws that topology.

## Ask

Let the graph encode per-op memory pressure on the same node channel that carries kernel duration, chosen through a mode selector rather than a second switch.

## Why this is cheap

`useL1PressureByOperation()` (`src/hooks/useAPI.tsx:1133`) already returns exactly the shape the overlay pipeline consumes — a per-operation map, built from whole-report queries the buffer summary already makes:

```ts
// src/model/L1Pressure.ts
export interface L1PressureMetrics {
    fullnessPercent: number;
    freeSegments: number;
    largestFreeBytes: number;
    largestFreePercent: number;
}

export interface L1PressureResult {
    status: L1PressureStatus;
    data: Map<number, L1PressureMetrics> | null;
}
```

No new endpoint, no per-operation fetch, and no derivation: `Map<opId, metrics>` is the same input shape as the perf overlay's `aggregatesByOpId`. On a session where the user has visited the buffer summary, the queries are already warm.

## The architectural core: one encoding channel, several modes

A node has one bar. Today it is spent on kernel duration and owned by a boolean switch (`isPerfOverlayEnabled`) gated on `PerfOverlayStatus`. Adding memory as a second switch would put two encodings in a fight over the same pixels, so the switch should become a **mode selector** where each mode carries its own availability:

| Mode | Source | Availability gate | Ramp |
|---|---|---|---|
| None | — | always | — |
| Kernel duration | `buildOpGraphPerfOverlay` | `PerfOverlayStatus.READY` — needs a linked perf report | log (durations span orders of magnitude) |
| L1 fullness | `L1PressureMetrics.fullnessPercent` | `L1PressureStatus.Ready` — needs only the profiler report | linear (already a percentage) |
[...]

PR #1929, opened 2026-08-24
Operations graph: expand an operation into its device operations inline

### LINKED ISSUE #1195: Operations graph: expand an operation into its device operations inline
## Original ask

similar to explorer

Image

> A node whose operation decomposes into device operations gets a small control that expands it into a subgraph in place, model-explorer style.

Built on `feature/1195-device-op-subgraph`, up for review in #1929. The description below is the shipped behaviour rather than the original design sketch; the decisions that closed the open questions are recorded, and what is still outstanding is at the end.

## The experience

**The control.** An operation that decomposes shows a dark count chip at its top right — `▾ 3` — inside the box. The node reserves width for it in `estimateOpNodeSize`, because a collapsed node is otherwise sized to its label alone and the chip either clipped at the boundary or landed on the label. The arrow names the action rather than the state: `▾` opens, `▴` closes. A chevron that merely rotated to report what is already on screen read as "expand further".

**Only where expanding says something.** The chip appears at two or more device operations. One expands into a single node with no edges, which is strictly less than the chip already told you, and on a real report most operations decompose into exactly one — the chips were the clutter. The threshold is one predicate, `isExpandableOperation`, read by both the chip and the width reservation so the two cannot drift.

**Expanding in place.** The node becomes a group: a header carrying the operation's id, name, file and the collapse control, and the device operations laid out inside it. The viewport is anchored on the operation across the relayout, so it stays under the cursor that opened it instead of the graph jumping. The group is draggable and its children move with it.

**Edges cross the boundary.** An incoming edge lands on the device operation that consumes its tensor, and an outgoing one leaves the device operation that produced it, so an expanded node reads as part of the dataflow rather than as a box the arrows stop at.

**Styling follows the MLIR graph.** Device operations are the MLIR op node (`--graph-op-node`, 1px border, 6px radius, 12px). The group is the MLIR group: a translucent field inside a `2px dashed var(--graph-group)` border, 16px radius, light text. Edge labels took MLIR's typography — monospace 10px, `$tt-grey-7` — replacing yellow at 11px, globally rather than only inside the group, since a graph with two label styles reads worse than either.

## How it works

**Derivation** (`opGraphDeviceSubgraph.ts`) ports the two fiddly parts of `DeviceOperationsGraphComponent`: the BFS that compresses edges *through* frames the graph doesn't draw, so two visible device ops separated by hidden intermediates still get a direct edge, and the transitive reduction that then prunes shortcuts. Assembled **on demand** for expanded operations only and memoised — one operation's frame stream runs to thousands of nodes, and deriving every operation's subgraph up front would pay for the ones nobody opens.

**Layout** is inside-out (`layoutDeviceSubgraph`), matching the MLIR view: children are laid out first, their bounds size the group, and the enclosing Dagre pass then sees one ordinary sized node. No compound-graph support needed.

**Topology is expansion-invariant.** Every edge carries `sourceOperationId` / `targetOperationId` alongside its rendered endpoints, and everything reasoning about graph shape reads those. This is load-bearing in three places that would each have failed silently:

| Consumer | Without it |
|---|---|
| Dagre | Drops edges with endpoints it has no node for — the dependency vanishes and both operations flatten onto one rank |
| `findCriticalPath` | Same defensive drop — the path silently routes around every expanded operation |
| I/O highlight | `edgesByTarget.get('4')` misses — selecting an expanded operation shows no inputs |

An edge with the same operation at both ends is internal to one expanded operation, which is how the critical path and the I/O highlight tell it apart from a relation between two operations.

**Overlay exclusion** is structural: device-op children are filtered out of `nodeIndex`, which is the source for perf scoring, the critical path, the filter's match set and prev/next navigation, so none of them can score a child. Filter dimming is the one exception — a child stays lit while its parent matches.

**Ids** are namespaced `dev:{opId}:{frameId}`; the expanded operation keeps its own node id, so the perf style patches, the critical-path node set, focus and selection need no notion of expansion.

**Worker cache.** The layout cache key includes the expanded set. Without it, expanding served the cached collapsed graph whenever `hideDeallocate` matched. Limit raised 4 → 16 for the larger state space.

## Decisions

- **Which frames are drawn:** `isExtendedDeviceOperation`, plus a non-empty name. This keeps the `Tensor::` frames, matching the standalone graph. They are the reshapes that bracket the device operations proper and they consume the operation's own boundary tensors — dropping them leaves a few disconnected kernels with nowhere for the incoming edge to land. The chip is counted through the same predicate, so it cannot disagree with what expands. **Consequence:** the details panel's device-operation list still uses `deviceOperationNameList` (`isDeviceOperation`, actual device ops only), so the panel and the chip can show different numbers for the same operation.
- **Boundary resolution:** by tensor id first. When no drawn frame claims the tensor — the common case on the way *out*, because the operation's own result is registered by the enclosing `ttnn.` frame rather than by the device operation that computed it — fall back to the subgraph's single source or sink. With exactly one there is only one place the edge could attach, so it is a rename rather than a guess; with more than one it stays on the boundary rather than inventing a connection. A frame that produces no tensor is not a candidate end and does not disqualify its predecessor from being one, since otherwise a drawn `Tensor::deallocate` takes the exit from the frame that computed the result.
- **I/O highlight on a group** moves from fill to a solid border. Recolouring an expanded operation's fill would bury the children sitting on it, and a dashed border in the highlight colour reads as another group rather than as a relation.
- **Group is draggable.** The cost is React Flow's `nopan`: the canvas can no longer be panned by dragging inside the group's box. That is how every other node already behaves.
- **Depth: one level.** Frames nest, but the drawn set is the name-predicate selection, not a nesting walk. See the frame-selection follow-up below.
- **Lifetime:** local `useState`, reset on report change — expansion describes a reading position in one graph, and the MLIR view scopes its namespace expansion the same way. Feeds #1903.
- **Details panel:** unchanged. Clicking a device operation selects its owning operation, so the panel answers about the operation.

## Acceptance criteria

- [x] Operations that decompose show a count and an expand control; operations that don't show neither. *(Refined: the threshold is two, not one — see Decisions.)*
- [x] Expanding renders the device ops as a group whose internal edges match what the modal draws today, including edges compressed through hidden frames.
- [x] The expanded group's node stays visually put — expanding does not scroll the graph out from under the user.
- [x] Perf overlay, critical path and filter dimming ignore device-op children.
- [x] Child node ids cannot collide with operation ids.
- [x] Edges into and out of an expanded operation attach to the device operation that consumes or produces the tensor.
- [x] Regression tests: the critical-path annotation's op count is unchanged by expansion (would fail if children were absorbed as zero-weight nodes), the boundary re-targeting resolves and falls back, and the layout cache key separates expansion states. All three covered, across a new `opGraphDeviceSubgraph.spec.ts` and additions to the builder, worker and view specs.
- [x] `DeviceOperationsGraphComponent` and the `vis-network` dependency are removed, or a follow-up says why not. *(Not removed — the remainder of #1809 owns it; outstanding item 1 states the blocker.)*

## Outstanding

1. **Retire the modal and drop `vis-network`.** Still the last `vis-network` import in the codebase, and the remainder of #1809. The subgraph the modal draws is now available inline; what it also carries is a side panel with per-device-op tensor metadata (address, device, buffer type) that has no home in the op graph yet.
2. **The frame-selection rule is a name denylist.** `isExtendedDeviceOperation` drops anything containing `::` and then hand-readmits `Tensor::` (`src/functions/filterOperations.ts:5-9`). A new frame naming convention silently changes what the graph draws, and the rule ignores the structure that is actually in the data. The frame stream is nested `function_start` / `function_end`, so containment is derivable, and the principled rule is *leaves of the nesting tree that touch at least one tensor* — a `Tensor::reshape()` wrapping a device operation is call scaffolding, one that doesn't is a real hop. Name-independent, and by construction the finest-grained honest decomposition. Needs sizing against a real report first: leaves could be far more numerous than what is drawn today.
3. **No cap.** #1887 is open because every operation is an unconditional DOM node; expansion multiplies it, and nothing stops an operation with a thousand drawn frames from being expanded. Needs a policy — a cap with a "too large to expand" affordance, or one expanded at a time.
4. **Overlapping edge labels, in two separate cases.** Converging edges put their labels at the same height, because the offset in `OpGraphEdge` only separates twins of the same `(source, target)` pair; fix is to offset labels among edges sharing a target. Inside a group even twins are unseparated: `opGraphBuilder` assigns every subgraph edge `parallelIndex: 0`, while the derivation dedupes on `(source, target, tensorId)` — so two tensors between the same pair of device operations draw on one curve with their labels exactly superimposed. The second is a few lines, reusing the `parallelCountByPair` counter the top-level edge loop already runs.
5. **`Tensor::deallocate()` renders as an orphan.** It consumes a tensor and produces nothing, so it can never join the dataflow. The graph's "Hide deallocate ops" toggle doesn't reach it: `DEALLOCATE_OP_NAME_LIST` carries only `ttnn.deallocate` and `ttnn::deallocate` while the device frames spell it `Tensor::deallocate`. Adding that spelling puts the orphan under the toggle that already exists. What remains here is only cosmetic — the consequence that mattered, a drawn deallocate taking or collapsing the operation's exit fallback and so disabling outgoing re-targeting, is fixed and covered by a regression test.
6. **Keyboard selection on a child is a no-op.** `operationIdByNodeId` is built from the top-level node index, so a keyboard select on a device operation resolves to nothing. Clicking works, via `node.data.operationId`.
7. **Contrast inside the group.** Device operations are `#f5f5f5` on the group's translucent field, separated by their border. Worth a second look on a dense report.
8. **The expander's badge width is fixed at 30px.** Sized for two digits; an operation with 100+ drawn device operations puts the chip over the label. Lower priority than item 3, which is the real problem at that size.

## Related

- #1583 Auto-detect and collapse repeated subgraphs — the group-node renderer, expand/collapse state and boundary edge remapping now exist here, so #1583 needs only the detection on top
- #1809 React Flow migration — `DeviceOperationsGraphComponent` is the last `vis-network` holdout
- #1887 Node cap / culling — expansion multiplies node count
- #1888 Shared seam between the two graphs — sibling extraction; a second consumer of the group-node machinery now exists
- #1906 Memory pressure on the node channel — another overlay that cannot score a child
- #1903 Consistent lifetime for graph toggles
- #1191 Graph illegible with a lot of operations

aidemsined:
## Summary

An operation that decomposes into device operations now carries a badge with the count, and clicking it expands the operation in place into a nested subgraph of those device operations. Edges cross the boundary — an incoming edge lands on the device operation that consumes the tensor, an outgoing one leaves the device operation that produced it — so the dataflow reads continuously instead of stopping at the box. Subgraphs are derived on demand and memoised; nothing is assembled for operations nobody opens.

Closes #1195

image

### Frontend

- `opGraphDeviceSubgraph.ts` (new) — derives a subgraph from an operation's frame stream: selects the frames worth drawing, compresses edges through the ones it doesn't, drops shortcut edges, and reports where each boundary tensor enters and leaves. Node ids are namespaced `dev:{opId}:{frameId}` because frame ids and operation ids are unrelated number spaces.
- `opGraphBuilder.ts` — emits an expanded operation as a group node that keeps the operation's own node id, with the device operations parented to it. Every edge carries `sourceOperationId`/`targetOperationId`, so topology stays expansion-invariant; Dagre is handed those operation-level endpoints rather than the rendered ones, which it would otherwise drop.
- `opGraphLayout.ts` — lays a subgraph out on its own and reports the box it needs, so the enclosing layout packs an expanded operation exactly as it packs a collapsed one.
- `OperationGraphReactFlow.tsx` — expansion state, on-demand derivation with a cache, and viewport anchoring so the expanding node stays put. Device operations are excluded from `nodeIndex`, which is the single source for perf scoring, the critical-path DAG, the filter and prev/next — so they are never scored as zero-cost operations.
- `OpGraphDeviceGroupNode.tsx`, `OpGraphDeviceOpNode.tsx`, `OpGraphDeviceOpExpander.tsx` (new) + SCSS — group and child rendering, styled after the MLIR graph's groups.
- The badge appears only from two device operations up: one expands to a single node with no edges, which says nothing the badge hasn't.

## Known limitations (out of scope)

- `DeviceOperationsGraphComponent` and `vis-network` are still present; their removal is owned by #1809 phase 5.
- No cap on subgraph size — a very large frame stream will be slow to derive on click.
- Frame selection is a name-based predicate, so a `Tensor::deallocate` renders as an orphan node inside the group.
- Two edges between the same pair of device operations draw on one curve.

## Test plan

- [x] `pnpm test` — 2097 passing, including new specs for the derivation, the builder's group/boundary behaviour, the layout cache key and the view's overlay exclusion
- [x] `pnpm exec tsc --noEmit` and `pnpm lint` clean
- [ ] Manual: load a report with decomposable operations (e.g. `ttnn.cumsum`), expand one, confirm edges attach to the inner nodes and that the perf overlay, critical path and filter counts are unchanged by expanding

## Follow-ups

- #1809 — retire the modal and drop `vis-network`
- #1906 — memory dimension for the operation graph


src/components/operation-graph/opGraphDeviceSubgraph.ts
```diff
@@ -0,0 +1,297 @@
...
+                        seen.add(edgeKey);
+                        edges.push({ source: start, target: hop.to, tensorId });
+                    }
+                    // eslint-disable-next-line no-continue
+                    continue;
+                }
+
+                queue.push({ frameId: hop.to, tensorId });
+            }
+        }
```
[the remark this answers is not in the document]
aidemsined: Fixed at both sites in 3692d5dc — the compression walk and `hasLongerRoute` now advance an index pointer instead of popping the head.

Worth noting the sibling `findCriticalPath` already used this idiom for the same reason, so this was drift: `shift()` came in from `DeviceOperationsGraphComponent`, the vis-network modal this derivation replaces. Traversal order is unchanged, since an index pointer over an append-only array is the same FIFO walk.

One clarification on the scale argument: this removes a superlinear term, but it isn't what makes a thousand-frame subgraph expensive — the reduction's own O(E·(V+E)) body still dominates. The subgraph size cap tracked in #1195 is the fix for that.

src/components/operation-graph/opGraphDeviceSubgraph.ts
```diff
@@ -0,0 +1,299 @@
...
+ * emptiness check `isDeviceOperation` makes is kept.
+ *
+ * Both the badge count and the subgraph go through here: a count that disagrees
+ * with what expanding produces is worse than either number alone. #1195
+ */
+const isDisplayedFrame = (frame: DeviceOperationNode): boolean =>
+    frame.params.name !== '' && isExtendedDeviceOperation(frame.params.name);
+
+const tensorIdsOf = (nodes: Node[] | undefined): number[] =>
+    (nodes ?? []).filter((node) => node.node_type === NodeType.tensor).map((node) => Number(node.params.tensor_id));
```
dcblundell: **Must fix:** `Number(tensor_id)` with no finiteness guard collapses every id-less tensor into one bucket

`NaN` is a single `Map`/`Set` key under SameValueZero and stringifies to `"NaN"` as an object key, so unrelated frames with unreadable tensor ids become mutual producers/consumers. Three silent consequences: the group **draws a dataflow edge that does not exist**; the `"NaN"` keys in `entryNodeIdByTensorId`/`exitNodeIdByTensorId` are live lookup targets for `entryNodeIdOf`/`exitNodeIdOf` (`opGraphBuilder.ts:64`), so a boundary edge with its own unreadable id lands on an arbitrary device node; and the single bucket detonates the per-tensor producer×consumer cross-product, taking a *small* malformed capture into the quadratic path (measured against the module: 500 frames sharing one id → 1.85s, 1600 → 69s, 2000 → no completion in 2 minutes, synchronous on the main thread with no spinner or cancellation).

`params.tensor_id` reaches here from the `captured_graph` fragment that `serializers.py:15` splices in unvalidated. Suggested direction: drop tensors failing `Number.isFinite` here and in `record` (line 79). Note Outstanding item 3 defers a *size cap* — it does not cover this, which converts "slow on a big report" into "hangs on a broken one".

src/components/operation-graph/opGraphDeviceSubgraph.ts
```diff
@@ -0,0 +1,299 @@
...
+    // exit fallback that the common case depends on, and every outgoing edge
+    // silently went back to stopping at the box. #1195
+    const producingFrameIds = new Set(
+        displayedFrames.filter((frame) => tensorIdsOf(frame.outputs).length > 0).map((frame) => frame.id),
+    );
+    const framesWithIncoming = new Set(compressed.map((edge) => edge.target));
+    const framesWithOutgoing = new Set(
+        compressed.filter((edge) => producingFrameIds.has(edge.target)).map((edge) => edge.source),
+    );
+    const sources = displayedFrames.filter((frame) => !framesWithIncoming.has(frame.id));
```
dcblundell: **Should fix:** the source-side boundary gate is missing — exact mirror of the bug `aced1231` just fixed

`sinks` (line 275) is gated by `producingFrameIds`; `sources` has no consuming equivalent. The issue states the rule symmetrically — "A frame that produces no tensor is not a candidate end and does not disqualify its predecessor from being one" — but only the exit side implements it. A drawn frame with no internal incoming edge and no outputs (a `Tensor::deallocate` of a tensor whose producer isn't drawn, which Outstanding item 5 confirms occurs) counts as a second source, collapsing `entryFallbackNodeId` to `null` and sending every incoming edge back to stopping at the box.

Suggested direction: build a `consumingFrameIds` set alongside `producingFrameIds` and gate `sources` identically. Worth the same regression test shape as the exit-side one.

src/components/operation-graph/OpGraphDeviceOpExpander.tsx
```diff
@@ -0,0 +1,55 @@
...
+    // the bare number.
+    const action = isExpanded ? 'Hide device operations' : `Show ${count} device operations`;
+
+    return (
+        <button
+            type='button'
+            // `nodrag`/`nopan` keep the button from doubling as a drag or pan surface.
+            className='op-graph-node-expander nodrag nopan'
+            title={action}
+            aria-label={action}
```
dcblundell, line 39: **Should fix:** the expansion state machine and its only affordance are untested

No spec anywhere renders this component, and none of the three node types is ever mounted — `OperationGraphRebuildDiscipline.spec.tsx:80` stubs `ReactFlow` to render `children` only, so `nodeTypes` is registered and never invoked. Every expansion test drives `buildOpGraph` with hand-written subgraphs, so nothing asserts a click rebuilds at all, that the payload is what `buildDeviceOperationSubgraph` derives, or that a second click collapses.

Two fixes on this branch shipped unguarded: `24613c2c` (this `aria-label` — without it the accessible name degrades to a bare number, since `▾` is `aria-hidden`) and the `onMouseDown` interception that stops expansion from also selecting. The review comment on #1195 also warned that group nodes without explicit handles lose their edges *silently* and asked for it as a criterion; the handles exist, nothing tests them.

Concrete unblock: `flowApi` (`tests/OperationGraphRebuildDiscipline.spec.tsx:~67`) supplies only `setCenter`, while the view now destructures `getViewport`/`setViewport` at `OperationGraphReactFlow.tsx:205`. Add both to the mock and the toggle path becomes drivable. Assert by role name (`getByRole('button', { name: 'Show 3 device operations' })`) so removing `aria-label` fails; `title` alone would not.

src/components/operation-graph/opGraphDeviceSubgraph.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    return edges;
+}
+
+/**
+ * Drops an edge whose endpoints are already joined by a longer route, so a chain
+ * of three device operations reads as a chain rather than a triangle. Only the
+ * shortcut goes: the two-hop route carries the same dependency in more detail.
+ */
+function withoutShortcutEdges(edges: CompressedEdge[]): CompressedEdge[] {
```
dcblundell, line 203: **Should fix:** `withoutShortcutEdges` runs a fresh BFS per edge

`edges.filter((edge) => !hasLongerRoute(...))` allocates a new `Set` and queue per iteration, giving O(E·(V+E)) — and the compressed graph is not sparse by construction, since `collectTensorHops` takes the full producer×consumer cross-product per tensor. Suggested direction: compute reachability once per source node and reuse it across that node's outgoing edges. Independent of the deferred cap policy; this is unforced cost.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -482,21 +602,30 @@ const OperationGraphInner = ({
...
-                outgoing.push(edge);
-            } else {
-                bySource.set(edge.source, [edge]);
-            }
-            const incoming = byTarget.get(edge.target);
-            if (incoming) {
-                incoming.push(edge);
-            } else {
-                byTarget.set(edge.target, [edge]);
+            const sourceId = String(edge.data?.sourceOperationId);
```
dcblundell: **Should fix:** the edge→operation idiom is open-coded six times, and the internal-edge test three different ways

`String(edge.data?.sourceOperationId)` / `targetOperationId` appears six times across four memos, and the "same operation at both ends" test is written as a string compare here (line 615), a number compare in `topologyEdges` (line 698), and implicitly by omission in the styling pass (line 827) — each a separate chance for the next consumer of edge topology to forget the exclusion. `String(undefined)` yields `"undefined"`, so a missing `data` degrades into a phantom node id rather than a type error.

Suggested direction: two module-scope helpers (`isInternalEdge`, plus one for the stringified endpoints) give the four memos one home for the reasoning. Having the builder emit the stringified ids once into `edge.data` would also drop two allocations per edge from `styledEdges`, which re-runs on every filter keystroke.

src/components/operation-graph/opGraphDeviceSubgraph.ts
```diff
@@ -0,0 +1,299 @@
...
+/**
+ * The device operations an operation decomposes into, as a subgraph ready to
+ * nest inside its node. Assembled on expansion rather than on report load: one
+ * operation's frame stream runs to thousands of nodes, and deriving every
+ * operation's subgraph up front would pay for the ones nobody opens.
+ *
+ * Returns `null` when nothing would be drawn, so the caller can leave the
+ * operation collapsed rather than expanding it into an empty box.
+ */
+export function buildDeviceOperationSubgraph(operation: OperationDescription): OpGraphDeviceSubgraph | null {
```
dcblundell, line 253: **Should fix:** AC2's "matches what the modal draws today" is unverified at the input

The BFS compression and transitive reduction are faithful ports, but the two derivations read differently-computed feedstock. This consumes `processedConnections` from `processInputsOutputs`, where inputs come from the backend's `input_tensors` list; the modal consumes `OperationDetails.preprocessConnections`, where a `function_start`'s inputs are derived in reverse from any node naming it. Outputs are equivalent on both paths; inputs are not demonstrably so, and inputs are half of the adjacency every compressed edge is built from. Both views ship simultaneously, so a user can open them on the same operation and see any discrepancy.

Either add one fixture-driven test asserting the two agree on a real frame stream, or amend AC2 to claim the modal's compression *semantics* rather than its rendered edge set. The test seems worth it — it is also the evidence #1809 needs before retiring the modal.

tests/opGraphDeviceSubgraph.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+                frame({ id: 3, name: 'BetaDeviceOperation', consumes: [11] }),
+            ]);
+
+            expect(endpointsOf(subgraph!)).toEqual([[nodeId(1), nodeId(3)]]);
+            // The tensor the walk started on, so the label still names something
+            // the user can find on the operation.
+            expect(subgraph?.edges[0].label).toBe('T10 [1, 32]');
+        });
+
+        // The port of this reduction discards edge identity, where the modal's
```
dcblundell: **Should fix:** this comment invents a conformance gap that doesn't exist

The comment says the modal "would let one twin justify dropping the other, since arriving over the sibling edge counted as a longer route". It wouldn't: `DeviceOperationsGraphComponent.tsx:189` does `if (next === to) { continue; }` on *any* direct arrival, before `skipEdgeId` can matter — so the modal keeps both twins too.

The assertion below is correct and the behaviour conforms; only the comment is wrong. Worth fixing because it will mislead whoever next audits this port against the modal, or decides under #1809 whether the modal is retirable.

tests/opGraphBuilder.spec.ts
```diff
@@ -4,23 +4,95 @@
...
 
 const build = (operations: OpGraphSourceOperation[], hideDeallocate: boolean) =>
-    buildOpGraph(operations, { hideDeallocate });
+    buildOpGraph(operations, { hideDeallocate, deviceSubgraphs: [] });
+
+const HEAD_FRAME_ID = 1;
+const TAIL_FRAME_ID = 2;
+
+interface SubgraphSpec {
+    operationId: number;
```
dcblundell, line 45: **Should fix:** the device id format is hand-built across four spec files

~24 literal `dev:`/`dev-edge:` strings here and in `opGraphDeviceSubgraph.spec.ts:54`, `opGraphLayoutWorker.spec.ts:67` and `OperationGraphRebuildDiscipline.spec.tsx:289` restate what `DEVICE_NODE_PREFIX` / `DEVICE_EDGE_PREFIX` / `getDeviceNodeId` own privately. Renaming the prefix compiles clean and fails four suites with unhelpful diffs — while the mocked-subgraph tests keep passing through a real namespacing break, which is the prefix's entire purpose. AGENTS.md §Naming lists this as a promote-and-import case.

Suggested direction: export the id helpers (the edge id is currently assembled inline in `buildDeviceOperationSubgraph`) and build one shared fixture. The `OPERATIONS` literal duplicated across the two worker specs — which this branch had to edit identically — belongs there too.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -237,11 +267,43 @@ const OperationGraphInner = ({
                 outputs: operation.outputs.map((tensor) => ({
                     edgeLabel: toReadableShape(tensor.shape),
                     consumers: tensor.consumers,
+                    tensorId: tensor.id,
                 })),
+                // Counted with the same predicate the subgraph is built from rather
+                // than from `deviceOperationNameList`, which the details panel uses
+                // and which screens out the `Tensor::` frames the graph draws.
+                deviceOperationCount: countDeviceOperations(operation),
```
dcblundell, line 308: **Should fix:** `deviceOperationCount` is recomputed on every mount

The `sourceOperations` memo is keyed on `operationList`, so this O(total frames) pass — four `String.includes` per frame via `isExtendedDeviceOperation` — is repaid on every mount of the graph tab, on the critical path to first paint, rather than once per report. `countDeviceOperations` also allocates a filtered array purely to read `.length`.

Suggested direction: compute it once in `fetchOperations` alongside `deviceOperationNameList`, which already walks the same frames, and count in a loop instead of `filter().length`.

src/components/operation-graph/opGraphDeviceSubgraph.ts
```diff
@@ -0,0 +1,299 @@
...
+    shapeByTensorId: Map<number, string>;
+} {
+    const shapeByTensorId = new Map<number, string>();
+    const producersByTensorId = new Map<number, number[]>();
+    const consumersByTensorId = new Map<number, number[]>();
+
+    const record = (frameId: number, tensors: Node[], target: Map<number, number[]>) => {
+        for (const tensor of tensors) {
+            if (tensor.node_type !== NodeType.tensor) {
+                // eslint-disable-next-line no-continue
```
dcblundell: **Should fix:** six `no-continue` suppressions carry no reason

Also at lines 102, 150 and 161 here, plus `opGraphBuilder.ts:103` and `opGraphLayout.ts:125`. AGENTS.md:119 requires an explanatory comment on the same or preceding line; `src/functions/perfOverlay.ts:76` is the in-repo form. Add a `-- reason` clause, or restructure the two tightest loops (`record`, the hop fan-out) so the guard reads as a positive `if` and no suppression is needed.

src/components/operation-graph/opGraphBuilder.ts
```diff
@@ -48,52 +55,154 @@ export function buildOpGraph(
         connected.add(candidate.target);
     }
 
+    const subgraphByOperationId = new Map(deviceSubgraphs.map((subgraph) => [subgraph.operationId, subgraph]));
```
dcblundell: **Should fix:** `Map` constructed without type parameters

Here and at line 107 (`new Map(subgraph.nodes.map(...))`). CONVENTIONS.md:123 requires spelling out generics on third-party containers, and this file's own `new Map<string, number>()` for `parallelCountByPair` already does. Writing them out — `Map<number, OpGraphDeviceSubgraph>` and `Map<string, { width: number; height: number }>` — also stops the two `childSizeById.get(child.id)!` assertions at line 111 being trusted blind.

src/components/operation-graph/opGraphLayout.ts
```diff
@@ -11,11 +11,20 @@ const NODE_MIN_WIDTH = 108;
...
+// The device-operation badge, kept in step with `.op-graph-node-expander`.
+const EXPANDER_WIDTH = 30;
 // Wide enough for an edge to carry its shape label between two ranks without
 // colliding with the neighbouring column.
 const NODE_SEP = 30;
 const RANK_SEP = 80;
 
+// The header carries the operation's own label and file, so it is sized like a
+// collapsed node; the rest is breathing room around the nested subgraph.
+const GROUP_HEADER_HEIGHT = 46;
```
dcblundell: **Should fix:** `GROUP_HEADER_HEIGHT = 46` restates `NODE_HEIGHT_WITH_FILE`

The comment states the intent is a derivation ("sized like a collapsed node"), so write it as one: `const GROUP_HEADER_HEIGHT = NODE_HEIGHT_WITH_FILE;`. As a bare literal, a font or padding change that moves `NODE_HEIGHT_WITH_FILE` leaves every expanded group's header reservation stale, and the symptom — children overlapping the header, or a gap under it — surfaces nowhere near the edit.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -322,6 +398,50 @@ const OperationGraphInner = ({
         focusOperation(target);
     }, [nodes, focusOperation]);
 
+    // Read from the committed array rather than `getNode`, which trails `setNodes`
+    // by a tick and would translate against the pre-rebuild position.
+    useEffect(() => {
+        const anchor = pendingViewportAnchorRef.current;
```
dcblundell, line 437: **Should fix:** the viewport anchor ref outlives its report

The effect returns on `nodes.length === 0` *before* clearing the ref at line 407, and the report-scope reset block (line ~229) clears `expandedOperationIds` and `deviceSubgraphCache` but not this anchor. An expand followed by a report switch can spend a stale anchor translating the new report's first non-empty commit against a same-id node from the old report.

Suggested direction: clear the ref before the empty-nodes return, and alongside the other expansion state on scope change. The precedence contract with `pendingFocusRef` (line 356) also exists only as a comment and is worth pinning with a test.

src/components/operation-graph/opGraphLayout.ts
```diff
@@ -32,9 +41,20 @@ export interface LayoutPosition {
...
+ */
+export function estimateOpNodeSize(
+    label: string,
+    fileIdentifier: string,
+    hasExpander = false,
+): { width: number; height: number } {
     const widestLine = Math.max(label.length * CHAR_WIDTH, fileIdentifier.length * FILE_CHAR_WIDTH);
-    const width = Math.ceil(Math.min(NODE_MAX_WIDTH, Math.max(NODE_MIN_WIDTH, widestLine + NODE_PADDING_X)));
+    const contentWidth = widestLine + NODE_PADDING_X + (hasExpander ? EXPANDER_WIDTH : 0);
+    const width = Math.ceil(Math.min(NODE_MAX_WIDTH, Math.max(NODE_MIN_WIDTH, contentWidth)));
```
dcblundell: **Should fix:** the `NODE_MAX_WIDTH` clamp swallows the badge reservation

`hasExpander` adds `EXPANDER_WIDTH` to `contentWidth` on the line above, but this `Math.min` against `NODE_MAX_WIDTH` discards it entirely at or past the clamp — so a long-labelled expandable node loses the 30px reservation, which is the exact badge-over-label overlap the parameter's docstring says it exists to prevent. Distinct from Outstanding item 8, which covers 100+ counts rather than long labels.

Suggested direction: clamp the label width first and reserve the badge after, or confirm the overhang is intended. Either way it wants a boundary test — `estimateOpNodeSize('x'.repeat(400), '', true)` currently equals the `false` case.

dcblundell (commented):
Conductor review — six lenses (DRY, TEST, PERF, STANDARDS, SECURITY, REQUIREMENTS) against #1195 plus its review comment. Overall risk moderate: the derivation, builder re-targeting and worker cache key are well covered, and the expansion-invariant topology design holds up.

Posted below: **1 must fix**, **13 should fix**, and **1 decision** on group draggability where the issue body and its review comment disagree and the code silently picks one.

Two themes dominate — an unguarded `Number(tensor_id)` that both fabricates edges and turns a malformed capture into an indefinite main-thread hang, and no coverage at all for the click-to-rebuild path including the feature's only affordance.

Consider/nit findings were kept out of the PR and are in the chat summary. All 8 acceptance criteria are implemented; Outstanding items 1–8 are genuinely deferred and match the code. `pnpm lint` and `pnpm lint:spdx` are clean at this head.

src/components/operation-graph/opGraphBuilder.ts
```diff
@@ -48,52 +55,154 @@ export function buildOpGraph(
...
+                subgraph.nodes.map((child) => ({ id: child.id, ...childSizeById.get(child.id)! })),
+                subgraph.edges,
+                collapsedSize.width,
+            );
+
+            // The operation keeps its node id when expanded. Everything keyed by
+            // node id — the perf style patches, the critical path's node set, focus
+            // and selection — then needs no notion of expansion at all, and the
+            // edges already pointing here stay pointing here. #1195
             nodes.push({
```
dcblundell: **Decision needed:** is the expanded group draggable?

The issue body decides "**Group is draggable.** The cost is React Flow's `nopan` … That is how every other node already behaves." The review comment on #1195 argued the opposite with specifics: `draggable: false` on the wrapper, `nopan nodrag` on the header, `dragHandle` scoping the drag, because RF's pan filter resolves `.nopan` via `closest()` from any descendant.

This code implements the body — by omission. There is no `draggable` or `dragHandle` anywhere in the builder, so RF's default applies and injects `nopan` on the group wrapper: `@xyflow/react/dist/esm/index.js:2332` sets `[noPanClassName]: isDraggable`, which kills drag-to-pan over the **entire box** of an expanded group. `mlirGraphBuilder.ts:447-457` documents this exact trap and deliberately does the opposite.

The weak step is the body's "how every other node already behaves" equivalence: a collapsed node is 108–560px of mostly text, whereas an expanded group is a large translucent field that is mostly empty space — precisely where someone reads a subgraph and then grabs to pan.

Recommendation: adopt the MLIR pattern, which already exists and is documented (`createMoveGroupBatcher` also already solves moving children). Trade-off: dragging then only works from the header, a smaller target than the whole box. If you'd rather keep whole-box dragging, amend the Decisions entry to record that the comment's advice was considered and rejected, and why — the issue currently carries both positions and the code silently picks one.

src/components/operation-graph/opGraphDeviceSubgraph.ts
aidemsined: Fixed in 309b2b59. `tensorRecordOf` drops anything that isn't a tensor with a finite id, so unreadable ids never enter the producer/consumer maps. Regression: two frames that share only `'nope'` as `tensor_id` draw no edge; a readable tensor sitting next to an unreadable one still joins.

src/components/operation-graph/OpGraphDeviceOpExpander.tsx
aidemsined: Fixed in 309b2b59. The React Flow stub now mounts `nodeTypes` and supplies `getViewport`/`setViewport`, so the expander is a real button.

`getByRole('button', { name: 'Show 3 device operations' })` drives a rebuild whose payload is what `buildDeviceOperationSubgraph` derived; the second click is `Hide device operations` and the next `runBuild` has `deviceSubgraphs: []`. That also mounts `OpGraphDeviceGroupNode`, which is where the handles live.

src/components/operation-graph/opGraphDeviceSubgraph.ts
aidemsined: Fixed in 309b2b59, but the gate is `producingFrameIds` on both ends, not a new `consumingFrameIds` set.

A `Tensor::deallocate` of an undrawn producer *does* consume, so gating sources on consume would still count it as a second start and collapse the fallback. It would also drop a producer-only entry. The issue's rule is "a frame that produces no tensor is not a candidate end"; applying that same producing test to the start is what actually keeps the unique-entry fallback. Regression added for a leading deallocate of tensor 99 alongside the cumsum chain.

src/components/operation-graph/opGraphDeviceSubgraph.ts
aidemsined: Fixed in 309b2b59. One BFS per source records nodes reachable in two or more hops; an edge is a shortcut iff its target is in that set. Twin edges and cycles still keep every edge — same assertions as before, cheaper walk.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in 309b2b59. `operationBoundaryOf` / `isInternalDeviceEdge` / `sourceOperationIdOf` / `targetOperationIdOf` are the one home; neighbourhood, topology, highlight and filter styling all go through them. Missing `data` is now `null` rather than `String(undefined)`.

tests/opGraphDeviceSubgraph.spec.ts
aidemsined: Fixed in 309b2b59. The modal skips any direct arrival (`next === to`) before `skipEdgeId` can matter, so it keeps both twins too. Comment now says that.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Counted in a loop now (no `filter().length` allocation). Not moving this into `fetchOperations`.

`deviceOperationNameList` is the details-panel predicate (`isDeviceOperation`), which deliberately drops the `Tensor::` frames the graph draws. The badge has to share `isDisplayedFrame` with the subgraph or it lies about what expanding produces. Putting that graph-display predicate on the fetch path is how those two numbers would drift again. The `sourceOperations` memo is keyed on `operationList`, so the walk is once per report identity, not per render — remounting the graph tab does repay it, which I'm willing to live with against coupling fetch to the expander's name test.

tests/opGraphBuilder.spec.ts
aidemsined: Fixed in 309b2b59. `getDeviceNodeId` / `getDeviceEdgeId` are exported; the four specs build ids through them. Left the duplicated `OPERATIONS` literals in the two worker specs — they aren't the namespacing hazard, and folding them into a shared fixture is a follow-up I'm happy to take if you want it on this PR.

src/components/operation-graph/opGraphDeviceSubgraph.ts
aidemsined: Fixed in 309b2b59. The six `no-continue` sites are now positive `if`s (and the collapsed-op `continue` in the builder is an `if`/`else`).

src/components/operation-graph/opGraphLayout.ts
aidemsined: Fixed in 309b2b59. `const GROUP_HEADER_HEIGHT = NODE_HEIGHT_WITH_FILE`.

src/components/operation-graph/opGraphBuilder.ts
aidemsined: Fixed in 309b2b59. `Map<number, OpGraphDeviceSubgraph>` and `Map<string, { width: number; height: number }>`.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: Fixed in 309b2b59. A render-phase `ref.current = null` is illegal (`react-hooks/refs`), so the anchor now carries the `reportScope` it was armed for. The spend effect drops it on an empty node array or a scope mismatch instead of translating against a same-id node from the new report.

The expand-then-navigate-before-rebuild race (Copilot's suppressed sibling of this) is still open — that's a different lifetime than report scope.

src/components/operation-graph/opGraphBuilder.ts
aidemsined: Keeping whole-box dragging. That was an explicit product call after the first pass shipped `draggable: false` and the group couldn't be moved — "the group isn't draggable when expanded" was a bug report, not an oversight.

The MLIR pattern is the right read for a large empty field you pan across. An expanded operation here is still one node in a column of operations; grabbing the box to reposition it is the same gesture as grabbing a collapsed node. `nopan` on the wrapper is the same cost every other node already pays, just over a larger hit target.

I'll record that the review-comment advice was considered and rejected in #1195's Decisions so the issue doesn't keep both positions. Header-only drag can be a follow-up if the larger `nopan` surface turns out to be the thing people hit.

src/components/operation-graph/opGraphLayout.ts
aidemsined: Fixed in 309b2b59. Clamp the label to `NODE_MAX_WIDTH` first, then add `EXPANDER_WIDTH`. `estimateOpNodeSize('x'.repeat(400), '', true)` is now strictly wider than the `false` case.

src/components/operation-graph/opGraphDeviceSubgraph.ts
aidemsined: Agreed the two feedstocks differ, and I don't want to claim they render the same edge set.

`processInputsOutputs` takes `input_tensors` from the backend; `preprocessConnections` reverse-derives a `function_start`'s inputs from any node that names it. Outputs are the same idea on both paths; inputs are not. A fixture that asserted agreement would either be tautological on synthetic data or flaky on a real stream, and it belongs with #1809 when the modal actually goes away.

Happy to amend AC2 to "the modal's compression semantics" rather than "what the modal draws today." Treating that as the honest claim unless you'd rather have the fixture on this PR anyway.

PR #1935, opened 2026-08-26
Add Github Copilot instructions

dcblundell:
This pull request adds a new documentation file, `.github/copilot-instructions.md`, providing clear instructions for contributors and reviewers. The file directs users to consult the main architecture, security, coding standards, and review guidelines before making changes.

Documentation improvements:
* Added `.github/copilot-instructions.md` to instruct contributors and reviewers to read `AGENTS.md` and `CONVENTIONS.md` for repository standards and best practices.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,5 @@
+# TT-NN Visualizer Copilot Instructions
+
+Before making or reviewing changes, read and follow [AGENTS.md](../AGENTS.md) and [CONVENTIONS.md](../CONVENTIONS.md). These are the canonical repository instructions for architecture, security, coding standards, testing, and review priorities.
+
+Check the relevant sections of both files rather than relying on assumptions. Preserve unrelated user changes, keep edits focused, and validate changed behaviour with the repository's documented checks.
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,5 +1,3 @@
 # TT-NN Visualizer Copilot Instructions
 
 Before making or reviewing changes, read and follow [AGENTS.md](../AGENTS.md) and [CONVENTIONS.md](../CONVENTIONS.md). These are the canonical repository instructions for architecture, security, coding standards, testing, and review priorities.
-
-Check the relevant sections of both files rather than relying on assumptions. Preserve unrelated user changes, keep edits focused, and validate changed behaviour with the repository's documented checks.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,5 @@
+# TT-NN Visualizer Copilot Instructions
+
+Before making or reviewing changes, read and follow [AGENTS.md](../AGENTS.md) and [CONVENTIONS.md](../CONVENTIONS.md). These are the canonical repository instructions for architecture, security, coding standards, testing, and review priorities.
+
+Check the relevant sections of both files rather than relying on assumptions. Preserve unrelated user changes, keep edits focused, and validate changed behaviour with the repository's documented checks.
```
aidemsined: > [!WARNING]
> This line makes the file carry rules of its own, which is the one thing the repo's wrapper convention says not to do — and all three are already in `AGENTS.md`, more specifically.

`CLAUDE.md` states the convention for exactly this situation:

> This file exists only so that Claude Code picks up that guidance; it is intentionally a thin wrapper and **holds no rules of its own**. Anything Claude-specific still belongs in `AGENTS.md`, alongside the guidance every other agent reads — **put new instructions there, not here**.

Line 3 follows that faithfully. Line 5 doesn't — and the three clauses it adds are already covered upstream, in sharper form:

- *"Preserve unrelated user changes"* / *"keep edits focused"* → `AGENTS.md:69` (*"Format only the code you touched; don't reformat unrelated paths"*) and `AGENTS.md:118` (*"Pre-existing lints in code you didn't touch are not yours to fix in unrelated PRs… don't sprawl scope"*).
- *"validate changed behaviour with the repository's documented checks"* → `AGENTS.md:69` and `:73` name them: `pnpm lint`, `pnpm lint:fix`, `pnpm lint:spdx`.

So this is a weaker restatement of guidance that already exists, in a file that by convention shouldn't hold guidance. That's the setup for drift: someone tightens the scope rule in `AGENTS.md` and this copy silently keeps saying something vaguer.

**Suggested action:** drop line 5 so the file is a pure pointer like `CLAUDE.md`. If any of it says something `AGENTS.md` doesn't, add it there instead — where Claude, Cursor and Copilot all read it — rather than in one tool's wrapper.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,5 @@
+# TT-NN Visualizer Copilot Instructions
+
+Before making or reviewing changes, read and follow [AGENTS.md](../AGENTS.md) and [CONVENTIONS.md](../CONVENTIONS.md). These are the canonical repository instructions for architecture, security, coding standards, testing, and review priorities.
```
aidemsined, line 3: > [!WARNING]
> Worth confirming the "or reviewing" half actually reaches Copilot's PR reviewer, because that's where this repo would get the most out of it.

The path is right for repo-wide custom instructions, and the relative links resolve correctly from `.github/`. My uncertainty is scope: I don't know that `.github/copilot-instructions.md` is consumed by Copilot **code review** the same way it is by Chat and completions. GitHub also has a separate path-scoped mechanism — `.github/instructions/*.instructions.md` with `applyTo` frontmatter — and `copilot-pull-request-reviewer` may only honour one of them. I'm flagging it as unverified rather than asserting it doesn't work.

It matters because the reviewer is a real participant here — it has commented on nearly every PR in this batch — and its accuracy is uneven in ways this file could plausibly fix. Two from the last few days:

- On #1889 it reported the `beta` badge as a sibling of `.bp6-button-text` and therefore stranded in the collapsed rail. Blueprint 6.6.1 renders `children` *inside* that span (`buttons.js:87`), so `display: none` already takes it. It was misled by a comment in the stylesheet that says "sibling" — a reviewer told to check `AGENTS.md`/`CONVENTIONS.md` first wouldn't have caught that one either, so it's a fair test of the limits.
- Also on #1889 it flagged banner copy that was carried over unchanged from the deleted `MainNavigation.tsx`. Pointing it at the repo's own scope rule (`AGENTS.md:118`, pre-existing issues aren't the PR's to fix) is exactly the kind of steer that would suppress that class of comment.

**Suggested action:** verify the reviewer picks this file up — a PR with a deliberate convention violation is the cheapest test. If it turns out review needs `.github/instructions/*.instructions.md`, adding that alongside is a small follow-up and would make the "or reviewing" wording true.

aidemsined (commented):
Right idea, and the mechanics are correct — `.github/copilot-instructions.md` is the documented location for repo-wide custom instructions, and the `../AGENTS.md` / `../CONVENTIONS.md` links resolve properly from `.github/`. Pointing a fourth agent at the same canonical pair rather than writing it a private rulebook is the right instinct, and it matches what `CLAUDE.md` already does.

Two things inline: the second paragraph makes this a rulebook rather than a pointer, and the "or reviewing" claim is worth verifying before it's relied on.

One observation beyond those. The repo now has four entry points into the same guidance — `AGENTS.md` (canonical), `CONVENTIONS.md` (expanded), `CLAUDE.md` (wrapper), and this. That's fine while the wrappers stay empty, which is precisely why `CLAUDE.md` spends a paragraph insisting on it. Worth deciding now whether wrapper files are allowed to grow at all, because the answer determines whether the next one — a Cursor rules file, say — arrives as three lines or as a competing style guide. `CLAUDE.md`'s framing ("put new instructions there, not here") already answers it; this PR is the moment to either follow it or amend it deliberately.

.github/copilot-instructions.md
dcblundell: Agreed — dropped in 38b56f0c, so the file is now just the heading and the pointer line.

Checked each clause against `AGENTS.md` before removing it:

- *keep edits focused* → `AGENTS.md:69` ("Format only the code you touched; don't reformat unrelated paths") and `:118` ("Pre-existing lints in code you didn't touch are not yours to fix… don't sprawl scope").
- *validate with the documented checks* → `AGENTS.md` names them concretely (`pnpm lint`, `pnpm lint:fix`, `pnpm lint:spdx`, `pnpm test`, `flask:lint`/`flask:mypy`), which the paraphrase did not.
- *check the relevant sections rather than assuming* → already implied by line 3's "read and follow".

Nothing in the removed line said something `AGENTS.md` doesn't, so there was nothing to promote upstream.

.github/copilot-instructions.md
dcblundell: Verified against GitHub's own support matrix ([docs.github.com/en/copilot/reference/custom-instructions-support](https://docs.github.com/en/copilot/reference/custom-instructions-support)) rather than by experiment. For **Copilot code review on GitHub.com**, all three are supported:

| | `.github/copilot-instructions.md` | `.github/instructions/**.instructions.md` | `AGENTS.md` |
|---|---|---|---|
| GitHub.com | yes | yes | yes |

So "or reviewing" is accurate as written, and the path-scoped mechanism is an *additional* layer rather than the one the reviewer requires — no follow-up needed on that front.

Two caveats worth recording:

1. Code review honours repository custom instructions only when the reviewer's personal "use custom instructions" setting is enabled, so it's per-account, not purely repo-side.
2. The same table shows Copilot code review on GitHub.com reads `AGENTS.md` **directly**. This wrapper therefore earns its keep mainly for IDE Chat/completions — VS Code, Visual Studio, JetBrains — where `AGENTS.md` is not read. Worth knowing if anyone later wonders why the file exists at all.

I've left the empirical test (a PR with a deliberate convention violation) undone; happy to run one if you'd rather not take the docs at face value.

Issue #1942, opened 2026-08-27
Operations graph: tidy the filter and toggle chrome


aidemsined:
## Problem

The operations-graph toolbar has outgrown a single row. Prev/next, four labelled switches, and (when repeats exist) Unroll/Fold now sit in a wrap-prone cluster over the canvas.

Today that row is:

- prev / selected id / next
- Hide deallocate ops
- Dim unrelated edges
- Perf overlay (`N/M` when linked)
- Highlight critical path
- Repeats · Unroll / Fold (second row, only when blocks exist)

Filter is already its own row above. Adding Dim unrelated edges is what made the second row feel crowded; the next switch will make it worse. This is follow-up chrome, not part of the #1583 / dim-edges change.

## Ask

Redesign the graph chrome so filters and toggles stay scannable at a typical viewport without covering the graph or wrapping into a soup.

Directions worth considering (pick one, don't ship all):

- Split **navigate** from **view options** (two compact rows, or a popover / overflow for the switches).
- Icon-only or short labels for the switches, full name in the tooltip.
- A single “View” menu that holds deallocate / dim / overlay / path, leaving prev-next + filter always visible.

## Acceptance criteria

- Filter, prev/next, and the current four switches (plus Repeats when present) remain reachable without hunting.
- The cluster does not cover the first rank of nodes at a 1280-wide window.
- Disabled / gated states (perf overlay and critical path without a linked report; mid-build) stay explained on hover.
- No behaviour change to the toggles themselves.

## Out of scope

- Changing what any switch does.
- The #1583 collapse / naming work and the dim-unrelated-edges behaviour.

Issue #1192, opened 2026-01-29
Graph: Identifying groups of operations


dcblundell:
Analyzing groups of operations (like an NN layer) requires significant manual action. This is currently a prime use of the graph view.

“Operation graph offered a nice visualization of the model execution flow, but I did not find any use of it other than visually finding the range of operation ids for a single NN layer.”

“I found operation graphs only useful to determine the range of operation ids for a particular NN layer"

Issue #1191, opened 2026-01-29
Graph: Illegible with a lot of operations


dcblundell:
The operation graph quickly becomes illegible as the number of operations increases.

_Very recent changes to how we render the nodes may render this feedback obsolete._

PR #1944, opened 2026-08-27
Operations graph: collapse repeated subgraphs

### LINKED ISSUE #1583: Operations graph: Auto-detect and collapse repeated subgraphs

aidemsined:
image

## Summary

A 12-layer model is hundreds of near-identical ops on the graph. This detects strict contiguous repeated subgraphs, folds each copy into a block, and names the block from the modules (or ops) that identify the tile.

Closes #1583

### Frontend
- `opGraphRepeatBlocks.ts` — smallest window that repeats (`L ≥ 2`), structural check including leaving edges, skip above `MAX_DETECT_OPS`. Labels from unique source-file stems; a file that covers most of the graph (e.g. `ttnn_functional_resnet50.py`) is dropped so the name becomes `add_ + conv2d` rather than the model file. `lazy_weight` is plumbing. A block of a single operation type is named by that operation rather than its file (two `ttnn.squeeze` read `squeeze`, not `tt_routed_expert`), and a leading prefix every stem shares is dropped — one SentenceBERT block spans four modules and was named in 124 characters, 72 of them `ttnn_sentencebert_` four times over, so it now reads `self_attention + self_output + intermediate + output`. Stripping stops while every stem keeps a segment of its own, and is abandoned if a remainder would fall under three characters, so `mlp_up + mlp_down` does not become `up + down`.
- `opGraphBuilder.ts` / layout worker — collapsed `blockNode`s, one unlabelled edge per rendered pair, detection cached per deallocate filter, layout cache keyed on the unrolled-instance set.
- `OperationGraphReactFlow.tsx` — folded by default; Unroll/Fold on the toolbar and per-block chip; double-click a block to unroll, double-click a member to fold that instance. Selecting a block opens the block panel (aggregates, types, boundary I/O), not the first member. Filter treats a member-name hit as a match (`+N inside`). Device-op expand is dropped when that instance is folded.
- Perf overlay and critical path sum member weights; critical path counts member ops.
- **Dim unrelated edges** toolbar switch (off by default): with a node selected, non-I/O edges fade to 0.35 via a container class, same rule as critical-path off-path dim. Filter still wins. No-op with nothing selected.

### Keeping your place when a block opens
Unrolling repacks the layout, which moved the graph under the reader — the reported symptom was "I opened one and I can't even see it on the screen anymore."

- A structural toggle holds the viewport anchor, and pans the revealed ops back into view when the repack put them outside it.
- Those ops are ringed: three slow pulses for attention, then a faint ring that **rests until the next toggle**, so "which ones did I just open" stays answerable while they are open. Exactly one group is ever marked; the ring clears on a report swap or a deallocate-filter repack.
- Labels are the other half of not losing your place, hence the naming rules above. Node labels cap at `NODE_MAX_WIDTH` and both panel headings ellipsise, so all three now carry the full string as a native tooltip — the file identifier beside them already did.

- Repeated windows longer than `MAX_REPEAT_WINDOW` (64 ops) are not detected — a layer above that stays unfolded — #1956. Descending from ⌊N/2⌋ made the no-repeat case O(N³) and folded 12 layers into two half-model blocks. The time half is now largely gone: hoisting the O(1) window discriminator ahead of the two O(length) consumed walks takes detection on a 10k-op no-repeat graph from 40.4ms to 4.0ms. What is left to settle is the half-model-block case, which is why the cap stays here
- No keyboard-accessible way to fold a single unrolled instance; the toolbar folds all. #1942 covers the cramped chrome rather than this, so the a11y gap needs its own issue — adding a fold affordance to unrolled members is new UI and was out of scope for a behaviour-preserving pass
- Toolbar chrome is cramped after the new switch — #1942
- No fuzzy match, no user rename (`Transformer layer`), no shared identity on the op list / perf table (as in #1583)

## Test plan
- [ ] Frontend: `pnpm test tests/opGraphRepeatBlocks.spec.ts tests/opGraphBuilder.spec.ts tests/opGraphLayoutWorker.spec.ts tests/OperationGraphRebuildDiscipline.spec.tsx tests/OpGraphToolbarSwitches.spec.tsx tests/opGraphNodeStyles.spec.ts tests/opGraphRevealPan.spec.ts tests/opGraphNavigation.spec.ts tests/OpGraphPanelLabelTooltip.spec.tsx tests/OpGraphBlockPanelStats.spec.tsx tests/opGraphBlockBoundary.spec.ts tests/opGraphBlockMeta.spec.ts`
- [ ] `pnpm lint` clean on touched files
- [ ] Manual — bge_m3: 24 copies named `norm + attention + encoder + mlp × 24`; unroll one instance; filter a member name; double-click to fold
- [ ] Manual — resnet50: blocks named `add_ + conv2d` (not `ttnn_functional_resnet50`); Dim unrelated edges on/off with a node selected
- [ ] Manual — MoE setup loop: blocks read `squeeze × 2` (they read `tt_routed_expert × 2` before the label fix)
- [ ] Manual — sentence_bert: the block reads `self_attention + self_output + intermediate + output`; hover a truncated label for the full name
- [ ] Manual — unroll a block low in the viewport: it stays in view, pulses, and holds a faint ring until the next unroll

## Review pass

David's review (4 Must fix, 21 Should fix) is addressed in `3fc7a5b2`..`884e6762`, constrained to changes that alter neither behaviour nor UX. Summary in the PR comment; the two deliberate exceptions are #1956 (raising the window cap changes what folds) and the keyboard fold affordance (new UI). Suite went 2285 → 2318.

## Follow-ups
- #1942 — tidy the filter and toggle chrome
- #1191 — graph illegible at scale (this is the collapse half)
- #1192 — identifying groups of operations

### DIFF
--- a/src/components/operation-graph/OpGraphBlockExpander.tsx
+++ b/src/components/operation-graph/OpGraphBlockExpander.tsx
@@ -2,7 +2,8 @@
 //
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
-import { type MouseEvent as ReactMouseEvent, memo, useContext } from 'react';
+import { memo, useContext } from 'react';
+import OpGraphNodeExpander from './OpGraphNodeExpander';
 import { OpGraphBlockExpansionContext } from './opGraphExpansionContext';
 
 interface OpGraphBlockExpanderProps {
@@ -14,44 +15,13 @@ interface OpGraphBlockExpanderProps {
 const OpGraphBlockExpander = memo(({ instanceId, opCount, isExpanded }: OpGraphBlockExpanderProps) => {
     const toggleBlock = useContext(OpGraphBlockExpansionContext);
 
-    // React Flow reads selection off `mousedown`, so without this the expander
-    // would also select the operation and start the centring tween that goes with
-    // it — the graph would jump before it expanded. Structural navigation and
-    // selection stay separate gestures.
-    const handleMouseDown = (event: ReactMouseEvent) => event.stopPropagation();
-
-    const handleClick = (event: ReactMouseEvent) => {
-        event.stopPropagation();
-        toggleBlock(instanceId);
-    };
-
-    // Also the accessible name: `title` only supplies one when the element has no
-    // content, and the count is content — so without this the button announced as
-    // the bare number.
-    const action = isExpanded ? `Fold ${opCount} operations` : `Unroll ${opCount} operations`;
-
     return (
-        <button
-            type='button'
-            // `nodrag`/`nopan` keep the button from doubling as a drag or pan surface.
-            className='op-graph-node-expander nodrag nopan'
-            title={action}
-            aria-label={action}
-            aria-expanded={isExpanded}
-            onMouseDown={handleMouseDown}
-            onClick={handleClick}
-        >
-            {/* Down opens, up closes — a chevron that only reported the current
-                state read as "expand further". Same glyph, rotated: `▴` sits
-                high in the em-box and the count looks dropped beside it. */}
-            <span
-                aria-hidden='true'
-                className='op-graph-node-expander-icon'
-            >
-                ▾
-            </span>
-            {opCount}
-        </button>
+        <OpGraphNodeExpander
+            action={isExpanded ? `Fold ${opCount} operations` : `Unroll ${opCount} operations`}
+            count={opCount}
+            isExpanded={isExpanded}
+            onToggle={() => toggleBlock(instanceId)}
+        />
     );
 });
 
--- a/src/components/operation-graph/OpGraphBlockExpander.tsx
+++ b/src/components/operation-graph/OpGraphBlockExpander.tsx
@@ -21,7 +21,7 @@ const OpGraphBlockExpander = memo(({ instanceId, opCount, isExpanded }: OpGraphB
         toggleBlock(instanceId);
     };
 
-    const action = isExpanded ? 'Collapse repeated block' : `Expand ${opCount} operations`;
+    const action = isExpanded ? `Fold ${opCount} operations` : `Unroll ${opCount} operations`;
 
     return (
         <button
--- a/src/components/operation-graph/OpGraphBlockExpander.tsx
+++ b/src/components/operation-graph/OpGraphBlockExpander.tsx
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { type MouseEvent as ReactMouseEvent, memo, useContext } from 'react';
+import { OpGraphBlockExpansionContext } from './opGraphExpansionContext';
+
+interface OpGraphBlockExpanderProps {
+    instanceId: string;
+    opCount: number;
+    isExpanded: boolean;
+}
+
+const OpGraphBlockExpander = memo(({ instanceId, opCount, isExpanded }: OpGraphBlockExpanderProps) => {
+    const toggleBlock = useContext(OpGraphBlockExpansionContext);
+
+    const handleMouseDown = (event: ReactMouseEvent) => event.stopPropagation();
+
+    const handleClick = (event: ReactMouseEvent) => {
+        event.stopPropagation();
+        toggleBlock(instanceId);
+    };
+
+    const action = isExpanded ? 'Collapse repeated block' : `Expand ${opCount} operations`;
+
+    return (
+        <button
+            type='button'
+            className='op-graph-node-expander nodrag nopan'
+            title={action}
+            aria-label={action}
+            aria-expanded={isExpanded}
+            onMouseDown={handleMouseDown}
+            onClick={handleClick}
+        >
+            <span
+                aria-hidden='true'
+                className='op-graph-node-expander-icon'
+            >
+                ▾
+            </span>
+            {opCount}
+        </button>
+    );
+});
+
+OpGraphBlockExpander.displayName = 'OpGraphBlockExpander';
+
+export default OpGraphBlockExpander;
--- a/src/components/operation-graph/OpGraphBlockExpander.tsx
+++ b/src/components/operation-graph/OpGraphBlockExpander.tsx
@@ -14,6 +14,10 @@ interface OpGraphBlockExpanderProps {
 const OpGraphBlockExpander = memo(({ instanceId, opCount, isExpanded }: OpGraphBlockExpanderProps) => {
     const toggleBlock = useContext(OpGraphBlockExpansionContext);
 
+    // React Flow reads selection off `mousedown`, so without this the expander
+    // would also select the operation and start the centring tween that goes with
+    // it — the graph would jump before it expanded. Structural navigation and
+    // selection stay separate gestures.
     const handleMouseDown = (event: ReactMouseEvent) => event.stopPropagation();
 
     const handleClick = (event: ReactMouseEvent) => {
@@ -21,11 +25,15 @@ const OpGraphBlockExpander = memo(({ instanceId, opCount, isExpanded }: OpGraphB
         toggleBlock(instanceId);
     };
 
+    // Also the accessible name: `title` only supplies one when the element has no
+    // content, and the count is content — so without this the button announced as
+    // the bare number.
     const action = isExpanded ? `Fold ${opCount} operations` : `Unroll ${opCount} operations`;
 
     return (
         <button
             type='button'
+            // `nodrag`/`nopan` keep the button from doubling as a drag or pan surface.
             className='op-graph-node-expander nodrag nopan'
             title={action}
             aria-label={action}
@@ -33,6 +41,9 @@ const OpGraphBlockExpander = memo(({ instanceId, opCount, isExpanded }: OpGraphB
             onMouseDown={handleMouseDown}
             onClick={handleClick}
         >
+            {/* Down opens, up closes — a chevron that only reported the current
+                state read as "expand further". Same glyph, rotated: `▴` sits
+                high in the em-box and the count looks dropped beside it. */}
             <span
                 aria-hidden='true'
                 className='op-graph-node-expander-icon'
--- a/src/components/operation-graph/OpGraphBlockNode.tsx
+++ b/src/components/operation-graph/OpGraphBlockNode.tsx
@@ -4,21 +4,8 @@
 
 import { Handle, type NodeProps, Position } from '@xyflow/react';
 import { memo } from 'react';
-import { formatMemorySize, formatSize } from '../../functions/math';
 import OpGraphBlockExpander from './OpGraphBlockExpander';
 import type { OpGraphFlowNode } from './opGraphTypes';
-
-const blockMeta = (opCount: number, durationSeconds: number, memoryDeltaBytes: number): string => {
-    const parts = [`${opCount} ops`];
-    if (durationSeconds > 0) {
-        parts.push(`${formatSize(durationSeconds, 2)} s`);
-    }
-    if (memoryDeltaBytes !== 0) {
-        const sign = memoryDeltaBytes > 0 ? '+' : '-';
-        parts.push(`${sign}${formatMemorySize(Math.abs(memoryDeltaBytes), 0)}`);
-    }
-    return parts.join(' · ');
-};
 
 const OpGraphBlockNode = memo(({ data }: NodeProps<OpGraphFlowNode>) => {
     const instanceId = data.blockInstanceId;
@@ -34,9 +21,7 @@ const OpGraphBlockNode = memo(({ data }: NodeProps<OpGraphFlowNode>) => {
                 position={Position.Top}
             />
             <div className='op-graph-node-label'>{data.label}</div>
-            <div className='op-graph-node-file'>
-                {blockMeta(opCount, data.durationSeconds ?? 0, data.memoryDeltaBytes ?? 0)}
-            </div>
+            <div className='op-graph-node-file'>{data.fileIdentifier}</div>
             <OpGraphBlockExpander
                 instanceId={instanceId}
                 opCount={opCount}
--- a/src/components/operation-graph/OpGraphBlockNode.tsx
+++ b/src/components/operation-graph/OpGraphBlockNode.tsx
@@ -29,7 +29,7 @@ const OpGraphBlockNode = memo(({ data }: NodeProps<OpGraphFlowNode>) => {
             >
                 {data.label}
             </div>
-            <div className='op-graph-node-file'>{data.fileIdentifier}</div>
+            <div className='op-graph-node-meta'>{data.metaLine}</div>
             <OpGraphBlockExpander
                 instanceId={instanceId}
                 opCount={opCount}
--- a/src/components/operation-graph/OpGraphBlockNode.tsx
+++ b/src/components/operation-graph/OpGraphBlockNode.tsx
@@ -20,7 +20,15 @@ const OpGraphBlockNode = memo(({ data }: NodeProps<OpGraphFlowNode>) => {
                 type='target'
                 position={Position.Top}
             />
-            <div className='op-graph-node-label'>{data.label}</div>
+            {/* The label is capped at `NODE_MAX_WIDTH` and ellipsised, and a block
+                naming several modules can outrun it. The native tooltip is the only
+                place the whole name is readable, on the node or in the panel. */}
+            <div
+                className='op-graph-node-label'
+                title={data.label}
+            >
+                {data.label}
+            </div>
             <div className='op-graph-node-file'>{data.fileIdentifier}</div>
             <OpGraphBlockExpander
                 instanceId={instanceId}
--- a/src/components/operation-graph/OpGraphBlockNode.tsx
+++ b/src/components/operation-graph/OpGraphBlockNode.tsx
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { Handle, type NodeProps, Position } from '@xyflow/react';
+import { memo } from 'react';
+import { formatMemorySize, formatSize } from '../../functions/math';
+import OpGraphBlockExpander from './OpGraphBlockExpander';
+import type { OpGraphFlowNode } from './opGraphTypes';
+
+const blockMeta = (opCount: number, durationSeconds: number, memoryDeltaBytes: number): string => {
+    const parts = [`${opCount} ops`];
+    if (durationSeconds > 0) {
+        parts.push(`${formatSize(durationSeconds, 2)} s`);
+    }
+    if (memoryDeltaBytes !== 0) {
+        const sign = memoryDeltaBytes > 0 ? '+' : '-';
+        parts.push(`${sign}${formatMemorySize(Math.abs(memoryDeltaBytes), 0)}`);
+    }
+    return parts.join(' · ');
+};
+
+const OpGraphBlockNode = memo(({ data }: NodeProps<OpGraphFlowNode>) => {
+    const instanceId = data.blockInstanceId;
+    const opCount = data.opCount ?? 0;
+    if (instanceId === undefined) {
+        return null;
+    }
+
+    return (
+        <>
+            <Handle
+                type='target'
+                position={Position.Top}
+            />
+            <div className='op-graph-node-label'>{data.label}</div>
+            <div className='op-graph-node-file'>
+                {blockMeta(opCount, data.durationSeconds ?? 0, data.memoryDeltaBytes ?? 0)}
+            </div>
+            <OpGraphBlockExpander
+                instanceId={instanceId}
+                opCount={opCount}
+                isExpanded={false}
+            />
+            {data.buriedMatchCount ? (
+                <span
+                    className='op-graph-node-buried-badge'
+                    title={`${data.buriedMatchCount} filter ${
+                        data.buriedMatchCount === 1 ? 'match' : 'matches'
+                    } inside`}
+                >
+                    {`+${data.buriedMatchCount}`}
+                </span>
+            ) : null}
+            <Handle
+                type='source'
+                position={Position.Bottom}
+            />
+        </>
+    );
+});
+
+OpGraphBlockNode.displayName = 'OpGraphBlockNode';
+
+export default OpGraphBlockNode;
--- a/src/components/operation-graph/OpGraphInfoPanel.tsx
+++ b/src/components/operation-graph/OpGraphInfoPanel.tsx
@@ -20,7 +20,6 @@ import MemoryConfigRow from '../MemoryConfigRow';
 import MemoryTag from '../MemoryTag';
 import SourceFileButton from '../operation-details/SourceFileButton';
 import PerfOverlayOpMetric from '../perf-overlay/PerfOverlayOpMetric';
-import { sumOptional } from './opGraphRepeatBlocks';
 import type { OpGraphBlockSummary } from './opGraphTypes';
 
 interface ConnectedOpGroup {
@@ -72,9 +71,6 @@ const getConnectedOpGroups = (
 
     return Array.from(groupsByKey.values());
 };
-
-const tensorBytes = (tensors: { size: number | null }[]): number =>
-    tensors.reduce((total, tensor) => total + (tensor.size ?? 0), 0);
 
 const uniqueNameCounts = (names: readonly string[]): Array<{ name: string; count: number }> => {
     const counts: Array<{ name: string; count: number }> = [];
@@ -273,10 +269,9 @@ const OpGraphBlockPanel = ({
         [members, memberIds, operationNamesById],
     );
     const typeCounts = useMemo(() => uniqueNameCounts(members.map((member) => member.name)), [members]);
-    const durationSeconds = sumOptional(members.map((member) => member.duration));
-    const memoryDeltaBytes = sumOptional(
-        members.map((member) => tensorBytes(member.outputs) - tensorBytes(member.inputs)),
-    );
+    // Carried on the summary, not re-derived: the node's meta line is on screen at
+    // the same time and is formatted from these same two numbers.
+    const { durationSeconds, memoryDeltaBytes } = block;
     const firstOpId = block.operationIds[0];
     const lastOpId = block.operationIds[block.operationIds.length - 1];
     const inputCount = inputGroups.reduce((total, group) => total + group.tensors.length, 0);
--- a/src/components/operation-graph/OpGraphInfoPanel.tsx
+++ b/src/components/operation-graph/OpGraphInfoPanel.tsx
@@ -290,7 +290,12 @@ const OpGraphBlockPanel = ({
         >
             <header className='op-graph-panel-header'>
                 <div className='op-graph-panel-titles'>
-                    <h2 className='op-graph-panel-label'>{block.label}</h2>
+                    <h2
+                        className='op-graph-panel-label'
+                        title={block.label}
+                    >
+                        {block.label}
+                    </h2>
                     <p className='op-graph-panel-id'>
                         {`ops ${firstOpId}–${lastOpId} · instance ${block.instanceIndex + 1} of ${block.instanceCount}`}
                     </p>
@@ -434,7 +439,10 @@ const OpGraphInfoPanel = memo(
             >
                 <header className='op-graph-panel-header'>
                     <div className='op-graph-panel-titles'>
-                        <h2 className='op-graph-panel-label'>
+                        <h2
+                            className='op-graph-panel-label'
+                            title={`${operationId} ${operation?.name ?? ''}`.trim()}
+                        >
                             {operationId} {operation?.name}
                         </h2>
                         <p
--- a/src/components/operation-graph/OpGraphInfoPanel.tsx
+++ b/src/components/operation-graph/OpGraphInfoPanel.tsx
@@ -231,7 +231,7 @@ const ConnectedOpGroupList = ({ groups, keyPrefix, onLocate }: ConnectedOpGroupL
 
 interface OpGraphBlockPanelProps {
     block: OpGraphBlockSummary;
-    operationList: OperationDescription[];
+    operationById: Map<number, OperationDescription>;
     operationNamesById: Map<number, string>;
     onLocateOperation: (operationId: number) => void;
     isPerfOverlayActive: boolean;
@@ -240,7 +240,7 @@ interface OpGraphBlockPanelProps {
 
 const OpGraphBlockPanel = ({
     block,
-    operationList,
+    operationById,
     operationNamesById,
     onLocateOperation,
     isPerfOverlayActive,
@@ -250,9 +250,9 @@ const OpGraphBlockPanel = ({
     const members = useMemo(
         () =>
             block.operationIds
-                .map((id) => operationList.find((operation) => operation.id === id))
+                .map((id) => operationById.get(id))
                 .filter((operation): operation is OperationDescription => operation !== undefined),
-        [block.operationIds, operationList],
+        [block.operationIds, operationById],
     );
     const inputGroups = useMemo(
         () =>
@@ -377,7 +377,7 @@ const OpGraphBlockPanel = ({
 
 interface OpGraphInfoPanelProps {
     operationId: number;
-    operationList: OperationDescription[];
+    operationById: Map<number, OperationDescription>;
     operationNamesById: Map<number, string>;
     onLocateOperation: (operationId: number) => void;
     isPerfOverlayActive: boolean;
@@ -391,7 +391,7 @@ interface OpGraphInfoPanelProps {
 const OpGraphInfoPanel = memo(
     ({
         operationId,
-        operationList,
+        operationById,
         operationNamesById,
         onLocateOperation,
         isPerfOverlayActive,
@@ -400,7 +400,7 @@ const OpGraphInfoPanel = memo(
         block = null,
     }: OpGraphInfoPanelProps) => {
         const navigate = useNavigate();
-        const operation = operationList.find((op) => op.id === operationId);
+        const operation = operationById.get(operationId);
         const operationSourceData = operation ? extractOperationSourceData(operation) : null;
 
         const inputGroups = useMemo(
@@ -418,7 +418,7 @@ const OpGraphInfoPanel = memo(
             return (
                 <OpGraphBlockPanel
                     block={block}
-                    operationList={operationList}
+                    operationById={operationById}
                     operationNamesById={operationNamesById}
                     onLocateOperation={onLocateOperation}
                     isPerfOverlayActive={isPerfOverlayActive}
--- a/src/components/operation-graph/OpGraphInfoPanel.tsx
+++ b/src/components/operation-graph/OpGraphInfoPanel.tsx
@@ -12,6 +12,7 @@ import ROUTES from '../../definitions/Routes';
 import { StackTraceLanguage } from '../../definitions/StackTrace';
 import { toReadableShape, toReadableType } from '../../functions/formatting';
 import { formatMemorySize, formatSize } from '../../functions/math';
+import { getBlockBoundaryTensors } from './opGraphBlockBoundary';
 import { extractOperationSourceData } from '../../functions/stackTraceSource';
 import type { OperationDescription, Tensor } from '../../model/APIData';
 import { BufferTypeLabel } from '../../model/BufferType';
@@ -85,45 +86,6 @@ const uniqueNameCounts = (names: readonly string[]): Array<{ name: string; count
         }
     }
     return counts;
-};
-
-const withExternalEndpoints = (tensor: Tensor, memberIds: ReadonlySet<number>, direction: NodeRelation): Tensor => {
-    if (direction === NodeRelation.Input) {
-        const producers = tensor.producers ?? [];
-        const producerNames = tensor.producerNames ?? [];
-        const keep = producers
-            .map((id, index) => ({ id, name: producerNames[index] ?? '' }))
-            .filter((entry) => !memberIds.has(entry.id));
-        return { ...tensor, producers: keep.map((entry) => entry.id), producerNames: keep.map((entry) => entry.name) };
-    }
-    const consumers = tensor.consumers ?? [];
-    const consumerNames = tensor.consumerNames ?? [];
-    const keep = consumers
-        .map((id, index) => ({ id, name: consumerNames[index] ?? '' }))
-        .filter((entry) => !memberIds.has(entry.id));
-    return { ...tensor, consumers: keep.map((entry) => entry.id), consumerNames: keep.map((entry) => entry.name) };
-};
-
-const getBlockBoundaryTensors = (
-    members: OperationDescription[],
-    memberIds: ReadonlySet<number>,
-    direction: NodeRelation,
-): Tensor[] => {
-    const tensors: Tensor[] = [];
-    const seenIds = new Set<number>();
-    for (const member of members) {
-        const list = (direction === NodeRelation.Input ? member.inputs : member.outputs) ?? [];
-        for (const tensor of list) {
-            if (!seenIds.has(tensor.id)) {
-                const counterparts = (direction === NodeRelation.Input ? tensor.producers : tensor.consumers) ?? [];
-                if (counterparts.length === 0 || counterparts.some((id) => !memberIds.has(id))) {
-                    seenIds.add(tensor.id);
-                    tensors.push(withExternalEndpoints(tensor, memberIds, direction));
-                }
-            }
-        }
-    }
-    return tensors;
 };
 
 interface TensorDetailsProps {
--- a/src/components/operation-graph/OpGraphInfoPanel.tsx
+++ b/src/components/operation-graph/OpGraphInfoPanel.tsx
@@ -11,6 +11,7 @@ import { NodeRelation } from '../../definitions/NodeRelation';
 import ROUTES from '../../definitions/Routes';
 import { StackTraceLanguage } from '../../definitions/StackTrace';
 import { toReadableShape, toReadableType } from '../../functions/formatting';
+import { formatMemorySize, formatSize } from '../../functions/math';
 import { extractOperationSourceData } from '../../functions/stackTraceSource';
 import type { OperationDescription, Tensor } from '../../model/APIData';
 import { BufferTypeLabel } from '../../model/BufferType';
@@ -19,6 +20,8 @@ import MemoryConfigRow from '../MemoryConfigRow';
 import MemoryTag from '../MemoryTag';
 import SourceFileButton from '../operation-details/SourceFileButton';
 import PerfOverlayOpMetric from '../perf-overlay/PerfOverlayOpMetric';
+import { sumOptional } from './opGraphRepeatBlocks';
+import type { OpGraphBlockSummary } from './opGraphTypes';
 
 interface ConnectedOpGroup {
     key: string;
@@ -68,6 +71,63 @@ const getConnectedOpGroups = (
     });
 
     return Array.from(groupsByKey.values());
+};
+
+const tensorBytes = (tensors: { size: number | null }[]): number =>
+    tensors.reduce((total, tensor) => total + (tensor.size ?? 0), 0);
+
+const uniqueNameCounts = (names: readonly string[]): Array<{ name: string; count: number }> => {
+    const counts: Array<{ name: string; count: number }> = [];
+    const indexByName = new Map<string, number>();
+    for (const name of names) {
+        const existing = indexByName.get(name);
+        if (existing === undefined) {
+            indexByName.set(name, counts.length);
+            counts.push({ name, count: 1 });
+        } else {
+            counts[existing].count += 1;
+        }
+    }
+    return counts;
+};
+
+const withExternalEndpoints = (tensor: Tensor, memberIds: ReadonlySet<number>, direction: NodeRelation): Tensor => {
+    if (direction === NodeRelation.Input) {
+        const producers = tensor.producers ?? [];
+        const producerNames = tensor.producerNames ?? [];
+        const keep = producers
+            .map((id, index) => ({ id, name: producerNames[index] ?? '' }))
+            .filter((entry) => !memberIds.has(entry.id));
+        return { ...tensor, producers: keep.map((entry) => entry.id), producerNames: keep.map((entry) => entry.name) };
+    }
+    const consumers = tensor.consumers ?? [];
+    const consumerNames = tensor.consumerNames ?? [];
+    const keep = consumers
+        .map((id, index) => ({ id, name: consumerNames[index] ?? '' }))
+        .filter((entry) => !memberIds.has(entry.id));
+    return { ...tensor, consumers: keep.map((entry) => entry.id), consumerNames: keep.map((entry) => entry.name) };
+};
+
+const getBlockBoundaryTensors = (
+    members: OperationDescription[],
+    memberIds: ReadonlySet<number>,
+    direction: NodeRelation,
+): Tensor[] => {
+    const tensors: Tensor[] = [];
+    const seenIds = new Set<number>();
+    for (const member of members) {
+        const list = (direction === NodeRelation.Input ? member.inputs : member.outputs) ?? [];
+        for (const tensor of list) {
+            if (!seenIds.has(tensor.id)) {
+                const counterparts = (direction === NodeRelation.Input ? tensor.producers : tensor.consumers) ?? [];
+                if (counterparts.length === 0 || counterparts.some((id) => !memberIds.has(id))) {
+                    seenIds.add(tensor.id);
+                    tensors.push(withExternalEndpoints(tensor, memberIds, direction));
+                }
+            }
+        }
+    }
+    return tensors;
 };
 
 interface TensorDetailsProps {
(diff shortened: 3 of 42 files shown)


src/components/operation-graph/opGraphPerfOverlay.ts
```diff
@@ -139,17 +139,54 @@ export const buildOpGraphPerfOverlay = (
...
+    const tOfNs = (ns: number): number | null => {
+        if (ns <= 0) {
+            return null;
+        }
+        if (overlay.minNs === overlay.maxNs) {
+            return 0;
+        }
+        const logMin = Math.log10(overlay.minNs);
+        const range = Math.log10(overlay.maxNs) - logMin;
+        return Math.min(1, Math.max(0, (Math.log10(ns) - logMin) / range));
```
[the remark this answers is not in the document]
aidemsined: Fixed in 3c6f151f.

Confirmed both failure modes. The ramp normalised node totals against the per-operation `minNs`/`maxNs`, so a block — which outruns the slowest single op almost by construction — clamped to 1, and every block came out equally hot. With all member rows on one duration the range collapsed to a point and forced every block to 0.

The scale is now derived from the totals of what is actually drawn. The legend moved onto the same range, since it is the key for those bars and would otherwise print endpoints the ramp no longer uses.

Two new cases fail on the old scale and pass on the new one: two blocks of different weight now separate to 0 and 1 instead of both clamping to 1, and a block still outranks an ordinary node when every row shares a duration.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -366,20 +444,43 @@ const OperationGraphInner = ({
...
             // deallocate), so selection falls back rather than point at nothing.
             const desired = selectedOperationIdRef.current;
-            const isPresent = desired !== null && graph.nodes.some((node) => node.data.operationId === desired);
+            const isPresent =
+                desired !== null &&
+                graph.nodes.some(
+                    (node) =>
+                        node.data.operationId === desired ||
+                        (node.data.memberOperationIds !== undefined && node.data.memberOperationIds.includes(desired)),
+                );
```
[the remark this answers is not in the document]
aidemsined: Fixed in 3c6f151f.

Reachable exactly as described — the rebuild deliberately keeps a member id as the selection when a block closes over it, so the exact-id lookup missed the entry and `next` fell back to `nodeIndex[0]`.

The lookup now matches `memberOperationIds` as well. It moved to `opGraphNavigation.ts` so the behaviour could be covered directly rather than only through a full graph render; the block case fails without the member match.

src/components/operation-graph/opGraphRepeatBlocks.ts
```diff
@@ -0,0 +1,334 @@
...
+/* eslint-disable no-continue -- window scan skips consumed spans instead of nesting four levels */
+
+import type { OpGraphSourceOperation, RepeatBlockInstance } from './opGraphTypes';
+
+export const MIN_REPEAT_WINDOW = 2;
+export const MIN_REPEAT_COUNT = 2;
+// A repeated subgraph is a layer, not half the graph. Descending from ⌊N/2⌋
+// made the no-repeat case — the common one — O(N³) and folded 12 layers into
+// two half-model blocks. #1583
+export const MAX_REPEAT_WINDOW = 64;
```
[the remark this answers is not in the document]
aidemsined: Accepted as a limitation, now documented in the PR description.

One correction to the framing: the 64 is `MAX_REPEAT_WINDOW`, not `MAX_DETECT_OPS` — the latter is 10,000 and guards whether detection runs at all. The constant on this line caps window length, so the effect you describe is real: `maxLength = Math.min(MAX_REPEAT_WINDOW, ...)` means a repeated window over 64 ops is never found.

The cap is deliberate rather than an oversight. Scanning down from ⌊N/2⌋ made the no-repeat case — the common one — O(N³), and it folded 12 layers into two half-model blocks. Raising it wants a scalable detector, which is a bigger change than this PR.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -875,6 +1101,22 @@ const OperationGraphInner = ({
...
+            if (node.data.blockInstanceId !== undefined) {
+                toggleBlockExpansion(node.data.blockInstanceId);
+                return;
+            }
+            const instance = detectedBlocks.find(
+                (block) => expandedBlockIds.has(block.instanceId) && block.operationIds.includes(node.data.operationId),
+            );
+            if (instance !== undefined) {
+                toggleBlockExpansion(instance.instanceId);
+            }
```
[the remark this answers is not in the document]
aidemsined: Agreed, and noted in the PR description as a limitation.

Confirmed: the expander is a real button with `aria-expanded`, but it only renders on the folded block node, so once a block is unrolled `onNodeDoubleClick` is the only per-instance fold. Keyboard users can fold everything from the toolbar and unroll individually, but cannot fold one instance.

Leaving it for #1942, which already covers the toolbar and toggle chrome, rather than adding a control here.

tests/OperationGraphRebuildDiscipline.spec.tsx
```diff
@@ -192,11 +202,16 @@ import type { ReportFolder } from '../src/definitions/Reports';
...
+    fileIdentifier = `model.py:${id}`,
+): OperationDescription =>
     ({
         id,
         name,
-        operationFileIdentifier: `model.py:${id}`,
+        operationFileIdentifier: fileIdentifier,
         // Tensor ids matter once an operation expands: they are how an edge finds
         // the device operation at the far end of the boundary.
         outputs: [{ id: id * 10, shape: 'Shape([1, 32])', consumers }],
```
dcblundell, line 245: **Must fix:** integration fixture omits the three fields this branch added

Anchored on the fixture factory; the finding is `sourceFor` at `OperationGraphRebuildDiscipline.spec.tsx:232-243`. It reimplements the component's `OperationDescription -> OpGraphSourceOperation` mapping but drops `inputShapes`, `durationSeconds` and `memoryDeltaBytes` (compare `OperationGraphReactFlow.tsx:357-375`). Detection fingerprints on `inputShapes`, so every integration assertion folds a source shape production never produces - deleting `inputShapes` from the real mapping would over-fold live graphs with the suite green. The hook mock at `:163` already discards the array as `_operations`: capture it into `harness`, assert the mapping carries all three, and have `deliver` build from the captured value instead of `sourceFor`.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -480,6 +589,80 @@ const OperationGraphInner = ({
...
+
+    const toggleBlockExpansion = useCallback(
+        (instanceId: string) => {
+            const block = detectedBlocks.find((entry) => entry.instanceId === instanceId);
+            const firstOpId = block?.operationIds[0];
+            armViewportAnchor(instanceId, firstOpId === undefined ? instanceId : String(firstOpId));
+            setExpandedBlockIds((previous) => {
+                const next = new Set(previous);
+                if (next.delete(instanceId)) {
+                    if (block !== undefined) {
```
dcblundell: **Must fix:** `setState` called from inside a state updater

`setExpandedOperationIds` runs inside the `setExpandedBlockIds` updater. Updaters must be pure - under StrictMode this fires twice, harmless today only because `delete` is idempotent, and any non-idempotent change here becomes a heisenbug. The fold/unroll decision is already knowable outside (`expandedBlockIds.has(instanceId)`, and `block` is in the closure), so branch in the callback body and call both setters as siblings; that also lets this share the member-dropping loop with `collapseAllBlocks` at `:640-650`, currently a third copy.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -321,21 +387,35 @@ const OperationGraphInner = ({
     // Only the expanded operations are assembled: one operation's frame stream
```
dcblundell: **Must fix:** stale comment

This comment ("Only the expanded operations are assembled...") describes `deviceSubgraphs`, but `collapsedMemberIds` was inserted between them, so it now heads a block it does not describe - `deviceSubgraphs` starts at `:402`. CONVENTIONS.md treats stale comments as review blockers. Move it down to immediately precede `const deviceSubgraphs`.

src/components/operation-graph/OpGraphInfoPanel.tsx
```diff
@@ -70,6 +73,63 @@ const getConnectedOpGroups = (
...
+            indexByName.set(name, counts.length);
+            counts.push({ name, count: 1 });
+        } else {
+            counts[existing].count += 1;
+        }
+    }
+    return counts;
+};
+
+const withExternalEndpoints = (tensor: Tensor, memberIds: ReadonlySet<number>, direction: NodeRelation): Tensor => {
```
dcblundell: **Must fix:** boundary-tensor logic has no test coverage

`getBlockBoundaryTensors` / `withExternalEndpoints` filter two index-aligned array pairs (`producers`/`producerNames`, `consumers`/`consumerNames`) in lockstep - the most intricate new logic in the branch. The only block-panel test (`OperationGraphRebuildDiscipline.spec.tsx:1357`) uses fixtures with `inputs: []` and outputs carrying no `producerNames`, so the Input direction and the name realignment never run at all. An off-by-one between ids and names would silently attribute a boundary tensor to the wrong operation, with nothing failing.

src/components/operation-graph/opGraphRepeatBlocks.ts
dcblundell, line 19: **Should fix:** `MAX_REPEAT_WINDOW` is an unmet acceptance criterion, untested, and the reason for it looks removable

#1583 asks for the smallest repeating window "extended as far as it matches" with no bound, so a tile over 64 kept ops is never folded - and the issue's headline case is a transformer layer, which routinely decomposes into more than 64 traced ops. The PR lists this under "tracked separately" but names no issue, and the only in-code reference is #1583 itself, i.e. the criterion it does not meet. `MAX_DETECT_OPS` has a test; this cap has none. The O(N^3) motive also looks avoidable: the window scan runs two O(length) `isRangeConsumed` walks *before* the O(1) discriminator, and inserting `ids[start] !== ids[start + length] -> continue` ahead of them measured ~10x on that step (25ms -> 2.5ms at 10k ops). Land that, then either raise/remove the cap or file the tracking issue and cite it here.

src/components/operation-graph/opGraphRepeatBlocks.ts
```diff
@@ -0,0 +1,334 @@
...
+    const stems = new Set<string>();
+    for (const [stem, count] of counts) {
+        if (count > threshold) {
+            stems.add(stem);
+        }
+    }
+    return stems;
+};
+
+export function formatRepeatPatternLabel(
```
dcblundell, line 203: **Should fix (decision):** block labels contradict the spec

#1583 says each copy is "labelled `Block A x N` (then `Block B`, ...)"; this derives the label from source-file stems instead. `blockLetter` survives only as the anonymous fallback at `:195` (unreachable for real reports - it needs every member to have a blank name *and* a blank `fileIdentifier`) and the collision suffix at `:203`; neither is tested. My recommendation is to **amend the issue, not the code** - a derived name is strictly more informative, and the out-of-scope list forbids only *user* rename. But the letters bought two things now missing: a stable short handle per pattern (collisions degrade to `... - A`, so the letter scheme appears only in the failure case), and bounded label width (four parts plus an ellipsis pushes the node toward `NODE_MAX_WIDTH` 560). Carrying both, e.g. `A - norm + attention x 24`, keeps them. Six spec assertions currently pin the derived scheme with no hint the issue disagrees.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -480,6 +589,80 @@ const OperationGraphInner = ({
...
+        setExpandedOperationIds((previous) => {
+            const next = new Set(previous);
+            for (const memberOpId of collapsedMembers) {
+                next.delete(memberOpId);
+            }
+            return next;
+        });
+    }, [detectedBlocks]);
+
+    const handleHideDeallocateChange = useCallback((next: boolean) => {
```
dcblundell: **Should fix:** toggling hide-deallocate now discards *all* device-op expansions, not just folded blocks'

The criterion is only "toggling that filter folds every block again", but `setExpandedOperationIds(NOTHING_EXPANDED)` also clears expansions on ops belonging to no detected block. On the merge-base this handler was a bare `setHideDeallocate`, so this silently changes pre-existing behaviour: a user with several ops open loses all of them on a deallocate toggle. `collapseAllBlocks` at `:640` already has the narrower form - mirror it and drop only members of detected blocks.

dcblundell (commented):
Combined review (DRY / TEST / PERF / STANDARDS / SECURITY / REQUIREMENTS lenses), posted as line-anchored comments: 4 Must fix, 21 Should fix. SECURITY found nothing exploitable - the diff is frontend-only and touches no trust boundary. REQUIREMENTS scored 27/30 criteria against #1583: one contradicted (block labels), two partial (`MAX_REPEAT_WINDOW`, in-graph locate). Two items are decisions rather than fixes and are flagged as such in the relevant threads: the label scheme (recommend amending #1583) and whether "Dim unrelated edges" belongs in this PR. Consider/nit findings were reviewed but not posted.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -179,6 +196,16 @@ const OpGraphToolbar = memo(
                     disabled={isDisabled}
                 />
 
+                <Switch
```
dcblundell, line 202: **Should fix (scope):** "Dim unrelated edges" answers no criterion in #1583

It is default-off, container-class only and cheap, so this is bookkeeping rather than risk - but it is a new user-visible toolbar control in a PR that closes a subgraph-collapse issue, and neither #1583 nor #1942 records that it shipped. Either split it into its own issue/PR or add it to #1583's Behaviour list so the issue and the code agree on what closing it means. Unlike its two neighbours it also carries no tooltip explaining what it does.

src/components/operation-graph/OpGraphInfoPanel.tsx
```diff
@@ -169,30 +229,178 @@ const ConnectedOpGroupList = ({ groups, keyPrefix, onLocate }: ConnectedOpGroupL
...
+                            />
+                        </Tooltip>
+                    </div>
+                )}
+            </header>
+
+            <dl className='op-graph-panel-stats'>
+                <dt>Operations</dt>
+                <dd>{block.operationIds.length}</dd>
+                {durationSeconds > 0 ? (
```
dcblundell, line 281: **Should fix:** a block's duration and memory delta are derived twice, by independent paths, and this one is untested

The node's meta line comes from `OpGraphSourceOperation.durationSeconds`/`memoryDeltaBytes` summed in `opGraphBuilder.ts:129-131` and formatted by `formatBlockMeta` (`opGraphLayout.ts:68`); the panel re-derives both from `OperationDescription` and re-inlines the same format expressions. Both are on screen at once, so drift shows as a node and a panel disagreeing about one block. `formatBlockMeta` is unit-tested but this path is not - both gates (`durationSeconds > 0`, `memoryDeltaBytes !== 0`) and the sign branch are uncovered. Carry the sums on `OpGraphBlockSummary` (the builder already computes them), share the formatters, and add one test asserting node and panel agree.

src/components/operation-graph/opGraphLayoutWorker.ts
```diff
@@ -29,19 +31,36 @@ const layoutCache = new Map<string, OpGraphBuiltGraph>();
 
 let sourceVersion = -1;
 let operations: OpGraphSourceOperation[] = [];
+const detectionByDeallocate = new Map<boolean, RepeatBlockInstance[]>();
```
dcblundell: **Should fix:** detection cache is keyed on the boolean alone, and its invalidation is untested

`detectionByDeallocate` depends entirely on the `.clear()` at `:134` for correctness. The layout cache ten lines below *does* key on `sourceVersion`, with a comment saying a stale entry "can never be served if that clear is ever moved or missed" - the same reasoning applies here. If the clear moves, report B is folded using report A's block instances, whose member op ids exist in B but mean unrelated operations. Add the version to this key, and add a `setGraph(1)` -> build -> `setGraph(2)` -> build test asserting the second `detectedBlocks` is not the first's.

tests/OperationGraphRebuildDiscipline.spec.tsx
```diff
@@ -34,19 +34,23 @@ const harness: {
...
     setEdges: null,
     onNodeClick: null,
+    onNodeDoubleClick: null,
     onNodesChange: null,
     onNodeMouseEnter: undefined,
     onNodeMouseLeave: undefined,
+    onPaneClick: null,
 };
 
 // What `useNodesState` hands the view as its change applier. Stable, because the
```
dcblundell, line 62: **Should fix:** `setCenter` / `setViewport` are inert stubs, so the anchor mechanism this branch extends is unobservable

Anchored near the mock block; the finding is the `useReactFlow` mock at `OperationGraphRebuildDiscipline.spec.tsx:73-76`. Both are no-ops and `getNode` at `:74` returns a node for *any* id, which actively defeats the `nodeId ?? fallbackNodeId` fallback that exists precisely because the anchor id flips between the block id and its first member id across a fold. Make both `vi.fn()` and have `getNode` return `undefined` for unknown ids, then assert one `setViewport` after a block toggle with `setCenter` *not* called in the same pass (the "graph jumps to op 0" regression). That also unblocks testing filter Next/Prev stepping onto a folded block (`OperationGraphReactFlow.tsx:814-828`), which the last commit fixed for the toolbar steppers but not the filter ones.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -201,6 +228,28 @@ const OpGraphToolbar = memo(
                     isDisabled={isDisabled}
                 />
             </div>
+
+            {hasBlocks ? (
+                <div className='op-graph-toolbar-row'>
+                    <span className='op-graph-toolbar-group-label'>Repeats</span>
+                    <Button
+                        variant={ButtonVariant.OUTLINED}
+                        disabled={isDisabled || areAllBlocksExpanded}
```
dcblundell, line 240: **Should fix:** the mid-build disabled branch is untested for both Repeats buttons

Each button has two independent disable reasons; `OperationGraphRebuildDiscipline.spec.tsx:1290-1319` covers only the all-expanded / all-collapsed one, and `OpGraphToolbarSwitches.spec.tsx` never passes `hasBlocks` at all. Mid-build is the branch that matters - an unroll queued against a node set about to be replaced. Extend `renderToolbar` with the block props and add `hasBlocks: false` plus the two `isDisabled: true` cases there.

src/components/operation-graph/OpGraphBlockExpander.tsx
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: Apache-2.0
```
dcblundell, line 1: **Should fix:** near-verbatim fork of `OpGraphDeviceOpExpander`

Identical `handleMouseDown` body, identical `className='op-graph-node-expander nodrag nopan'`, identical `title`/`aria-label`/`aria-expanded` wiring - and both multi-line rationale comments copied word for word. A future fix to the mousedown behaviour has to land twice or one comment starts lying. Extract one presentational expander taking `action`, `isExpanded`, `count` and `onToggle`; the two node components supply the label and the context-bound toggle. Note both call sites hard-code `isExpanded={false}`, so that branch is currently dead in both - see the separate comment on `OpGraphBlockNode.tsx` before deleting it.

src/components/operation-graph/opGraphBuilder.ts
```diff
@@ -42,19 +52,28 @@ function collectCandidateEdges(operations: OpGraphSourceOperation[]): CandidateE
...
     operations: OpGraphSourceOperation[],
-    { hideDeallocate, deviceSubgraphs }: OpGraphBuildOptions,
-): OpGraphBuiltGraph {
-    const candidates = collectCandidateEdges(operations);
-
+    hideDeallocate: boolean,
+): OpGraphSourceOperation[] {
     // Connectivity is decided before the deallocate filter so hiding deallocate
     // ops cannot drop their neighbours or pull in ops that were always isolated.
+    const candidates = collectCandidateEdges(operations);
```
dcblundell: **Should fix:** `collectCandidateEdges` now runs two to three times per build

`buildOpGraph` calls it directly at `:76` and again via `getKeptOperations` at `:95`, and the worker pays a third pass through `detectedBlocksOf` -> `getKeptOperations` (`opGraphLayoutWorker.ts:61`). Before the split, connectivity came from the single `candidates` pass. On the 4k-8k op graphs this feature explicitly targets that is a doubled ops x outputs x consumers walk on the build hot path. Have `getKeptOperations` accept the precomputed `candidates` (or export `connectedIdsOf(candidates)`) and pass `buildOpGraph`'s own array in.

src/components/operation-graph/opGraphBuilder.ts
```diff
@@ -73,13 +92,65 @@ export function buildOpGraph(
...
+                const size = estimateBlockNodeSize(collapsedInstance.label, meta);
+                nodes.push({
+                    id: collapsedInstance.instanceId,
+                    type: OpGraphNodeType.BLOCK,
+                    position: { x: 0, y: 0 },
+                    ...size,
+                    data: {
+                        operationId: collapsedInstance.operationIds[0],
+                        label: collapsedInstance.label,
+                        fileIdentifier: meta,
```
dcblundell: **Should fix:** `fileIdentifier` is overloaded to carry the block's stats string

`fileIdentifier: meta` puts `"12 ops - 1.2 s - +4 MB"` into a field `opGraphTypes.ts:71` documents as the source-file identifier, and `OpGraphBlockNode.tsx` then renders it through `.op-graph-node-file`, so the class lies too. Anyone grepping `fileIdentifier` gets a false positive. Add an explicit `metaLine?: string` to `OpGraphNodeData` (it already gained six optional block fields) with a matching class. Note `durationSeconds`/`memoryDeltaBytes` are also written into node data at `:148-149` but read by nothing except `opGraphBuilder.spec.ts` - either drop those or drop `meta` and format in the component, not both.

src/components/operation-graph/OpGraphBlockNode.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <div className='op-graph-node-file'>{data.fileIdentifier}</div>
+            <OpGraphBlockExpander
+                instanceId={instanceId}
+                opCount={opCount}
+                isExpanded={false}
+            />
+            {data.buriedMatchCount ? (
+                <span
+                    className='op-graph-node-buried-badge'
+                    title={`${data.buriedMatchCount} filter ${
```
dcblundell, line 41: **Should fix (a11y):** no keyboard path to fold a single unrolled instance

`isExpanded={false}` is hardcoded, so a block expander never renders `aria-expanded="true"`, and once unrolled the button disappears with the block node. The only ways back are `onNodeDoubleClick` (`OperationGraphReactFlow.tsx:1104-1117`), which is pointer-only, and the toolbar's Fold-all - so a keyboard or screen-reader user can unroll one instance and cannot re-fold just that one. Either render a fold affordance on the unrolled members (which makes the existing `isExpanded` branch live and testable) or bind a key handler alongside the double-click. Worth confirming #1942 covers this a11y gap specifically and not just the cramped chrome.

src/components/operation-graph/opGraphLayoutWorker.ts
```diff
@@ -29,19 +31,36 @@ const layoutCache = new Map<string, OpGraphBuiltGraph>();
 
 let sourceVersion = -1;
 let operations: OpGraphSourceOperation[] = [];
```
dcblundell, line 40: **Should fix:** `LAYOUT_CACHE_LIMIT` was sized before this branch added a third key dimension

The comment above it (`:26-29`) explains the limit in terms of `hideDeallocate` x expanded device-op set. `expandedBlockIds` is now a third, independently-toggleable dimension: every detected instance folds and unrolls on its own, so exploring a handful of blocks mints a distinct key per toggle where the cache was previously rarely full. Each entry is a fully laid-out graph - every node plus its `memberNames` array, every edge, plus its own copy of `graph.blocks` - so a filled cache on an 8k-op report retains tens of MB in the worker for the whole session, freed only on `SET_GRAPH`. Either drop the limit to ~6-8 now that keys churn faster, or budget on total cached nodes rather than entry count.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -179,6 +196,16 @@ const OpGraphToolbar = memo(
                     disabled={isDisabled}
                 />
 
+                <Switch
+                    className='op-graph-toolbar-switch'
+                    checked={focusUnrelatedEdges}
```
dcblundell: **Should fix (naming):** `focusUnrelatedEdges` says the opposite of what it does

`true` *dims* unrelated edges - `OperationGraphReactFlow.scss:397` sets `opacity: 0.35` on every edge that is not critical-path/input/output - and the visible label here is 'Dim unrelated edges'. The prop is declared at `OpGraphToolbar.tsx:452`. Rename to `isDimUnrelatedEdges` / `onDimUnrelatedEdgesChange` (AGENTS.md asks `is*`/`has*` for boolean predicates) and the container class to `op-graph-dim-unrelated-edges`, so the label, the state and the selector all read the same.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -73,6 +75,12 @@ interface OpGraphToolbarProps {
     linkedOpCount: number;
     totalOpCount: number;
     isDisabled: boolean;
+    hiddenMatchCount?: number;
+    hasBlocks?: boolean;
+    areAllBlocksExpanded?: boolean;
+    areAllBlocksCollapsed?: boolean;
+    onExpandAllBlocks?: () => void;
```
dcblundell: **Should fix:** these five block props are optional but the sole caller always supplies them

`hasBlocks`, `areAllBlocksExpanded`, `areAllBlocksCollapsed`, `onExpandAllBlocks` and `onCollapseAllBlocks` all have defaults, yet `OperationGraphReactFlow.tsx:1266-1276` passes every one. The optionality buys nothing and actively hides a bug class: `<Button disabled={isDisabled || areAllBlocksExpanded} onClick={onExpandAllBlocks}>` can render an *enabled* button with `onClick === undefined`, a silent no-op the type system would otherwise catch. Make the five required.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -480,6 +589,80 @@ const OperationGraphInner = ({
         [getNode, getViewport, reportScope],
     );
 
+    const armViewportAnchor = useCallback(
```
dcblundell, line 647: **Should fix:** `armViewportAnchor` was extracted but the pre-existing caller was not migrated

`toggleOperationExpansion` at `:566-579` still inlines the same 12-line anchor body (get viewport, `position * zoom + offset`, stash with `reportScope`) - and this branch edited *that* copy to add `fallbackNodeId: nodeId`. Two copies of the anchor maths will drift. Have it call `armViewportAnchor(nodeId, nodeId)`.

src/components/operation-graph/opGraphRepeatBlocks.ts
```diff
@@ -0,0 +1,334 @@
...
+                instanceIndex,
+                instanceCount: run.repeatCount,
+            });
+        }
+    });
+
+    return instances;
+}
+
+export function sumOptional(values: readonly (number | undefined)[]): number {
```
dcblundell: **Should fix:** generic numeric helpers living in feature modules

`sumOptional` is a general reducer over `(number | undefined)[]` exported from the repeat-block *detector* and imported by both `opGraphBuilder.ts:14` and `OpGraphInfoPanel.tsx:23` - so the info panel now depends on the detection module for arithmetic. Separately `tensorBytes` is defined identically in `OpGraphInfoPanel.tsx:76` and `OperationGraphReactFlow.tsx:128`. AGENTS.md (File organization) asks for pure helpers in `src/functions/math.ts`, where the sibling `formatSize`/`formatMemorySize` already live; move both there. Same theme: `formatBlockMeta` is display formatting sitting in the Dagre geometry module (`opGraphLayout.ts:68`).

src/components/operation-graph/OpGraphInfoPanel.tsx
```diff
@@ -169,30 +229,178 @@ const ConnectedOpGroupList = ({ groups, keyPrefix, onLocate }: ConnectedOpGroupL
...
         operationNamesById,
         onLocateOperation,
         isPerfOverlayActive,
         perfDeviceTimeNs,
         perfColor,
+        block = null,
     }: OpGraphInfoPanelProps) => {
         const navigate = useNavigate();
-        const operation = operationList.find((op) => op.id === operationId);
+        const operation = operationById.get(operationId);
```
dcblundell, line 366: **Should fix:** the block branch pays for the operation panel before returning

`operationById.get`, `extractOperationSourceData` and both `getConnectedOpGroups` memos run and are discarded on every block selection, because the block early-return sits after them. The file's own comment at `:389-390` says every prop must be stable because this re-renders per drag frame, so paying for the unused half undercuts the stated contract. Make `OpGraphInfoPanel` a thin dispatcher (`block !== null ? <OpGraphBlockPanel/> : <OpGraphOperationPanel/>`) and move the operation-side hooks into the new sibling - hooks-before-return then stops forcing the wasted work.

src/components/operation-graph/opGraphCriticalPath.ts
```diff
@@ -84,15 +88,26 @@ export const findCriticalPath = (
         }
     }
 
-    const weightOf = (nodeId: string) => deviceTimeNsByOpId.get(opIdByNodeId.get(nodeId) ?? UNKNOWN_OP_ID) ?? 0;
+    const membersByNodeId = new Map<string, number[]>();
+    for (const node of nodes) {
+        membersByNodeId.set(node.id, node.memberOperationIds ?? [node.operationId]);
+    }
+
+    const weightOf = (nodeId: string) => {
```
dcblundell: **Should fix:** member device-time is summed three times, and recomputed once per edge

"Sum `deviceTimeNs` over a node's member ops" now exists here, in `opGraphPerfOverlay.ts:145-152` (`sumNsByRenderedNode`) and in `OperationGraphReactFlow.tsx:1206-1218`, all three sharing the `memberOperationIds ?? [operationId]` fallback shape. Worse, `weightOf` is called inside the edge relaxation loop (`:127`), so each node's member sum is recomputed once per incoming edge instead of once per node, and `:128` pays a second map lookup for the same length. Build `weightByNodeId` and `memberCountByNodeId` in the existing node loop at `:108-111` and read those in both places - which also removes the need to retain `membersByNodeId`. Then export one `sumDeviceTimeNs` for the other two call sites.

src/components/operation-graph/opGraphPerfOverlay.ts
```diff
@@ -131,6 +131,71 @@ export const buildOpGraphPerfOverlay = (
...
+
+export const getRenderedPerfRange = (
+    overlay: OpGraphPerfOverlay,
+    nodes?: readonly RenderedPerfNode[],
+): { minNs: number; maxNs: number } => {
+    const perOperationRange = { minNs: overlay.minNs, maxNs: overlay.maxNs };
+    if (nodes === undefined) {
+        return perOperationRange;
+    }
+    return spanOfTotals(sumNsByRenderedNode(overlay, nodes), perOperationRange);
```
dcblundell: **Should fix:** two full passes per rebuild, one dead branch, and no gate when the overlay is off

`sumNsByRenderedNode` + `spanOfTotals` run twice over the whole `nodeIndex` - once from `getRenderedPerfRange`, once from `buildPerfNodeStyleByNodeId` - from two memos with effectively the same dependency set, each allocating its own map and a one-element array per non-block node. The `renderedPerfRange` memo is also ungated by `isPerfOverlayActive` (`OperationGraphReactFlow.tsx:981-988`), so it does the full pass even with no perf report loaded, when its only consumer `PerfOverlayLegend` is not rendered. Separately the `nodes === undefined` branch at `:212-220` is unreachable in production - only `opGraphPerfOverlay.spec.ts` exercises it - and it re-derives the log10 normalisation already in `scoreOps`, comment included. Return the range alongside the style map, gate on active, make `nodes` required, and extract the ramp as a shared `logScore`.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -366,20 +446,43 @@ const OperationGraphInner = ({
...
             // An op can drop out between builds (isolated, or filtered as a
             // deallocate), so selection falls back rather than point at nothing.
             const desired = selectedOperationIdRef.current;
-            const isPresent = desired !== null && graph.nodes.some((node) => node.data.operationId === desired);
+            const isPresent =
+                desired !== null &&
+                graph.nodes.some(
+                    (node) =>
+                        node.data.operationId === desired ||
+                        (node.data.memberOperationIds !== undefined && node.data.memberOperationIds.includes(desired)),
```
dcblundell, line 524: **Should fix:** five linear "which block owns this op id?" scans

Here and at `:660`, `:1111`, `:1192`, plus `opGraphNavigation.ts:29-36` - each a `detectedBlocks.find(... .operationIds.includes(id))` or `memberOperationIds.includes(desired)`. Two sit inside per-render memos and one is in the render body. The component already builds `nodeIdByOperationId` for exactly this shape of lookup: memoise a `blockByMemberOperationId: Map<number, OpGraphBlockSummary>` and use it at all four sites, passing the same map into `getAdjacentOperationIds` instead of re-scanning.

src/components/operation-graph/OpGraphToolbar.tsx
aidemsined: > [!NOTE]
> **Decision: it stays in this PR.** Recorded in #1583's Behaviour list, so the issue and the code now agree on what closing it means.

It was added because the collapsed graph needed it, not alongside it. Folding concentrates many edges onto a single block node, so tracing what one block connects to needs the unrelated ones out of the way — without it the collapsed graph reads worse than the unfolded one at exactly the points collapsing was meant to clarify. Splitting it out would land a collapse feature that is harder to read than what it replaced.

On the tooltip: its immediate neighbour `Hide deallocate ops` is a plain `Switch` with no tooltip either. The two that do have one are `PerfGatedSwitch`es, where `tooltipByStatus` explains why the control is *disabled* given the perf-report status, rather than describing the feature. So adding one here alone would create a different inconsistency. There is a real point buried in it though — the switch is a no-op with nothing selected, which the label does not convey — so if it gets a tooltip, `Hide deallocate ops` should get one in the same pass. Happy to do both if you think it is worth it.

src/components/operation-graph/opGraphRepeatBlocks.ts
aidemsined: > [!NOTE]
> **Decision: amend the issue, not the code.** #1583's Behaviour list now describes derived naming, so the spec and the six assertions agree. Taking your reading that a derived name is strictly more informative and that the out-of-scope list rules out only *user* rename.

You were right that the letters bought two things. Handling them separately:

**Bounded width — fixed.** Independently reported from the SentenceBERT graph, where a block spans four modules and was named in 124 characters, 72 of them `ttnn_sentencebert_` four times over, so the node rendered three parts and clipped the fourth. Two rules now apply before `joinCapped`:

- a block of a single operation type is named by that operation rather than its file — two `ttnn.squeeze` read `squeeze`, not `tt_routed_expert`
- a leading `_`-separated prefix every stem shares is dropped, giving `self_attention + self_output + intermediate + output` (52 chars). Stripping stops while every stem keeps a segment of its own, and is abandoned if a remainder would fall under three characters, so `mlp_up + mlp_down` does not degrade to `up + down`.

Measured against the real stems from that report: 124 → 52 characters, comfortably inside `NODE_MAX_WIDTH`. Separately, node labels and both panel headings now carry the full string as a `title`, since all three ellipsise and the whole name was previously readable nowhere.

**Stable short handle — deliberately not added.** A letter prefix would be index-based and computed per build, so it is not stable in the way the name implies: detection order shifts when the deallocate filter toggles, and `A` could then denote a different pattern before and after. A handle that silently re-points seemed worse than none. If it turns out to be wanted, it should key on the pattern fingerprint rather than discovery order — happy to file that if you think it earns its keep.

aidemsined:
> [!NOTE]
> All 4 Must fix and 19 of 21 Should fix are addressed across 8 commits (`3fc7a5b2`..`884e6762`). Two are deliberately not, for reasons below. Gates: `tsc` 0, `eslint` 0, `stylelint` 0, build clean, **2318 tests / 206 files** (from 2285).

The pass was constrained to changes that alter neither behaviour nor UX, so anything that would have has been separated out rather than folded in.

## Must fix

- **`setState` inside a state updater** — `isUnrolling` already decided the branch outside, so the two setters are siblings now. The member-dropping loop existed three times and is one helper.
- **Stale comment** — moved down to `deviceSubgraphs`, which it describes.
- **Integration fixture dropped three fields** — builds now run against the array the view actually hands the worker, captured in the harness, and one test asserts it carries `inputShapes`, `durationSeconds` and `memoryDeltaBytes`. Verified the way you framed it: with `inputShapes` deleted from the real mapping, **that test alone fails and every folding assertion still passes**.
- **Boundary tensors untested** — `getBlockBoundaryTensors` / `withExternalEndpoints` moved to their own module with 10 tests. Two of them fail if the ids and names are filtered independently, which is the off-by-one you described.

## Performance

- **`collectCandidateEdges` 2–3× per build** — callers that hold it pass it in; the worker caches one pass per source, asserted by a test.
- **Window scan discriminator** — landed. Measured on a 10k-op no-repeat graph: **40.4ms → 4.0ms** median of five, matching your ~10×.
- **`weightOf` inside the edge loop** — `weightByNodeId` and `memberCountByNodeId` are built in the existing node loop; `membersByNodeId` is gone. One `sumDeviceTimeNs` now serves all three call sites.
- **Perf overlay double pass** — one pass returns styles and range together, gated on the overlay being active. The `nodes === undefined` branch is deleted and the log ramp extracted.
- **Five linear block scans** — one memoised `blockByMemberOperationId`. The prev/next steppers get a position map instead of `findIndex`. One exception below.
- **`LAYOUT_CACHE_LIMIT`** — 16 → 8, with the third key dimension written into the comment.

## Correctness, naming, structure

- **Detection cache key** — versioned. Demonstrated: with the `SET_GRAPH` clear removed *and* the version dropped, the new test fails; restoring just the version makes it pass with the clear still missing, so the key now carries the guarantee on its own.
- **`focusUnrelatedEdges`** — `isDimUnrelatedEdges` / `onDimUnrelatedEdgesChange`, class `op-graph-dim-unrelated-edges`.
- **Five optional toolbar props** — required. Making them so immediately caught the test harness missing all five, which is the bug class you named. Mid-build disabled coverage added for both Repeats buttons.
- **`fileIdentifier` carrying the stats line** — its own `metaLine` field and class; a stylesheet test pins the two to the same declarations so the rename stayed invisible. The node-data `durationSeconds`/`memoryDeltaBytes` were read by one assertion only and are gone.
- **Block stats derived twice** — the summary carries the sums, the panel reads them, and a test pins the node's meta line to the summary the panel renders. Both gates and the sign branch now covered.
- **Expander fork** — one presentational `OpGraphNodeExpander`; the callers supply only the wording and the toggle.
- **`armViewportAnchor` not migrated** — `toggleOperationExpansion` calls it.
- **Helpers in feature modules** — `sumOptional` and `tensorBytes` to `functions/math.ts`, `formatBlockMeta` out of the Dagre module into its own.
- **Panel pays for the unused half** — `OpGraphInfoPanel` is a dispatcher; the operation side owns its hooks.
- **Inert `setCenter`/`setViewport`** — real spies, and `getNode` only knows delivered ids, so the `nodeId ?? fallbackNodeId` fallback is falsifiable. An unroll is asserted to hold the viewport with `setCenter` not called.

## Not done, deliberately

- **`MAX_REPEAT_WINDOW`** — the discriminator landed, but raising the cap changes *what folds*, and the half-model-block failure needs its own answer. Filed #1956 with the measurements, and the constant now cites it instead of citing the criterion it fails.
- **Keyboard fold affordance** — a genuine a11y gap, but adding a fold control to unrolled members is new UI. #1942 covers the cramped chrome, not this, so it needs its own issue or an explicit decision. Flagging rather than quietly leaving it.

## One partial

The `graph.nodes.some(...)` selection-survival scan in `onBuilt` is not on the ownership map. It runs against the freshly built graph before `setDetectedBlocks` commits, so a memo derived from state would be a version behind — and it runs once per build, not per render, so a local map would only move the same work. Left as is; happy to be argued out of it.

## Separately

Three UX fixes landed ahead of this from a report on the built branch: the reveal ring now rests faint until the next toggle instead of timing out; a single-op-type block is named by its op; and a shared stem prefix is dropped, which is the bounded-width half of your label thread — one SentenceBERT block went from 124 characters to 52. Details on that thread.

tests/OperationGraphRebuildDiscipline.spec.tsx
aidemsined: > [!NOTE]
> Fixed in `bddb72cb`. Builds now run against the array the view hands the worker, captured in the harness, and a test asserts it carries all three fields.

Checked it the way you framed it: with `inputShapes` deleted from the real mapping, that test alone fails and every folding assertion still passes — which is the gap you described.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: > [!NOTE]
> Fixed in `3fc7a5b2`. The two setters are siblings, branching on `isUnrolling` outside the updater as you suggested, and the member-dropping loop is one helper shared with `collapseAllBlocks` and the deallocate handler.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: > [!NOTE]
> Fixed in `3fc7a5b2` — moved down to immediately precede `const deviceSubgraphs`.

src/components/operation-graph/OpGraphInfoPanel.tsx
aidemsined: > [!NOTE]
> Fixed in `bddb72cb`. `getBlockBoundaryTensors` / `withExternalEndpoints` moved to `opGraphBlockBoundary.ts` with 10 tests.

Two of them fail if the ids and names are filtered independently, so the off-by-one you described is now caught rather than merely covered. Both directions and the short-names case are exercised.

src/components/operation-graph/opGraphRepeatBlocks.ts
aidemsined: > [!WARNING]
> Split: the discriminator landed in `3b0d268d`, the cap did not, and #1956 carries the rest.

Your optimisation reproduced almost exactly — detection on a 10k-op no-repeat graph goes **40.4ms → 4.0ms** median of five, matching your ~10x. Control flow is unchanged; every rejected `start` still advances by one.

I did not raise the cap. That changes *what folds*, and the half-model-block failure needs its own answer rather than riding along in a review pass. #1956 has the measurements, and the constant now cites it instead of citing the criterion it fails.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: > [!NOTE]
> Fixed in `3fc7a5b2` — mirrors `collapseAllBlocks` and drops only members of detected blocks.

Worth noting this is the one change in the pass that alters behaviour, and deliberately so: it restores what `dev` did for ops belonging to no block, which this branch had silently changed.

src/components/operation-graph/OpGraphInfoPanel.tsx
aidemsined: > [!NOTE]
> Fixed in `ad7ccf70` — the summary carries the sums and the panel reads them.

The panel keeps its two stats rows rather than adopting `formatBlockMeta`, since that is its presentation; what is shared is the derivation. A test pins the node meta line to the summary the panel renders, and both gates plus the sign branch now have coverage.

src/components/operation-graph/opGraphLayoutWorker.ts
aidemsined: > [!NOTE]
> Fixed in `c44b0192` — keyed on the source version, with the `setGraph(1)` → build → `setGraph(2)` → build test you asked for.

Demonstrated rather than assumed: with the `SET_GRAPH` clear removed **and** the version dropped from the key, the new test fails. Restoring only the version makes it pass with the clear still missing — so the key now carries the guarantee on its own, which was the point.

tests/OperationGraphRebuildDiscipline.spec.tsx
aidemsined: > [!NOTE]
> Fixed in `bddb72cb`. Both are real spies, and `getNode` returns `undefined` for ids no delivered graph holds.

That last part mattered as you said — answering for any id made the `nodeId ?? fallbackNodeId` fallback unfalsifiable. There is now a test that an unroll calls `setViewport` once with `setCenter` not called in the same pass, which is the "graph jumps to op 0" regression.

The filter Next/Prev stepping onto a folded block is still untested; noting it rather than claiming it.

src/components/operation-graph/OpGraphToolbar.tsx
aidemsined: > [!NOTE]
> Fixed in `ed89dec4` — `renderToolbar` takes the block props and the mid-build case is covered for both buttons.

Getting there was your other finding paying off: making the five props required immediately failed `tsc` on this harness for missing all five.

src/components/operation-graph/OpGraphBlockExpander.tsx
aidemsined: > [!NOTE]
> Fixed in `3b0d268d` — one presentational `OpGraphNodeExpander` taking `action`, `count`, `isExpanded` and `onToggle`.

The two callers now supply only the wording and the context-bound toggle, which is all they ever differed in. Left `isExpanded` in place rather than deleting the dead branch, per your note on the `OpGraphBlockNode` thread.

src/components/operation-graph/opGraphBuilder.ts
aidemsined: > [!NOTE]
> Fixed in `c44b0192` — `getKeptOperations` takes the precomputed candidates and the worker caches one pass per source.

A test asserts one `collectCandidateEdges` call across two builds of the same source, and a second call after `SET_GRAPH`.

src/components/operation-graph/opGraphBuilder.ts
aidemsined: > [!NOTE]
> Fixed in `ed89dec4` — `metaLine` with a matching class, and the dead sums dropped.

Took both halves of your last sentence: `durationSeconds`/`memoryDeltaBytes` on node data were read by exactly one assertion, and the summary carries them now, so they are gone rather than kept alongside `meta`. A stylesheet test pins `.op-graph-node-meta` to the same declarations as `.op-graph-node-file` so the rename stayed invisible.

src/components/operation-graph/OpGraphBlockNode.tsx
aidemsined: > [!WARNING]
> Not fixed — a real gap, but it is new UI and needs its own issue rather than a quiet addition here.

You are right on the substance: `isExpanded={false}` is hardcoded, the button disappears with the block, and the only ways back are pointer-only. A keyboard or screen-reader user can unroll one instance and cannot re-fold just that one.

To your closing question — no, #1942 does not cover this. It covers the cramped toolbar chrome. I have said so in the PR description rather than leaving the impression it is tracked. Adding a fold affordance to unrolled members is a UI change that deserves its own review, so I would rather file it than fold it into a cleanup pass. Happy to file it now if you agree with that shape.

src/components/operation-graph/opGraphLayoutWorker.ts
aidemsined: > [!NOTE]
> Fixed in `c44b0192` — dropped to 8, with the third key dimension written into the comment above it.

Took the entry-count option rather than budgeting on total cached nodes: it is the smaller change and the comment now explains what made 16 wrong. If you think the node budget is worth the complexity, say and I will do it properly.

src/components/operation-graph/OpGraphToolbar.tsx
aidemsined: > [!NOTE]
> Fixed in `ed89dec4` — `isDimUnrelatedEdges` / `onDimUnrelatedEdgesChange`, and the container class is `op-graph-dim-unrelated-edges`.

Label, state and selector now read the same, and the two specs that asserted on the old class name moved with it.

src/components/operation-graph/OpGraphToolbar.tsx
aidemsined: > [!NOTE]
> Fixed in `ed89dec4` — all five required.

Your bug-class argument proved itself immediately: making them required failed `tsc` on `OpGraphToolbarSwitches.spec.tsx`, which was passing none of them, which is exactly the enabled-button-with-undefined-onClick shape the optionality was hiding.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: > [!NOTE]
> Fixed in `3fc7a5b2` — `toggleOperationExpansion` calls `armViewportAnchor(nodeId, nodeId)`.

`armViewportAnchor` moved above it so the declaration order works out. An op node is its own fallback, since unlike a block toggle the anchor id never flips to a member.

src/components/operation-graph/opGraphRepeatBlocks.ts
aidemsined: > [!NOTE]
> Fixed in `ad7ccf70` — `sumOptional` and `tensorBytes` to `functions/math.ts`, `formatBlockMeta` out of the geometry module.

All three parts. `formatBlockMeta` went to its own `opGraphBlockMeta.ts` rather than `formatting.ts`, since it is specific to a block's stats line and the panel now shares it. Their tests moved with them.

src/components/operation-graph/OpGraphInfoPanel.tsx
aidemsined: > [!NOTE]
> Fixed in `884e6762` — `OpGraphInfoPanel` is a thin dispatcher and the operation side owns its own hooks.

Your framing was the persuasive part: the file's own comment says every prop must be stable because it re-renders per drag frame, and paying for the discarded half undercut that.

src/components/operation-graph/opGraphCriticalPath.ts
aidemsined: > [!NOTE]
> Fixed in `80fca79b` — `weightByNodeId` and `memberCountByNodeId` are built in the existing node loop, and `membersByNodeId` is gone.

They have to be separate maps from `costByNodeId`/`opCountByNodeId` rather than reusing those, since relaxation overwrites them with chain totals. One `sumDeviceTimeNs` in `opGraphMemberTime.ts` now serves this and the perf overlay.

The third call site — `selectedPerfDeviceTimeNs` — kept its own loop: it tracks a `found` flag to distinguish "no linked rows" from zero, which the shared sum cannot express. Flagging rather than counting it as done.

src/components/operation-graph/opGraphPerfOverlay.ts
aidemsined: > [!NOTE]
> Fixed in `80fca79b` — one pass returning styles and range together, gated on active, dead branch deleted, ramp extracted.

All four parts. `buildRenderedPerfStyling` replaces both functions and `nodes` is required. The spec was leaning on the `nodes === undefined` branch in several places; those now pass an explicit one-node-per-op list, which is what the component actually does.

src/components/operation-graph/OperationGraphReactFlow.tsx
aidemsined: > [!WARNING]
> Fixed in `ed4778ee` for four of the five — the fifth is deliberate, and I think it is the wrong change.

`blockByMemberOperationId` now serves the render-body lookup, the double-click handler and the selected-block memo, and `getAdjacentOperationIds` takes a `positionByOperationId` map instead of `findIndex` — that one needed a position map rather than the block map, since it wants render order.

The exception is the `graph.nodes.some(...)` selection-survival scan inside `onBuilt`. It runs against the freshly built graph *before* `setDetectedBlocks` commits, so a memo derived from state would be one version stale, and it runs once per build rather than per render — so a local map would move the same work rather than remove it. Happy to be argued out of it.

Issue #1941, opened 2026-08-27
Device log reader breaks on current tt-metal profiler CSVs (15 columns vs 13 hardcoded)


aidemsined:
`DeviceLogProfilerQueries` overwrites the device log's column names positionally with a hardcoded list of 13:

```python
self.runner.df.columns = self.DEVICE_LOG_COLUMNS
```

`profile_log_device.csv` from a current tt-metal has **15** columns, so the assignment raises before any query runs:

```
ValueError: Length mismatch: Expected axis has 15 elements, new values have 13 elements
```

Both endpoints that use it are dead as a result:

- `GET /performance/device-log`
- `GET /performance/device-log/zone/<zone>`

### Column drift

The file gained `trace id`, `trace id counter`, `type` and `meta data`, and renamed `stat value` to `data`. Positions 0–5 still line up; everything after that is wrong:

| pos | file | hardcoded list names it |
|---|---|---|
| 6 | `data` | `stat value` |
| 7 | `run host ID` | `run ID` |
| 8 | `trace id` | `run host ID` |
| 9 | `trace id counter` | `zone name` |
| 10 | `zone name` | `zone phase` |
| 11 | `type` | `source line` |
| 12 | `source line` | `source file` |
| 13–14 | `source file`, `meta data` | *dropped* |

Note position 7: `run host ID` would be renamed to `run ID`. That is the join key back to `ops_perf_results`, so even if the lengths happened to match, the zone queries would join on the wrong column and silently return nonsense rather than failing.

### Suggested fix

Read the header by name instead of overwriting it positionally — the CSV has a real header row, and `OpsPerformanceQueries` already does the right thing with `df.columns.str.strip()`. That also makes the reader tolerant of the next column addition.

Reproduced against `profile_log_device.csv` from the `2026_08_26_22_21_37` profiler run (Blackhole, 839,744 rows), with `skiprows=1` matching `LocalCSVQueryRunner(offset=1)`.

aidemsined:
### Unaffected by the upstream subdevice work — still needs fixing now

For anyone reading this alongside #1940: the tt-metal branch `smountenay-tt/subdevice_reports` does not touch this. It changes graph capture and the report DB schema; `profile_log_device.csv` and this reader are a separate path entirely. Nothing upstream is going to resolve this one.

[...]

Issue #1940, opened 2026-08-27
Surface subdevice fields from tt-perf-report in the performance report


aidemsined:
Models are starting to partition a chip's grid into subdevices so two ops can run concurrently on disjoint core strips. A DeepSeek-V3 MoE prefill run (`test_ttnn_moe_aug26_2217`, 32-chip Blackhole mesh) creates two:

| SUB DEVICE ID | Available cores | Ops |
|---|---|---|
| `1` shared expert | 108 | 3× Matmul, BinaryNg, ReduceScatterMinimalAsync |
| `0` dispatch | 12 | DispatchDeviceOperation |
| *(blank)* | 120 | everything else — full grid |

The performance report can't show any of this today.

### The data is there

`ops_perf_results` (v2.1 format, 289 columns) carries `SUB DEVICE ID` and `AVAILABLE WORKER CORE COUNT` per op.

### tt-perf-report drops both

Its output is 24 columns and neither is among them. Worse, `AVAILABLE WORKER CORE COUNT` is read only to build one global `ArchitectureSpec` for the whole file — it takes the first value and warns when they disagree:

```
Warning: Multiple AVAILABLE WORKER CORE COUNT values found: [120.0, 108.0, 12.0]. Using first value: 120.
Architecture: blackhole, Worker cores: 120
```

So every subdevice op is measured against a 120-core budget it never had. The dispatch op uses 6 of its subdevice's 12 cores — half of what it was given — but is scored as 6/120.

### What we want

Primary, per op:

- `SUB DEVICE ID`
- `AVAILABLE WORKER CORE COUNT` (per-op, not the collapsed global)

Secondary, to make a subdevice view actually answer "did these two grids overlap?":

- `DEVICE FW START CYCLE` / `DEVICE FW END CYCLE` — the only way to tell concurrent from sequential
- `METAL TRACE ID` / `METAL TRACE REPLAY SESSION ID` — overlap only happens under trace replay

### Two ways in

1. **Back-fill locally.** `PASSTHROUGH_COLUMNS` in `csv_queries.py` already pulls fields tt-perf-report doesn't return, by indexing the raw CSV at `ID - 2`. That alignment holds on this file even with 32 devices merged (33 output rows against 808 raw rows, op codes match on every spot-check). Adding the fields is a small change and needs nothing from upstream.
[...]

PR #1945, opened 2026-08-27
Link memory and performance reports across signposts

### LINKED ISSUE #1943: Memory↔performance reports fail to link when the model emits signposts
A memory report and performance report from the same run fail to link when the model emits signposts. The two reports agree completely — the match is rejected purely because signpost rows sit in the perf sequence and nothing in the matcher accounts for them.

Sibling of #1810: same symptom, different cause. That one was per-device duplication, fixed by the collapse fallback; this is signpost rows.

### Repro

DeepSeek-V3 MoE prefill run — memory report `test_ttnn_moe_aug26_2217` plus its profiler run `2026_08_26_22_21_37` (32-chip Blackhole mesh).

| | perf rows | matched | result |
|---|---|---|---|
| as shipped | 33 | 0 | UNLINKED |
| signpost rows dropped | 25 | 25 | **LINKED** |

Verified by calling the real `matchDeviceOperationsToPerf` against both reports, not a reimplementation.

### Cause

`matchDeviceOperationsToPerf` in `src/functions/deviceOperationMatching.ts` requires `perfRows[i].raw_op_code === deviceOperations[i].name` at every index, all-or-nothing. Trailing perf rows are tolerated; leading and interleaved ones are not.

The linked report is fetched with `LINKED_PERFORMANCE_REPORT_FILTERS` (`hideHostOps: true`, `mergeDevices: true`, no signpost window), which yields **33 rows: 25 `tt_dnn_device` + 8 `signpost`**. The memory report has **25** device operations. The signposts are interleaved, and index 0 is one of them:

| idx | memory device op | perf `raw_op_code` |
|---|---|---|
| 0 | `TilizeDeviceOperation` | `TtMoe PCC test - mesh MeshShape([8; 4]); …` ← mismatch |

So it fails on the first comparison. The multi-device fallback can't rescue it either: `collapseMultideviceOperations` keeps only keys seen exactly `numDevices` times, and this memory report records each device op once rather than once per device, so the collapsed list is empty and alignment returns immediately.

`hideHostOps: true` does not help — signposts carry `OP TYPE = signpost`, not a host-op type. This capture has no host ops at all (808 raw rows = 25 device ops × 32 devices + 8 signposts), so that filter is a no-op here.

### Why it matters

Signposts only appear when a model calls `signpost()`. That is exactly what instrumented model bring-up does, so the runs most likely to need the memory↔perf link are the ones that currently can't form it.

### Suggested fix

Filter `OP TYPE == 'signpost'` out of the perf sequence before aligning, the same way host ops already are. The alignment itself needs no change — with the signposts gone these two reports match at every one of the 25 positions.

Worth confirming the ids consumers join on still behave: `useOpToPerfIdFiltered` notes that a row id survives host-op and signpost filtering, so dropping the rows for alignment purposes should not disturb it.

aidemsined:
Closes #1943. Targets `dev`.

## Summary

A memory report and a performance report from the same run would not link if the model called `signpost()`.

Device operations are matched to perf rows by strict positional alignment, all-or-nothing. A signpost row is a marker, not an operation, so the memory report has no device operation to pair it with — and because signposts are their own op type rather than a host op, the pinned `hideHostOps` filter never removed them. One signpost anywhere but the tail shifted every later row out of position and rejected the whole report.

Signpost rows are now dropped before alignment. The alignment logic itself is unchanged.

This only ever affected models that emit signposts — which is precisely the instrumented bring-up work most likely to need the link. A model whose signposts all sit at the end was already linking, since trailing extra rows were tolerated.

The filtering happens inside the matcher rather than at fetch time on purpose: the performance tab still wants its signpost rows, and the matcher is the only consumer that aligns positionally. Every other consumer reads the matcher's output and inherits the fix.

## Test plan

Verified against a real pair that reproduced the bug — memory report `test_ttnn_moe_aug26_2217` with profiler run `2026_08_26_22_21_37`, a DeepSeek-V3 MoE prefill on a 32-chip Blackhole mesh, whose perf report is 25 device ops plus 8 signposts:

| | perf rows | matched | result |
|---|---|---|---|
| before | 33 | 0 | UNLINKED |
| after | 33 | 25 | **LINKED** |

Perf row ids are the unfiltered report's own (`23, 51, 86, … 782`), so consumers still join against the full perf table.

New cases in `deviceOperationMatching.spec.ts` cover signposts leading, interleaved and trailing; signposts on the collapsed multi-device path; and a negative case confirming a genuinely mismatched report is still rejected once its signposts are dropped.

`pnpm test` green (2212 tests, 199 files), `tsc --noEmit` and eslint clean.

## Follow-ups

Not addressed here, both filed separately: #1940 (surface the subdevice fields tt-perf-report drops) and #1941 (the device log reader breaks on current profiler CSVs).

### DIFF
--- a/src/functions/deviceOperationMatching.ts
+++ b/src/functions/deviceOperationMatching.ts
@@ -4,6 +4,7 @@
 
 import { PerfTableRow } from '../model/PerfTable';
 import { DeviceOperationMapping } from '../model/DeviceOperationMapping';
+import { OpType } from '../definitions/Performance';
 
 /**
  * TODO: remove once memory and performance reports carry a shared run id (#1800)
@@ -72,6 +73,17 @@ const alignToPerfRows = (
 };
 
 /**
+ * @description Drop the signpost rows a model emits with `signpost()`. They are
+ * markers rather than operations, so the memory report has no device operation
+ * to pair them with, and they are their own op type rather than a host op — so
+ * the pinned `hideHostOps` filter leaves them in place. Alignment is positional,
+ * so one signpost anywhere but the tail shifts every later row out of position
+ * and fails the whole report. See #1943.
+ */
+const alignableRowsOf = (perfRows: PerfTableRow[]): PerfTableRow[] =>
+    perfRows.filter((perfRow) => perfRow.op_type !== OpType.SIGNPOST);
+
+/**
  * @description Match a memory report's device operations against a performance
  * report's rows, returning the mappings when the two sequences describe the same
  * run and [] when they don't (the report-link UNLINKED signal).
@@ -85,7 +97,8 @@ export const matchDeviceOperationsToPerf = (
     perfRows: PerfTableRow[],
     numDevices: number,
 ): DeviceOperationMapping[] => {
-    const directMatch = alignToPerfRows(deviceOperations, perfRows);
+    const alignableRows = alignableRowsOf(perfRows);
+    const directMatch = alignToPerfRows(deviceOperations, alignableRows);
 
     if (directMatch.length > 0) {
         return directMatch;
@@ -97,5 +110,5 @@ export const matchDeviceOperationsToPerf = (
         return [];
     }
 
-    return alignToPerfRows(collapseMultideviceOperations(deviceOperations, numDevices), perfRows);
+    return alignToPerfRows(collapseMultideviceOperations(deviceOperations, numDevices), alignableRows);
 };
--- a/tests/deviceOperationMatching.spec.ts
+++ b/tests/deviceOperationMatching.spec.ts
@@ -6,6 +6,7 @@ import { describe, expect, it } from 'vitest';
 import { collapseMultideviceOperations, matchDeviceOperationsToPerf } from '../src/functions/deviceOperationMatching';
 import { PerfTableRow } from '../src/model/PerfTable';
 import { DeviceOperationMapping } from '../src/model/DeviceOperationMapping';
+import { OpType } from '../src/definitions/Performance';
 
 const mapping = (name: string, id: number): DeviceOperationMapping => ({
     name,
@@ -17,6 +18,10 @@ const perfRow = (rawOpCode: string, id: number): PerfTableRow =>
     ({ id: String(id), raw_op_code: rawOpCode, device_time: '1' }) as unknown as PerfTableRow;
 
 const perfRowsFor = (names: string[]): PerfTableRow[] => names.map((name, index) => perfRow(name, index));
+
+/** A row for a `signpost()` marker, which no device operation corresponds to. */
+const signpostRow = (label: string, id: number): PerfTableRow =>
+    ({ id: String(id), raw_op_code: label, op_type: OpType.SIGNPOST }) as unknown as PerfTableRow;
 
 /** Each op recorded once per device, as `numDevices` consecutive entries. */
 const duplicatedPerDevice = (names: string[], numDevices: number): DeviceOperationMapping[] =>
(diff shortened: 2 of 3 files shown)


tests/deviceOperationMatching.spec.ts
```diff
@@ -18,6 +19,10 @@ const perfRow = (rawOpCode: string, id: number): PerfTableRow =>
 
 const perfRowsFor = (names: string[]): PerfTableRow[] => names.map((name, index) => perfRow(name, index));
 
+/** A row for a `signpost()` marker, which no device operation corresponds to. */
+const signpostRow = (label: string, id: number): PerfTableRow =>
```
dcblundell: **Should fix:** fixture doesn't model real device rows, so the new predicate is only exercised against `undefined`

`perfRow` (lines 17-18, just above) omits `op_type`, and `as unknown as PerfTableRow` hides that it is required and non-nullable on the model. So every non-signpost row in this spec is `op_type: undefined` — a shape the backend never emits — and the filter is only ever tested as `undefined !== 'signpost'`. Nothing pins that rows carrying real op types are *retained*. Worth caring about because `set_op_type_from_results` writes `op_type = None` when a row id falls outside the ops-perf CSV, so `null` is a production shape the filter must pass through too.

Suggested direction: default `perfRow` to `OpType.DEVICE_OP` and derive this helper from it (`{ ...perfRow(label, id), op_type: OpType.SIGNPOST }`) — one cast, one shape, and it collapses the two near-duplicate builders at the same time.

src/functions/deviceOperationMatching.ts
```diff
@@ -71,6 +72,17 @@ const alignToPerfRows = (
...
+/**
+ * @description Drop the signpost rows a model emits with `signpost()`. They are
+ * markers rather than operations, so the memory report has no device operation
+ * to pair them with, and they are their own op type rather than a host op — so
+ * the pinned `hideHostOps` filter leaves them in place. Alignment is positional,
+ * so one signpost anywhere but the tail shifts every later row out of position
+ * and fails the whole report. See #1943.
+ */
+const alignableRowsOf = (perfRows: PerfTableRow[]): PerfTableRow[] =>
+    perfRows.filter((perfRow) => perfRow.op_type !== OpType.SIGNPOST);
```
dcblundell, line 84: **Should fix:** nothing pins the backend seam this filter depends on (true location: `backend/ttnn_visualizer/tests/test_perf_report.py:101`)

The whole fix rests on signpost rows arriving as `op_type === 'signpost'`. `test_signpost_rows_have_no_kernel_durations` filters `report` for that value and then loops the result, so it passes vacuously if the report contains none — and no test anywhere asserts a signpost row *reaches* `report` with that spelling. If tt-perf-report or the raw `OP TYPE` column ever changes it, this PR silently reverts to the #1943 bug with a fully green suite.

Suggested direction: a synthetic-CSV case in that file via the existing `_fake_generate_perf_report` stub, asserting a raw `OP TYPE=signpost` row reaches `report` with `op_type == "signpost"`.

dcblundell (commented):
Combined review (DRY / TEST / PERF / STANDARDS / SECURITY / REQUIREMENTS). No must-fix findings; two should-fix comments below, both about test fidelity rather than the fix itself. All six criteria in #1943 are implemented, and the ids-are-the-unfiltered-report's-own claim checks out against all three consumers of the linked report.

dcblundell:
### Spec vs code: where should signposts be filtered?

Raising this as a decision rather than a finding, since it could equally be resolved by amending the issue.

#1943's suggested fix says to filter signposts *"the same way host ops already are"* — i.e. at fetch time, via `LINKED_PERFORMANCE_REPORT_FILTERS`. This PR filters inside the matcher instead. Two ways to close that gap:

**Amend the issue (recommended).** `LINKED_PERFORMANCE_REPORT_FILTERS` exposes signpost *windowing* (`startSignpost` / `endSignpost`), `hideHostOps`, `mergeDevices`, `tracingMode` and `groupBy` — there is no signpost-suppression parameter, so the literal suggestion would need a new backend flag. And the linked report is shared with consumers that must keep their signpost rows, so suppressing at fetch time would trade this bug for a different one. The matcher is the only positional consumer, which makes it the correct place. Trade-off: the two mechanisms end up asymmetric (host ops server-side, signposts client-side), which reads as accidental unless it is written down somewhere.

**Change the code.** Add a fetch-time suppression flag and filter there for symmetry. Costs a backend change and needs the linked report separated from the consumers that still want signposts — a much larger change for a bug fix, and no current consumer benefits.

If the recommendation stands, worth a closing note on #1943 recording *why* the mechanisms differ, so whoever next touches `matchDeviceOperationsToPerf` doesn't read the asymmetry as a deviation and "fix" it.

For what it's worth, the safety condition for filtering in the matcher does hold: of the three consumers of the linked report, `useReportLinkMatch` reads fetch state only, and `GraphView` / `useTopNAnnotations` use it for a loaded/non-empty boolean while taking every id-space value from `useGetDeviceOperationListPerf()`. None indexes it positionally, and row object identity is preserved, so `perfData.id` stays the unfiltered report's own — as the PR body claims.

tests/deviceOperationMatching.spec.ts
aidemsined: Fixed in 5b7c707a, taking the suggested direction.

`perfRow` now defaults to `OpType.DEVICE_OP` and `signpostRow` derives from it, so there is one cast and one shape, and every existing case in the file now pins that a real device row is *retained* rather than only testing `undefined !== 'signpost'`.

Added the null case too — confirmed `set_op_type_from_results` has an explicit `else: processed_row["op_type"] = None`, so that is a production shape and not just a theoretical one.

src/functions/deviceOperationMatching.ts
aidemsined: Fixed in 5b7c707a — and the vacuity is worse than "if the report contains none". I checked the fixture: `n300-llama.zip`'s ops perf CSV is 68 rows, all `tt_dnn_device`, zero signposts. So `test_signpost_rows_have_no_kernel_durations` iterates an empty list every run today.

Added `TestPerfReportSignpostOpType` driven from a synthetic CSV via the existing stub, asserting a raw `OP TYPE=signpost` row reaches `report` as `op_type == "signpost"`, plus that the device row beside it keeps its own value — a filter that swept up real operations would be just as broken.

Verified it actually guards the seam: renaming the source column in `set_op_type_from_results` fails the new test, while the old one still passes.

aidemsined:
On the filtering-location question — agreed, and taking the recommendation to amend the issue rather than the code, for the reasons you set out. I have added a closing note to #1943 recording why the two mechanisms differ.

Thanks for independently checking the id-space claim across all three consumers; that was the part I was least able to prove from the matcher alone.

dcblundell (approved):

Issue #1241, opened 2026-02-18
Communicate ranks metadata to frontend


smountenay-tt:
There should be a way for the frontend to quickly determine whether the report is from a single host or multiple hosts.

It's not clear the best way to communicate that to the frontend right now. It could be inferred by calling the `/api/mesh-descriptor` endpoint and seeing multiple mesh descriptors returned, once that change is made. But we may want to explicitly communicate whether it is a single or multihost report, and the total number of ranks somewhere.

For example, it could be tracked on the instance and communicated to the frontend explicitly that it is multi-host report / has multiple ranks in the response from `/api/instance`. Maybe with a `rank_count` field or similar.

PR #1947, opened 2026-08-27
Treat a lone unsuffixed cluster descriptor as one host

### LINKED ISSUE #1939: Topology clones a single UBB cluster descriptor into 32 Galaxy boards
## Problem

Opening `test_ttnn_moe_aug26_2217` (MoE report, 26 Aug) in Topology renders **32 Galaxy-style boards** instead of the one 32-chip UBB that the descriptor actually describes.

The report is a single-host Blackhole UBB:

- `arch`: 32 × `blackhole` (chips 0–31)
- `boards`: one entry, `board_type: ubb_blackhole`, all 32 chips on that board
- `chips: {}` (no mesh coords in the cluster descriptor)
- `chip_unique_ids`: populated for all 32
- `ethernet_connections_to_remote_devices`: populated (scale-out / fabric links to chips **not** in this report)
- File name: unsuffixed `cluster_descriptor.yaml` — not `cluster_descriptor_n_of_m.yaml`

Expected: one condensed (or mesh) board of 32 chips, `worldSize === 1`.
Observed: `multi-host · 32 ranks` — the same board drawn 32 times, one per probed rank.

## Mechanism

Two independently reasonable defaults compose into a 32× clone:

1. **Frontend treats the descriptor as a per-rank slice.** `looksLikeRankedDescriptor` (`src/functions/clusterTopology.ts:159`) is true whenever *both* `ethernet_connections_to_remote_devices` and `chip_unique_ids` are non-empty. The comment and the spec (`tests/clusterTopology.spec.ts:378-379`) assumed a galaxy/UBB would have an *empty* remote-eth list, so it “can't trigger a 32-rank probe.” This UBB has remote-eth entries (links to unique ids that are not in this capture), so the probe fires.

2. **Backend serves the same unsuffixed file for every rank.** `_pick_single_or_ranked_report_path` (`backend/ttnn_visualizer/utils.py:126-127`) returns `cluster_descriptor.yaml` as soon as it exists, with **no** `rank_out_of_range` / `missing_rank_file`. The probe in `fetchClusterTopology` (`src/hooks/useAPI.tsx:607-623`) therefore never 400/404s and walks `MAX_PROBE_RANKS` (32).

`stitchClusterTopology` then builds 32 hosts, each a copy of the same 32-chip descriptor. Inter-host `remote_chip_id`s don't resolve against that union (`unresolvedRemoteCount` climbs); intra-host links duplicate 32 times.

This is the failure mode #1510's single-host short-circuit was meant to prevent, and the reason #1241 wanted a real `world_size` / rank-count from the backend instead of probing.

## Proposed direction

Either (or both):

- **Stop probing when the unsuffixed file is the only descriptor.** Ranked world size is encoded in `cluster_descriptor_<n>_of_<world>.yaml`. An unsuffixed file is one host, even if it lists remote unique ids.
- **Tighten `looksLikeRankedDescriptor`:** remote unique ids that don't appear in *this* descriptor's `chip_unique_ids` are fabric/scale-out, not another rank of the same report. Ranked slices resolve their remotes against other ranks' unique ids.

#1241 remains the durable fix (don't infer world size from YAML shape).

## Acceptance criteria

- Open `test_ttnn_moe_aug26_2217`. Topology shows **one** 32-chip UBB, not 32 boards. No `multi-host · 32 ranks` badge.
- True multi-host reports that ship `cluster_descriptor_<n>_of_<world>.yaml` still stitch to `world` hosts.
- A galaxy / legacy descriptor with an *empty* remote-eth list still short-circuits (existing spec).
- Add a fixture shaped like this report: unsuffixed yaml, 32 BH chips, one `ubb_blackhole` board, remote-eth to unique ids *not* in `chip_unique_ids`. Assert `looksLikeRankedDescriptor` is false *or* that a single-file backend path never returns rank 1…31 as success.

## Out of scope

- Rendering the unresolved scale-out remotes as other machines (those chips are not in the report).
- Replacing the probe with #1241's rank-metadata endpoint — still wanted, not required to stop this clone.

aidemsined:
Closes #1939. Targets `dev`.

## Summary

Opening `test_ttnn_moe_aug26_2217` in Topology rendered 32 boards instead of the one 32-chip UBB the descriptor describes.

The topology probe walks ranks until the backend refuses one, but `_pick_single_or_ranked_report_path` returned the unsuffixed `cluster_descriptor.yaml` for **every** rank — including ranks past the world, verified up to 99. The probe's break condition was unreachable, so it collected `MAX_PROBE_RANKS` copies of one descriptor and stitched 32 hosts out of them.

World size is encoded in the ranked filenames, so a report shipping only the unsuffixed descriptor now answers rank 0 and reports `rank_out_of_range` beyond it. The probe stops after rank 0 and gets one host.

Reports that ship `cluster_descriptor_<n>_of_<world>.yaml` are untouched, and the existing precedence of an unsuffixed file over ranked siblings is preserved for every rank — no report on disk mixes the two, and the two tests pinning that precedence still pass unchanged.

### Why the backend rather than `looksLikeRankedDescriptor`

The issue offered tightening the frontend heuristic as an alternative: treat remote unique ids absent from this descriptor's own `chip_unique_ids` as fabric rather than another rank. That does not discriminate. A genuine rank-0 slice also points its remotes at chips it does not own — in the two-host fixture at the top of `clusterTopology.spec.ts`, rank 0's `remote_chip_id: 200` is likewise absent from its own `chip_unique_ids`. Adopting it would have classified real multi-host rank slices as single-host and broken stitching, failing the issue's own second acceptance criterion.

No property of a lone descriptor separates a single UBB from a rank slice, so the filename is the only authority. `looksLikeRankedDescriptor` is left as-is and a spec case now records that it returns `true` for this shape deliberately, with the backend gating the probe.

## Test plan

Verified against both real reports through the running backend, driving the same probe loop the frontend runs and stitching the result with the shipped `stitchClusterTopology`:

| report | rank slices collected | hosts | worldSize | isMultiHost | interHostLinks |
|---|---|---|---|---|---|
| `test_ttnn_moe_aug26_2217` (one unsuffixed) | 1 | **1** | 1 | **false** | 0 |
| `multihost_poc_jun24_2321` (`1_of_2`, `2_of_2`) | 2 | 2 | 2 | true | 6 |

Before the change the MoE report produced 32 hosts with `unresolvedRemoteCount: 1248` (39 remote links × 32 clones). It is now 1 host with 39 — the genuine off-capture fabric links, which #1939 puts out of scope.

Endpoint behaviour, single-descriptor report: rank 0 → 200, ranks 1/2/31 → 400. Ranked report: ranks 0 and 1 → 200 with *different* descriptors, rank 2 → 400.

New coverage:
- `test_pick_cluster_descriptor_single_file_answers_rank_zero_only` — rank 0 resolves, ranks 1/2/31/99 all return `rank_out_of_range`.
- `test_pick_mesh_descriptor_single_file_answers_rank_zero_only` — same for the mesh descriptor, which shares the picker.
- A `clusterTopology.spec.ts` fixture shaped like this report (32 Blackhole chips, one `ubb_blackhole` board, no mesh coords, off-capture remote-eth), recording why the heuristic cannot decide it.

Also corrected a spec comment that claimed a galaxy "can't trigger a 32-rank probe" — this UBB is the counterexample.

Backend 1135 passing, frontend 2210 passing, `tsc`, eslint, isort and black all clean.

## Follow-ups

Not addressed here:

- #1241 remains the durable fix — take world size from report metadata instead of probing YAML shape. This stops the clone without it.
- Real `remote_chip_id` values are 64-bit and exceed `Number.MAX_SAFE_INTEGER`, so they are already rounded by `JSON.parse` before any matching happens. Noted in the new fixture; worth its own issue if remote-id matching is ever tightened.

### DIFF
--- a/backend/ttnn_visualizer/tests/test_utils.py
+++ b/backend/ttnn_visualizer/tests/test_utils.py
@@ -2,6 +2,7 @@
 #
 # SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
+import os
 from pathlib import Path
 from unittest.mock import mock_open, patch
 
@@ -693,6 +694,128 @@ def test_pick_mesh_descriptor_rank_out_of_range(tmp_path):
     path, err = pick_mesh_descriptor_path(tmp_path, logical_rank=2)
     assert path is None
     assert err == "rank_out_of_range"
+
+
+def _write(path: Path, text: str, mtime_ns: int | None = None) -> Path:
+    path.write_text(text, encoding="utf-8")
+    if mtime_ns is not None:
+        os.utime(path, ns=(mtime_ns, mtime_ns))
+    return path
+
+
+# The producer reuses the output directory and writes each descriptor only
+# `if not path.exists()`, so the import that runs first owns its filenames and a
+# later import of a different world size adds the other family. Which family is
+# stale therefore depends on import order, and both orders are reachable. #1947
+OLD_NS = 1_600_000_000_000_000_000
+NEW_NS = 1_700_000_000_000_000_000
+
+
+def _cluster_families(tmp_path, single_ns, ranked_ns, world=2):
+    _write(tmp_path / "cluster_descriptor.yaml", "chips: []\n", single_ns)
+    for index in range(1, world + 1):
+        _write(
+            tmp_path / f"cluster_descriptor_{index}_of_{world}.yaml",
+            f"chips: [{index}]\n",
+            ranked_ns,
+        )
+
+
+def _mesh_families(tmp_path, single_ns, ranked_ns, world=2):
+    _write(
+        tmp_path / "physical_chip_mesh_coordinate_mapping.yaml",
+        "chips: []\n",
+        single_ns,
+    )
+    for index in range(1, world + 1):
+        _write(
+            tmp_path / f"physical_chip_mesh_coordinate_mapping_{index}_of_{world}.yaml",
+            f"chips: [{index}]\n",
+            ranked_ns,
+        )
+
+
+def test_ranked_family_wins_when_it_is_the_newer_import(tmp_path):
+    # world-1 imported first, then world-2: the unsuffixed file is the leftover.
+    _cluster_families(tmp_path, single_ns=OLD_NS, ranked_ns=NEW_NS)
+
+    for rank, expected in (
+        (0, "cluster_descriptor_1_of_2.yaml"),
+        (1, "cluster_descriptor_2_of_2.yaml"),
+    ):
+        path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=rank)
+        assert err is None
+        assert path is not None and path.name == expected
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=2)
+    assert path is None and err == "rank_out_of_range"
+
+
+def test_unsuffixed_wins_when_it_is_the_newer_import(tmp_path):
+    # The reverse order, which the producer permits just as readily: world-2
+    # imported first, then world-1, so the ranked family is the leftover.
+    _cluster_families(tmp_path, single_ns=NEW_NS, ranked_ns=OLD_NS)
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+    assert err is None
+    assert path is not None and path.name == "cluster_descriptor.yaml"
+
+    # One host, so nothing past rank 0 — not the stale ranked family's world.
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=1)
+    assert path is None and err == "rank_out_of_range"
+
+
+def test_a_partial_ranked_family_does_not_mask_an_unsuffixed_descriptor(tmp_path):
+    # `ranked_report_basenames` calls a lone `_1_of_2` a consistent group, which
+    # made rank 1 fail with `missing_rank_file` while a usable descriptor sat
+    # beside it. An incomplete family is not eligible to win, whatever its mtime.
+    _write(tmp_path / "cluster_descriptor.yaml", "chips: []\n", OLD_NS)
+    _write(tmp_path / "cluster_descriptor_1_of_2.yaml", "chips: [1]\n", NEW_NS)
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+    assert err is None
+    assert path is not None and path.name == "cluster_descriptor.yaml"
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=1)
+    assert path is None and err == "rank_out_of_range"
+
+
+def test_a_partial_ranked_family_is_still_used_when_it_is_all_there_is(tmp_path):
+    # Completeness governs precedence, not usability: with no unsuffixed file to
+    # defer to, rank 0 is still serviceable and only the absent rank errors.
+    _write(tmp_path / "cluster_descriptor_1_of_2.yaml", "chips: [1]\n")
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+    assert err is None
+    assert path is not None and path.name == "cluster_descriptor_1_of_2.yaml"
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=1)
+    assert path is None and err == "missing_rank_file"
+
+
+@pytest.mark.parametrize(
+    "single_ns,ranked_ns,expect_ranked",
+    [(OLD_NS, NEW_NS, True), (NEW_NS, OLD_NS, False)],
+)
+def test_cluster_and_mesh_choose_the_same_generation(
+    tmp_path, single_ns, ranked_ns, expect_ranked
+):
+    # The failure this guards: cluster selecting the current ranked descriptors
+    # while mesh returned the stale unsuffixed mapping, which then overrode
+    # `descriptor.chips` for every host. Family selection is shared, so the two
+    # cannot disagree; only the per-rank rule differs. #1947
+    _cluster_families(tmp_path, single_ns, ranked_ns)
+    _mesh_families(tmp_path, single_ns, ranked_ns)
+
+    cluster, cluster_err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+    mesh, mesh_err = pick_mesh_descriptor_path(tmp_path, logical_rank=0)
+
+    assert cluster_err is None and mesh_err is None
+    assert cluster is not None and mesh is not None
+    assert ("_1_of_2" in cluster.name) is expect_ranked
+    assert (
+        "_1_of_2" in mesh.name
+    ) is expect_ranked, "mesh chose a different generation"
 
 
 def test_pick_cluster_descriptor_ranked_set_outranks_a_stale_unsuffixed_file(tmp_path):
--- a/backend/ttnn_visualizer/tests/test_utils.py
+++ b/backend/ttnn_visualizer/tests/test_utils.py
@@ -695,15 +695,33 @@ def test_pick_mesh_descriptor_rank_out_of_range(tmp_path):
     assert err == "rank_out_of_range"
 
 
-def test_pick_cluster_descriptor_prefers_unsuffixed_file(tmp_path):
+def test_pick_cluster_descriptor_ranked_set_outranks_a_stale_unsuffixed_file(tmp_path):
+    # tt-metal reuses the output directory and writes each descriptor only
+    # `if not path.exists()`, so a world-1 import followed by a world-2 import into
+    # the same directory leaves the unsuffixed file beside the ranked pair. Serving
+    # that leftover for every rank reproduced #1939 bounded by the world size, so
+    # the ranked set wins and its bound applies. #1947
     (tmp_path / "cluster_descriptor.yaml").write_text("chips: []\n", encoding="utf-8")
     (tmp_path / "cluster_descriptor_1_of_2.yaml").write_text(
         "chips: [1]\n", encoding="utf-8"
     )
+    (tmp_path / "cluster_descriptor_2_of_2.yaml").write_text(
+        "chips: [2]\n", encoding="utf-8"
+    )
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+    assert err is None
+    assert path is not None and path.name == "cluster_descriptor_1_of_2.yaml"
+
     path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=1)
     assert err is None
-    assert path is not None
-    assert path.name == "cluster_descriptor.yaml"
+    assert path is not None and path.name == "cluster_descriptor_2_of_2.yaml"
+
+    # The probe stops at the world bound instead of walking to its cap.
+    for rank in (2, 3, 31):
+        path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=rank)
+        assert path is None, f"rank {rank} resolved to {path}"
+        assert err == "rank_out_of_range"
 
 
 def test_pick_cluster_descriptor_single_file_answers_rank_zero_only(tmp_path):
@@ -723,18 +741,20 @@ def test_pick_cluster_descriptor_single_file_answers_rank_zero_only(tmp_path):
         assert err == "rank_out_of_range", f"rank {rank} returned {err!r}"
 
 
-def test_pick_mesh_descriptor_single_file_answers_rank_zero_only(tmp_path):
+def test_pick_mesh_descriptor_single_file_serves_every_rank(tmp_path):
+    # Deliberately unlike the cluster rule: one unsuffixed mesh mapping covers the
+    # whole world, either reused as a legacy single doc or holding one `chips:`
+    # document per rank that the frontend selects from. Applying the cluster
+    # rank-0-only rule here left later hosts with no mesh data. #1947
     (tmp_path / "physical_chip_mesh_coordinate_mapping.yaml").write_text(
         "chips: []\n", encoding="utf-8"
     )
 
-    path, err = pick_mesh_descriptor_path(tmp_path, logical_rank=0)
-    assert err is None
-    assert path is not None
-
-    path, err = pick_mesh_descriptor_path(tmp_path, logical_rank=1)
-    assert path is None
-    assert err == "rank_out_of_range"
+    for rank in (0, 1, 5, 31):
+        path, err = pick_mesh_descriptor_path(tmp_path, logical_rank=rank)
+        assert err is None, f"rank {rank} returned {err!r}"
+        assert path is not None
+        assert path.name == "physical_chip_mesh_coordinate_mapping.yaml"
 
 
 def test_pick_cluster_descriptor_ranked_files(tmp_path):
--- a/backend/ttnn_visualizer/tests/test_utils.py
+++ b/backend/ttnn_visualizer/tests/test_utils.py
@@ -706,6 +706,37 @@ def test_pick_cluster_descriptor_prefers_unsuffixed_file(tmp_path):
     assert path.name == "cluster_descriptor.yaml"
 
 
+def test_pick_cluster_descriptor_single_file_answers_rank_zero_only(tmp_path):
+    # A report shipping one unsuffixed descriptor is a single host. Answering
+    # rank 1..N with the same file made the topology probe clone that host once
+    # per probed rank (#1939), so every rank past 0 is out of range.
+    (tmp_path / "cluster_descriptor.yaml").write_text("chips: []\n", encoding="utf-8")
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+    assert err is None
+    assert path is not None
+    assert path.name == "cluster_descriptor.yaml"
+
+    for rank in (1, 2, 31, 99):
+        path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=rank)
+        assert path is None, f"rank {rank} should not resolve to a file"
+        assert err == "rank_out_of_range", f"rank {rank} returned {err!r}"
+
+
+def test_pick_mesh_descriptor_single_file_answers_rank_zero_only(tmp_path):
+    (tmp_path / "physical_chip_mesh_coordinate_mapping.yaml").write_text(
+        "chips: []\n", encoding="utf-8"
+    )
+
+    path, err = pick_mesh_descriptor_path(tmp_path, logical_rank=0)
+    assert err is None
+    assert path is not None
+
+    path, err = pick_mesh_descriptor_path(tmp_path, logical_rank=1)
+    assert path is None
+    assert err == "rank_out_of_range"
+
+
 def test_pick_cluster_descriptor_ranked_files(tmp_path):
     (tmp_path / "cluster_descriptor_2_of_2.yaml").write_text(
         "chips: [2]\n", encoding="utf-8"
--- a/backend/ttnn_visualizer/tests/test_utils.py
+++ b/backend/ttnn_visualizer/tests/test_utils.py
@@ -10,12 +10,14 @@ import pytest
 from ttnn_visualizer.models import RemoteConnection
 from ttnn_visualizer.utils import (
     FALSE_VALUES,
+    MAX_RANKED_WORLD_SIZE,
     TRUE_VALUES,
     find_gunicorn_path,
     get_app_data_directory,
     get_mlir_path,
     get_report_data_directory,
     is_running_in_container,
+    is_valid_profiler_ranked_entry,
     parse_bool,
     pick_cluster_descriptor_path,
     pick_mesh_descriptor_path,
@@ -891,3 +893,46 @@ def test_pick_cluster_descriptor_ranked_files(tmp_path):
     assert err is None
     assert path is not None
     assert path.name == "cluster_descriptor_2_of_2.yaml"
+
+
+def test_ranked_world_size_is_bounded_against_a_hostile_filename(tmp_path):
+    """
+    ``world_size`` is parsed straight out of a filename, and the local upload path
+    preserves client basenames. Completeness used to be decided by comparing
+    against ``set(range(1, world_size + 1))``, so an 11-digit world size asked for
+    a multi-terabyte allocation inside the shared Flask process. #1947
+    """
+    assert is_valid_profiler_ranked_entry(1, MAX_RANKED_WORLD_SIZE) is True
+    assert is_valid_profiler_ranked_entry(1, MAX_RANKED_WORLD_SIZE + 1) is False
+    assert is_valid_profiler_ranked_entry(1, 99_999_999_999) is False
+
+    # End to end: an over-cap name is skipped, so a usable unsuffixed descriptor
+    # still answers rather than the request dying on the allocation.
+    (tmp_path / "cluster_descriptor.yaml").write_text("chips: []\n", encoding="utf-8")
+    (tmp_path / "cluster_descriptor_1_of_99999999999.yaml").write_text(
+        "chips: [1]\n", encoding="utf-8"
+    )
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+
+    assert err is None
+    assert path is not None
+    assert path.name == "cluster_descriptor.yaml"
+
+
+def test_ranked_family_completeness_counts_rather_than_enumerating(tmp_path):
+    # The count settles completeness because the indices are unique and already
+    # bounded to 1..world; an incomplete family must not outrank the unsuffixed file.
+    (tmp_path / "cluster_descriptor.yaml").write_text("chips: []\n", encoding="utf-8")
+    (tmp_path / "cluster_descriptor_1_of_3.yaml").write_text(
+        "chips: [1]\n", encoding="utf-8"
+    )
+    (tmp_path / "cluster_descriptor_2_of_3.yaml").write_text(
+        "chips: [2]\n", encoding="utf-8"
+    )
+
+    path, err = pick_cluster_descriptor_path(tmp_path, logical_rank=0)
+
+    assert err is None
+    assert path is not None
+    assert path.name == "cluster_descriptor.yaml"
--- a/backend/ttnn_visualizer/utils.py
+++ b/backend/ttnn_visualizer/utils.py
@@ -119,11 +119,17 @@ def _pick_single_or_ranked_report_path(
     Prefer ``single_basename`` when present; otherwise pick the ranked file for
     ``logical_rank``. Returns ``(path, None)`` or ``(None, error)`` where error
     is ``rank_out_of_range``, ``missing_rank_file``, or ``None`` when absent.
+
+    An unsuffixed file with no ranked siblings answers rank 0 only: world size is
+    encoded in the ranked filenames, so a report that ships one unsuffixed
+    descriptor is a single host whatever chips it lists. See ``logical_rank``
+    handling below and #1939.
     """
     if not report_dir.is_dir():
         return None, None
     single = report_dir / single_basename
-    if single.is_file():
+    single_exists = single.is_file()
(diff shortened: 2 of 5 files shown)


aidemsined:
Chip *positions* on the surviving board are a separate, pre-existing problem — filed as #1948. In the condensed fallback chips are placed at `(c % 4, c / 4)`, so 41% of this report's intra-host links are drawn between non-adjacent cells and 12 touching pairs have no link at all. Out of scope here; this PR only settles how many boards are rendered.

backend/ttnn_visualizer/utils.py
```diff
@@ -119,18 +119,33 @@ def _pick_single_or_ranked_report_path(
...
+    encoded in the ranked filenames, so a report that ships one unsuffixed
+    descriptor is a single host whatever chips it lists. See ``logical_rank``
+    handling below and #1939.
     """
     if not report_dir.is_dir():
         return None, None
     single = report_dir / single_basename
-    if single.is_file():
+    single_exists = single.is_file()
+    if single_exists and logical_rank == 0:
```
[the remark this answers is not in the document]
smountenay-tt: Confirmed. Current tt-metal clean imports name cluster and mesh files consistently from `world_size`, so this exact combination is not the normal output. It is still a supported compatibility state: the visualizer intentionally handles one mesh file containing all ranks, and tt-metal reuses output directories without clearing or overwriting descriptor files ([directory creation](https://github.com/tenstorrent/tt-metal/blob/31a9a54154d4ec5f372fbd1c8899d6ecdbf2270a/ttnn/ttnn/graph_report.py#L1924-L1929), [descriptor writes](https://github.com/tenstorrent/tt-metal/blob/31a9a54154d4ec5f372fbd1c8899d6ecdbf2270a/ttnn/ttnn/graph_report.py#L2091-L2109)). That makes mixed historical/current families possible. The rank-cardinality policy should be cluster-specific; keep the old unsuffixed mesh reuse behavior.

backend/ttnn_visualizer/utils.py
```diff
@@ -119,18 +119,33 @@ def _pick_single_or_ranked_report_path(
...
 
     ranked_names = ranked_report_basenames(
         (p.name for p in report_dir.iterdir() if p.is_file()),
         ranked_re,
         malformed_label,
     )
+    if single_exists:
+        # Ranked siblings present: the unsuffixed file keeps precedence over them
+        # for every rank, as it always has.
+        if ranked_names:
```
smountenay-tt: This still leaves the original runaway-probe failure reachable whenever an unsuffixed descriptor coexists with ranked siblings: every rank, including `rank >= world_size`, returns the same unsuffixed file. That mixed state is possible in the current producer because [`import_report` reuses the output directory](https://github.com/tenstorrent/tt-metal/blob/31a9a54154d4ec5f372fbd1c8899d6ecdbf2270a/ttnn/ttnn/graph_report.py#L1924-L1929) and [does not overwrite existing descriptors](https://github.com/tenstorrent/tt-metal/blob/31a9a54154d4ec5f372fbd1c8899d6ecdbf2270a/ttnn/ttnn/graph_report.py#L2091-L2109). A world-size-1 import followed by a world-size-N import into the same report directory therefore leaves exactly this layout. I reproduced ranks 0, 1, 2, and 31 all resolving to `cluster_descriptor.yaml` despite complete `_1_of_2` / `_2_of_2` siblings. Please make a consistent ranked set authoritative (or at minimum apply its world bound) and add this mixed-layout regression case; otherwise #1939 remains for stale-but-producer-reachable reports.

smountenay-tt (changes requested):
The core diagnosis and the lone-cluster fix are sound, and the new UBB fixture correctly captures why the frontend cannot infer world size from descriptor contents. I found two narrow backend resolution issues to address before merge:

1. Keep the rank-0-only rule cluster-specific; applying it to a lone unsuffixed mesh mapping regresses the existing global/multi-document compatibility path (see the existing thread).
2. When ranked cluster siblings coexist with an unsuffixed file, the current branch still serves the unsuffixed file for every rank and never reaches the world-size bound. tt-metal can produce this stale/mixed directory state because imports neither clear nor overwrite descriptor files (see my inline reproduction and source links).

Once those two cases are covered, I expect this to be ready. All CI checks are green, and I also ran the affected utility and endpoint suites locally: 90 passed.

backend/ttnn_visualizer/utils.py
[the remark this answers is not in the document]
aidemsined: Agreed and fixed in 5db49d14. This was a regression I introduced by putting the rule in the shared picker without checking the mesh contract.

Verified the contract before changing it: `clusterTopology.spec.ts` asserts `pickMeshDocForRank(single, 5)` returns the single doc, and the multi-doc branch does `sorted[rank]` — so one unsuffixed file legitimately answers every rank, either reused or indexed per rank. My change made rank 1+ return `rank_out_of_range`, which would have left later hosts on `descriptor.chips`.

The rule is now an opt-in `single_is_one_rank` parameter that only `pick_cluster_descriptor_path` sets; mesh keeps its previous behaviour. I also replaced the mesh test I had added with one asserting the opposite — a lone unsuffixed mapping resolves for ranks 0, 1, 5 and 31 — so the two policies are pinned as deliberately different rather than looking like drift.

backend/ttnn_visualizer/utils.py
aidemsined: Agreed and fixed in 5db49d14. Your reachability argument is better evidence than mine was — I had reasoned from "no report on disk mixes them", which only sampled my local reports, and the producer settles it.

Confirmed against the source you linked: `output_dir.mkdir(parents=True, exist_ok=True)` plus `if not cluster_path.exists():  # Only write once`, with the suffix keyed on `report_world_size`. A world-1 import followed by a world-N import into the same directory leaves exactly that layout.

One thing worth recording: I first considered only applying the ranked world bound and keeping the unsuffixed file as the response. That is insufficient — ranks 0..N-1 would all still receive the stale file, which is #1939 again bounded to N instead of the probe cap. So a consistent ranked set is now authoritative and the unsuffixed file is ignored when one exists.

Reproduced your case end to end before and after. With `cluster_descriptor.yaml` (32 chips, world-1 leftover) beside `_1_of_2` / `_2_of_2` (8 chips each):

| rank | before | after |
|---|---|---|
| 0 | 200, 32 chips | 200, **8 chips** |
| 1 | 200, 32 chips | 200, **8 chips** |
| 2 | 200, 32 chips | 400 `rank_out_of_range` |
| 31 | 200, 32 chips | 400 `rank_out_of_range` |

The chip count is the part that matters — it shows the ranked files win rather than just the bound applying. Added as `test_pick_cluster_descriptor_ranked_set_outranks_a_stale_unsuffixed_file`, which replaces the old `prefers_unsuffixed_file` test since that precedence is what made this reachable.

Backend 1135 passing, frontend 2210 passing.

backend/ttnn_visualizer/utils.py
```diff
@@ -114,16 +114,29 @@ def _pick_single_or_ranked_report_path(
...
+    Left ``False`` for mesh mappings, where one unsuffixed file legitimately
+    covers every rank — reused as a legacy single doc, or holding one ``chips:``
+    document per rank that the frontend selects from by rank. #1947
     """
     if not report_dir.is_dir():
         return None, None
     single = report_dir / single_basename
-    if single.is_file():
+    single_exists = single.is_file()
+    if single_exists and not single_is_one_rank:
```
smountenay-tt: This preserves unsuffixed mesh compatibility, but returning before examining ranked siblings makes cluster and mesh choose different generations in the producer-reachable world-1 → world-N case fixed below: cluster selects the new ranked descriptors while both mesh ranks receive the stale unsuffixed world-1 mapping, which then overrides `descriptor.chips` in `fetchRankSlice`. The family-selection rule needs to be consistent across cluster and mesh while still allowing an unsuffixed/global mesh file when it is the only/current family.

backend/ttnn_visualizer/utils.py
```diff
@@ -132,8 +145,20 @@ def _pick_single_or_ranked_report_path(
...
     )
     if not ranked_names:
+        if single_exists:
+            # The only descriptor in the report, and it stands for a single rank.
+            if logical_rank == 0:
+                return single, None
+            return None, "rank_out_of_range"
         return None, None
 
+    # A consistent ranked set is authoritative even when an unsuffixed file sits
```
smountenay-tt: I need to correct/narrow my previous recommendation: unconditional ranked precedence handles the world-1 → world-N import order, but the producer also permits the reverse. If world N is imported first and then world 1 into the same directory, the ranked files are stale and the unsuffixed descriptor is current; this branch still selects the ranked hosts. Also, `ranked_report_basenames` calls any non-empty same-world group “consistent”, so a lone stale `_1_of_2` masks a valid unsuffixed descriptor and makes rank 1 fail with `missing_rank_file`. I reproduced both outcomes. Please make ranked precedence conditional on a complete/current family (for example, require all `1..world` files and use family mtimes to resolve coexistence), with tests for both import orders and a partial ranked family. The current test covers only unsuffixed-stale → ranked-current.

smountenay-tt (changes requested):
The update correctly fixes both cases from the first review: the rank-0 rule is cluster-only, and a complete newer ranked cluster family wins over a stale unsuffixed file. The targeted backend suites remain green (90 passed), as is CI.

One family-selection issue remains before approval. Producer reuse works in either order, but ranked files now win unconditionally: a current unsuffixed world-1 descriptor is masked by stale or even partial ranked files from an earlier run. In the opposite order, cluster chooses the new ranked family while mesh still returns the stale unsuffixed mapping. I posted reproductions for both manifestations.

This needs one coherent complete/current-family rule across cluster and mesh, while retaining unsuffixed global-mesh fallback. I expect that to be the final blocker.

backend/ttnn_visualizer/utils.py
aidemsined: Agreed and fixed in e3bacd1a. You are right that the early return was the bug, not just an omission: `single_is_one_rank` was gating *whether the ranked files were looked at*, which conflated two separate questions.

Family selection is now shared — `_ranked_family` + `_prefer_ranked_family` decide which generation to read, identically for cluster and mesh — and `single_is_one_rank` governs only the per-rank question once a family is chosen. So mesh can still serve one whole-world mapping for every rank, while cluster still answers rank 0 only, and neither can end up on a different generation from the other.

Pinned with `test_cluster_and_mesh_choose_the_same_generation`, parametrised over both import orders and asserting the two pickers land on the same family. It fails on the previous commit in both directions.

backend/ttnn_visualizer/utils.py
aidemsined: Thanks for narrowing it — this is the part I had wrong, and the correction matters more than the original finding. Fixed in e3bacd1a.

Confirmed the reverse order is equally reachable: `if not cluster_path.exists()` means the first import owns its filenames and a later import of a different world size simply adds the other family, so whichever ran first is the survivor. Import world N then world 1 and the ranked files are the leftovers, which my unconditional precedence would have served. Nothing in the filenames records generation, so recency decides.

Also confirmed the completeness gap you describe: `ranked_report_basenames` takes the majority world and returns whatever matched without checking the indices, so a lone `_1_of_2` presents as a family, world_size 2, and rank 1 falls to `missing_rank_file` with a perfectly usable unsuffixed descriptor sitting beside it.

What e3bacd1a does:

- `_ranked_family` reports completeness — indices must be exactly `1..world` — and only a complete family is eligible to outrank an unsuffixed descriptor.
- `_prefer_ranked_family` compares the newest mtime in the family against the unsuffixed file's. Ties go to ranked, which carries per-rank detail.
- Completeness governs *precedence*, not usability. A partial family with no unsuffixed file to defer to is still served, so rank 0 keeps working and only the absent rank errors — I did not want to turn your finding into a regression for reports that have only a partial family.

Tests, all four failing on the previous commit:

| case | expected |
|---|---|
| world-1 then world-N (unsuffixed stale) | ranked wins, bounded at world |
| world-N then world-1 (ranked stale) | unsuffixed wins, rank 1 out of range |
| lone `_1_of_2` beside an unsuffixed file | unsuffixed wins; no `missing_rank_file` |
| lone `_1_of_2` alone | still served; only rank 1 errors |

One caveat worth recording against the mtime approach, since it is a heuristic rather than a fact in the data: mtimes do not survive every copy — an archive extract can flatten a whole directory to one timestamp — and then the tie-break is what applies. A generation marker from the producer would be better than inference; this is the best in-band signal available today. Noted in the docstring.

Both real reports re-verified end to end: MoE rank 0 → 200, ranks 1/31 → 400; multihost_poc ranks 0/1 → 200 on both cluster and mesh, rank 2 → 400. Backend 1141 passing, frontend 2210 passing.

backend/ttnn_visualizer/utils.py
```diff
@@ -106,6 +106,71 @@ def ranked_profiler_config_basenames(file_names: Iterable[str]) -> List[str]:
...
+    parsed = parse_ranked_basename(names[0], ranked_re)
+    if not parsed:
+        return [], 0, False
+    _, world_size = parsed
+    indices = {
+        parse_ranked_basename(name, ranked_re)[0]  # type: ignore[index]
+        for name in names
+        if parse_ranked_basename(name, ranked_re)
+    }
+    return names, world_size, indices == set(range(1, world_size + 1))
```
dcblundell: **Must fix:** unbounded allocation from a client-controlled filename

`world_size` is parsed straight out of a filename and `is_valid_profiler_ranked_entry` only checks `1 <= index <= world` — there is no upper bound on `world` (`views._MAX_RANK` bounds the *query param*, not this). Reproduced locally: cost is linear at ~85 MB / ~22 ms per million (world=5,000,000 → 423 MB), so an 11-digit filename attempts a multi-TB allocation and OOM-kills the shared Flask process; `/api/local/upload/profiler` is not `@local_only` and preserves client basenames, so it is reachable in the hosted posture. Newly introduced — on `dev` the picker returned at `single.is_file()` and never built a set, and the set here is built *before* the family is rejected as incomplete. Suggest `len(indices) == world_size` (indices are already unique and within `1..world`) plus a cap on `world_size` at parse time, dropping over-cap names via the existing `logger.warning` skip path.

backend/ttnn_visualizer/utils.py
```diff
@@ -106,6 +106,71 @@ def ranked_profiler_config_basenames(file_names: Iterable[str]) -> List[str]:
...
+
+    Nothing in the filenames records that, so recency decides, and only a complete
+    family is eligible to win. Ties go to the ranked family, which carries per-rank
+    detail. Note that mtimes do not survive every copy — an archive extract can
+    flatten them all — in which case the tie-break is what applies. #1947
+    """
+    if not ranked_is_complete:
+        return False
+    ranked_mtime = _newest_mtime_ns(report_dir / name for name in ranked_names)
+    return ranked_mtime >= _newest_mtime_ns([single])
```
dcblundell, line 181: **Must fix:** mtime recency can re-open acceptance criterion 2

This picks the family by newest mtime on the premise that mtime records which import ran later, but neither ingestion path preserves producer mtimes: local upload stamps at write time, and sftp/scp sync uses no `-p` and no `os.utime` (the `-p` in `sftp_operations.py` is ssh's *port* flag). So "newest family" means "whichever family's last file transferred last" — and when the unsuffixed file lands last, a genuine multi-host report returns `rank_out_of_range` at rank 1 and stitches to one host, silently discarding `world-1` real hosts. Recommend preferring a *complete* ranked family unconditionally (the rule from 5db49d14): the reverse failure — a stale complete ranked set rendering `world` stale hosts — is visible and bounded, whereas this one is silent and breaks a stated criterion. Flagging that this reverses the most recent review round, so worth settling with that reviewer rather than unilaterally.

backend/ttnn_visualizer/utils.py
```diff
@@ -114,36 +179,60 @@ def _pick_single_or_ranked_report_path(
...
+        )
+    else:
+        use_ranked = bool(ranked_names)
+
+    if not use_ranked:
+        if not single_exists:
+            return None, None
+        # A whole-world mapping answers every rank; a single-host descriptor
+        # answers rank 0 and nothing beyond it.
+        if single_is_one_rank and logical_rank != 0:
```
dcblundell, line 235: **Should fix:** the fix ships as an HTTP status, and nothing tests that mapping

This line is what stops the 32-host clone, but it only reaches the frontend as a 400. `tests/views/test_descriptor_endpoints.py` never passes a `rank` param at all, so nothing asserts `GET /api/cluster-descriptor?rank=1` returns 400 for a lone descriptor — a regression remapping `rank_out_of_range` to 500 or 200 would pass the whole suite while restoring the clone. The frontend half is equally unpinned: the probe loop in `useAPI.tsx:607` has no test that it stops at rank 0, and the new spec case explicitly documents that the pure function cannot decide it, which leaves the untested wiring as the sole guarantee.

backend/ttnn_visualizer/tests/test_utils.py
```diff
@@ -695,16 +696,189 @@ def test_pick_mesh_descriptor_rank_out_of_range(tmp_path):
...
+
+    assert cluster_err is None and mesh_err is None
+    assert cluster is not None and mesh is not None
+    assert ("_1_of_2" in cluster.name) is expect_ranked
+    assert (
+        "_1_of_2" in mesh.name
+    ) is expect_ranked, "mesh chose a different generation"
+
+
+def test_pick_cluster_descriptor_ranked_set_outranks_a_stale_unsuffixed_file(tmp_path):
```
dcblundell, line 823: **Should fix:** the documented tie-break is untested, and this test depends on write order

No test uses equal mtimes, so mutating `>=` to `>` in `_prefer_ranked_family` leaves all 91 tests green — despite the docstring naming archive-extract flattening (arguably the most common delivery path) as the case the tie-break exists for. This is also the one new test that sets no mtimes: it passes only because the unsuffixed file happens to be written first, and reordering the three `write_text` calls would flip it for reasons unrelated to its stated intent. Suggest repointing it at `_cluster_families(single_ns=OLD_NS, ranked_ns=OLD_NS)` and renaming it to the tie-break test it should be — it is otherwise a near-duplicate of `test_ranked_family_wins_when_it_is_the_newer_import`.

tests/clusterTopology.spec.ts
```diff
@@ -376,10 +376,49 @@ describe('looksLikeRankedDescriptor', () => {
...
+                    { remote_chip_id: 9001, chan: 9 },
+                ],
+            ],
+        } as unknown as ClusterModel;
+
+        // Deliberately true. A genuine rank-0 slice also points its remotes at
+        // chips it does not own, so no property of a lone descriptor separates
+        // the two — see the two-host case at the top of this file, where rank 0's
+        // remote id 200 is likewise absent from its own chip_unique_ids.
+        expect(looksLikeRankedDescriptor(ubb)).toBe(true);
```
dcblundell, line 415: **Should fix:** the disproved assumption still stands in the source docstring

This spec comment was corrected, but the `looksLikeRankedDescriptor` docstring itself (`src/functions/clusterTopology.ts:153-157`, not in this diff) still reads "single-host reports omit these and a more permissive check would trigger spurious world-size probes" — the exact assumption #1939 identifies as the mechanism. The next reader gets the wrong model of the check and may well read this deliberate `toBe(true)` as the bug. Suggest rewriting the docstring to say the heuristic cannot separate a UBB with scale-out links from a rank-0 slice, and that world size is bounded by the backend instead.

backend/ttnn_visualizer/utils.py
```diff
@@ -114,36 +179,60 @@ def _pick_single_or_ranked_report_path(
...
 
     if logical_rank < 0 or logical_rank >= world_size:
         return None, "rank_out_of_range"
 
     path = report_dir / ranked_basename_for(logical_rank + 1, world_size)
     if not path.is_file():
+        # Only reachable for an incomplete family that had no unsuffixed file to
+        # defer to, or a file removed since the directory was listed.
         return None, "missing_rank_file"
     return path, None
```
dcblundell, line 247: **Should fix:** profiler configs are left on the old rule in the same directory

`pick_profiler_config_paths` and `read_profiler_config_api_payload` (`utils.py:240` and `:289`, just below this hunk) still hand-roll "unsuffixed always wins", while descriptors now get completeness plus recency. Both read the same directory, so the staleness scenario `_prefer_ranked_family`'s docstring describes applies verbatim to `config.json` vs `config_<n>_of_<world>.json`: a report can serve rank-1 descriptor data from the ranked family while `/api/config?rank=1` answers from the stale single config. Either route configs through this picker, or add a comment recording that the omission is deliberate.

backend/ttnn_visualizer/utils.py
```diff
@@ -106,6 +106,71 @@ def ranked_profiler_config_basenames(file_names: Iterable[str]) -> List[str]:
...
+        malformed_label,
+    )
+    if not names:
+        return [], 0, False
+    parsed = parse_ranked_basename(names[0], ranked_re)
+    if not parsed:
+        return [], 0, False
+    _, world_size = parsed
+    indices = {
+        parse_ranked_basename(name, ranked_re)[0]  # type: ignore[index]
```
dcblundell: **Should fix:** avoidable suppression, and it costs a double parse

`# type: ignore[index]` carries no explanatory comment, against AGENTS.md *Lint discipline* and CONVENTIONS.md *Assess validity before suppressing* — and it is avoidable rather than merely undocumented. Hoisting the parse removes the mypy error entirely and drops the triple regex execution per filename:

```python
parsed_names = [parse_ranked_basename(name, ranked_re) for name in names]
indices = {p[0] for p in parsed_names if p}
```

Note `warn_unused_ignores` is unset in `pyproject.toml`, so a stale ignore here would rot silently.

dcblundell (commented):
Automated combined review (DRY / TEST / PERF / STANDARDS / SECURITY / REQUIREMENTS), deduplicated. Posting Must-fix and Should-fix findings only; lower-severity notes were delivered separately.

Checked against #1939: criteria 1, 3 and 4 met; criterion 2 is partial — see the comment on `_prefer_ranked_family`.

smountenay-tt (approved):

backend/ttnn_visualizer/utils.py
aidemsined: > [!NOTE]
> Fixed in a61ff691 — bounded at parse time, and completeness is a count.

`MAX_RANKED_WORLD_SIZE = 4096`, enforced in `is_valid_profiler_ranked_entry` so it flows through the `logger.warning` skip path you pointed at, and `len(indices) == world_size` replaces the `set(range(...))` comparison. Both halves, since either alone leaves a gap: the count stops the allocation, the cap stops a nonsense world size being treated as a real family.

Two tests: the predicate at and past the cap, and an end-to-end case where a `_1_of_99999999999` name sits beside a usable unsuffixed descriptor and the descriptor still answers.

backend/ttnn_visualizer/utils.py
aidemsined: > [!NOTE]
> Fixed in a61ff691 — `test_cluster_descriptor_rejects_a_rank_a_lone_descriptor_cannot_answer`.

Asserts 200 for rank 0 and 400 for ranks 1, 2 and 31. Checked it actually holds the mapping: neutralising the `err == "rank_out_of_range"` branch in the view fails it, where before the whole suite passed.

The frontend half of your point is still open — nothing tests that the probe loop in `useAPI.tsx:607` stops at rank 0. Left it out because it needs a hook-level test with a mocked fetch sequence rather than an addition to an existing file, so it is a bigger change than the rest of this pass. Say if you would rather it landed here than as a follow-up.

tests/clusterTopology.spec.ts
aidemsined: > [!NOTE]
> Fixed in a61ff691 — the docstring now says what the check can and cannot do.

You were right that the spec comment and the source had drifted apart, and the source was the one the next reader would trust. It now records that a UBB with scale-out links populates both fields while describing one host, that the heuristic cannot separate that from a rank-0 slice, and that the probe is bounded by the backend's 400 instead — so a false positive costs one rejected request rather than a fabricated host.

backend/ttnn_visualizer/utils.py
aidemsined: > [!NOTE]
> Fixed in a61ff691 — hoisted, so the suppression is gone rather than documented.

`parsed = [parse_ranked_basename(name, ranked_re) for name in names]` once, then `{entry[0] for entry in parsed if entry}`. Three regex executions per filename down to one, no `type: ignore`, and mypy is clean without it. Noted on `warn_unused_ignores` being unset — that is a good argument for removing ignores rather than annotating them, which is what this does.

backend/ttnn_visualizer/utils.py
aidemsined: > [!WARNING]
> **Not changed — this needs @smountenay-tt, as you said.** Your factual claim checks out, so the mtime premise is weaker than I wrote it. But the two reviews disagree about which failure to accept, and that is not mine to settle by picking whichever review came last.

**Your claim is correct.** Verified all three points:

- `sftp_operations.py:141` — `cmd.extend(["-p", str(remote_connection.port)])` is ssh's *port* flag, exactly as you said.
- `_scp_cmd_prefix` builds `["scp", "-O"]`, no `-p`.
- `file_uploads.py:159` is `file.save(dest_path)`, stamping at write time; there is no `os.utime` anywhere in the backend.

**Where I think the disagreement actually sits.** mtime does track import order *across* imports — a second import writes later, so it stamps later. What it cannot track is order *within* one import, where the sequence files land is arbitrary. So the hazard is narrower than "mtime is meaningless" but real, and it is the one that matters: a genuine multi-host report shipping both an unsuffixed and a ranked descriptor can have the unsuffixed file land last and collapse to one host.

That leaves two failures to choose between:

| rule | failure | visibility |
|---|---|---|
| complete ranked family always wins | a stale complete ranked set renders `world` stale hosts | visible, bounded |
| completeness + mtime recency | a real multi-host report silently stitches to one host | silent, breaks criterion 2 |

On that framing I agree with you, and it is also where 5db49d14 started. The reason I moved off it was @smountenay-tt's point that reverse import order is reachable and a lone `_1_of_2` should be treated consistently — which the completeness check now handles independently of mtime, so it may no longer need recency to hold.

**Scott** — does dropping mtime recency and preferring a complete ranked family unconditionally still cover the case you raised, given `_ranked_family` now reports a lone `_1_of_2` as incomplete? If so I will make that change and delete the recency path.

Holding two related findings until this settles, since both describe the mtime rule and would be rewritten with it:
- the untested tie-break and the write-order-dependent test (your `test_utils.py:821` comment)
- profiler configs being left on the old rule (your `utils.py:237` comment), which is a question about which rule they should adopt

backend/ttnn_visualizer/tests/test_utils.py
aidemsined: > [!NOTE]
> Held pending the mtime decision above — both halves of this describe the recency rule, and if that goes they go with it.

Both of your points are right on their own terms, and neither is in dispute: mutating `>=` to `>` in `_prefer_ranked_family` does leave all tests green, and that test does pass only because the unsuffixed file happens to be written first, which is unrelated to its stated intent.

Rather than write a tie-break test for a rule that may not survive the thread above, I would rather settle the rule first and then either repoint this at `_cluster_families(single_ns=OLD_NS, ranked_ns=OLD_NS)` as you suggest, or delete it along with the recency path. Flagging so it does not read as skipped.

backend/ttnn_visualizer/utils.py
aidemsined: > [!NOTE]
> Held pending the mtime decision above — this is a question about which rule configs should adopt, and that rule is what is under discussion.

The inconsistency you describe is real and I am not disputing it: `pick_profiler_config_paths` and `read_profiler_config_api_payload` hand-roll "unsuffixed always wins" over the same directory, so a report can serve rank-1 descriptor data from the ranked family while `/api/config?rank=1` answers from the stale single config.

Routing them through the shared picker is the right shape, but pointing them at completeness-plus-recency now would just propagate a rule that may be about to change. Once the thread above settles I will either route them through it or record the omission as deliberate with the reason, as you suggested.

PR #1949, opened 2026-08-27
Group condensed cluster chips by board slot

### LINKED ISSUE #1948: Condensed cluster layout places chips by device id, contradicting the link data
## Problem

In the condensed layout, chips are placed in **device-id order** rather than at physical positions, so the canvas contradicts the link data drawn on top of it.

`ClusterRenderer` places chip *c* at `(c % 4, c / 4)`:

```ts
x = localIndex % FALLBACK_PER_HOST_COLS;   // = 4
y = nextHostOffsetY + Math.floor(localIndex / FALLBACK_PER_HOST_COLS);
```

`localIndex` is the index into `chip_unique_ids` keys sorted ascending — no topology input at all.

Measured on `test_ttnn_moe_aug26_2217` (32-chip Blackhole UBB, single host):

| | |
|---|---|
| distinct intra-host links | 68 |
| drawn between **non-adjacent** grid cells | **28 (41%)** |
| grid-adjacent chip pairs in the render | 52 |
| …with **no link** between them | **12** |

Concretely: chips **0 and 8 are directly linked but drawn two rows apart**, while **4 and 8 are drawn touching and are not linked**. The 28 non-adjacent links are the long sweeping curves that dominate the view.

Device ids never track physical position. A report that *does* carry mesh coordinates proves it — `multihost_poc_jun24_2321` reads `6: [0, 0]`, `4: [0, 1]`, `5: [0, 2]`, `7: [0, 3]`.

## Why this report falls back

`chips: {}` is empty and there is no `physical_chip_mesh_coordinate_mapping.yaml`, so `hostHasMeshCoords` is false and `mesh` mode is unavailable. The layout toggle is gated on `isMultiHost && isMeshAvailable`, so it is hidden here — correct, but the user gets no indication that the positions carry no meaning.

Pre-existing behaviour (the #1510 fallback), not a regression. It became visible once #1939 stopped the report rendering as 32 cloned boards.

## The real topology

Decoded from the 140 `ethernet_connections` entries. Edge deltas are `{+1: 24, +4: 16, +8: 22, +24: 6}`:

- **4 groups of 8** chips (`chip // 8`)
- each group is a **2×4 mesh** — `+1` along a row of 4, `+4` between the two rows
- groups form a **ring** — `+8` to the next group, `+24` wrapping group 3 → 0
- 4 of the 32 group-ring links are absent, consistent with `eth_harvesting_mask: 288` in the same descriptor

`asic_locations` states the same structure independently: it cycles **1…8 four times**. That field is currently used **nowhere** in `src/` or `backend/`.

## Nuance: the grid shape is already right by accident

Stacking four 2×4 trays vertically gives `row = 2*(c/8) + (c%8)/4`, which is arithmetically identical to `c/4`. So 4-wide × 8-tall is not the wrong shape.

What is missing is any expression of the **tray boundaries** and the **ring dimension**. That is why 41% of links read as arbitrary curves and 12 touching pairs read as adjacency that does not exist. Widening or narrowing the grid will not help.

## Proposed direction

- Make the group structure explicit — visual grouping / gutters per tray, so ring links read as tray-to-tray and no adjacency is implied across a tray boundary. Drive it from `asic_locations`, falling back to `chip // 8`.
- Optionally derive placement from the `ethernet_connections` adjacency directly, which is arch-agnostic and needs nothing new from the report.
- Surface that condensed positions are not physical, since the toggle is hidden when there is no mesh data.

The durable fix is upstream: tt-metal emitting `physical_chip_mesh_coordinate_mapping.yaml` for single-host UBB captures would make this the `mesh` path rather than a fallback.

## Acceptance criteria

- On `test_ttnn_moe_aug26_2217`, no pair of chips drawn adjacent is unlinked, or tray boundaries are visually distinct enough that adjacency across them is not implied.
- Reports with mesh coordinates (`multihost_poc_jun24_2321`) are unchanged.
- A report in the condensed fallback communicates that positions are not physical.

## Out of scope

- Rendering the 39 unresolved scale-out remotes (those chips are not in the report) — as in #1939.

aidemsined:
Closes #1948. **Stacked on #1947** — based on `feature/1939-single-descriptor-one-host`, so this PR's diff is just the layout commit. Retarget to `dev` once #1947 merges.

## Summary

The condensed layout placed chip *c* at `(c % 4, c / 4)` — index order, no topology input. On the 32-chip UBB in `test_ttnn_moe_aug26_2217` that meant the canvas contradicted the links drawn on it: 12 pairs of chips were drawn touching with no link between them, and 41% of the links ran between non-adjacent cells as long sweeping curves.

`asic_locations` already states which slot each chip occupies within its board group — it cycles 1…8 four times on this report — and was used nowhere in the codebase. It now splits the chips into groups, each ordered by slot, with a gutter row between groups.

| | before | after |
|---|---|---|
| chip pairs drawn adjacent | 52 | 40 |
| …with **no link** between them | **12** | **0** |
| links drawn between non-adjacent cells | 28 | 28 |

The 28 are unchanged: they are the genuine group-to-group ring links. They now cross a gutter, which reads as a boundary, but they are still drawn straight *through* the chips between their endpoints — see Known limitations. This PR fixes placement, not legibility.

### It is a fallback tier, not an authority

Placement order is mesh coordinates → `asic_locations` → raw id order. Mesh mode is untouched.

Grouping is refused unless the slots repeat in equal consecutive runs that each cover every slot exactly once. That guard matters: `multihost_poc_jun24_2321` gives all eight of its chips the *same* `asic_location`, so keying on the distinct count would have shattered it into eight groups of one. Descriptors with one slot per chip, one slot for all chips, a non-divisible count, a repeated slot inside a run, or a missing/unparseable field all keep id order.

### Not-physical label

The layout toggle is gated on `isMultiHost && isMeshAvailable`, so a single-host report with no mesh data shows no control and previously gave no hint that the positions were arbitrary. A muted "Positions not to scale" note now appears whenever mesh coordinates are absent, with the detail in its tooltip.

## Test plan

Verified in the running app against `test_ttnn_moe_aug26_2217`: four separated blocks of eight chips, gutters between them, note present in the toolbar.

New coverage in `clusterTopology.spec.ts`:
- 32-chip UBB splits into four groups of eight, first and last group asserted.
- Groups ordered by slot, not chip id (slots deliberately reversed).
- **No pair drawn adjacent is unconnected** — hermetic, built from the real link deltas (`+1` within a row of four, `+4` within a group, `+8` to the next group, `+24` wrapping the ring).
- `null` for: every chip on one slot (the `multihost_poc` shape), one slot per chip, a run repeating a slot, absent field, unparseable values, empty chip list.

Frontend 2217 passing, backend 1135 passing, `tsc`, eslint, stylelint all clean.

## Known limitations

**Spanning links are still drawn over the chips they cross.** All 28 same-column links pass straight over unrelated chip boxes:

```
 1-9   passes over [5]
 0-8   passes over [4]
 0-24  passes over [4, 8, 12, 16, 20]
 2-26  passes over [6, 10, 14, 18, 22]
```

Selecting a chip highlights its links, and the highlight runs through chips that are not its neighbours — so the view is not yet readable even though the placement is right. Filed as #1951; not addressed here because it needs a change to the edge-routing pass rather than to placement.

It is also structural rather than a layout mistake: the ring links join *same-slot* chips, so `0↔8` runs from group 0's top row to group 1's top row and must cross the whole intervening group plus the gutter. Any linear arrangement of four groups has this property.

**What this PR does and does not establish.** Placement is consistent with the topology — every one of the 68 links is same-row or same-column, zero diagonals, and no two chips drawn adjacent are unconnected. It does not make the cluster view legible at 32 chips.

## Follow-ups

Chip unique ids shown in the tooltip were wrong while verifying this — a separate precision bug, fixed in #1952 (#1950).

The durable fix is still upstream: tt-metal emitting `physical_chip_mesh_coordinate_mapping.yaml` for single-host UBB captures would make this the `mesh` path rather than a fallback, at which point this grouping stops being used. Deriving placement from the `ethernet_connections` adjacency directly — arch-agnostic, needs nothing new from the report — remains an option if more shapes turn up that `asic_locations` cannot group.

### DIFF
--- a/src/components/cluster/ClusterRenderer.tsx
+++ b/src/components/cluster/ClusterRenderer.tsx
@@ -13,6 +13,7 @@ import { useGetClusterTopology } from '../../hooks/useAPI';
 import { getChipDesign } from '../../functions/getChipDesign';
 import {
     FALLBACK_PER_HOST_COLS,
+    getAsicLocationGroups,
     hostHasMeshCoords,
     sortHostsByConnectionProximity,
 } from '../../functions/clusterTopology';
@@ -47,6 +48,10 @@ const CLUSTER_CHIP_SIZE_SMALL = 150;
 // stays in sync). Used as a fallback when no mesh coords are present, or as
 // a user-selectable alternative to the literal mesh layout. #1510
 const FALLBACK_HOST_GUTTER_ROWS = 1;
+// Blank row between board groups in the condensed fallback, so a group-to-group
+// link visibly crosses a boundary and chips either side of one are not drawn
+// touching. #1948
+const FALLBACK_GROUP_GUTTER_ROWS = 1;
 // Top padding gives outward-edge curves from the topmost host room to arc
 // without colliding with the cluster panel chrome.
 const FALLBACK_TOP_PAD_ROWS = 2;
@@ -219,6 +224,21 @@ function buildClusterRenderModel(
 
         let usedFallback = false;
 
+        // Fallback tier between mesh coords and raw id order: group the chips by
+        // board slot so the grid can separate groups. `null` keeps id order. #1948
+        const fallbackGroups = useMeshCoords ? null : getAsicLocationGroups(host.descriptor, chipIdsForHost);
+        const fallbackRowsPerGroup = fallbackGroups ? Math.ceil(fallbackGroups[0].length / FALLBACK_PER_HOST_COLS) : 0;
+        const fallbackPositionByChip = new Map<number, readonly [number, number]>();
+        fallbackGroups?.forEach((group, groupIndex) => {
+            const groupTop = groupIndex * (fallbackRowsPerGroup + FALLBACK_GROUP_GUTTER_ROWS);
+            group.forEach((groupedChipId, slotIndex) => {
+                fallbackPositionByChip.set(groupedChipId, [
+                    slotIndex % FALLBACK_PER_HOST_COLS,
+                    groupTop + Math.floor(slotIndex / FALLBACK_PER_HOST_COLS),
+                ]);
+            });
+        });
+
         // Plain `for` to keep mutable counters out of a closure (no-loop-func).
         for (let localIndex = 0; localIndex < chipIdsForHost.length; localIndex += 1) {
             const chipId = chipIdsForHost[localIndex];
@@ -229,8 +249,14 @@ function buildClusterRenderModel(
                 [x, y] = meshCoord;
             } else {
                 usedFallback = true;
-                x = localIndex % FALLBACK_PER_HOST_COLS;
-                y = nextHostOffsetY + Math.floor(localIndex / FALLBACK_PER_HOST_COLS);
+                const grouped = fallbackPositionByChip.get(chipId);
+                if (grouped) {
+                    [x] = grouped;
+                    y = nextHostOffsetY + grouped[1];
+                } else {
+                    x = localIndex % FALLBACK_PER_HOST_COLS;
+                    y = nextHostOffsetY + Math.floor(localIndex / FALLBACK_PER_HOST_COLS);
+                }
             }
             totalCols = Math.max(totalCols, x + 1);
             totalRows = Math.max(totalRows, y + 1);
@@ -249,7 +275,12 @@ function buildClusterRenderModel(
         }
 
         if (usedFallback) {
-            const hostRows = Math.ceil(chipIdsForHost.length / FALLBACK_PER_HOST_COLS);
+            // Grouped layouts are taller by one gutter row per boundary; the last
+            // group does not get a trailing one.
+            const hostRows = fallbackGroups
+                ? fallbackGroups.length * (fallbackRowsPerGroup + FALLBACK_GROUP_GUTTER_ROWS) -
+                  FALLBACK_GROUP_GUTTER_ROWS
+                : Math.ceil(chipIdsForHost.length / FALLBACK_PER_HOST_COLS);
             nextHostOffsetY += hostRows + FALLBACK_HOST_GUTTER_ROWS;
         }
     }
@@ -837,6 +868,11 @@ function ClusterRenderer() {
                         />
                     </Tooltip>
                 </ButtonGroup>
+                {!isMeshAvailable && (
+                    <Tooltip content='This report carries no mesh coordinates, so chips are tiled by board slot rather than placed at their physical positions. Links are accurate; positions are not.'>
+                        <span className='cluster-view-layout-note'>Positions not to scale</span>
+                    </Tooltip>
+                )}
                 {isMultiHost && isMeshAvailable && (
                     <Tooltip content='Switch between honouring the report mesh coordinates and a condensed 4-wide grid'>
                         <Switch
--- a/src/components/cluster/ClusterRenderer.tsx
+++ b/src/components/cluster/ClusterRenderer.tsx
@@ -13,7 +13,7 @@ import { useGetClusterTopology } from '../../hooks/useAPI';
 import { getChipDesign } from '../../functions/getChipDesign';
 import {
     FALLBACK_PER_HOST_COLS,
-    getAsicLocationGroups,
+    getCondensedHostLayout,
     hostHasMeshCoords,
     sortHostsByConnectionProximity,
 } from '../../functions/clusterTopology';
@@ -48,10 +48,6 @@ const CLUSTER_CHIP_SIZE_SMALL = 150;
 // stays in sync). Used as a fallback when no mesh coords are present, or as
 // a user-selectable alternative to the literal mesh layout. #1510
 const FALLBACK_HOST_GUTTER_ROWS = 1;
-// Blank row between board groups in the condensed fallback, so a group-to-group
-// link visibly crosses a boundary and chips either side of one are not drawn
-// touching. #1948
-const FALLBACK_GROUP_GUTTER_ROWS = 1;
 // Top padding gives outward-edge curves from the topmost host room to arc
 // without colliding with the cluster panel chrome.
 const FALLBACK_TOP_PAD_ROWS = 2;
@@ -114,6 +110,8 @@ interface ClusterRenderModel {
     layoutMode: ClusterLayoutMode;
     // Mesh layout is viable for this topology (drives the UI toggle's visibility).
     isMeshAvailable: boolean;
+    /** True when at least one host's condensed layout fell through to chip id order. */
+    anyHostFellToIdOrder: boolean;
 }
 
 type ClusterRenderResult = { status: 'ready'; model: ClusterRenderModel } | { status: 'unsupported' };
@@ -200,6 +198,10 @@ function buildClusterRenderModel(
 
     const renderChips: RenderChip[] = [];
     let nextHostOffsetY = useMeshCoords ? 0 : FALLBACK_TOP_PAD_ROWS;
+    // Which condensed tier the layout actually reached. The caveat copy has to name
+    // the tier it got: board-slot grouping needs the descriptor to support it, and
+    // plenty of reports — the 2-chip n300 among them — fall through to chip id order.
+    let anyHostFellToIdOrder = false;
     let totalCols = 0;
     let totalRows = 0;
 
@@ -224,20 +226,12 @@ function buildClusterRenderModel(
 
         let usedFallback = false;
 
-        // Fallback tier between mesh coords and raw id order: group the chips by
-        // board slot so the grid can separate groups. `null` keeps id order. #1948
-        const fallbackGroups = useMeshCoords ? null : getAsicLocationGroups(host.descriptor, chipIdsForHost);
-        const fallbackRowsPerGroup = fallbackGroups ? Math.ceil(fallbackGroups[0].length / FALLBACK_PER_HOST_COLS) : 0;
-        const fallbackPositionByChip = new Map<number, readonly [number, number]>();
-        fallbackGroups?.forEach((group, groupIndex) => {
-            const groupTop = groupIndex * (fallbackRowsPerGroup + FALLBACK_GROUP_GUTTER_ROWS);
-            group.forEach((groupedChipId, slotIndex) => {
-                fallbackPositionByChip.set(groupedChipId, [
-                    slotIndex % FALLBACK_PER_HOST_COLS,
-                    groupTop + Math.floor(slotIndex / FALLBACK_PER_HOST_COLS),
-                ]);
-            });
-        });
+        // Fallback tier between mesh coords and raw id order: board slots when the
+        // descriptor supports them, id order when it does not. #1948
+        const condensed = useMeshCoords ? null : getCondensedHostLayout(host.descriptor, chipIdsForHost);
+        if (condensed && !condensed.isGrouped) {
+            anyHostFellToIdOrder = true;
+        }
 
         // Plain `for` to keep mutable counters out of a closure (no-loop-func).
         for (let localIndex = 0; localIndex < chipIdsForHost.length; localIndex += 1) {
@@ -249,14 +243,12 @@ function buildClusterRenderModel(
                 [x, y] = meshCoord;
             } else {
                 usedFallback = true;
-                const grouped = fallbackPositionByChip.get(chipId);
-                if (grouped) {
-                    [x] = grouped;
-                    y = nextHostOffsetY + grouped[1];
-                } else {
-                    x = localIndex % FALLBACK_PER_HOST_COLS;
-                    y = nextHostOffsetY + Math.floor(localIndex / FALLBACK_PER_HOST_COLS);
-                }
+                const placed = condensed?.positionByChip.get(chipId) ?? [
+                    localIndex % FALLBACK_PER_HOST_COLS,
+                    Math.floor(localIndex / FALLBACK_PER_HOST_COLS),
+                ];
+                [x] = placed;
+                y = nextHostOffsetY + placed[1];
             }
             totalCols = Math.max(totalCols, x + 1);
             totalRows = Math.max(totalRows, y + 1);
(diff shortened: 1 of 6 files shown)


smountenay-tt (approved):
Looks good to me. No noteworthy findings in my review.

dcblundell (commented):
Conductor review — six lenses (DRY, TEST, PERF, STANDARDS, SECURITY, REQUIREMENTS). Posting Must/Should findings only.

Placement logic checks out: I traced the grouped geometry by hand and the `hostRows` arithmetic and multi-host offset accumulation are off-by-one clean. SECURITY found nothing — `asic_locations` arrives from `yaml.safe_load` unfiltered, and every hostile shape (scalar, array, string/`NaN` slots, `__proto__` keys) degrades to `null` and id order. PERF found nothing blocking; the grouping sits inside the existing `useMemo`.

Two themes in the comments below: the user-visible caveat asserts a mechanism the code often does not use, and the headline #1948 regression test does not exercise the shipped layout.

Two items with no line to anchor to:

- **Rebase onto the base tip.** HEAD is missing `e3bacd1a` and `5db49d14` from `feature/1939-single-descriptor-one-host`. Nothing merges wrong — this three-dot diff is clean — but the "frontend 2217 passing" figure predates both commits.
- **Spec decision.** #1948's proposed direction says to fall back to `chip // 8`; this falls back to raw id order instead. That looks like the better call — `chip // 8` would mis-group any non-8-per-board topology, and a confidently-wrong grouping never beats honest id order — but the decision is undocumented in both the JSDoc and the PR body. Recommend amending the issue and adding a line to `getAsicLocationGroups`'s docstring so nobody re-adds it. Separately, the deferred `ethernet_connections`-derived placement carries no tracking issue number, unlike the spanning-links limitation which cites #1951.

src/components/cluster/ClusterRenderer.tsx
```diff
@@ -837,6 +868,11 @@ function ClusterRenderer() {
                         />
                     </Tooltip>
                 </ButtonGroup>
+                {!isMeshAvailable && (
+                    <Tooltip content='This report carries no mesh coordinates, so chips are tiled by board slot rather than placed at their physical positions. Links are accurate; positions are not.'>
```
dcblundell: **Must fix:** Tooltip claims board-slot tiling on reports that get id order

The note renders on `!isMeshAvailable`, but slot tiling additionally requires `getAsicLocationGroups` to return non-`null`. Every descriptor that falls to the raw-id tier — one slot per chip, one slot for all chips, non-divisible count, absent or unparseable field, including the 2-chip n300 fixtures — shows copy claiming a slot-driven layout it did not get. Surface which tier was used from `buildClusterRenderModel` (the model already carries `layoutMode`/`isMeshAvailable`) and word the note per tier, or drop the "by board slot" clause.

src/components/cluster/ClusterRenderer.tsx
```diff
@@ -837,6 +868,11 @@ function ClusterRenderer() {
...
+                        <span className='cluster-view-layout-note'>Positions not to scale</span>
```
dcblundell: **Must fix:** "Positions not to scale" names the wrong defect

"Not to scale" implies the arrangement is right and only the proportions are off; the actual situation is that the arrangement carries no physical meaning at all. The accurate sentence is hover-only, so the visible half is the misleading half. "Positions not physical" (or similar) would be true on its own.

src/components/cluster/ClusterRenderer.tsx
```diff
@@ -837,6 +868,11 @@ function ClusterRenderer() {
                         />
                     </Tooltip>
                 </ButtonGroup>
+                {!isMeshAvailable && (
```
dcblundell: **Should fix:** Note gating drifts from the layout toggle

This uses `!isMeshAvailable` while the toggle below uses `isMultiHost && isMeshAvailable`, so the two are mutually exclusive by construction: switching a mesh-bearing report to condensed yields equally unphysical positions and no caveat. Gating on `activeLayoutMode === 'condensed'` reads the effective mode the render model was built to expose.

src/components/cluster/ClusterRenderer.tsx
```diff
@@ -219,6 +224,21 @@ function buildClusterRenderModel(
 
         let usedFallback = false;
 
+        // Fallback tier between mesh coords and raw id order: group the chips by
+        // board slot so the grid can separate groups. `null` keeps id order. #1948
+        const fallbackGroups = useMeshCoords ? null : getAsicLocationGroups(host.descriptor, chipIdsForHost);
+        const fallbackRowsPerGroup = fallbackGroups ? Math.ceil(fallbackGroups[0].length / FALLBACK_PER_HOST_COLS) : 0;
```
dcblundell: **Should fix:** `buildClusterRenderModel` has no test coverage

`fallbackRowsPerGroup`, `groupTop` and the `hostRows` trailing-gutter subtraction are where placement actually happens, and nothing exercises them. `tests/ClusterUnknownArch.spec.tsx` already has the harness (mocked `useGetClusterTopology`, `ResizeObserver` stub), and positions are readable from `style.gridColumn`/`gridRow` plus `title="Device N"` — so this is cheap, not blocked on a refactor. Highest-value cases: mesh coords win over grouping (the guard protecting "mesh reports unchanged", currently untested), gutter rows are actually empty, and a mixed multi-host report exercising both `hostRows` branches in one pass.

src/components/cluster/ClusterRenderer.tsx
```diff
@@ -229,8 +249,14 @@ function buildClusterRenderModel(
...
             } else {
                 usedFallback = true;
-                x = localIndex % FALLBACK_PER_HOST_COLS;
-                y = nextHostOffsetY + Math.floor(localIndex / FALLBACK_PER_HOST_COLS);
+                const grouped = fallbackPositionByChip.get(chipId);
+                if (grouped) {
+                    [x] = grouped;
+                    y = nextHostOffsetY + grouped[1];
+                } else {
+                    x = localIndex % FALLBACK_PER_HOST_COLS;
```
dcblundell: **Should fix:** Two independent implementations of the condensed geometry

The grouped branch and this `else` compute the same layout, as do the two `hostRows` formulas below. The ungrouped case is algebraically the grouped case with one group: `groups = [chipIdsForHost]` gives `groupTop = 0`, `slotIndex = localIndex`, and `1 * (rows + gutter) - gutter === Math.ceil(n / 4)`. A pure `getCondensedHostLayout(descriptor, chipIds) -> { positionByChip, rows }` in `clusterTopology.ts` collapses both, moves `FALLBACK_GROUP_GUTTER_ROWS` next to `FALLBACK_PER_HOST_COLS` (hoisted there deliberately so renderer and heuristic stay in lockstep), and gives the adjacency test something real to assert against.

tests/clusterTopology.spec.ts
```diff
@@ -338,6 +340,118 @@ describe('pickMeshDocForRank', () => {
...
+        const links = new Set(
+            chipIds.flatMap((chip) =>
+                [1, 4, 8, 24].filter((delta) => isLinked(chip, delta)).map((delta) => `${chip}-${chip + delta}`),
+            ),
+        );
+
+        const groups = getAsicLocationGroups(withLocations(locations), chipIds);
+        expect(groups).toHaveLength(4);
+
+        const gutterRows = 1;
```
dcblundell: **Should fix:** The #1948 regression test validates its own copy of the layout

`gutterRows = 1` re-declares `FALLBACK_GROUP_GUTTER_ROWS` (module-private in `ClusterRenderer.tsx`) and the geometry is re-typed here, so changing the renderer's gutter or regressing `buildClusterRenderModel` leaves this test green. It also has no negative control — nothing asserts the ungrouped tiling produces the 12 unconnected pairs the docstring cites, so a trivially-passing property is indistinguishable from a real one. And because the fixture's slots are `chip % 8 + 1`, slot order equals id order: swapping `getAsicLocationGroups` for a naive `chunk(chipIds, 8)` would still pass. Assert against a shared layout helper and add the 12-pair baseline.

src/scss/components/ClusterView.scss
```diff
@@ -58,6 +58,16 @@ $mmio-fill: rgba(colours.$tt-yellow, 0.18);
...
             margin: 0;
         }
+
+        // Shown when the report has no mesh coordinates, so the tiling is not a
+        // physical arrangement. Muted: a caveat, not a control. #1948
+        .cluster-view-layout-note {
+            font-size: 12px;
+            font-style: italic;
+            white-space: nowrap;
+            cursor: help;
```
dcblundell, line 68: **Should fix:** Caveat detail is unreachable without a mouse

`cursor: help` on a non-focusable `<span>` inside a Blueprint `Tooltip` means keyboard and assistive-tech users get only the four-word label, never the sentence explaining it — and that label is the inaccurate half. Every other tooltip target in this toolbar (three zoom `Button`s, the layout `Switch`) carries an explicit `aria-label`, so this is inconsistent with the file's own convention. Add `tabIndex={0}` plus `aria-label`/`aria-describedby`, or render the explanation as visible text.

src/components/cluster/ClusterRenderer.tsx
aidemsined: > [!NOTE]
> Fixed in 75dd9764 — the render model reports the tier and the copy is worded per tier.

`buildClusterRenderModel` now sets `anyHostFellToIdOrder` when any host's condensed layout falls through, and the note reads either "tiled by board slot" or "tiled in id order … this report carries no usable board-slot information".

src/components/cluster/ClusterRenderer.tsx
aidemsined: > [!NOTE]
> Fixed in 75dd9764 — "Positions not physical".

Agreed on the reasoning: "not to scale" concedes the proportions while asserting the arrangement, and the arrangement is the part that carries no meaning. The tooltip sentence ends "positions carry no physical meaning" rather than "positions are not", for the same reason.

src/components/cluster/ClusterRenderer.tsx
aidemsined: > [!NOTE]
> Fixed in 75dd9764 — gated on `activeLayoutMode === 'condensed'`.

You were right that the two were mutually exclusive by construction, so a mesh report switched to condensed got unphysical positions and no caveat at all.

src/components/cluster/ClusterRenderer.tsx
aidemsined: > [!NOTE]
> Fixed in 75dd9764 — `getCondensedHostLayout(descriptor, chipIds) -> { positionByChip, rows, isGrouped }` in `clusterTopology.ts`, with `FALLBACK_GROUP_GUTTER_ROWS` moved next to `FALLBACK_PER_HOST_COLS`.

The algebra held exactly as you set it out: `groups = [chipIds]` gives `groupTop` 0, `slotIndex === localIndex`, and `1 * (rows + gutter) - gutter === ceil(n / cols)`. Both branches and both `hostRows` formulas collapse into it. It returns `isGrouped` as well, which is what the caveat copy above needed to name its tier.

src/components/cluster/ClusterRenderer.tsx
aidemsined: > [!NOTE]
> Fixed in 75dd9764 — `tests/ClusterCondensedPlacement.spec.tsx`, four tests on the harness you pointed at.

Your three highest-value cases, plus id-order tiling. Each was checked against a deliberate regression rather than just observed passing:

- **gutter rows are actually empty** — asserts chip 8 sits three rows below chip 0, not two, and that the row between holds nothing. Setting `FALLBACK_GROUP_GUTTER_ROWS = 0` fails this and three others.
- **both `hostRows` branches in one pass** — a grouped 16-chip host beside a slotless 6-chip one; asserts the blocks do not overlap, that exactly one row divides them, and that the spans are 5 and 2.
- **mesh coords win over grouping** — worth a correction on your framing: this passes with `useMeshCoords ? null : …` removed, because that guard only avoids computing a layout the `meshCoord` branch never consults. It is a work-saving guard, not the thing enforcing precedence. What enforces it is the `meshCoord` lookup, and stubbing *that* to `undefined` does fail the test.

One incidental finding: host render order is not rank order — with no inter-host links the proximity sort put rank 1 first — so the multi-host test asserts the order-independent invariant instead.

tests/clusterTopology.spec.ts
aidemsined: > [!NOTE]
> Fixed in 75dd9764 — it asserts against `getCondensedHostLayout` now, and gained the baseline.

All three of your points were live:

- the re-declared `gutterRows = 1` meant changing the renderer's gutter left the test green; it now shares the helper, and setting the constant to 0 fails it.
- the negative control is there: the same 32 chips laid out without slots produce exactly the **12 unconnected adjacent pairs** the docstring cites, against 0 when grouped. So the property is earned rather than trivially true.
- on `chip % 8 + 1` making slot order equal id order — added a separate case with the slots reversed within each board (`8 - (chip % 8)`), where a naive `chunk(chipIds, 8)` gives the wrong answer and the helper puts chip 0 last in its group.

src/scss/components/ClusterView.scss
aidemsined: > [!WARNING]
> Partly fixed in 75dd9764 — the assistive-tech half is done, the focusability half is not, and I would rather flag that than quietly claim it.

The caveat sentence is now the span's `aria-label` with `role='note'`, so a screen reader gets the whole explanation instead of the three visible words.

I did not add `tabIndex={0}`: on a non-interactive element it trips `jsx-a11y/no-noninteractive-tabindex`, and suppressing an a11y rule to satisfy an a11y finding seemed like the wrong trade. The alternatives both have costs — giving it `role='button'` would be a lie about what it does, and turning it into a real `Button` changes a text note into a control in the toolbar.

So the remaining gap is a sighted keyboard-only user, who gets the summary but not the sentence. Your other suggestion — rendering the explanation as visible text — would close it properly. Happy to do that if you think the toolbar can carry the extra line; it is your call rather than mine.

PR #1952, opened 2026-08-27
Carry chip unique ids as strings

### LINKED ISSUE #1950: Chip unique ids lose precision on parse, corrupting labels and inter-host link resolution
## Problem

Chip unique ids are 64-bit and exceed `Number.MAX_SAFE_INTEGER`, so `JSON.parse` rounds them the moment the cluster descriptor reaches the frontend.

Observed on `test_ttnn_moe_aug26_2217`, chip 9:

| | |
|---|---|
| `cluster_descriptor.yaml` | `5313998941933517939` |
| shown in the chip tooltip | `5313998941933518000` |

Off by 61. Every uid above 2^53 in that report is wrong, and the tooltip is only where it becomes visible.

## It is not only a display bug

`stitchClusterTopology` keys its owner index by the uid **as a number**:

```ts
const uniqueIdToOwner = new Map<number, { rank: number; chip: number }>();
```

`remote_chip_id` is rounded by the same parse. Two distinct 64-bit ids that round to the same double therefore collide in that map, and an inter-host link resolves to the wrong owner — or a link that should resolve is counted as unresolved. Silent, and it gets more likely the more chips a report has.

## Cause

`/api/cluster-descriptor` does `yaml.safe_load` then `jsonify`. Python ints are arbitrary precision, so the JSON on the wire carries the exact digits; the loss is entirely on the JS side. No frontend-only fix is possible — the value is already damaged before any of our code sees it.

## Proposed direction

Serialise the big-integer fields as **strings** at the API boundary and treat them as opaque strings throughout:

- `chip_unique_ids` values
- `ethernet_connections_to_remote_devices[].remote_chip_id`

Nothing does arithmetic on them — they are only map keys and label text — so string is a straight substitution that also makes the owner index exact.

## Acceptance criteria

- The tooltip for chip 9 of `test_ttnn_moe_aug26_2217` reads `5313998941933517939`.
- Two uids differing only below the float64 rounding threshold resolve to distinct owners.
- Reports whose uids are small enough to be exact as doubles are unaffected.

aidemsined:
Closes #1950. **Stacked on #1949** — based on `feature/1948-group-condensed-chip-layout`, so this PR's diff is just the uid commit. Retarget down the stack as the others merge.

## Summary

Chip unique ids are 64-bit and exceed `Number.MAX_SAFE_INTEGER`, so the browser rounded them on parse. Chip 9 of `test_ttnn_moe_aug26_2217` is `5313998941933517939` in the descriptor and was displayed as `5313998941933518000`.

Not only a label. `stitchClusterTopology` keys its unique-id-to-owner index by the parsed value:

```ts
const uniqueIdToOwner = new Map<number, { rank: number; chip: number }>();
```

`remote_chip_id` was rounded by the same parse, so two ids differing below the float64 threshold collided in that map — an inter-host link resolved to the wrong chip, or was counted unresolved. Silent, and more likely the more chips a report has.

The loss happens inside `JSON.parse`, before any frontend code runs, so no frontend-only fix exists. `chip_unique_ids` values and `remote_chip_id` are now serialised as strings at the API boundary and treated as opaque strings throughout. Nothing did arithmetic on them — they are map keys and label text — so it is a straight substitution that also makes the owner index exact.

## Test plan

End to end against the real report through the running backend:

| | |
|---|---|
| `cluster_descriptor.yaml`, chip 9 | `5313998941933517939` |
| on the wire, before | `5313998941933518000` (number) |
| on the wire, after | `"5313998941933517939"` (string) |

All 32 uids remain distinct after the change, and `remote_chip_id` arrives as a string.

New coverage:
- `test_stringify_chip_unique_ids_preserves_64_bit_values` — uses the real chip-9 value, asserts adjacent ids stay distinct as strings while `float(exact) == float(exact + 1)` shows they would not as doubles, and that the local endpoint of a remote connection is untouched.
- `test_stringify_chip_unique_ids_tolerates_missing_and_odd_shapes` — `None`, a non-mapping, a descriptor lacking both fields, and remote entries that are not endpoint pairs all pass through instead of raising.
- `resolves inter-host links between uids that collide as doubles (#1950)` — asserts `Number(first) === Number(second)` to prove the collision is real, then that both links still resolve to their own owner.

Frontend 2218 passing, backend 1137 passing, `tsc`, eslint, stylelint clean.

## Note for reviewers

The bulk of the diff is mechanical: existing specs used numeric uid literals, so they move to strings. The behavioural change is the helper in `utils.py`, one call site in `views.py`, and three type changes.


tests/clusterTopology.spec.ts
```diff
@@ -340,6 +340,45 @@ describe('pickMeshDocForRank', () => {
...
+                        ],
+                    ],
+                }),
+                meshDescriptor: null,
+            },
+            {
+                rank: 1,
+                descriptor: makeDescriptor({ chip_unique_ids: { 0: second } }),
+                meshDescriptor: null,
+            },
```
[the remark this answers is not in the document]
aidemsined: Good catch, and it was worse than "would not expose the failure" — the test passed against the exact behaviour it was written to catch. I reverted the map to number keys and ran it: **37 passed**, including this one. Vacuous.

Cause is as you describe: `second` is indexed last, so with a collision the surviving entry is `(rank 1, chip 0)`, which is precisely what rank 0's lookup of `second` needs. It resolved correctly by luck.

Fixed in 7e7427eb by making the fixture symmetric — both hosts now declare the link, so both directions get looked up. With string keys they dedupe to one link. With number keys rank 1's lookup of `first` resolves to rank 1 itself and emits a second link with both endpoints on one host. Re-ran with number keys to confirm it now fails:

```
× resolves inter-host links between uids that collide as doubles (#1950)
AssertionError: expected [ …(2) ] to have a length of 1 but got 2
```

Also added a comment recording why the fixture has to be two-sided, so it does not get simplified back.

tests/ClusterUnknownArch.spec.tsx
```diff
@@ -56,7 +56,7 @@ const twoHostTopology = (arch: Record<number, string>): ClusterTopology =>
             rank,
             descriptor: {
                 arch,
-                chip_unique_ids: { 0: 100 + rank },
+                chip_unique_ids: { 0: `100${rank}` },
```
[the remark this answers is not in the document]
aidemsined: Correct, and self-inflicted: I converted the uid literals with a regex and `eslint --fix` turned `String(100 + rank)` into a template literal, which silently changed the values to `1000`/`1001` while `interHostLinks` still named `100`/`101`. Restored to `String(100 + rank)` in 7e7427eb.

tests/clusterTopology.spec.ts
```diff
@@ -512,12 +551,12 @@ describe('looksLikeRankedDescriptor', () => {
                     chips: Array.from({ length: 32 }, (_, chip) => chip),
                 },
             ],
-            chip_unique_ids: Object.fromEntries(Array.from({ length: 32 }, (_, chip) => [chip, 1000 + chip])),
+            chip_unique_ids: Object.fromEntries(Array.from({ length: 32 }, (_, chip) => [chip, String(1000 + chip)])),
```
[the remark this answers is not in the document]
aidemsined: Fixed in 7e7427eb — the comment now says the ids arrive as strings and references #1950, rather than documenting the rounding this PR removes.

smountenay-tt (approved):
Looks good. No issues found.

dcblundell (commented):
Combined review (DRY / TEST / PERF / STANDARDS / SECURITY / REQUIREMENTS). No Must-fix findings — all six acceptance criteria of #1950 are met and no lens found a correctness, security or performance defect. Three Should-fix comments below.

backend/ttnn_visualizer/views.py
```diff
@@ -1716,7 +1717,8 @@ def get_cluster_descriptor(instance: Instance):
     try:
         with open(path, "r", encoding="utf-8") as cluster_desc_file:
             cluster_desc = yaml.safe_load(cluster_desc_file)
-        return jsonify(cluster_desc), HTTPStatus.OK
+        # Chip unique ids are 64-bit; the browser would round them on parse. #1950
+        return jsonify(stringify_chip_unique_ids(cluster_desc)), HTTPStatus.OK
```
dcblundell, line 1749: **Should fix:** nothing pins the endpoint's wire shape

Flagged independently by the TEST, SECURITY and REQUIREMENTS lenses. `test_descriptor_endpoints.py` only has a 404 case for `/api/cluster-descriptor` (its 200-path tests are all `mesh-descriptor`), so deleting `stringify_chip_unique_ids(...)` from this line leaves the whole suite green while the frontend's `chip_unique_ids: Record<ChipId, string>` becomes a lie with no runtime guard behind it. Acceptance criterion 1 is currently verified only by the manual end-to-end run in the PR body.

`_register_profiler_instance_with_dir` already exists in that file, so this is ~15 lines. Assert on `response.data` bytes rather than `get_json()` — Python's `json.loads` is lossless, so only the raw bytes prove the value was quoted on the wire.

tests/clusterTopology.spec.ts
```diff
@@ -579,8 +629,8 @@ describe('sortHostsByConnectionProximity', () => {
     const eightChips = (offset: number) => Object.fromEntries([0, 1, 2, 3, 4, 5, 6, 7].map((id) => [id, offset + id]));
```
dcblundell: **Should fix:** stale fixture contradicts the new contract

Flagged by four lenses (DRY, TEST, STANDARDS, REQUIREMENTS). This `eightChips` helper still emits numeric uids (`offset + id`), and `makeHost` just above (~line 615) still declares `chipUniqueIds: Record<number, number>`. The `as unknown as ClusterModel` cast at ~line 626 hides both from `tsc`. The sibling `link()` helper immediately below *was* migrated to `'0'`, so this reads as an oversight rather than a deliberate exemption.

Behaviour is unaffected — `sortHostsByConnectionProximity` reads only `Object.keys` — but the fixture now describes a payload the backend can no longer produce, so the next person editing it reintroduces numeric uids with no compiler complaint. Suggest `Record<number, string>` and `String(offset + id)`.

backend/ttnn_visualizer/tests/test_utils.py
```diff
@@ -706,6 +707,45 @@ def test_pick_cluster_descriptor_prefers_unsuffixed_file(tmp_path):
...
+
+def test_stringify_chip_unique_ids_tolerates_missing_and_odd_shapes():
+    assert stringify_chip_unique_ids(None) is None
+    assert stringify_chip_unique_ids("not a mapping") == "not a mapping"
+    # A descriptor without either field passes through unchanged.
+    assert stringify_chip_unique_ids({"arch": {0: "blackhole"}}) == {
+        "arch": {0: "blackhole"}
+    }
+    # Remote entries that are not endpoint pairs are left alone rather than raising.
+    assert stringify_chip_unique_ids(
```
dcblundell, line 889: **Should fix:** this comment overstates what the assertion proves

"The local endpoint is untouched" implies positional discrimination, but the implementation is purely key-based — it rewrites *any* endpoint dict carrying `remote_chip_id`, index 0 included. The fixture's local endpoint simply lacks the key, so the assertion says nothing about position and the real contract goes unpinned.

Either add an endpoint pair whose *first* element carries `remote_chip_id` (which pins the key-based behaviour), or reword to "endpoints without the key are untouched".

backend/ttnn_visualizer/views.py
aidemsined: > [!NOTE]
> Fixed in b952ea26 — `test_cluster_descriptor_quotes_chip_unique_ids_on_the_wire`, asserting on `response.data` bytes as you suggested.

Verified the way you framed it: with `stringify_chip_unique_ids(...)` deleted from the view, the new test fails and the other five pass. The failure prints the raw body, which is the useful part —

```
{"chip_unique_ids":{"9":5313998941933517939}, ...}
```

so the diagnostic shows the unquoted number rather than just a boolean. It also asserts the bare value appears nowhere unquoted, which covers `remote_chip_id` in the same pass.

backend/ttnn_visualizer/tests/test_utils.py
aidemsined: > [!NOTE]
> Fixed in b952ea26 — took the first option rather than the reword.

You are right that the assertion said nothing: the rewrite matches on the `remote_chip_id` key wherever it appears, and the fixture's first endpoint just happened not to carry one. The fixture now has a second, reversed pair whose index-0 endpoint carries the key, so both directions are pinned, and the comment says "endpoints without the key are untouched, whichever side they sit on".

tests/clusterTopology.spec.ts
aidemsined: > [!NOTE]
> Fixed in b952ea26 — `Record<number, string>` and `String(offset + id)`, as suggested.

Agreed it read as an oversight rather than an exemption; `link()` right below it was migrated and this was not. Behaviour is unaffected as you say, but the cast meant `tsc` had nothing to complain about, so the next edit would have reintroduced numeric uids silently.

PR #1955, opened 2026-08-28
Let an NPE report supply its own SoC descriptor

### LINKED ISSUE #1776: Support an optional SoC descriptor override in the NPE JSON (BOS request)
Unlike Cluster (see #1772), **NPE genuinely needs the full SoC grid** — it lands NoC transfers on actual node coordinates (`useNodeType` → cores/dram/eth/pcie/grid), and there is no report-side substitute for that grid. Today NPE resolves it from a **baked** `arch-*.json` keyed on `npeData.common_info.arch` (raw-cast to `DeviceArchitecture`); an unknown arch misses the `useArchitecture` switch → empty `{}` → "Unsupported architecture" toast. That doesn't scale to the Quasar-IP family (Grendel + licensee parts like Trinity / Horizon), whose complete descriptors are rolled by the customer and can't be baked into the app.

> Rescoped from the broad "report-supplied SoC descriptors" to the concrete, NPE-specific lever: let the NPE JSON **optionally carry its own SoC descriptor** as an override.

## Scope

- Allow the NPE JSON to include an optional complete SoC descriptor (`ChipDesign` shape) alongside `common_info`.
- Have NPE (`useNodeType` / `useArchitecture` in `src/hooks/useAPI.tsx`) consume the supplied descriptor when present, and fall back to the baked WH/BH lookup keyed on `common_info.arch` when absent.
- Validate the override shape (`grid` + `functional_workers` / `dram` / `eth` / `pcie`) and surface a clear error when malformed, instead of rendering an empty grid.
- Make the unknown-arch path defer to the override rather than immediately toasting "Unsupported architecture".

## Acceptance criteria

- An NPE JSON carrying a SoC-descriptor override renders its real tensix/dram/eth/pcie topology in NPE with no baked entry for that arch.
- NPE JSONs with no override still resolve via the baked WH/BH descriptors — no regression, no spurious "Unsupported" for known archs.
- A malformed override produces an explicit, user-visible error, not a silent empty grid.

## Notes

- Cluster is being decoupled from baked arch separately (#1772); this issue covers the NPE topology path that has no cluster-descriptor equivalent.
- `grendel` exists in UMD as an arch implementation but ships no SoC descriptor yaml, reinforcing that the complete descriptor must come from the report artifact, not the app bundle.

aidemsined:
Closes #1776. Targets `dev`.

## Summary

NPE lands NoC transfers on real node coordinates, so it needs the full SoC grid and resolved one from a baked `arch-*.json` keyed on `common_info.arch`. An arch with no baked entry fell through to `null` and drew nothing but an "Unsupported architecture" toast. That does not work for the Quasar-IP family — Grendel and the licensee parts have complete descriptors rolled by the customer that cannot be bundled into the app.

An NPE report may now carry its own descriptor in an optional top-level `soc_descriptor`. Resolution order:

1. a **valid** supplied descriptor wins;
2. **absent** falls back to the baked lookup, unchanged;
3. **malformed** is reported with the specific problems and does *not* silently fall back to "unsupported".

That third case is the point. A descriptor that was offered and rejected is a different failure from one that was never there, and collapsing them is how an unusable override renders as an empty grid.

### Also carried through the windowed path

Uploads take the windowed path locally (`isWindowedView` in `routes/NPE.tsx`), which assembles `NPEData` from the sqlite index rather than the source JSON — so the descriptor would have been dropped for exactly the reports people upload. The index now persists it and `INDEX_VERSION` moves to 5 so existing indexes rebuild. `read_summary` reads it with `.get`, so an older index is still readable.

### On validation

Empty `dram` / `eth` / `pcie` are accepted: a tensix-only emulation descriptor legitimately has none, and rejecting them would make the override unusable for the family it exists for. `functional_workers` is the one list that must be populated — without it there is nothing to draw. `arch_name` carries the descriptor's own label, falling back to the arch the report declared, never a baked one: a Grendel override reported as `wormhole` would be a worse answer than an unrecognised name.

## Test plan

Driven end to end in the running app with three fixtures built from `report_with_zones/npe_viz/UnknownOP_ID1.npeviz.zst`, each with `common_info.arch` changed to `grendel_xyz` (no baked entry):

| fixture | result |
|---|---|
| valid override (Wormhole descriptor baked in) | renders the full topology, `Architecture: grendel_xyz`, **no** toast, no `Unsupported arch` console error |
| no override | unresolved as before — `Unsupported architecture` toast, unchanged behaviour |
| malformed override | explicit toast naming all four problems, no silent empty grid |

Confirmed the descriptor survives both server paths: `/api/npe` (whole-file) and `/api/npe/summary` (windowed index).

New coverage:
- `socDescriptorOverride.spec.ts` — 11 cases. The load-bearing one asserts **both baked descriptors validate**, so the gate cannot end up stricter than the data the app already renders. Plus tensix-only acceptance, arch-name labelling, and six malformed shapes.
- `useNodeTypeSocOverride.spec.ts` — 6 cases over the hook: unknown arch rendered from a supplied descriptor, baked fallback when nothing is supplied, supplied beating baked, unknown arch left unresolved, malformed reported rather than treated as absent, and a malformed override not masking a usable baked descriptor.

Frontend 2230 passing, backend 1148 passing; `tsc`, eslint, stylelint, isort and black clean.

## Notes for reviewers

- `soc_descriptor` is typed `unknown` on `NPEData` deliberately — it arrives from a file we do not control, so it is validated rather than cast.
- The malformed toast appears twice in dev. That is React's strict-mode double effect; the pre-existing "Unsupported architecture" toast behaves identically, so it is not introduced here.
- Fixtures are not committed (5.4 MB each). They live at `~/Documents/repos/npe-1776-fixtures/` and the generator is three lines of `zstd`/`orjson` if they need rebuilding.

### DIFF
--- a/backend/ttnn_visualizer/npe_index.py
+++ b/backend/ttnn_visualizer/npe_index.py
@@ -42,7 +42,9 @@ _BUILD_LOCK = threading.Lock()
 INDEX_SUFFIX = ".npeidx.sqlite"
 
 # Bump when the schema or build logic changes so stale caches are rebuilt.
-INDEX_VERSION = 4
+# 5: carries `soc_descriptor` in meta, so a report-supplied SoC descriptor
+#    survives the windowed path as well as the whole-file one. #1776
+INDEX_VERSION = 5
 
 _TRANSFER_INSERT_BATCH = 5000
 
@@ -241,6 +243,9 @@ def _write_index(
                 ("source_mtime_ns", str(source_mtime_ns)),
                 ("n_timesteps", str(len(timesteps))),
                 ("common_info", orjson.dumps(obj.get("common_info", {}))),
+                # Optional; `null` when the report supplies none. Stored rather than
+                # derived because the windowed path never re-reads the source JSON.
+                ("soc_descriptor", orjson.dumps(obj.get("soc_descriptor"))),
                 ("chips", orjson.dumps(obj.get("chips", {}))),
                 ("zones", orjson.dumps(obj.get("zones", []))),
             ],
@@ -359,6 +364,9 @@ def read_summary(db_path: Path) -> dict:
 
     return {
         "common_info": orjson.loads(meta["common_info"]),
+        # `.get` rather than `[...]`: an index written before this key existed is
+        # still readable, and a rebuild is only triggered by INDEX_VERSION.
+        "soc_descriptor": orjson.loads(meta.get("soc_descriptor") or b"null"),
         "chips": orjson.loads(meta["chips"]),
         "zones": orjson.loads(meta["zones"]),
         "n_timesteps": int(meta["n_timesteps"]),
--- a/src/functions/socDescriptorOverride.ts
+++ b/src/functions/socDescriptorOverride.ts
@@ -75,8 +75,10 @@ export const parseSocDescriptorOverride = (raw: unknown, reportedArch?: string):
         ...gridProblems(raw.grid),
         ...coordinateProblems('functional_workers', raw.functional_workers),
         // `dram` is nested one level deeper than the rest: a bank per channel.
-        ...(Array.isArray(raw.dram)
-            ? raw.dram.flatMap((channel, index) => coordinateProblems(`dram[${index}]`, channel))
+        // Optional on the same terms as `eth` / `pcie`, so `grid` and
+        // `functional_workers` are the only two a producer must write.
+        ...(Array.isArray(raw.dram ?? [])
+            ? ((raw.dram ?? []) as unknown[]).flatMap((channel, index) => coordinateProblems(`dram[${index}]`, channel))
             : ['`dram` must be an array of channels']),
         ...coordinateProblems('eth', raw.eth ?? []),
         ...coordinateProblems('pcie', raw.pcie ?? []),
@@ -106,6 +108,7 @@ export const parseSocDescriptorOverride = (raw: unknown, reportedArch?: string):
             ...raw,
             arch_name: (raw.arch_name ?? reportedArch ?? 'unknown') as DeviceArchitecture,
             arc: raw.arc ?? [],
+            dram: raw.dram ?? [],
             eth: raw.eth ?? [],
             pcie: raw.pcie ?? [],
             router_only: raw.router_only ?? [],
--- a/src/functions/socDescriptorOverride.ts
+++ b/src/functions/socDescriptorOverride.ts
@@ -21,31 +21,87 @@ export type SocDescriptorOverride =
 /** Node lists are `"x-y"` coordinate strings, as the baked descriptors write them. */
 const COORDINATE = /^\d+-\d+$/;
 
+/**
+ * Renderer-safe ceilings. `EmptyChipRenderer` materialises one element per cell,
+ * so the grid is a direct multiplier on DOM size: a report asking for
+ * 1,000,000 x 1,000,000 would hang the tab before anything drew. The largest real
+ * part is ~17 x 12, so these leave roughly two orders of magnitude of headroom
+ * while keeping a nonsense value from reaching React. #1776
+ */
+const MAX_GRID_AXIS = 256;
+const MAX_GRID_CELLS = 16_384;
+
+interface GridExtent {
+    xSize: number;
+    ySize: number;
+}
+
 const isRecord = (value: unknown): value is Record<string, unknown> =>
     typeof value === 'object' && value !== null && !Array.isArray(value);
 
-const coordinateProblems = (label: string, value: unknown): string[] => {
+/**
+ * Problems with one node list, including any coordinate outside `extent`.
+ *
+ * The bounds check is the difference between rejecting a descriptor and rendering
+ * an empty grid from it: `EmptyChipRenderer` walks x < width and y < height and
+ * looks each cell up by coordinate, so a worker at `99-99` on a 4x4 grid is never
+ * visited and silently disappears — the exact failure this validation exists to
+ * prevent. `extent` is null only when `grid` itself failed, where a bound would
+ * be meaningless and the grid problem is already reported.
+ */
+const coordinateProblems = (label: string, value: unknown, extent: GridExtent | null): string[] => {
     if (!Array.isArray(value)) {
         return [`\`${label}\` must be an array of "x-y" coordinates`];
     }
-    const bad = value.filter((entry) => typeof entry !== 'string' || !COORDINATE.test(entry));
-    if (bad.length === 0) {
+
+    const malformed = value.filter((entry) => typeof entry !== 'string' || !COORDINATE.test(entry));
+    if (malformed.length > 0) {
+        const count = malformed.length === 1 ? '1 entry that is' : `${malformed.length} entries that are`;
+        return [`\`${label}\` has ${count} not an "x-y" coordinate`];
+    }
+
+    if (extent === null) {
         return [];
     }
-    const count = bad.length === 1 ? '1 entry that is' : `${bad.length} entries that are`;
-    return [`\`${label}\` has ${count} not an "x-y" coordinate`];
+
+    const outside = (value as string[]).filter((entry) => {
+        const [x, y] = entry.split('-').map(Number);
+        return x >= extent.xSize || y >= extent.ySize;
+    });
+    if (outside.length === 0) {
+        return [];
+    }
+    const shown = outside.slice(0, 3).join(', ');
+    const rest = outside.length > 3 ? `, and ${outside.length - 3} more` : '';
+    return [`\`${label}\` has coordinates outside the ${extent.xSize}x${extent.ySize} grid: ${shown}${rest}`];
 };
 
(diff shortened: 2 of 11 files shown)


src/functions/socDescriptorOverride.ts
```diff
@@ -0,0 +1,114 @@
...
+const COORDINATE = /^\d+-\d+$/;
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+    typeof value === 'object' && value !== null && !Array.isArray(value);
+
+const coordinateProblems = (label: string, value: unknown): string[] => {
+    if (!Array.isArray(value)) {
+        return [`\`${label}\` must be an array of "x-y" coordinates`];
+    }
+    const bad = value.filter((entry) => typeof entry !== 'string' || !COORDINATE.test(entry));
```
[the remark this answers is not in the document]
aidemsined: Right, and it defeats the purpose rather than merely being incomplete — I had listed this as a known limitation in the PR notes, which was the wrong call. Fixed in e4ea0faa.

Confirmed the mechanism in `EmptyChipRenderer`: it walks `x < width` and `y < height` and resolves each cell via `nodeTypeByCoord.get(`${y}-${x}`)`, so a node the grid never reaches is simply never asked for. A `99-99` worker on a 4x4 grid produces an empty topology from a descriptor that passed validation — the failure #1776 exists to prevent, reached *through* the validation.

Every node list is now bounds-checked, not just the workers, and the message names the offenders:

```
`functional_workers` has coordinates outside the 4x4 grid: 99-99
```

Two details worth flagging:

- A grid that is itself invalid skips the bounds check. Otherwise a `0x0` grid would emit "outside the 0x0 grid" for every node and bury the one problem that matters.
- Only the first three offenders are named, then `and N more`, so a wholly-transposed descriptor produces a readable message rather than a wall of coordinates.

src/functions/socDescriptorOverride.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const gridProblems = (grid: unknown): string[] => {
+    if (!isRecord(grid)) {
+        return ['`grid` must be an object with `x_size` and `y_size`'];
+    }
+    return (['x_size', 'y_size'] as const).flatMap((axis) => {
+        const size = grid[axis];
+        return typeof size === 'number' && Number.isInteger(size) && size > 0
+            ? []
+            : [`\`grid.${axis}\` must be a positive integer`];
```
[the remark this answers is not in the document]
aidemsined: Agreed, fixed in e4ea0faa. `EmptyChipRenderer` materialises one element per cell, so the grid is a direct multiplier on DOM size and a report file could hang the tab before anything drew.

Capped at **256 per axis** and **16384 cells** total. Both, because either alone has a hole: a per-axis cap lets `200 x 200` through at 40,000 cells, and a cell cap alone lets a `1 x 16384` sliver through. The largest real part is ~17x12, so this leaves roughly two orders of magnitude of headroom — the numbers and that reasoning are in the constant's docstring rather than left implicit.

One consequence worth naming: the axis cap is expressible in draft-07 and is now in the published schema, but the **cell-count cap is not**, so it joins the bounds check as runtime-only. See my reply on the schema below.

backend/ttnn_visualizer/npe_index.py
```diff
@@ -241,6 +243,9 @@ def _write_index(
                 ("source_mtime_ns", str(source_mtime_ns)),
                 ("n_timesteps", str(len(timesteps))),
                 ("common_info", orjson.dumps(obj.get("common_info", {}))),
+                # Optional; `null` when the report supplies none. Stored rather than
+                # derived because the windowed path never re-reads the source JSON.
+                ("soc_descriptor", orjson.dumps(obj.get("soc_descriptor"))),
```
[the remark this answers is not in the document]
aidemsined: Correct, and this is the one I would have been most likely to regress — I verified the round-trip by hand through the endpoint and then did not pin it. Fixed in e4ea0faa.

Three cases in `test_npe_index.py`, driving `ensure_index` → `read_summary`:

- a descriptor round-trips unchanged from a plain `.json` report;
- the same from a `.npeviz.zst` report, which is the format uploads actually arrive in;
- a report *without* one yields `soc_descriptor: None` rather than a `KeyError`, since absent means "fall back to the baked lookup" and a missing key here would be a 500 for every legacy report.

Checked they actually bite: deleting the `("soc_descriptor", ...)` meta row fails the first two and leaves the third passing, which is the shape you would expect from a dropped-field regression.

aidemsined:
Following up on the two new validation rules, because they change what the published schema can promise.

`grid.x_size` / `y_size` now carry `maximum: 256` in `npe-soc-descriptor.schema.json`. The other two rules cannot be expressed in draft-07 at all:

- **coordinates inside the grid** — cross-field comparison;
- **total cell count ≤ 16384** — arithmetic across two fields.

So the schema is now a *necessary but not sufficient* condition, and I would rather say that plainly than let it quietly drift. Three changes:

1. The schema `description` states both runtime-only rules explicitly, so a producer reading only the schema knows they exist.
2. `socDescriptorSchema.spec.ts` is now directional. A `runtimeOnly` table lists descriptors the schema accepts and the validator rejects, asserted in that direction, plus a case asserting the schema **never rejects what the validator accepts** — the direction that would actually hurt a producer.
3. Verified the sample in the tt-npe request (tenstorrent/tt-npe#132) still passes both, along with the two baked descriptors.

Frontend 2256 passing, backend 1151 passing.

src/functions/socDescriptorOverride.ts
```diff
@@ -0,0 +1,176 @@
...
+            return [`\`grid.${axis}\` must be a positive integer`];
+        }
+        return size > MAX_GRID_AXIS ? [`\`grid.${axis}\` is ${size}, above the ${MAX_GRID_AXIS} limit`] : [];
+    });
+    if (problems.length > 0) {
+        return problems;
+    }
+
+    const cells = (grid.x_size as number) * (grid.y_size as number);
+    return cells > MAX_GRID_CELLS ? [`\`grid\` asks for ${cells} cells, above the ${MAX_GRID_CELLS} limit`] : [];
```
smountenay-tt, line 96: [blocking] This ceiling does not bound the renderer's actual work. `NPEView` mounts one `EmptyChipRenderer` per report-controlled chip, and each instance both materialises `width × height` cells and builds its lookup from every supplied node entry. A descriptor with a valid 128×128 grid plus 64 chips creates 1,048,576 React cells; alternatively, millions of duplicate valid `"0-0"` workers pass validation and are processed once per chip. A crafted uploaded report can therefore still freeze the tab despite this check. Please cap the total workload (`cells × chip count`), cap aggregate node-list entries, and reject duplicates (or otherwise enforce an equivalent renderer-safe global bound).

src/functions/socDescriptorOverride.ts
```diff
@@ -0,0 +1,176 @@
...
+        // `functional_workers` are the only two a producer must write.
+        ...(Array.isArray(raw.dram ?? [])
+            ? ((raw.dram ?? []) as unknown[]).flatMap((channel, index) =>
+                  coordinateProblems(`dram[${index}]`, channel, extent),
+              )
+            : ['`dram` must be an array of channels']),
+        ...coordinateProblems('eth', raw.eth ?? [], extent),
+        ...coordinateProblems('pcie', raw.pcie ?? [], extent),
+        ...coordinateProblems('arc', raw.arc ?? [], extent),
+        ...coordinateProblems('router_only', raw.router_only ?? [], extent),
```
smountenay-tt, line 145: [blocking] The lists are validated independently, so conflicting node types are accepted. For example, `{ grid: { x_size: 2, y_size: 2 }, functional_workers: ['1-1'], dram: [['1-1']] }` returns `valid`; `EmptyChipRenderer` then uses first-match-wins (`cores` before `dram`) and silently hides the DRAM node. Since this override is the source of truth for an otherwise unknown device, that produces a materially wrong topology with no error. Please reject coordinates that occur in more than one rendered node-type list (and add a regression case).

smountenay-tt (changes requested):
Backend-focused review complete: the index versioning, descriptor round-trip, legacy-index fallback, route serialization, and JSON/zstd coverage look solid. I found two frontend blockers in the untrusted descriptor validation path; both can still produce a frozen or materially incorrect topology from a descriptor that is reported as valid.

src/functions/socDescriptorOverride.ts
aidemsined: this should be addressed on the SoC provision side. we should not be responsible for validating or accounting for deliberately malicious SoC. we are ok with the info that can be rendred. data correctness is NPE's responsibility

src/functions/socDescriptorOverride.ts
aidemsined: this should be addressed on the SoC provision side. we should not be responsible for validating or accounting for deliberately malicious SoC. we are ok with the info that can be rendred. data correctness is NPE's responsibility

smountenay-tt (approved):

Issue #1897, opened 2026-08-19
Operation details: CB and buffer legend columns don't align across nesting depths


aidemsined:
## Symptom

<img width="1024" height="556" alt="Image" src="https://github.com/user-attachments/assets/1d0b13d9-d6cd-463b-89b1-68ac569646a8" />

In the device operations view, the CB rows and buffer-allocate rows don't line up with each other. Address and size columns sit at different x positions depending on how deeply the row's DeviceOp is nested, so a list of CB allocations reads as stair-stepped columns rather than a table. The `CBs / Buffers / Total` gutter on the right stays put, which makes the drift on the left more obvious.

## Reproduction

1. Load a memory report (measured on `resnet50_may08_1841`).
2. Open any operation with device operations — e.g. operation 5 `ttnn.fold` — and switch to the device operations graph view.
3. Compare the left edge of the address column across the `Buffer allocate` and `CBs` rows.

Measured left edge of the address column, same view, no scrolling:

| Viewport | Distinct address-column positions |
|---|---|
| 1280px | 75px, 125px |
| 1024px | 45px, 75px, 125px |

At 1024px that's three different column positions among 34 rows of one view.

## What is *not* wrong

Worth stating so a fix doesn't regress it: the 320px numeric gutter is correct. Every row's gutter and the header's measure identically (`left: 945`, `right: 1265` at 1280px), so the `.memory-info` / `.memory-info-placeholder` pairing introduced for that purpose is doing its job. Nothing is clipped or overflowing either — `.device-operations-full-render` has `scrollWidth === clientWidth` and no size cell is truncated.

## Root cause

Each legend row is an independent grid, so nothing ties columns across rows:

- `.legend-item` is `display: inline-grid` per row (`src/scss/components/MemoryLegendElement.scss:30-32`), and `.memory-legend-row` (`src/scss/components/DeviceOperationFullRender.scss:165-216`) flexes one such grid against the fixed gutter.
- The only elastic track is the address, `minmax($legend-address-column, 1fr)` (`DeviceOperationFullRender.scss:180`, floor `90px` at `src/scss/definitions/_variables.scss:52`). Alignment across rows only holds while that single track can absorb every per-row difference. Once it bottoms out, the differences surface as raw offsets.
[...]

PR #1957, opened 2026-08-31
Collapse a mesh op's circular buffers into one L1 legend row

### LINKED ISSUE #1879: L1 plot legend repeats each circular buffer once per device on multi-host reports
Split out of #1844 review ([#1877](https://github.com/tenstorrent/ttnn-visualizer/pull/1877)), raised by @dcblundell.

## Problem

#1844 collapsed the per-device CB fan-out on two surfaces: the per-DeviceOp memory legend and the circular-buffer pressure modal. A third list was left untouched.

`DeviceOperationNode.cbList` is appended once per `circular_buffer_allocate` node with no device collapse (`src/model/OperationDetails.ts`, in the captured-graph walk). `L1Plots` then flattens every op's `cbList` into `memoryReportWithCB`:

```ts
const memoryReportWithCB: FragmentationEntry[] = [
    ...memoryReport,
    ...operationDetails.deviceOperations.map((op) => op.cbList.map((cb) => ({ ...cb, markerType: MarkerType.CB, ... }))).flat(),
].sort((a, b) => a.address - b.address);
```

So with the "Circular buffers" toggle on, the multihost capture shows each CB **eight times** in that legend — unlabelled, and interleaved by address with the buffer fragmentation entries.

## Scope

Row repetition only. The **numbers are safe**: `calculateCondensed` is envelope-based, so the duplicated entries do not inflate any total. This is the cosmetic half of #1844's criterion 3, for the one surface that criterion did not name.

## Why it was not folded into #1877

The grouping context differs from the other two surfaces. Entries here are sorted by address and merged with fragmentation entries rather than listed per DeviceOp, so the collapse needs its own design decision (which address wins, how a collapsed marker renders against the plot) rather than reuse of `CBDeviceFanout`. Bundling it would have grown an already-large PR.

## Suggested approach

Collapse at the `L1Plots` boundary. Reuse `getDeviceCountLabel()` (`src/functions/formatting.ts`) for the label so the wording stays consistent with the legend and modal — `MemoryLegendElement` already takes a `deviceCount` prop and renders it, from #1844.

**Correction to the original suggestion:** sharing a helper with `CBDeviceFanout` is not viable as written. `cbFanout` is built inside `DeviceOperationsFullRender`'s `useMemo` and is keyed by graph **node id**, and `cbList` carries neither the node id nor a way to reach that memo.

What makes the collapse straightforward anyway: **`device_id` is already available and already typed.** It is present on 100% of `circular_buffer_allocate` nodes (2560 of them across 14 ops in `test_ttnn_moe_aug26_2217`), the backend splices the graph blob through verbatim via `orjson.Fragment`, `BaseMemoryParams.device_id` is declared, and `Chunk.device_id` exists on `CircularBuffer` already — it is simply never populated by the `cbList.push()`. So the fix is to populate it and group locally, mirroring #1844's identity-and-repeat scheme and sharing its `SINGLE_DEVICE_KEY` sentinel so the three surfaces cannot disagree.

Two behaviours worth stating up front:

- a second allocation of the same CB **on one device** is a real allocation, not fan-out, and keeps its own row (this is why the repeat index is part of the key)
- a capture with **no** `device_id` cannot distinguish fan-out from a repeat, so its rows are preserved rather than merged on a guess

## Scale on a 32-device report

Op 104 of `test_ttnn_moe_aug26_2217` (`ttnn.matmul`) has **96** CB nodes = 3 distinct CBs x 32 devices, all byte-identical but for `device_id`:

| address | size | devices |
|---|---|---|
| 111616 | 8192 | 32 |
| 119808 | 487424 | 32 |
| 607232 | 917504 | 32 |

## Scope that grew during implementation

Recorded here because neither this issue nor #1897 named it, and it touches a surface #1844 shipped:

- **The multipliers changed cells.** Collapsing put `x N devices` beside the size, but that cell is `text-align: right`, so it right-aligned `8 KiB x 32 devices` as one string and the size stopped lining up with the plain sizes around it (three distinct size right-edges at 1900px, against one). Both multipliers therefore moved to the left-aligned description cell — including the pre-existing `x N cores`, which never appears on this issue's surface (`L1Plots` passes no `numCores`) and so only affects `DeviceOperationsFullRender`. Leaving `cores` behind would have split the two multipliers across different cells.
- **80px moved between two fixed columns.** `$legend-size-column` 300px → 220px and `$legend-tensor-column` 180px → 260px. The pair still totals 480px deliberately: that width places the numeric gutter beside it, which #1897 records as correct today.

### Known consequence

On a **mesh aliased-CB row** — one carrying both multipliers and a resolved tensor identifier — `.legend-multipliers` is `flex: 0 0 auto` and takes 202px of the 260px column, leaving **58px** for the tensor identifier, down from 180px before. The identifier ellipsises much earlier on those rows.

Not addressed here. It needs the multipliers to become shrinkable or the identifier to move, and both are judgement calls about which half matters more. No report in the local corpus produces such a row — the MoE report's CBs are all `globally_allocated=0` — so it is unexercised as well as undecided.

## Reproduce

Open the `multihost_poc_jun24_2321` report, any op with CBs, enable the "Circular buffers" toggle on the L1 plot, and count the legend rows for one address.

aidemsined:
image

## Summary

A mesh op emits the same circular buffer once per device, and the L1 plot's legend listed every copy. Op 104 (`ttnn.matmul`) of `test_ttnn_moe_aug26_2217` has **96** `circular_buffer_allocate` nodes for **3** distinct CBs across **32** devices, so the legend drew 96 unlabelled rows interleaved by address with the fragmentation entries. Each CB is now one row carrying its device count.

Fixing that exposed a second problem: the size cell is `text-align: right`, so appending the count produced `8 KiB x 32 devices` as a single right-aligned string and pushed the size out of line with the plain sizes above and below it. Both multipliers — `x N devices` and the pre-existing `x N cores` — move to the left-aligned description cell.

Closes #1879

### The row repetition

Before — every device's copy got its own unlabelled row, so three CBs filled 96 of them:

```
 0111616    L1 START
 0111616       8 KiB
 0111616       8 KiB
 0111616       8 KiB
                          ... 29 more identical rows
 0119808     476 KiB
 0119808     476 KiB
                          ... 30 more
 0607232     896 KiB
                          ... 31 more
 1524736   40.13 KiB  Empty space
```

After — one row per CB, saying how many devices it stands for:

```
 0111616    L1 START
 0111616       8 KiB  x 32 devices
 0119808     476 KiB  x 32 devices
 0607232     896 KiB  x 32 devices
 1524736   40.13 KiB  Empty space
 1565824       432 B  95 ttnn.to_layout tt_moe.py:687 : Tensor 470
```

### Why the count is not in the size cell

The obvious place for the label is beside the size, but that cell is `text-align: right`, so it right-aligns `8 KiB x 32 devices` as one string and the size no longer lines up with the plain sizes around it. Measured right-edge of the size cell across the first 14 legend rows, same report, 1900px viewport — the misalignment is a pixel effect that a code block cannot honestly show, so these are measurements rather than ASCII:

| label placement | distinct size right-edges |
|---|---|
| in the size cell | **3** — 417 for every buffer row, 451 and 468 for the three CB rows |
| in the description cell (shipped) | **1** — 417 for all 14 |

### Frontend

- `src/functions/cbDeviceSlots.ts` *(new)* — the device key and slot allocator that decide how many devices a row stands for, shared by `processMemoryAllocations` (#1844's legend and pressure modal) and the collapse below, so the three CB surfaces cannot disagree. `cbDeviceKey` normalises before bucketing, so a raw `params.device_id` and an already-normalised field land together.
- `src/functions/collapseCbDeviceRows.ts` *(new)* — collapses one device operation's CBs on top of that shared bookkeeping. A second allocation of the same CB **on one device** is a real allocation rather than fan-out and keeps its own row.
- `src/model/OperationDetails.ts` — the `cbList` push records `device_id`. The field was already declared on `Chunk` and simply never populated, which is why this surface had nothing to collapse on.
- `src/components/operation-details/L1Plots.tsx` — collapse per device operation before flattening into `memoryReportWithCB`; pass `deviceCount` through.
- `src/components/operation-details/MemoryLegendElement.tsx` — multipliers render in the description cell instead of the numeric one. No new prop: `deviceCount` already existed here.
- `src/scss/components/MemoryLegendElement.scss` — the description cell clips rather than spilling into the buffer-type column, and the ellipsis sits on a wrapped `.legend-description-text` rather than the flex container, which cannot pass it to bare text. The multiplier keeps its width and the tensor identifier truncates first: a truncated identifier is recoverable from the row, a truncated count is not.
- `src/scss/definitions/_variables.scss` — 80px moves from `$legend-size-column` to `$legend-tensor-column` on the device-operations view. Both were sized for the old split, where the size column carried `40.13 KiB x 50 cores x 32 devices`; afterwards it was three times wider than a size alone needs and the tensor column was 22px short of the labels it inherited. The pair still totals 480px, so the numeric gutter that view aligns against does not move.

Totals are untouched — `calculateCondensed` is envelope-based, so the duplicate rows never inflated anything.

- `src/components/operation-details/L1Plots.tsx` — the collapse is memoised on `[operationDetails.deviceOperations]`, since it allocates per CB per device operation and this component re-renders on every legend click. The surrounding concatenation and sort stay in the render body: `[...].sort()` mutates in place, so a `useMemo` over it cannot be preserved by the compiler and `react-hooks/preserve-manual-memoization` fails the build.

- **Columns still drift at narrow viewports, before and after this change.** Each legend row is its own `inline-grid`, so alignment only holds while the elastic tracks clear their min-content floors — at 800px the plain sizes already spanned 324–375px on `dev`. That is #1897, which diagnoses it and proposes subgrid or moving the nesting indent out of the aligned area.
- **The plot still draws one rectangle per device per CB.** `cbChartDataByOperation` also flattens `cbList`, so 32 identical rectangles overlap. #1879 scopes to row repetition; whether the fill opacity compounds is unverified.
- **A capture with no `device_id` keeps its rows.** Without it, fan-out and a genuine repeat are indistinguishable, so merging would be a guess. All 2560 CB nodes in the reference report carry the field.
- **A mesh aliased-CB row leaves the tensor identifier 58px**, down from 180px, because `.legend-multipliers` takes 202px of the 260px column and does not shrink. Recorded on #1879; unexercised by any report in the corpus, since the reference report's CBs are all `globally_allocated=0`.

## Test plan

- [ ] Frontend: `pnpm test tests/collapseCbDeviceRows.spec.ts tests/cbDeviceSlots.spec.ts tests/MemoryLegendElement.spec.tsx tests/L1PlotsCbDeviceRows.spec.tsx tests/OperationDetails.globallyAllocatedCB.spec.ts tests/memoryLegendColumnBudget.spec.ts`
- [ ] `tsc`, `eslint`, `stylelint` clean on touched files
- [ ] Manual — open `test_ttnn_moe_aug26_2217` op 104, enable **Show Circular Buffers Details** on the L1 plot: three CB rows, each `x 32 devices`, sizes right-aligned with the buffer rows below
- [ ] Manual — device operations view on any report with CBs: `x N cores` reads in the description column and does not spill into the buffer-type column
- [ ] Manual — a single-device report still shows one row per CB with no device label

## Follow-ups

- #1897 — legend columns don't align across rows/nesting depths, the structural half of the alignment problem

### DIFF
--- a/src/components/operation-details/L1Plots.tsx
+++ b/src/components/operation-details/L1Plots.tsx
@@ -104,26 +104,34 @@ function L1Plots({
     const groupedMemoryReport = getGroupedMemoryReport(BufferType.L1);
     const isLengthyLegend = memoryReport.length > MAX_LEGEND_LENGTH;
 
-    const memoryReportWithCB: FragmentationEntry[] = [
-        ...memoryReport,
-        // Collapsed per device op: a mesh op emits the same CB once per device, which
-        // put 32 identical unlabelled rows in this legend interleaved with the
-        // fragmentation entries. One row now carries the device count. #1879
-        ...operationDetails.deviceOperations
-            .map((op) =>
+    // Only the collapse is memoised, not the concatenation and sort around it. The
+    // collapse is the part that allocates per CB per device operation (a Map, a Set
+    // per slot, two joined identity strings), and this component re-renders on
+    // `selectedAddressAtom` — every legend click — plus the region toggle and the
+    // scroll shades, none of which change CB data.
+    //
+    // The sort stays outside deliberately: `[...].sort()` mutates in place, and a
+    // `useMemo` wrapping it cannot be preserved by the compiler
+    // (`react-hooks/preserve-manual-memoization`), which is an error here. `toSorted`
+    // would fix that but needs an es2023 lib target. #1879
+    const collapsedCbEntries: FragmentationEntry[] = useMemo(
+        () =>
+            operationDetails.deviceOperations.flatMap((op) =>
                 collapseCbDeviceRows(op.cbList).map(
                     (cb) =>
                         ({
                             ...cb,
                             markerType: MarkerType.CB,
                             colorVariance: op.id,
-                            globallyAllocated: cb.globallyAllocated,
-                            deviceCount: cb.deviceCount,
                         }) as FragmentationEntry,
                 ),
-            )
-            .flat(),
-    ].sort((a, b) => a.address - b.address);
+            ),
+        [operationDetails.deviceOperations],
+    );
+
+    const memoryReportWithCB: FragmentationEntry[] = [...memoryReport, ...collapsedCbEntries].sort(
+        (a, b) => a.address - b.address,
+    );
 
     // keeping for now, to make sure nothing breaks
     // const bufferZoomRangeStart = Math.min(...bufferMemory.map((chunk) => chunk.address));
--- a/src/components/operation-details/L1Plots.tsx
+++ b/src/components/operation-details/L1Plots.tsx
@@ -27,6 +27,7 @@ import { FragmentationEntry, MarkerType } from '../../model/APIData';
 import { MemoryLegendGroup } from './MemoryLegendGroup';
 import { useGetL1SmallMarker, useGetL1StartMarker } from '../../hooks/useAPI';
 import useScrollShade from '../../hooks/useScrollShade';
+import { collapseCbDeviceRows } from '../../functions/collapseCbDeviceRows';
 
 interface L1PlotsProps {
     operationDetails: OperationDetails;
@@ -105,15 +106,19 @@ function L1Plots({
 
     const memoryReportWithCB: FragmentationEntry[] = [
         ...memoryReport,
+        // Collapsed per device op: a mesh op emits the same CB once per device, which
+        // put 32 identical unlabelled rows in this legend interleaved with the
+        // fragmentation entries. One row now carries the device count. #1879
         ...operationDetails.deviceOperations
             .map((op) =>
-                op.cbList.map(
+                collapseCbDeviceRows(op.cbList).map(
                     (cb) =>
                         ({
                             ...cb,
                             markerType: MarkerType.CB,
                             colorVariance: op.id,
                             globallyAllocated: cb.globallyAllocated,
+                            deviceCount: cb.deviceCount,
                         }) as FragmentationEntry,
                 ),
             )
@@ -318,6 +323,7 @@ function L1Plots({
                             colorVariance={chunk.colorVariance}
                             userL1ZoomRange={userL1ZoomRange}
                             isGloballyAllocated={chunk.globallyAllocated === true}
+                            deviceCount={chunk.deviceCount}
                         />
                     ))}
 
--- a/src/functions/collapseCbDeviceRows.ts
+++ b/src/functions/collapseCbDeviceRows.ts
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import type { CircularBuffer } from '../model/APIData';
+import { SINGLE_DEVICE_KEY } from './processMemoryAllocations';
+
+/**
+ * @description Normalise a graph node's `device_id` for use as a number.
+ * `undefined` for a capture with no device dimension, or a value that is not a
+ * finite number — the field is absent in older captures and emitted as a string
+ * by at least one other. #1844
+ */
+export const normaliseDeviceId = (deviceId: number | string | undefined): number | undefined => {
+    if (deviceId === undefined || deviceId === null || deviceId === '') {
+        return undefined;
+    }
+    const asNumber = Number(deviceId);
+    return Number.isFinite(asNumber) ? asNumber : undefined;
+};
+
+export interface CollapsedCircularBuffer extends CircularBuffer {
+    /** Devices this row stands for. 1 when the capture carries no device dimension. */
+    deviceCount: number;
+}
+
+/**
+ * @description Collapse one device operation's circular buffers so a mesh op's
+ * fan-out is a single row carrying a device count, instead of the same CB listed
+ * once per device. #1879
+ *
+ * The identity and repeat scheme mirrors `processMemoryAllocations`, which already
+ * does this for the per-DeviceOp legend and the pressure modal (#1844) — the three
+ * surfaces are on screen together and must not disagree about how many devices a CB
+ * spans. Op scoping is implicit here because `cbList` is already per operation.
+ *
+ * A second allocation of the same CB *on one device* is a real allocation rather
+ * than fan-out, so it keeps its own row: keying on the device's repeat count pairs
+ * it with the other devices' repeats instead of letting it collapse onto the first.
+ * The first occurrence in graph order is the row the rest fold onto.
+ */
+export const collapseCbDeviceRows = (cbList: readonly CircularBuffer[]): CollapsedCircularBuffer[] => {
+    const repeatsByDevice = new Map<string, number>();
+    const slotOrder: string[] = [];
+    const bySlot = new Map<string, { circularBuffer: CircularBuffer; deviceKeys: Set<string> }>();
+
+    for (const circularBuffer of cbList) {
+        const identity = [
+            circularBuffer.address,
+            circularBuffer.size,
+            circularBuffer.core_range_set,
+            circularBuffer.globallyAllocated === true,
+        ].join('|');
+        const deviceKey = String(circularBuffer.device_id ?? SINGLE_DEVICE_KEY);
+        const repeatKey = `${identity}|${deviceKey}`;
+        const repeat = (repeatsByDevice.get(repeatKey) ?? 0) + 1;
+        repeatsByDevice.set(repeatKey, repeat);
+
+        const slot = `${identity}|${repeat}`;
+        const existing = bySlot.get(slot);
+        if (existing) {
+            existing.deviceKeys.add(deviceKey);
+        } else {
+            bySlot.set(slot, { circularBuffer, deviceKeys: new Set([deviceKey]) });
+            slotOrder.push(slot);
+        }
+    }
+
+    return slotOrder.map((slot) => {
+        const entry = bySlot.get(slot);
+        return { ...(entry?.circularBuffer as CircularBuffer), deviceCount: entry?.deviceKeys.size ?? 1 };
+    });
+};
--- a/src/functions/collapseCbDeviceRows.ts
+++ b/src/functions/collapseCbDeviceRows.ts
@@ -3,21 +3,9 @@
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
 import type { CircularBuffer } from '../model/APIData';
-import { SINGLE_DEVICE_KEY } from './processMemoryAllocations';
+import { cbDeviceKey, createCbSlotKeys } from './cbDeviceSlots';
 
-/**
- * @description Normalise a graph node's `device_id` for use as a number.
- * `undefined` for a capture with no device dimension, or a value that is not a
- * finite number — the field is absent in older captures and emitted as a string
- * by at least one other. #1844
- */
-export const normaliseDeviceId = (deviceId: number | string | undefined): number | undefined => {
-    if (deviceId === undefined || deviceId === null || deviceId === '') {
-        return undefined;
-    }
-    const asNumber = Number(deviceId);
-    return Number.isFinite(asNumber) ? asNumber : undefined;
-};
+export { normaliseDeviceId } from './cbDeviceSlots';
 
 export interface CollapsedCircularBuffer extends CircularBuffer {
     /** Devices this row stands for. 1 when the capture carries no device dimension. */
@@ -40,7 +28,7 @@ export interface CollapsedCircularBuffer extends CircularBuffer {
  * The first occurrence in graph order is the row the rest fold onto.
  */
 export const collapseCbDeviceRows = (cbList: readonly CircularBuffer[]): CollapsedCircularBuffer[] => {
-    const repeatsByDevice = new Map<string, number>();
+    const slotKeyOf = createCbSlotKeys();
     const slotOrder: string[] = [];
     const bySlot = new Map<string, { circularBuffer: CircularBuffer; deviceKeys: Set<string> }>();
 
@@ -51,12 +39,8 @@ export const collapseCbDeviceRows = (cbList: readonly CircularBuffer[]): Collaps
             circularBuffer.core_range_set,
             circularBuffer.globallyAllocated === true,
         ].join('|');
-        const deviceKey = String(circularBuffer.device_id ?? SINGLE_DEVICE_KEY);
-        const repeatKey = `${identity}|${deviceKey}`;
-        const repeat = (repeatsByDevice.get(repeatKey) ?? 0) + 1;
-        repeatsByDevice.set(repeatKey, repeat);
-
-        const slot = `${identity}|${repeat}`;
+        const deviceKey = cbDeviceKey(circularBuffer.device_id);
+        const slot = slotKeyOf(identity, deviceKey);
         const existing = bySlot.get(slot);
         if (existing) {
             existing.deviceKeys.add(deviceKey);
--- a/src/model/OperationDetails.ts
+++ b/src/model/OperationDetails.ts
@@ -24,6 +24,7 @@ import getChartData from '../functions/getChartData';
 import { calculateCondensed } from '../functions/calculateCondensed';
 import { L1_DEFAULT_MEMORY_SIZE, L1_NUM_CORES } from '../definitions/L1MemorySize';
 import { TensorDeallocationReport } from './BufferSummary';
+import { normaliseDeviceId } from '../functions/collapseCbDeviceRows';
 
 export interface OperationDetailsOptions {
     renderPattern: boolean;
@@ -262,6 +263,11 @@ export class OperationDetails implements Partial<OperationDetailsData> {
                                 num_cores: getCoresInRange(node.params.core_range_set),
                                 colorVariance: deviceOp.id,
                                 globallyAllocated,
+                                // Normalised here rather than read as a number: absent in
+                                // older captures and emitted as a string by at least one
+                                // other. Without it every device's copy of a CB became its
+                                // own legend row. #1844 / #1879
+                                device_id: normaliseDeviceId(node.params.device_id),
                             });
                         }
                     }
--- a/src/scss/components/MemoryLegendElement.scss
+++ b/src/scss/components/MemoryLegendElement.scss
@@ -65,6 +65,29 @@
         white-space: nowrap;
     }
 
+    // `min-width: 0` and the clip keep this cell's contents inside its own track:
+    // the device-operations view caps it at `$legend-tensor-column`, and without
+    // these the multipliers below would spill into the buffer-type column rather
+    // than truncate. Matches what #1897 asks of these labels. #1879
+    .legend-description {
+        display: flex;
+        min-width: 0;
+        align-items: center;
+        overflow: hidden;
+        white-space: nowrap;
+        text-overflow: ellipsis;
+    }
+
+    // Sits in the left-aligned description cell, ahead of any tensor label. Dimmed
+    // because it qualifies the size rather than naming the row, and never shrinks —
+    // if the cell runs out of room the tensor identifier truncates instead, since
+    // the multiplier is the shorter and less recoverable of the two. #1879
+    .legend-multipliers {
+        flex: 0 0 auto;
+        margin-right: 8px;
+        color: $tt-grey-5;
+    }
+
     .legend-marker-swatch {
         width: 15px;
         height: 15px;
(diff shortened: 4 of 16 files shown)


dcblundell (commented):
Combined multi-lens review (DRY / TEST / PERF / STANDARDS / SECURITY / REQUIREMENTS). Core of #1879 is correctly implemented and verified against the report the issue cites (op 104: 96 CB nodes -> 3 rows). SECURITY found nothing in scope. Posting the Must fix and Should fix items only; Consider/nit items are in the chat summary.

src/model/OperationDetails.ts
```diff
@@ -262,6 +263,11 @@ export class OperationDetails implements Partial<OperationDetailsData> {
                                 num_cores: getCoresInRange(node.params.core_range_set),
                                 colorVariance: deviceOp.id,
                                 globallyAllocated,
+                                // Normalised here rather than read as a number: absent in
+                                // older captures and emitted as a string by at least one
+                                // other. Without it every device's copy of a CB became its
+                                // own legend row. #1844 / #1879
+                                device_id: normaliseDeviceId(node.params.device_id),
```
dcblundell, line 270: **Must fix:** the `device_id` plumbing has no test

I removed this property and ran `collapseCbDeviceRows`, `OperationDetails.globallyAllocatedCB`, `DeviceOperationsFullRender` and `MemoryLegendElement`: 49/49 still green, feature dead. `collapseCbDeviceRows` falls into its own "no device id -> keep the rows" branch, so the regression is silent. Lint only catches a whole-line deletion via the unused import, not a renamed param key or a wrong value.

`tests/OperationDetails.globallyAllocatedCB.spec.ts` already builds `cbList` from graph nodes, so this is ~10 lines: 8 allocate nodes at `device_id: 0..7`, assert `cbList.map(cb => cb.device_id)`, then `collapseCbDeviceRows(cbList)` has length 1 with `deviceCount === 8`.

src/components/operation-details/L1Plots.tsx
```diff
@@ -105,15 +106,19 @@ function L1Plots({
 
     const memoryReportWithCB: FragmentationEntry[] = [
         ...memoryReport,
+        // Collapsed per device op: a mesh op emits the same CB once per device, which
+        // put 32 identical unlabelled rows in this legend interleaved with the
+        // fragmentation entries. One row now carries the device count. #1879
         ...operationDetails.deviceOperations
             .map((op) =>
-                op.cbList.map(
+                collapseCbDeviceRows(op.cbList).map(
```
dcblundell, line 120: **Should fix:** no test covers the user-visible outcome

Coverage stops at the pure function and at the leaf renderer. This seam - `collapseCbDeviceRows` being called, `deviceCount` surviving the `as FragmentationEntry` cast, and reaching `deviceCount={chunk.deviceCount}` - is untested. `tests/DeviceOperationsFullRender.spec.tsx` already does exactly this integration test for the sibling #1844 path, and `tests/mocks/plotComponent.ts` exists for stubbing `MemoryPlotRenderer`.

src/functions/collapseCbDeviceRows.ts
```diff
@@ -0,0 +1,73 @@
...
+ * does this for the per-DeviceOp legend and the pressure modal (#1844) — the three
+ * surfaces are on screen together and must not disagree about how many devices a CB
+ * spans. Op scoping is implicit here because `cbList` is already per operation.
+ *
+ * A second allocation of the same CB *on one device* is a real allocation rather
+ * than fan-out, so it keeps its own row: keying on the device's repeat count pairs
+ * it with the other devices' repeats instead of letting it collapse onto the first.
+ * The first occurrence in graph order is the row the rest fold onto.
+ */
+export const collapseCbDeviceRows = (cbList: readonly CircularBuffer[]): CollapsedCircularBuffer[] => {
```
dcblundell, line 30: **Should fix:** the collapse scheme is duplicated, so the "must not disagree" invariant is only aspirational

Identity, device key, repeat counter, slot fold and Set-size count are all reimplemented from `processMemoryAllocations`. Only `SINGLE_DEVICE_KEY` is shared - the sentinel, not the logic that determines the count. They already diverge: this module keys on the *normalised* id, the sibling on raw `node.params.device_id`. For absent/numeric/numeric-string ids they agree, but mixed unparseable values (`''` alongside `'abc'`) give 2 rows of 1 device here and 1 row of 2 devices in the modal.

Contrived on real captures, so the divergence itself is minor - the duplication is the risk. Extracting just the slot derivation (a `cbDeviceKey()` plus a `createCbSlotKeys()` closure) would make the docstring structurally true and put both existing suites over the shared code.

src/scss/components/MemoryLegendElement.scss
```diff
@@ -65,6 +65,29 @@
...
+    // the device-operations view caps it at `$legend-tensor-column`, and without
+    // these the multipliers below would spill into the buffer-type column rather
+    // than truncate. Matches what #1897 asks of these labels. #1879
+    .legend-description {
+        display: flex;
+        min-width: 0;
+        align-items: center;
+        overflow: hidden;
+        white-space: nowrap;
+        text-overflow: ellipsis;
```
dcblundell, line 86: **Should fix:** `text-overflow: ellipsis` is inert here

`.legend-description` is `display: flex` and the tensor text is an unwrapped fragment child (`MemoryLegendElement.tsx:212`), so it becomes an anonymous flex item and never receives `text-overflow`. The result is a hard mid-glyph clip, which contradicts this rule's own comment and the `.legend-multipliers` comment below ("the tensor identifier truncates instead").

Either wrap the tensor text in its own span carrying `min-width: 0; overflow: hidden; text-overflow: ellipsis`, or drop the declaration and reword both comments to say "clips".

src/components/operation-details/L1Plots.tsx
```diff
@@ -105,15 +106,19 @@ function L1Plots({
 
     const memoryReportWithCB: FragmentationEntry[] = [
```
dcblundell: **Should fix:** `memoryReportWithCB` is rebuilt on every render

No `useMemo` here, and this change moves per-CB work into the render body: a Map, a Set per slot, and two joined identity strings per CB per device operation. `L1Plots` re-renders on `selectedAddressAtom` (every legend click), the region toggle and the scroll shades - none of which change CB data.

Net the branch is still a large win (32x fewer legend subtrees), so this is insurance rather than a regression. Memoising on `[memory, fragmentation, operationDetails.deviceOperations]` covers it, since the collapse is pure over `cbList`.

tests/opGraphNodeStyles.spec.ts
```diff
@@ -136,3 +136,39 @@ describe('op graph critical path styling', () => {
...
     });
 });
+
+// The two device-operation legend tracks are a fixed pair, and #1879 moved the
+// `x N cores x N devices` labels from the first into the second. Sizing them for
+// the old split clipped the labels at 180px, and widening the second without
+// narrowing the first would shift the numeric gutter this view aligns against —
+// which #1897 records as correct today. Both halves are asserted here because
+// geometry cannot be measured in jsdom.
+describe('device operation legend column budget', () => {
```
dcblundell: **Should fix:** the legend column-budget tests are in the wrong spec

This file's premise is the module-level `STYLESHEET` const reading `OperationGraphReactFlow.scss`. This block reads `_variables.scss` and asserts on the device-operation memory legend, which has no relationship to op-graph node styling - whoever next re-tunes those variables will hit a CI failure in a file with no apparent connection. Worth its own spec next to the legend tests.

Separately, `expect(size + tensor).toBe(480)` hard-codes a total that exists only in a comment and in this assertion, so a deliberate re-tune fails with a message that reads like a bug.

src/scss/definitions/_variables.scss
```diff
@@ -40,11 +40,16 @@ $active-transfer-width: 380px;
...
+//
+// The size column held `40.13 KiB x 50 cores x 32 devices` until the multipliers
+// moved to the tensor column (#1879), which left it three times wider than the
+// size alone needs and the tensor column 22px short of the labels it inherited.
+// 80px moves across; the pair still totals 480px, so the numeric gutter this view
+// aligns against does not shift. #1897 tracks the wider alignment problem.
+$legend-size-column: 220px;
 $legend-address-column: 90px;
-$legend-tensor-column: 180px;
+$legend-tensor-column: 260px;
```
dcblundell, line 52: **Should fix (a decision, not a defect):** record the scope that moved beyond #1879

#1879 scopes to "row repetition only", but this PR also moves the pre-existing `x N cores` multiplier and shifts 80px between the two columns. That label never appears on the #1879 surface - `L1Plots` passes no `numCores`, so only `DeviceOperationsFullRender` renders it - so this changes a shipped #1844 surface the issue does not name.

The alignment argument is sound, and leaving `cores` behind would split the two multipliers across cells, so my recommendation is keep the code and amend the issue: neither #1879 nor #1897 currently records that the multipliers changed cells or that 80px moved. One consequence worth capturing - on a mesh aliased-CB row `.legend-multipliers` is `flex: 0 0 auto` and takes ~202px of the 260px, leaving ~58px for the tensor identifier, down from ~180px.

src/model/OperationDetails.ts
aidemsined: > [!CAUTION]
> **Confirmed and fixed in 9f1c4460.** I reproduced your experiment before fixing it: deleting the `device_id` line left 41/41 green. The same deletion now fails three tests.

Added to `tests/OperationDetails.globallyAllocatedCB.spec.ts` as you suggested — eight allocate nodes at `device_id: 0..7`, asserting `cbList.map(cb => cb.device_id)`, and a second case taking it to the end of the chain: `collapseCbDeviceRows` returning one row with `deviceCount === 8`. Plus a string-id case (`'3'` → `3`) and an absent-id case, since those are the two shapes the normaliser exists for.

This is the gap I keep leaving — coverage either side of a seam and nothing on the seam. The pure function and the leaf renderer were both tested, which is exactly what made it feel covered.

src/components/operation-details/L1Plots.tsx
aidemsined: > [!NOTE]
> Fixed in 9f1c4460 — `tests/L1PlotsCbDeviceRows.spec.tsx`, using the `plotComponent` mock you pointed at.

Three cases: a 32-device CB rendering one `.legend-multipliers` reading `x 32 devices`, no CB rows at all with the toggle off, and the size cell staying free of `devices` (the alignment half).

Checked it catches a bypass: replacing `collapseCbDeviceRows(op.cbList)` with `op.cbList` fails the first test.

One thing that fell out — dropping `deviceCount: cb.deviceCount` from the cast changed nothing, because `...cb` already carries it. Same for `globallyAllocated`. Both were dead, so they are gone rather than covered by a test asserting a no-op.

src/functions/collapseCbDeviceRows.ts
aidemsined: > [!NOTE]
> Fixed in 9f1c4460 — `src/functions/cbDeviceSlots.ts` now owns `SINGLE_DEVICE_KEY`, `normaliseDeviceId`, `cbDeviceKey` and `createCbSlotKeys`, and both consumers use it.

You were right that the docstring claimed an invariant the code did not enforce — sharing only the sentinel while reimplementing the logic that produces the count.

On the divergence: rather than pick one side's convention I made `cbDeviceKey` normalise internally, so a raw `params.device_id` and an already-normalised field land in the same bucket. That closes your `''` / `'abc'` case in both directions — both now bucket to `single`. It does change `processMemoryAllocations` for unparseable ids (previously `'abc'` was its own bucket); no real capture has them, and the full suite is green, so #1844's behaviour is otherwise untouched.

`tests/cbDeviceSlots.spec.ts` covers the shared code directly, including that `cbDeviceKey` is idempotent and that the nth repeat pairs across devices rather than displacing the first.

src/scss/components/MemoryLegendElement.scss
aidemsined: > [!NOTE]
> Fixed in 9f1c4460 — took the first option. The tensor text is wrapped in `.legend-description-text`, which carries `min-width: 0; overflow: hidden; text-overflow: ellipsis`.

Your diagnosis was exact: bare text inside a flex container is an anonymous flex item and cannot receive `text-overflow`, so the declaration on the container did nothing and the result was a hard mid-glyph clip — contradicting both comments. The container keeps `overflow: hidden` (it still has to stop the multipliers spilling into the buffer-type column); the ellipsis moved to the element that can actually use it. `emptyChunkLabel` is wrapped the same way.

src/components/operation-details/L1Plots.tsx
aidemsined: > [!WARNING]
> Memoised in 9f1c4460, but not where you suggested, and the reason is worth recording.

Wrapping the whole `memoryReportWithCB` expression cannot be done here: `[...].sort()` mutates in place, so the compiler refuses to preserve the memo and `react-hooks/preserve-manual-memoization` fails the build — "Compilation Skipped: Existing memoization could not be preserved". I tried your dependency list, then `[memory, fragmentation]` with `memoryReport` memoised too, then depending on `operationDetails` rather than its field; all still tripped it. `toSorted` fixes it cleanly but needs an es2023 lib target, which is not mine to change in this PR.

So the memo wraps just the collapse — `collapsedCbEntries`, keyed on `[operationDetails.deviceOperations]` — and the concatenation and sort stay in the render body where they already were. That covers the cost you actually named (a Map, a Set per slot and two joined identity strings per CB per device operation); what is left unmemoised is a spread and a sort over the same array count as before this branch.

The constraint is recorded in a comment above it so the next person does not repeat the three attempts.

tests/opGraphNodeStyles.spec.ts
aidemsined: > [!NOTE]
> Fixed in 9f1c4460 — moved to `tests/memoryLegendColumnBudget.spec.ts`.

You are right that it had no business in a spec whose premise is a module-level `STYLESHEET` const reading `OperationGraphReactFlow.scss`.

On the hard-coded total: kept, but it is now a named constant with the reasoning attached and a failure message that names the constraint — "these two columns place the numeric gutter beside them, trade width between them rather than growing the pair". The widest-observed multiplier is a named constant too. A deliberate re-tune now reads as a constraint to update rather than a bug to chase.

src/scss/definitions/_variables.scss
aidemsined: > [!NOTE]
> Took your recommendation — code kept, #1879 amended. Your 58px figure checks out exactly.

Added a **Scope that grew during implementation** section to #1879 recording both parts: that the multipliers changed cells (and why `cores` had to come along — leaving it behind would split the two across different cells), and that 80px moved between the two fixed columns with the pair total held at 480px so the gutter does not shift.

Verified the consequence rather than repeating it: multiplier 194px + 8px gap = 202px of the 260px column, leaving **58px** for the tensor identifier against 180px before. Recorded as a known consequence, explicitly not addressed — fixing it means either making the multipliers shrinkable or moving the identifier, and that is a judgement call about which half matters more rather than something to settle inside this PR.

Also noted there that it is unexercised: no report in the local corpus produces a mesh aliased-CB row, since the MoE report's CBs are all `globally_allocated=0`. So the regression is real but currently unobservable, which is worth knowing before someone picks it up.

src/scss/components/MemoryLegendElement.scss
```diff
@@ -65,6 +65,29 @@
...
+    }
+
+    // Sits in the left-aligned description cell, ahead of any tensor label. Dimmed
+    // because it qualifies the size rather than naming the row, and never shrinks —
+    // if the cell runs out of room the tensor identifier truncates instead, since
+    // the multiplier is the shorter and less recoverable of the two. #1879
+    .legend-multipliers {
+        flex: 0 0 auto;
+        margin-right: 8px;
+        color: $tt-grey-5;
```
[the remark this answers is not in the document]
aidemsined: > [!NOTE]
> Already addressed before this review landed — `11326aed`, one commit after the one Copilot read.

The 180px cap is gone: `$legend-tensor-column` is 260px now, with 80px traded from `$legend-size-column` so the pair still totals 480px and the numeric gutter does not move.

Measured on op 104 of the report the issue cites, at 1900px:

- before: description cell 180px against 202px of content, `scrollWidth > clientWidth`, clipped mid-word
- after: **260px against 260px, not clipped**

The `flex: 0 0 auto` point was right on the earlier commit — the multiplier was being clipped rather than protected. Separately, `text-overflow: ellipsis` on that container turned out to be inert for a different reason (bare text in a flex parent is an anonymous flex item), fixed in `9f1c4460`, so the tensor identifier now genuinely ellipsises rather than hard-clipping.

dcblundell (approved):

PR #1959, opened 2026-08-31
Say that report_name is required for a report to be written

aidemsined:
## Summary

`docs/src/installing.md` describes when a memory report is created automatically and names two conditions, `enable_logging` and `enable_graph_report`, leaving out `report_name`. Setting exactly what that sentence lists produces **no report and no error** — not an empty folder, nothing.

All three conditions that gate report generation fail the same way — `enable_logging` at tt-metal's root `conftest.py:1271`, the `enable_graph_report`/`enable_comparison_mode` pair at `:1277`, and `report_name` at `graph_report.py:82-85` — each a bare `yield; return` with no log. `report_name` is the one this sentence omitted.

The config table above already said `report_name` "must have a value", but the two statements read independently, and this sentence is the one people skim when asking why they have no report.

Refs tenstorrent/tt-metal#54956 — surfaced while reviewing that fix, which corrects the equivalent trap in tt-metal's own `ttnn/tutorials/ttnn_visualizer.md`. **Our docs were already correct about `enable_graph_report`** (`true`, marked "Must be enabled"); this is the other half of the same failure, on our side.

### Docs

- `docs/src/installing.md` — the automatic-creation sentence now lists all three conditions (and notes that `enable_comparison_mode` also triggers generation, so the list is exhaustive rather than true-given-the-recommended-block), followed by a caveat that every one of them fails silently and that `report_name` should always be set explicitly. Written as plain prose to match the existing caveat style (`remote-sync.md:38`) rather than a MyST admonition: `myst_parser` is enabled so a directive would build, but it renders as a code block on GitHub, where these pages are also read. The paragraph deliberately describes no tt-metal internals, so there is no cross-repo claim here to rot.
- Same file — fixes a stray character in the `report_name` table row, which read "Must have a value for data8 to be output to disk".

## Verification

Claims about tt-metal were checked against its source rather than taken from the linked PR:

- all three gates bail with a bare `yield; return` and no log (root `conftest.py:1271`, `:1277`; `graph_report.py:82-85`)
- `report_path.mkdir()` sits *after* the gate (`graph_report.py:110`), which is why the doc can say "not even an empty folder" rather than hedging
- generation triggers on `enable_graph_report` **or** `enable_comparison_mode` (root `conftest.py:1277`)

## Test plan

- [ ] `pnpm run build:docs` succeeds
- [ ] Rendered `docs/output/src/installing.html` contains the amended condition and the caveat, and no longer contains `data8`
- [ ] The caveat renders as prose on GitHub as well as in the Sphinx build

### DIFF
--- a/docs/src/installing.md
+++ b/docs/src/installing.md
@@ -27,7 +27,7 @@ export TTNN_CONFIG_OVERRIDES='{
 |----------------------|-------------|
 | **enable_fast_runtime_mode** | Disable fast runtime mode to ensure all operations are properly traced. **Must be disabled to enable logging**. |
 | **enable_logging** | Synchronizes main thread after every operation and logs the operation. **Must be enabled**. |
-| **report_name** | Prefix of the folder name where the memory report is output. **Must have a value for data8 to be output to disk**. |
+| **report_name** | Prefix of the folder name where the memory report is output. **Must have a value for data to be output to disk**. |
 | **enable_detailed_buffer_report** | Enable to visualize the detailed buffer report after every operation. **Needed for full buffer information**. |
 | **enable_detailed_tensor_report** | Enable to visualize the values of input and output tensors of every operation. **Data not used by the visualizer**. |
 | **enable_graph_report** | Generates an SVG visualization of the computation graph and enables automatic report generation with pytest. **Must be enabled**. |
@@ -40,7 +40,9 @@ To run a test with custom input data, you can use the following command with sui
 pytest --disable-warnings --input-path="path/to/input.json" path/to/test_file.py::test_function[param]
 ```
 
-The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report will be created automatically when running TT-Metal model demos and tests with `pytest`, when `enable_logging` and `enable_graph_report` are `true`.
+The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report will be created automatically when running TT-Metal model demos and tests with `pytest`, when `enable_logging` and `enable_graph_report` are `true` **and `report_name` has a value**.
+
+Important: `report_name` is the option that fails silently. With it unset you get no report and no error — not an empty folder, nothing at all. Only tests collected under `tests/ttnn/` derive a name from the test id for you (in `tests/ttnn/conftest.py`); model demos under `models/` do not, so set `report_name` explicitly for those.
 
 <img width="909" alt="Memory report files" src="https://github.com/user-attachments/assets/ab31892a-2779-4fe1-9ad5-0f35f8329f9a" />
 
--- a/docs/src/installing.md
+++ b/docs/src/installing.md
@@ -40,9 +40,9 @@ To run a test with custom input data, you can use the following command with sui
 pytest --disable-warnings --input-path="path/to/input.json" path/to/test_file.py::test_function[param]
 ```
 
-The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report will be created automatically when running TT-Metal model demos and tests with `pytest`, when `enable_logging` and `enable_graph_report` are `true` **and `report_name` has a value**.
+The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report is created automatically when running TT-Metal model demos and tests with `pytest`, provided `enable_logging` is `true`, `report_name` has a value, and `enable_graph_report` is `true`. (`enable_comparison_mode` also triggers generation, but it is not a substitute — see the table above.)
 
-Important: `report_name` is the option that fails silently. With it unset you get no report and no error — not an empty folder, nothing at all. Only tests collected under `tests/ttnn/` derive a name from the test id for you (in `tests/ttnn/conftest.py`); model demos under `models/` do not, so set `report_name` explicitly for those.
+Important: every one of those conditions fails silently. With any of them unset you get no report and no error, and not even an empty folder. `report_name` is the one most often missed — always set it explicitly, for tests as well as model demos.
 
 <img width="909" alt="Memory report files" src="https://github.com/user-attachments/assets/ab31892a-2779-4fe1-9ad5-0f35f8329f9a" />
 


dcblundell (commented):
Conductor review (six lenses). Branch diff is docs-only; the substantive verification was done against tt-metal source. `report_name` **is** required — confirmed at the source of truth, and the silent-failure claim is precise. Two accuracy issues in the new prose, one of which inverts the guidance for `tests/ttnn/`.

docs/src/installing.md
```diff
@@ -40,7 +40,9 @@ To run a test with custom input data, you can use the following command with sui
 pytest --disable-warnings --input-path="path/to/input.json" path/to/test_file.py::test_function[param]
 ```
 
-The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report will be created automatically when running TT-Metal model demos and tests with `pytest`, when `enable_logging` and `enable_graph_report` are `true`.
+The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report will be created automatically when running TT-Metal model demos and tests with `pytest`, when `enable_logging` and `enable_graph_report` are `true` **and `report_name` has a value**.
+
+Important: `report_name` is the option that fails silently. With it unset you get no report and no error — not an empty folder, nothing at all. Only tests collected under `tests/ttnn/` derive a name from the test id for you (in `tests/ttnn/conftest.py`); model demos under `models/` do not, so set `report_name` explicitly for those.
```
dcblundell: **Should fix:** the `tests/ttnn/` carve-out is inverted — the derived name arrives too late to matter.

The autouse fixture that decides whether a report is written is `ttnn_graph_report` in tt-metal's **root** `conftest.py:1246`, and it reads `report_name` at setup (`ttnn/ttnn/graph_report.py:82-85`). `tests/ttnn/conftest.py:66` sets the derived name in a *subdirectory* conftest. pytest sets up root-conftest autouse fixtures before subdirectory ones, so the gate has already bailed by the time the name is assigned. I confirmed the ordering with a standalone two-conftest repro: the root fixture reads `None`, then the subdir fixture sets the value.

So `report_name` must be set explicitly under `tests/ttnn/` too — this sentence currently tells readers the one place they can skip it is a place they cannot. Suggest dropping the carve-out entirely and saying `report_name` is always required. (Traced from source and proven for the ordering rule; not run on hardware, so worth a confirming run.)

dcblundell: **Should fix:** "**the** option that fails silently" over-claims — all three conditions do.

tt-metal root `conftest.py:1257` returns silently when `enable_logging` is false, and `:1262-1265` returns silently when neither `enable_graph_report` nor `enable_comparison_mode` is set — the same `yield; return` shape as the `report_name` gate, with no log in any of them. `report_name` is just the one this sentence used to omit.

A reader with `enable_logging` off is actively misdirected by the definite article. "`report_name` fails silently: with it unset you get no report and no error" loses nothing the PR set out to say.

docs/src/installing.md
```diff
@@ -40,7 +40,9 @@ To run a test with custom input data, you can use the following command with sui
 pytest --disable-warnings --input-path="path/to/input.json" path/to/test_file.py::test_function[param]
 ```
 
-The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report will be created automatically when running TT-Metal model demos and tests with `pytest`, when `enable_logging` and `enable_graph_report` are `true`.
+The final output should be a folder within `${TT_METAL_HOME}/generated/ttnn/reports/` which should include at least a `db.sqlite` file (config.json is optional for the visualizer). The report will be created automatically when running TT-Metal model demos and tests with `pytest`, when `enable_logging` and `enable_graph_report` are `true` **and `report_name` has a value**.
```
dcblundell: **Consider / nit** (bundled — all minor, none blocking):

1. **Unqualified tt-metal paths** (line 45): `tests/ttnn/`, `tests/ttnn/conftest.py` and `models/` are tt-metal-relative but stated without naming the repo, in a doc whose own tree has a `tests/`. The `${TT_METAL_HOME}` sentence makes it inferable, but the parenthetical is the caveat's one actionable pointer — "in tt-metal's `tests/ttnn/conftest.py`" would fix it.
2. **The claim can't be kept honest from here** (line 45): it asserts the contents of a file in another repository, which will rot unnoticed. An "at time of writing" hedge, or naming the tt-metal commit it was checked against, gives it an expiry.
3. **The condition list is still not exhaustive** (line 43): root `conftest.py:1262-1265` triggers report generation on `enable_graph_report` **or** `enable_comparison_mode`, so `enable_comparison_mode: true` alone also produces a report. Not false as written, since the recommended block sets it `false` — but since this PR exists to make this sentence's condition list trustworthy, worth deciding now rather than reopening the line later. My suggestion: leave it and record the reasoning here.
4. **`report_name` is now required in three places within fifteen lines** (lines 30, 43, 45). That looks deliberate — the point is that people skim the sentence, not the table — but the bold clause at line 43 is the weakest of the three now the paragraph exists.
5. **Follow-up worth filing against tt-metal:** the silent gate is `ttnn/ttnn/graph_report.py:82`, and `report_path` is *derived* from `report_name` (`ttnn/core/config.cpp:51-109` returns `nullopt` without it; `ttnn/api/ttnn/config.hpp:42` defaults it to `nullopt`). A `log_warning` there is the durable fix; prose in two repos is the workaround.

Confirmed while checking the PR's central claim, which holds exactly as stated: `report_path.mkdir()` sits *after* the gate (`graph_report.py:110`), so there is genuinely no folder at all — "not an empty folder, nothing" is precise.

docs/src/installing.md
aidemsined: > [!CAUTION]
> **You are right, and this was worse than the omission it replaced.** Fixed in b99e9c20 — the carve-out is gone and `report_name` is now stated as always required.

I reproduced the ordering independently rather than take it on the description, with a two-conftest case standing in for `ttnn_graph_report` and `pre_and_post`:

```
ROOT gate setup      -> reads name = None
SUBDIR fixture setup -> sets name
at test body: name = 'derived-from-node-id'
```

Root-conftest autouse sets up first and reads `None`; the derived name lands after the gate has already returned. Confirmed the two halves in tt-metal too — `tests/ttnn/conftest.py:57-68` is an autouse `pre_and_post` that only wraps `manage_config` and yields, and the gate it needed to beat is `graph_report.py:82-85`.

Worth recording why I got this wrong: I verified that the derivation *exists* and stopped there, without checking that it takes effect before the thing that reads it. Half a verification presented as a whole one, and it put actively misleading guidance into the docs — a straight omission would have been safer.

Dropping the carve-out also resolves your first two nits for free: no unqualified tt-metal paths remain, and there is no longer a cross-repo claim in our docs to rot unnoticed.

aidemsined: > [!NOTE]
> Agreed — fixed in b99e9c20. It now reads "every one of those conditions fails silently".

Verified all three take the same `yield; return` with no log: `enable_logging` at root `conftest.py:1271`, the `enable_graph_report`/`enable_comparison_mode` pair at `:1277`, and `report_name` at `graph_report.py:82-85`.

Your point about the definite article actively misdirecting is the part that convinced me. Someone whose `enable_logging` is off reads "`report_name` is *the* option that fails silently", concludes their problem is elsewhere, and the sentence has cost them time rather than saved it.

docs/src/installing.md
aidemsined: > [!NOTE]
> Four of the five are addressed in b99e9c20; the fifth is a tt-metal follow-up I will file. Taking them in order.

**1. Unqualified tt-metal paths** — resolved by removing the carve-out. No tt-metal-relative path remains in the paragraph.

**2. The claim cannot be kept honest from here** — same. Rather than hedge a cross-repo assertion with "at time of writing", the assertion is gone: the docs now state the requirement (`report_name` is always needed) without describing another repository's internals. That is the version that cannot rot.

**3. The condition list is still not exhaustive** — I went the other way from your suggestion and made it exhaustive, since the sentence being trustworthy is the whole reason this PR exists. It now reads: `enable_logging` is `true`, `report_name` has a value, and `enable_graph_report` is `true`, with a parenthetical that `enable_comparison_mode` also triggers generation but is not a substitute. One clause, and the sentence is now true rather than true-given-the-recommended-block.

**4. `report_name` in three places within fifteen lines** — the bold clause is dropped. Restructuring the sentence to list all three conditions plainly made it redundant, so it is now table + paragraph rather than table + bold + paragraph.

**5. `log_warning` at `graph_report.py:82`** — agreed that prose in two repos is the workaround and the warning is the fix. Note dgomezTT's tenstorrent/tt-metal#54956 adds exactly that for the fixture path, warning once per process when `report_name` is unset. It does not cover the other two gates you found at root `conftest.py:1271` and `:1277`, which are still silent, nor the non-pytest path. I will file that against tt-metal and link it back here.

Thanks for confirming the central claim at the source — `report_path.mkdir()` sitting after the gate is what let the doc say "not even an empty folder" rather than hedging it.

smountenay-tt (approved):

PR #1961, opened 2026-09-01
Record report load usage signals

### LINKED ISSUE #1831: Usage signals: record report_loaded and report_load_failed

dcblundell:
## Summary
- record successful report loads with bounded report kind and source values across local, remote, NPE, and MLIR flows
- classify report load failures without recording response bodies or stack traces
- preserve exact-once semantics for user-initiated NPE and MLIR loads while excluding boot restoration
- add regression coverage for success, failure, cancellation, stale attempts, and multi-file cardinality

## Testing
- `pnpm lint`
- `pnpm lint:ts`
- `pnpm test` (211 files, 2,411 tests)

Closes #1831

### DIFF
--- a/src/components/mlir/MlirFileResultsOverlay.tsx
+++ b/src/components/mlir/MlirFileResultsOverlay.tsx
@@ -29,6 +29,12 @@ import getResponseError from '../../functions/getResponseError';
 import { getActiveMlirServer } from '../../functions/mlirServer';
 import { MlirFileResult, MlirLoadedReport } from '../../model/MLIRJsonModel';
 import mapConvertedMlirServerResult from '../../functions/mapConvertedMlirServerResult';
+import { ReportKind, ReportLoadFailureReason, ReportSource } from '../../definitions/UsageEvent';
+import {
+    getReportLoadFailureReason,
+    recordReportLoadFailed,
+    recordReportLoaded,
+} from '../../functions/reportLoadUsage';
 import 'styles/components/MlirFileResultsOverlay.scss';
 
 const MAX_MLIR_FILE_SELECTION = 2;
@@ -173,11 +179,13 @@ const MlirFileResultsOverlay = () => {
                 return;
             }
 
+            const mappedResult = mapConvertedMlirServerResult(retried, result.host ?? null);
+            if (mappedResult.status === ConnectionTestStates.FAILED) {
+                recordReportLoadFailed(ReportKind.MLIR, ReportLoadFailureReason.OTHER);
+            }
             setResults(
                 (current) =>
-                    current?.map((entry, entryIndex) =>
-                        entryIndex === index ? mapConvertedMlirServerResult(retried, entry.host ?? null) : entry,
-                    ) ?? current,
+                    current?.map((entry, entryIndex) => (entryIndex === index ? mappedResult : entry)) ?? current,
             );
         } catch (err: unknown) {
             // Skip writeback for user-triggered aborts (close, unmount, or per-row cancel).
@@ -206,6 +214,7 @@ const MlirFileResultsOverlay = () => {
                     ) ?? current,
             );
             createToastNotification('MLIR', message, ToastType.ERROR);
+            recordReportLoadFailed(ReportKind.MLIR, getReportLoadFailureReason(err));
         } finally {
             // Only delete the controller if it matches the one we stored for this retry.
             // Prevents a stale (completed) retry's finally block from evicting a new controller
@@ -250,8 +259,6 @@ const MlirFileResultsOverlay = () => {
             // the same peer name while staying on the MLIR route.
             setSplitViewEpoch((epoch) => epoch + 1);
         }
-        setMlirLoadedReports(loadedReports);
-
         // Local JSON loads live only in memory; only server uploads are stored
         // on disk and can be recorded as the instance's active MLIR so a reload
         // restores them. Persist index 0 only.
@@ -260,9 +267,13 @@ const MlirFileResultsOverlay = () => {
                 await setActiveMlir(primary.name, primary.host);
             } catch (err: unknown) {
                 createToastNotification('MLIR', getResponseError(err, 'Unable to set active MLIR'), ToastType.ERROR);
+                selectedResults.forEach(() => recordReportLoadFailed(ReportKind.MLIR, getReportLoadFailureReason(err)));
                 return;
             }
         }
+
+        setMlirLoadedReports(loadedReports);
+        selectedResults.forEach(() => recordReportLoaded(ReportKind.MLIR, ReportSource.UPLOAD));
 
         const toastDetail = comparison ? `${primary.filename} / ${comparison.filename}` : primary.filename;
         createToastNotification('MLIR', toastDetail, ToastType.SUCCESS);
--- a/src/components/mlir/MlirFileResultsOverlay.tsx
+++ b/src/components/mlir/MlirFileResultsOverlay.tsx
@@ -267,8 +267,6 @@ const MlirFileResultsOverlay = () => {
                 await setActiveMlir(primary.name, primary.host);
             } catch (err: unknown) {
                 createToastNotification('MLIR', getResponseError(err, 'Unable to set active MLIR'), ToastType.ERROR);
-                selectedResults.forEach(() => recordReportLoadFailed(ReportKind.MLIR, getReportLoadFailureReason(err)));
-                return;
             }
         }
 
--- a/src/components/mlir/MlirFileResultsOverlay.tsx
+++ b/src/components/mlir/MlirFileResultsOverlay.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
-import { useEffect, useRef, useState } from 'react';
+import { MouseEvent, useEffect, useRef, useState } from 'react';
 import { useAtom, useAtomValue, useSetAtom } from 'jotai';
 import { useLocation, useNavigate } from 'react-router';
 import axios from 'axios';
@@ -62,6 +62,7 @@ const MlirFileResultsOverlay = () => {
     const [retryingIndices, setRetryingIndices] = useState<Set<number>>(new Set<number>());
     const retrySessionRef = useRef(0);
     const retryAbortControllersRef = useRef<Map<number, AbortController>>(new Map<number, AbortController>());
+    const [isViewing, setIsViewing] = useState(false);
 
     // Abort all in-flight retries on unmount to prevent setResults writebacks
     // on unmounted tree. Complements the axios.isCancel guard in the catch block.
@@ -233,8 +234,8 @@ const MlirFileResultsOverlay = () => {
         }
     };
 
-    const handleView = async () => {
-        if (!results || selectedIndices.length === 0) {
+    const handleView = async (event: MouseEvent<HTMLElement>) => {
+        if (isViewing || !results || selectedIndices.length === 0) {
             return;
         }
 
@@ -248,6 +249,10 @@ const MlirFileResultsOverlay = () => {
         if (selectedResults.length === 0 || selectedResults.length !== selectedIndices.length) {
             return;
         }
+
+        // State does not commit between the two click events of a double-click.
+        event.currentTarget.setAttribute('disabled', '');
+        setIsViewing(true);
 
         const [primary, comparison] = selectedResults;
         // Server graphs are relabelled on the backend; local JSON at load time.
@@ -280,6 +285,7 @@ const MlirFileResultsOverlay = () => {
         if (location.pathname !== ROUTES.MLIR) {
             void navigate(ROUTES.MLIR);
         }
+        setIsViewing(false);
     };
 
     return (
@@ -325,7 +331,7 @@ const MlirFileResultsOverlay = () => {
             <div className={Classes.DIALOG_FOOTER_ACTIONS}>
                 <Button
                     intent={Intent.PRIMARY}
-                    disabled={selectedIndices.length === 0}
+                    disabled={selectedIndices.length === 0 || isViewing}
                     onClick={handleView}
                 >
                     View
--- a/src/components/npe/NPEFileLoader.tsx
+++ b/src/components/npe/NPEFileLoader.tsx
@@ -4,7 +4,7 @@
 
 import React, { useState } from 'react';
 import { useAtom } from 'jotai';
-import { useQueryClient } from '@tanstack/react-query';
+import { type Query, useQueryClient } from '@tanstack/react-query';
 import { FileInput, FormGroup, Icon, IconName, Intent } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import useLocalConnection from '../../hooks/useLocal';
@@ -15,6 +15,8 @@ import createToastNotification from '../../functions/createToastNotification';
 import { ToastType } from '../../definitions/ToastType';
 import getResponseError from '../../functions/getResponseError';
 import sanitiseFileName from '../../functions/sanitiseFileName';
+import { ReportKind, ReportLoadFailureReason } from '../../definitions/UsageEvent';
+import { getReportLoadFailureReason, recordReportLoadFailed } from '../../functions/reportLoadUsage';
 import 'styles/components/FileLoader.scss';
 
 const ICON_MAP: Record<ConnectionTestStates, IconName> = {
@@ -33,7 +35,14 @@ const INTENT_MAP: Record<ConnectionTestStates, Intent> = {
     [ConnectionTestStates.WARNING]: Intent.WARNING,
 };
 
-const NPEFileLoader = () => {
+const NPE_QUERY_KEYS = new Set<string>([NPE_SUMMARY_QUERY_KEY, NPE_WINDOW_QUERY_KEY, NPE_QUERY_KEY]);
+const isNpeQuery = (query: Query): boolean => NPE_QUERY_KEYS.has(String(query.queryKey[0]));
+
+interface NPEFileLoaderProps {
+    onUploadAccepted: (fileName: string) => void;
+}
+
+const NPEFileLoader = ({ onUploadAccepted }: NPEFileLoaderProps) => {
     const [statusMessage, setStatusMessage] = useState<string | null>(null);
     const { uploadNpeFile } = useLocalConnection();
     const queryClient = useQueryClient();
@@ -41,12 +50,12 @@ const NPEFileLoader = () => {
     const [uploadStatus, setUploadStatus] = useState<ConnectionTestStates>(ConnectionTestStates.IDLE);
 
     const handleFileChange = async (event: React.ChangeEvent<HTMLInputElement>) => {
-        setStatusMessage('Uploading...');
-        setUploadStatus(ConnectionTestStates.PROGRESS);
-
-        if (!event.target.files) {
+        if (!event.target.files?.length) {
             return;
         }
+
+        setStatusMessage('Uploading...');
+        setUploadStatus(ConnectionTestStates.PROGRESS);
 
         const file = event.target.files?.[0];
 
@@ -56,16 +65,18 @@ const NPEFileLoader = () => {
             if (response?.data?.status !== ConnectionTestStates.OK) {
                 setUploadStatus(ConnectionTestStates.FAILED);
                 setStatusMessage(response?.data?.message ?? 'Upload failed');
+                recordReportLoadFailed(ReportKind.NPE, ReportLoadFailureReason.OTHER);
             } else {
                 const fileName = file.name;
                 // Re-uploading a same-named file reuses the NPE query keys, and the
                 // windowed hooks are staleTime: Infinity — drop the cached summary /
                 // windows so the freshly-rebuilt server index is refetched instead of
                 // serving the previous report's data.
-                queryClient.removeQueries({ queryKey: [NPE_SUMMARY_QUERY_KEY] });
-                queryClient.removeQueries({ queryKey: [NPE_WINDOW_QUERY_KEY] });
-                queryClient.removeQueries({ queryKey: [NPE_QUERY_KEY] });
-                setActiveNpe(sanitiseFileName(fileName));
+                await queryClient.cancelQueries({ predicate: isNpeQuery });
+                queryClient.removeQueries({ predicate: isNpeQuery });
+                const sanitisedFileName = sanitiseFileName(fileName);
+                onUploadAccepted(sanitisedFileName);
+                setActiveNpe(sanitisedFileName);
                 createToastNotification('Active NPE', fileName, ToastType.SUCCESS);
                 setUploadStatus(ConnectionTestStates.OK);
                 setStatusMessage(`${fileName} uploaded successfully`);
@@ -73,6 +84,7 @@ const NPEFileLoader = () => {
         } catch (err: unknown) {
             setUploadStatus(ConnectionTestStates.FAILED);
             setStatusMessage(getResponseError(err, 'Unable to upload file'));
+            recordReportLoadFailed(ReportKind.NPE, getReportLoadFailureReason(err));
         }
     };
 
--- a/src/components/npe/NPEFileLoader.tsx
+++ b/src/components/npe/NPEFileLoader.tsx
@@ -4,7 +4,7 @@
 
 import React, { useState } from 'react';
 import { useAtom } from 'jotai';
-import { type Query, useQueryClient } from '@tanstack/react-query';
+import { type QueryFilters, useQueryClient } from '@tanstack/react-query';
 import { FileInput, FormGroup, Icon, IconName, Intent } from '@blueprintjs/core';
 import { IconNames } from '@blueprintjs/icons';
 import useLocalConnection from '../../hooks/useLocal';
@@ -36,7 +36,9 @@ const INTENT_MAP: Record<ConnectionTestStates, Intent> = {
 };
 
 const NPE_QUERY_KEYS = new Set<string>([NPE_SUMMARY_QUERY_KEY, NPE_WINDOW_QUERY_KEY, NPE_QUERY_KEY]);
-const isNpeQuery = (query: Query): boolean => NPE_QUERY_KEYS.has(String(query.queryKey[0]));
+const NPE_QUERY_FILTER: QueryFilters<readonly unknown[]> = {
+    predicate: (query) => NPE_QUERY_KEYS.has(String(query.queryKey[0])),
+};
 
 interface NPEFileLoaderProps {
     onUploadAccepted: (fileName: string) => void;
@@ -72,8 +74,8 @@ const NPEFileLoader = ({ onUploadAccepted }: NPEFileLoaderProps) => {
                 // windowed hooks are staleTime: Infinity — drop the cached summary /
                 // windows so the freshly-rebuilt server index is refetched instead of
                 // serving the previous report's data.
-                await queryClient.cancelQueries({ predicate: isNpeQuery });
-                queryClient.removeQueries({ predicate: isNpeQuery });
+                await queryClient.cancelQueries(NPE_QUERY_FILTER);
+                queryClient.removeQueries(NPE_QUERY_FILTER);
                 const sanitisedFileName = sanitiseFileName(fileName);
                 onUploadAccepted(sanitisedFileName);
                 setActiveNpe(sanitisedFileName);
--- a/src/components/npe/NPEFileLoader.tsx
+++ b/src/components/npe/NPEFileLoader.tsx
@@ -41,7 +41,7 @@ const NPE_QUERY_FILTER: QueryFilters<readonly unknown[]> = {
 };
 
 interface NPEFileLoaderProps {
-    onUploadAccepted: (fileName: string) => void;
+    onUploadAccepted: () => void;
 }
 
 const NPEFileLoader = ({ onUploadAccepted }: NPEFileLoaderProps) => {
@@ -77,7 +77,7 @@ const NPEFileLoader = ({ onUploadAccepted }: NPEFileLoaderProps) => {
                 await queryClient.cancelQueries(NPE_QUERY_FILTER);
                 queryClient.removeQueries(NPE_QUERY_FILTER);
                 const sanitisedFileName = sanitiseFileName(fileName);
-                onUploadAccepted(sanitisedFileName);
+                onUploadAccepted();
                 setActiveNpe(sanitisedFileName);
                 createToastNotification('Active NPE', fileName, ToastType.SUCCESS);
                 setUploadStatus(ConnectionTestStates.OK);
--- a/src/components/report-selection/LocalFolderSelector.tsx
+++ b/src/components/report-selection/LocalFolderSelector.tsx
@@ -51,6 +51,12 @@ import {
 import { TEST_IDS } from '../../definitions/TestIds';
 import { DBVersionValidation } from '../../definitions/Versions';
 import { evaluateDbVersion } from '../../functions/compareDbVersion';
+import { ReportKind, ReportLoadFailureReason, ReportSource } from '../../definitions/UsageEvent';
+import {
+    getReportLoadFailureReason,
+    recordReportLoadFailed,
+    recordReportLoaded,
+} from '../../functions/reportLoadUsage';
 
 const ICON_MAP: Record<ConnectionTestStates, IconName> = {
     [ConnectionTestStates.IDLE]: IconNames.DOT,
@@ -166,7 +172,7 @@ const LocalFolderOptions = () => {
         useReportLinkBadgeIds();
 
     const handleReportDirectoryOpen = async (e: ChangeEvent<HTMLInputElement>) => {
-        if (!e.target.files) {
+        if (!e.target.files?.length) {
             return;
         }
 
@@ -175,6 +181,7 @@ const LocalFolderOptions = () => {
 
         if (!checkRequiredReportFiles(files)) {
             setProfilerFolder(invalidReportStatus);
+            recordReportLoadFailed(ReportKind.PROFILER, ReportLoadFailureReason.MISSING_FILE);
             return;
         }
 
@@ -200,9 +207,11 @@ const LocalFolderOptions = () => {
             createToastNotification(ACTIVE_MEMORY_REPORT_TOAST_TITLE, updatedReport.reportName, ToastType.SUCCESS);
             setProfilerReportLocation(ReportLocation.LOCAL);
             setProfilerFolder(connectionOkStatus);
+            recordReportLoaded(ReportKind.PROFILER, ReportSource.UPLOAD);
         } catch (err: unknown) {
             const message = getResponseError(err, 'Unable to upload selected directory');
             setProfilerFolder({ status: ConnectionTestStates.FAILED, message });
+            recordReportLoadFailed(ReportKind.PROFILER, getReportLoadFailureReason(err));
         } finally {
             clearReportCaches(queryClient);
             setIsUploadingReport(false);
@@ -210,7 +219,7 @@ const LocalFolderOptions = () => {
     };
 
     const handlePerformanceDirectoryOpen = async (e: ChangeEvent<HTMLInputElement>) => {
-        if (!e.target.files) {
+        if (!e.target.files?.length) {
             return;
         }
 
@@ -219,6 +228,7 @@ const LocalFolderOptions = () => {
 
         if (!checkRequiredPerformanceFiles(files)) {
             setPerformanceFolder(invalidProfilerStatus);
+            recordReportLoadFailed(ReportKind.PERFORMANCE, ReportLoadFailureReason.MISSING_FILE);
             return;
         }
 
@@ -230,6 +240,7 @@ const LocalFolderOptions = () => {
 
             if (response?.data?.status !== ConnectionTestStates.OK) {
                 setPerformanceFolder(directoryErrorStatus);
+                recordReportLoadFailed(ReportKind.PERFORMANCE, ReportLoadFailureReason.MISSING_FILE);
             } else {
                 const fileName = getFolderName(files);
                 setPerformanceDataUploadLabel(`${files.length} files uploaded`);
@@ -237,10 +248,12 @@ const LocalFolderOptions = () => {
                 setActivePerformanceReport({ path: fileName, reportName: fileName });
                 createToastNotification(ACTIVE_PERFORMANCE_REPORT_TOAST_TITLE, fileName, ToastType.SUCCESS);
                 setPerformanceFolder(connectionOkStatus);
+                recordReportLoaded(ReportKind.PERFORMANCE, ReportSource.UPLOAD);
             }
         } catch (err: unknown) {
             const message = getResponseError(err, 'Unable to upload selected directory');
             setPerformanceFolder({ status: ConnectionTestStates.FAILED, message });
+            recordReportLoadFailed(ReportKind.PERFORMANCE, getReportLoadFailureReason(err));
         } finally {
             clearReportCaches(queryClient);
             setIsPerformanceUploading(false);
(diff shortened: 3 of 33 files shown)


tests/reportLoadUsage.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    it('records only the bounded kind and reason for a failed load', () => {
+        recordReportLoadFailed(ReportKind.NPE, ReportLoadFailureReason.PARSE_ERROR);
+
+        expect(recordUsage).toHaveBeenCalledWith({
+            event: UsageEvent.REPORT_LOAD_FAILED,
+            details: { kind: ReportKind.NPE, reason_class: ReportLoadFailureReason.PARSE_ERROR },
+        });
+        expect(JSON.stringify(recordUsage.mock.calls)).not.toContain('private response message');
```
aidemsined, line 54: > [!CAUTION]
> **Definite:** the privacy assertion cannot fail.

`'private response message'` is never in scope at the `recordReportLoadFailed` call — both arguments are bounded enum members, and that string is only ever constructed inside `getAxiosError`, which this test never invokes. There is no change to `reportLoadUsage.ts` that could make this assertion fail, including removing the classifiers entirely and forwarding a raw error through.

That matters because this is the only tripwire for the PR's central privacy claim, and #1831 states it categorically ("never record an error body or a stack trace"). The shipped code *is* correct — I traced every payload construction and only enum members reach `details` — but the test guarding it is inert.

**Failure scenario it should catch and does not:** someone widens the signature to `(kind, reason, error?)` and forwards `getResponseError(error)` into `details`. This spec stays green.

**Suggested action:** drive the assertion through the classifier so the string is genuinely in scope:

```ts
const error = getAxiosError(HttpStatusCode.InternalServerError, { error: 'private response message' });
recordReportLoadFailed(ReportKind.NPE, getNpeReportLoadFailureReason(NPEValidationError.DEFAULT, error));
expect(JSON.stringify(recordUsage.mock.calls)).not.toContain('private response message');
```

src/routes/NPE.tsx
```diff
@@ -79,6 +130,27 @@ const NPE = () => {
 
     const canShowView = !isLoading && errorCode === NPEValidationError.OK && npeData != null;
 
+    useEffect(() => {
```
aidemsined, line 96: > [!CAUTION]
> **Definite:** this effect cannot emit an event in any deployment where recording is enabled.

Recording requires `!SERVER_MODE` (`isUsageRecordingEnabled`). Reaching the record calls requires `isWindowedView === false`, and `isWindowedView = !isServerMode && !filepath && !!npeFileName`. Tracing the three ways to falsify it:

- `isServerMode` → recording is off, `recordUsage` returns at its first line
- `filepath` set → `NPEFileLoader` isn't rendered and `isDemoEnabled = isServerMode` is false, so no attempt is ever begun
- `!npeFileName` → but `onUploadAccepted` and `setActiveNpe` are issued in the same React batch (`NPEFileLoader.tsx:80-81`), so no render observes a pending attempt with a null name

So every live NPE event comes from `NpeWindowedView`'s callbacks. Three of the four new cases in the `NPE report-load recording` block run with `serverMode = true` and pin a path production cannot emit on. The comment at :103 covers the demo half only; a reader won't deduce the whole effect is inert.

There's a second edge here worth fixing at the same time: the live path's correctness rests entirely on that batching. If an `await` or a store write outside React's batch is ever introduced between those two lines, one render sees `isWindowedView === false` with `npeData == null`, `validateNpeData(undefined)` returns `INVALID_NPE_DATA`, and every successful upload inverts to `report_load_failed{parse_error}`.

**Suggested action:** extend the :103 comment to say the non-windowed branch is retained for #1802 and records nothing today, and make the invariant explicit rather than implicit — e.g. bail while nothing has settled: `if (!isNpeQueryEnabled && !isTimelineQueryEnabled && demoData == null) return;`.

tests/localFolderSelector.spec.tsx
```diff
@@ -83,6 +93,12 @@ vi.mock('../src/hooks/useAPI', async () => {
     };
 });
 
+vi.mock('../src/functions/reportLoadUsage', () => ({
+    getReportLoadFailureReason: () => ReportLoadFailureReason.OTHER,
```
aidemsined: > [!CAUTION]
> **Definite:** stubbing the classifier to a constant makes every `reason_class` assertion in the component specs vacuous.

This spec and three siblings (`MlirFileResultsOverlay.spec.tsx:46`, `NPEFileLoader.spec.tsx:20`, `remoteFolderSelector.spec.tsx:127`) all mock `getReportLoadFailureReason: () => ReportLoadFailureReason.OTHER`, and the real mapping is only exercised in isolation in `reportLoadUsage.spec.ts`. Nothing anywhere connects a real error to a reason at a call site.

**Concrete mutation that survives the entire suite:** replace every `getReportLoadFailureReason(err)` argument with the literal `ReportLoadFailureReason.OTHER` — `LocalFolderSelector.tsx:179,223,265`, `MlirJsonFileLoader.tsx:195`, `MlirFileResultsOverlay.tsx:217,270`, `NPEFileLoader.tsx:89`, `RemoteSyncConfigurator.tsx:462,481,506,538`. Result: `permission`, `missing_file` and `too_large` become unreachable for every non-NPE report, and every spec still passes. Those three buckets are a stated deliverable of #1831.

The layering is defensible in principle — the mapping is unit-tested separately — but it means the reason literal in these four files is decoration, and reviewers should know that. The hardcoded-literal sites (`MISSING_FILE` at `LocalFolderSelector:193,240,252`) *are* meaningfully asserted, because the reason is a literal in source rather than classifier output.

**Suggested action:** one spec per surface that mocks `recordUsage` rather than `reportLoadUsage` wholesale, rejects with a real `AxiosError` carrying a 404, and asserts `missing_file` end to end.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -442,8 +474,12 @@ const RemoteSyncConfigurator = () => {
 
                 return;
             }
-        } catch {
+            recordReportLoadFailed(kind, ReportLoadFailureReason.OTHER);
+        } catch (mountError: unknown) {
             // Mount itself failed; surface the original sync error below.
+            if (!axios.isCancel(mountError)) {
```
aidemsined: > [!CAUTION]
> **Definite:** the classify/record/skip-cancel triple is hand-written six times in this component, and is already inconsistent.

Guarded here and at :536 with `if (!axios.isCancel(err))`; **unguarded** at :460 and :506. So a user cancelling a sync records a `report_load_failed` on two of the four paths. That divergence appearing within one PR is the predicted cost of a manual pairing rather than a coincidence.

Cancel detection already has three homes in the repo: `notifyFolderSyncError.ts:18` and `:32` both return early on `axios.isCancel`, and `getRemoteSyncFailureAction.ts:18` models it as `RemoteSyncFailureAction.IGNORE_CANCEL`. Every one of the six new record calls sits immediately beside a `notifyFolderSyncError(...)` / `notifyRemoteFolderMountError(...)` that has just made the same decision. The PR also has to add `axios` to this component's imports to do it.

**Suggested action:** put the pairing where the toast half already lives — `notifyAndRecordFolderSyncError(kind, err)` in `notifyFolderSyncError.ts` — or give `reportLoadUsage` a cancel-aware `recordReportLoadFailure(kind, err)` that returns early on `axios.isCancel` (that module already imports axios). Six call sites become six one-liners, the guard has one home, and the `axios` import leaves the component.

src/functions/reportLoadUsage.ts
```diff
@@ -0,0 +1,62 @@
...
+        case HttpStatusCode.PayloadTooLarge:
+            return ReportLoadFailureReason.TOO_LARGE;
+        case HttpStatusCode.UnprocessableEntity:
+            return options.unprocessableEntityReason ?? ReportLoadFailureReason.OTHER;
+        default:
+            return ReportLoadFailureReason.OTHER;
+    }
+}
+
+export function getNpeReportLoadFailureReason(
```
aidemsined: > [!CAUTION]
> **Definite:** one feature's domain vocabulary has landed inside the cross-cutting instrumentation module, and the generic classifier's extension point exists only to serve it.

`NPEValidationError` (imported at :7) is the only feature-specific import in a module that otherwise knows nothing but axios and `UsageEvent`. This function's sole consumer is `NPE.tsx:92`. And `ReportLoadFailureOptions` / `options.unprocessableEntityReason` is passed by exactly one caller — this function itself, at :58. So an exported generic option has one internal user, and a generic module depends on a feature enum.

The asymmetry shows the concept is missing rather than misnamed: MLIR has `MLIRValidationError` with its own `INVALID_JSON` and gets no mapper at all — reasons are hand-picked at `MlirJsonFileLoader.tsx:100,166,176` and `MlirFileResultsOverlay.tsx:184`.

**Suggested action:** model "validation error → failure reason" once, as data beside each validation enum — `NPE_LOAD_FAILURE_REASONS: Partial<Record<NPEValidationError, ReportLoadFailureReason>>` in `definitions/NPEData.ts` (and the MLIR twin), with the route doing `NPE_LOAD_FAILURE_REASONS[errorCode] ?? getReportLoadFailureReason(error)`. `reportLoadUsage.ts` then keeps three exports and no feature imports, `ReportLoadFailureOptions` disappears, and a new `NPEValidationError` member becomes a `Record` completeness error instead of falling silently through the `default:` at :57 — which is the exhaustiveness argument CONVENTIONS.md makes for enums.

src/components/report-selection/LocalFolderSelector.tsx
```diff
@@ -165,8 +171,17 @@ const LocalFolderOptions = () => {
     const { linkedPerfIds, unlinkedPerfIds, linkedProfilerReportIds, unlinkedProfilerReportIds } =
         useReportLinkBadgeIds();
 
+    const activateLocalReport = async (kind: ReportKind, action: () => Promise<void>) => {
+        try {
+            await withActivatingReport(action);
+            recordReportLoaded(kind, ReportSource.LOCAL_TT_METAL);
```
aidemsined, line 179: > [!CAUTION]
> **Definite:** an incompatible report version is recorded as a successful load, so `unsupported_version` ends up NPE-only.

`activateLocalReport` records `report_loaded{kind, local_tt_metal}` as soon as `withActivatingReport` resolves. The DB-version check lives in a separate effect at :132-138 which only raises a toast (`'Incompatible report version'`) — it runs after the record and has no path back to it.

Reading the log, a profiler or performance report the app has just told the user it cannot open will be indistinguishable from a clean load. Meanwhile the NPE path *does* produce `unsupported_version` via `getNpeReportLoadFailureReason`, so the bucket looks populated while silently covering one report kind out of three.

**Suggested action:** either record `report_load_failed{unsupported_version}` when `evaluateDbVersion` returns a non-OK status — which needs the record moved after that check, or the effect given the ability to supersede its own success — or document in `usage.py`'s enum docstring that `unsupported_version` is NPE-only so the numbers are read correctly.

src/routes/NPE.tsx
```diff
@@ -79,6 +130,27 @@ const NPE = () => {
...
+        if (isWindowedView || pendingLoadAttemptId === null || isLoading) {
+            return;
+        }
+
+        if (canShowView) {
+            completeLoadAttempt(pendingLoadAttemptId);
+        } else if (errorCode !== NPEValidationError.OK) {
+            failLoadAttempt(pendingLoadAttemptId, errorCode, fetchError);
```
aidemsined: > [!CAUTION]
> **Definite:** the `fetchError` argument is unasserted, so dropping it is invisible.

`NPERoute.spec.tsx:106` stubs `getNpeReportLoadFailureReason: () => ReportLoadFailureReason.PARSE_ERROR`, which means the third argument never affects an assertion.

**Mutation that survives:** delete `, fetchError` from this call. Every HTTP-classified NPE failure collapses to the validation-derived reason, losing `permission`, `missing_file` and `too_large` for the NPE kind, and the suite stays green.

Worth noting `NpeWindowedView` is covered properly here — `NpeWindowedView.spec.tsx` asserts `expect.objectContaining({ message: 'index build failed' })` — which makes the route the odd one out rather than a systemic gap.

**Suggested action:** assert the third argument in the route spec, or stop stubbing the classifier there so the real mapping runs.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -260,10 +267,14 @@ const MlirFileResultsOverlay = () => {
                 await setActiveMlir(primary.name, primary.host);
             } catch (err: unknown) {
                 createToastNotification('MLIR', getResponseError(err, 'Unable to set active MLIR'), ToastType.ERROR);
+                selectedResults.forEach(() => recordReportLoadFailed(ReportKind.MLIR, getReportLoadFailureReason(err)));
                 return;
             }
         }
 
+        setMlirLoadedReports(loadedReports);
```
aidemsined, line 278: > [!WARNING]
> **Possible:** a persist failure now discards successfully-converted graphs — a behaviour change the PR body doesn't mention.

`setMlirLoadedReports(loadedReports)` moved from before the `setActiveMlir` try/catch to after it, so the `return` in the catch now fires first.

**Failure scenario:** a user uploads two MLIR files, both convert fine, clicks View; `PUT`ing the active MLIR fails (backend restarted, disk full). Previously the graphs were in `mlirLoadedReportsAtom` and viewable — only the restore-on-reload convenience was lost. Now they see "Unable to set active MLIR" and no report at all. The new specs codify this with `expect(getDefaultStore().get(mlirLoadedReportsAtom)).toEqual([])`.

`setSplitViewEpoch((epoch) => epoch + 1)` at :260 still runs on that path, so the failure leaves the epoch bumped with no peer loaded.

**Suggested action:** decide it explicitly. If in-memory viewing should survive a persist failure, keep `setMlirLoadedReports` before the persist and record `report_loaded` (the load did happen), with a separate signal for the persist failure if one is wanted. If the new behaviour is intended, say so in the PR body and move the epoch bump below the persist too.

src/components/report-selection/LocalFolderSelector.tsx
```diff
@@ -165,8 +171,17 @@ const LocalFolderOptions = () => {
     const { linkedPerfIds, unlinkedPerfIds, linkedProfilerReportIds, unlinkedProfilerReportIds } =
         useReportLinkBadgeIds();
 
+    const activateLocalReport = async (kind: ReportKind, action: () => Promise<void>) => {
```
aidemsined: > [!WARNING]
> **Possible:** a failed local activation is now completely silent to the user.

Commit `1698cc50` wrote this catch as `{ record…; throw err; }`; commit `0cd25a79` ("Address report load usage review findings") removed the `throw`. The result is that when `updateInstance` fails, the only trace anywhere is a usage event — no toast, no console rejection, and the picker keeps showing the old selection while the user believes the click worked.

Every sibling path in this same file surfaces it: `handleReportDirectoryOpen:222` and `handlePerformanceDirectoryOpen:264` both set a `FAILED` `ConnectionStatus` with a `getResponseError` message.

**Failure scenario:** `PUT /api/instance` 500s. User clicks a local profiler report; nothing visible happens; the app records `report_load_failed{profiler,other}` and moves on.

Worth flagging separately that this is a user-visible behaviour change made in a commit titled as a review-findings pass — it deserved its own commit or a line in the message.

**Suggested action:** surface it alongside the record, matching the upload handlers — `createToastNotification(title, getResponseError(err), ToastType.ERROR)` with the title threaded through the helper — or re-instate the `throw` if an outer boundary is meant to own it.

src/components/npe/NPEFileLoader.tsx
```diff
@@ -33,21 +35,30 @@ const INTENT_MAP: Record<ConnectionTestStates, Intent> = {
...
 };
 
-const NPEFileLoader = () => {
+const NPE_QUERY_KEYS = new Set<string>([NPE_SUMMARY_QUERY_KEY, NPE_WINDOW_QUERY_KEY, NPE_QUERY_KEY]);
+const NPE_QUERY_FILTER: QueryFilters<readonly unknown[]> = {
+    predicate: (query) => NPE_QUERY_KEYS.has(String(query.queryKey[0])),
+};
+
+interface NPEFileLoaderProps {
+    onUploadAccepted: (fileName: string) => void;
```
aidemsined: > [!WARNING]
> **Possible:** a report filename is threaded into the recording handshake and then discarded — a privacy leak one edit away.

The prop is typed `(fileName: string) => void` and called with `sanitisedFileName` at :80, but the consumer takes no parameter (`NPE.tsx:94`, `useCallback(() => {…})`). `NPEFileLoader.spec.tsx:78` even pins the value with `toHaveBeenCalledWith('report.npeviz')`.

Nothing leaks today — I traced every path into `recordUsage` and only enum members reach `details`. But `usage.py`'s module docstring is categorical that no report, file, directory, operation or host name may ever be written, and the only thing standing between a user filename and a payload on this seam is that the receiving callback happens to ignore its argument. This is the exact shape a future "add the report name for debugging" change would exploit, and the spec would not object.

**Suggested action:** narrow to `onUploadAccepted: () => void`, drop the argument at the call site, and change the assertion to `toHaveBeenCalledTimes(1)`. The type then makes the leak impossible rather than merely absent. Renaming the prop for what it means to the parent — a "user-initiated load is starting" signal, not a filename delivery — would help too.

src/components/mlir/MlirFileResultsOverlay.tsx
```diff
@@ -260,10 +267,14 @@ const MlirFileResultsOverlay = () => {
...
+        selectedResults.forEach(() => recordReportLoaded(ReportKind.MLIR, ReportSource.UPLOAD));
```
aidemsined, line 279: > [!NOTE]
> **Possible:** `handleView` is re-entrant behind an `await`, so a double-click records two loads for one.

The View button is disabled only on an empty selection (`disabled={selectedIndices.length === 0}` at :328), and `handleView` awaits `setActiveMlir` at :267.

**Failure scenario:** a user double-clicks View on a persisted upload. Two invocations run concurrently with the same `selectedIndices`, both await, both resolve, both reach this line — one real load producing two `report_loaded{mlir,upload}` events, or four with a split-view peer.

The underlying double-submit predates this PR, but until now it had no measurement consequence. #1831 makes `report_loaded` the denominator every other figure is read against, so inflating it is worse here than the duplicate POST.

**Suggested action:** guard with an in-flight ref set and cleared around the body, and reflect it in `disabled` so the second click can't start.

tests/useRestoreInstance.spec.tsx
```diff
@@ -105,6 +112,33 @@ it('does not reset memory list state during initial instance hydration', async (
     expect(mockResetMemoryListStates).toHaveBeenCalledTimes(0);
 });
 
+it('does not record restored reports as user-initiated loads', async () => {
```
aidemsined, line 117: > [!NOTE]
> **Possible:** this is a weak tripwire for the AC it's guarding.

`useRestoreInstance.tsx` imports only React, jotai, `useAPI`, `useRemote`, `definitions/Reports`, `model/RemoteConnection` and `useRestoreScrollPosition` — none of which reach `reportLoadUsage` — and the harness renders `HookHarness` alone. So this negative holds *structurally* and can only break if a call is added inside that one hook.

It would not catch the failure mode #1831 actually warns about: recording wired to an atom-change effect somewhere else in the tree, which is precisely how a restore would get miscounted. `NPERoute.spec.tsx:202` is the strong version of this test for the NPE kind; there's no equivalent for profiler or performance.

Related and worth checking while you're here: boot restore can also *invert* a pending NPE attempt. `useRestoreInstance` writes `setActiveNpe` from a boot snapshot after a network round-trip; if an upload completes before that resolves, the overwrite flips `isWindowedView` with the attempt still pending, and a report that loaded fine records `parse_error`.

**Suggested action:** render the report-selection surface under a restored instance and assert zero events there, rather than asserting over a hook that cannot reach the module.

src/components/report-selection/LocalFolderSelector.tsx
```diff
@@ -200,17 +216,19 @@ const LocalFolderOptions = () => {
             createToastNotification(ACTIVE_MEMORY_REPORT_TOAST_TITLE, updatedReport.reportName, ToastType.SUCCESS);
```
aidemsined, line 225: > [!NOTE]
> **Possible:** the profiler upload records success without checking the response status, unlike its performance sibling.

`/local/upload/profiler` returns **HTTP 200** with `{status: 'failed', message: 'Invalid project directory.'}` when `validate_files(files, {"db.sqlite"})` fails (`views.py:1429-1432`). This handler goes straight to `normaliseReportFolder`, a success toast and `recordReportLoaded`. The performance handler at :250-252 does check the payload status.

Latent today, because the client pre-check mirrors the backend's depth rule — but the backend has one condition the client doesn't (`len(leading_segments) > 1` in `file_uploads.py:55`), so the two can diverge.

**Suggested action:** check `response?.data?.status` before recording, the way the performance handler does.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -493,9 +526,16 @@ const RemoteSyncConfigurator = () => {
                     if (hasBeenNormalised(folder)) {
                         createDataIntegrityWarning(folder);
                     }
+                } else {
+                    const mountError = new Error(UNEXPECTED_MOUNT_STATUS_MESSAGE);
```
aidemsined: > [!NOTE]
> **Possible:** this non-200 branch is unreachable on a resolved promise, but it adds a user-visible error toast.

`mountRemoteFolder` posts through `axiosInstance` with the default status validation (2xx resolve), and `/remote/use` returns `Response(status=HTTPStatus.OK)` on success and 400/404 otherwise (`views.py:2433-2486`). So `response.status !== 200` cannot be observed here — every real failure lands in the `catch` below. The same shape appears at :477 for `mountLocalFolderOnSyncFailure`.

That makes the new `notifyRemoteFolderMountError(new Error('Remote report could not be mounted'))` dead-but-noisy: if the backend ever *did* return 201 or 204 the user would get an error toast for a successful mount. No test covers it — `remoteFolderSelector.spec.tsx:719-763` rejects the mount instead of returning an odd 2xx.

This is also one of four behaviour changes not mentioned in the PR body (the others: MLIR load ordering, the windowed NPE version gate, and `cancelQueries` before eviction). All look like improvements, but a reviewer should know to test them.

**Suggested action:** drop the `else` and record the failure from the `catch` only, or keep the record and lose the new toast.

src/components/report-selection/RemoteSyncConfigurator.tsx
```diff
@@ -776,8 +824,7 @@ const RemoteSyncConfigurator = () => {
                         selectAndActivateFolder(folder, {
                             setSelected: setSelectedReportFolder,
                             syncFolder: syncSelectedReportFolder,
-                            mount: (connection, report) => remote.mountRemoteFolder(connection, report),
-                            activateWithToast: updateReportSelection,
+                            ...profilerReportActions,
                         })
                     }
                     type='profiler'
```
aidemsined: > [!NOTE]
> **Possible:** `kind` is threaded through five signatures to carry what the sibling prop on this same element already says.

`...profilerReportActions` (supplying `kind: ReportKind.PROFILER`) is spread onto the same `RemoteFolderSelector` that receives `type='profiler'` four lines down. And `RemoteFolderSelector.tsx:23` declares `type FolderTypes = 'performance' | 'profiler'` — a string-literal union whose values are character-for-character the `ReportKind.PROFILER` / `ReportKind.PERFORMANCE` members.

CONVENTIONS.md names this exact case ("prefer named enums over string-literal unions when the union has semantic meaning"). The threading costs five new parameters — `mountLocalFolderOnSyncFailure:454`, `handleSyncFailure:492`, `syncSelectedFolder:564`, plus two `Pick<RemoteReportActions, …>` unions at :505 and :659 — to carry one compile-time constant.

**Suggested action:** either narrow `FolderTypes` to `ReportKind.PROFILER | ReportKind.PERFORMANCE` and derive `kind` from the `type` already in scope, or — if coupling a UI prop to the wire vocabulary is unwanted — replace the `kind` field on `RemoteReportActions` with a bound `recordFailure: (err: unknown) => void`. The four downstream helpers then keep the bundle they already receive and gain no new parameter at all.

Minor, same area: both `Pick<...>` unions select three of the interface's four members while both call sites spread the whole bundle anyway, so they document nothing true about what arrives.

src/routes/NPE.tsx
```diff
@@ -48,9 +59,49 @@ const NPE = () => {
     } = useNPETimelineFile(filepath);
     const [demoData, setDemoData] = useState<NPEData | null>(null);
     const [selectedDemo, setSelectedDemo] = useState<NPEDemoData | null>(null);
+    const nextLoadAttemptIdRef = useRef(0);
```
aidemsined: > [!NOTE]
> **Possible:** the `loadAttemptId` handshake implements "settle each attempt at most once" twice, on both sides of a component boundary, plus a third settle path forked on a flag.

The parent holds three pieces of state for it (`nextLoadAttemptIdRef`, `pendingLoadAttemptRef`, `pendingLoadAttemptId`) and re-checks `attempt.id !== attemptId` in both `completeLoadAttempt` and `failLoadAttempt`. The child holds `settledLoadAttemptIdRef` and writes it five times inside one effect with an eleven-entry dependency array (`NpeWindowedView.tsx:79-111`). And the same success/failure decision is made independently a third time in this route's own effect at :133.

So one question — "did this NPE report load?" — has two implementations selected by a feature gate, and the id protocol exists to reconcile them. The correctness argument is sound (I traced the StrictMode and stale-attempt cases and both are genuinely covered, which is good), but it's a lot of machinery for one boolean.

**Suggested action:** `NpeWindowedView` is already keyed on the report (`key={npeFileName}`), so it remounts per load — give it ownership of its own attempt and pass only `source`. Or lift the attempt into a `useReportLoadAttempt(kind)` hook that both the route effect and the windowed view call, so the id guard lives in one function and the three new props collapse to zero.

aidemsined (commented):
The shipped privacy behaviour is correct — I traced every path into `recordUsage` and only closed-enum values reach `details`; no path, filename, host, report name, message or stack is reachable. Frontend/backend vocabulary matches `usage.py:220-243` member for member, and telemetry can't break a load (`recordUsage` is a predicate plus a bounded push). StrictMode double-invocation is genuinely covered, twice.

The findings are concentrated in two places: tests that would pass if the feature were removed, and one seam that is instrumented but unreachable.

## Unanchored findings

- **Definite — no call-site convention was added to `AGENTS.md` / `CONVENTIONS.md`.** This is the first code in the repo to *emit* a usage event (`grep recordUsage src/` previously hit only `Layout.tsx`'s `initUsageRecording`), and `CONVENTIONS.md:11` mandates updating both files together. Four load-bearing rules are invented here and invisible from any one file: record at the user-initiated activation site so boot restoration is excluded; one event per report, not per batch; cancels are not failures; `ReportSource.DEMO` is wired but permanently dropped on hosted. `VIEW_OPENED` / `VIEW_ENGAGED` are still unwired, so the next author — plausibly #1963 — faces these questions with nothing to read.
- **Possible — the seam the ticket asked about does exist here, but wasn't shared.** `activateLocalReport` is exactly a `withActivatingReport`-plus-recording wrapper, yet it's a local arrow inside `LocalFolderOptions`, so `RemoteSyncConfigurator:519,606` re-does the work by hand. The ticket's literal suggestion (instrument inside `useActivatingReport`) genuinely can't work — that hook covers 2 of ~24 record sites and can't know `source` — and the PR is right to reject it. But a middle option was available: `useReportLoad(kind)` returning `{ run, succeeded, failed }`, which gives the wrapper, the cancel guard and the classifier one home each and captures `kind` once instead of threading it.
- **Possible — `/npe/:filepath` deep links emit no `report_loaded`**, so NPE reports opened by link are invisible to the denominator. Outside #1831's stated wiring list, so likely deliberate — worth a sentence in the PR body or a follow-up either way.
- **Possible — two new decision points in `RemoteSyncConfigurator` are untested**: the `if (!connection)` record at :515, and the mid-sync abandonment branch at :602-606 which deliberately records *nothing*. That second one is worth pinning: adding a `recordReportLoadFailed` there would leave the suite green while every mid-sync report switch started logging a phantom failure.

## General observations

- `MlirJsonFileLoader.tsx:100` — a `relabelMlirGraphIds` throw is classified `parse_error` because the relabel sits inside the parse `try`; `other` would be honest.
- `MlirJsonFileLoader.tsx:174` — a short `results` array under-records: 3 files uploaded, 2 results returned, nothing recorded for the dropped one.
- `NpeWindowedView.tsx:61` — a summary missing `common_info` reads as `unsupported_version` rather than a shape error; `validateNpeSummary` is the existing seam for shape.
- `reportLoadUsage.ts:26` — the `error.status` limb of `error.status ?? error.response?.status` is never exercised; the fixture only sets `response.status`, and modern axios populates both.
- `tests/NpeWindowedView.spec.tsx:185` — the degrade-in-place case asserts only the negative; adding `expect(onInitialLoadSuccess).toHaveBeenCalledWith(9)` would pin that a degraded window still counts as loaded.
- `MlirJsonFileLoader.tsx:166` — `selectedFiles.forEach(() => record(...))` expresses cardinality by discarding the item; a `count` parameter on the recorder would say it, and would also keep "one report equals one event" in one place.
- `reportLoadUsage.ts:1` — no module header, in a subsystem whose siblings are documented unusually heavily (`recordUsage.ts` opens with 12 lines of rationale). The non-obvious decisions — why a 422 has no universal meaning, why cancels must never reach the classifier, why `default:` is a deliberate bucket — go unexplained.
- `ReportKind.CLUSTER_DESCRIPTOR` has no call site and `ReportSource.DEMO` is reachable only under `SERVER_MODE` where events are dropped. Both will read as permanent zeros to a collector; worth a note in the `usage.py` docstrings so they're read as "not instrumented" rather than "not used".
- The MLIR upload paths have no `axios.isCancel` guard, unlike the remote ones. Harmless today since those uploads pass no signal — but the day one becomes cancellable, a user cancel turns into N `report_load_failed`.
- Per-file cardinality can saturate the shared 50-event buffer: `MAX_MLIR_FILE_SELECTION` bounds viewing, not selection, so a large batch can push past `MAX_BUFFERED_EVENTS` and drop unrelated events until the next flush. Contained by design, so it costs telemetry rather than stability.
- PR body omits four behaviour changes worth listing for whoever tests this: the MLIR load ordering, the new windowed-NPE version gate, the new unexpected-mount-status toast, and `cancelQueries` before the cache eviction.

## Worth calling out as good

`NPERoute.spec.tsx:210` is the strongest test in the PR — it begins two attempts, fires the stale one's success and asserts nothing recorded, then fires attempt 2's success plus two replays and a failure and asserts exactly one record. That kills both the stale-attempt and the replay double-fire in one case. `remoteFolderSelector.spec.tsx:1026` pinning "sync failure with local fallback records exactly one loaded and zero failed" is the other one I'd have worried about, and it's nailed down.

tests/reportLoadUsage.spec.ts
dcblundell: Agreed — the string was never in scope. The assertion now runs through `getNpeReportLoadFailureReason` with a real Axios body, and `recordReportLoadFailure` has a matching 404 case. Both fail if a body ever reaches `details`.

src/routes/NPE.tsx
dcblundell: Agreed. Extended the comment to say the non-windowed branch is for hosted / #1802 and records nothing where recording is on. Also bail while nothing has settled (`!isNpeQueryEnabled && !isTimelineQueryEnabled && demoData == null`) so a pending attempt cannot be classified as `parse_error` from empty `npeData`.

tests/localFolderSelector.spec.tsx
dcblundell: Agreed. Specs now keep the real classifiers and only spy on `recordReportLoaded` / `recordReportLoadFailed`. Each surface has a 404 → `missing_file` case that would fail if the call site passed a literal `OTHER`.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Agreed on the duplicated pairing; extracted `recordReportLoadFailure` and `notifyAndRecordFolderSyncError` / `notifyAndRecordRemoteFolderMountError` so cancel-skip lives in one place.

Disagree that `:460` / `:506` recorded cancels: `handleSyncFailure` already returns on `IGNORE_CANCEL` before those lines, and fallback is not entered for a cancel. The helper still makes that harder to regress.

src/functions/reportLoadUsage.ts
dcblundell: Agreed this is a coupling smell. Leaving the NPE mapper in `reportLoadUsage` for this PR — the generic `unprocessableEntityReason` option is still only used here. Happy to move the maps next to the validation enums in a follow-up.

src/components/report-selection/LocalFolderSelector.tsx
dcblundell: Agreed the bucket would be misread. Documented instead of recording a failure: `evaluateDbVersion` is a warning, not a hard gate — the report still opens. `usage.py` (and a note on the TS enum) now say `unsupported_version` is NPE-only today. Recording `report_load_failed` here would disagree with the UI.

src/routes/NPE.tsx
dcblundell: Agreed. The route spec no longer stubs `getNpeReportLoadFailureReason`. Added a hosted 404 case that asserts `missing_file`, so dropping `fetchError` fails the suite.

src/components/mlir/MlirFileResultsOverlay.tsx
dcblundell: Agreed this was an unintended behaviour change. Persist failure now toasts, but converted graphs stay in `mlirLoadedReportsAtom` and still count as `report_loaded` — the load happened. Epoch bump stays where it is because split view still opens.

src/components/report-selection/LocalFolderSelector.tsx
dcblundell: Agreed — swallowing the error left the picker looking unchanged. `activateLocalReport` now toasts `Unable to open memory/performance report` with `getResponseError` alongside the usage record.

src/components/npe/NPEFileLoader.tsx
dcblundell: Agreed. Narrowed to `onUploadAccepted: () => void` and dropped the filename at the call site. The spec now only asserts it was called.

src/components/mlir/MlirFileResultsOverlay.tsx
dcblundell: Agreed this now inflates the `report_loaded` denominator. Leaving the in-flight guard for a follow-up — the double-submit predates this PR and is not specific to usage recording.

tests/useRestoreInstance.spec.tsx
dcblundell: Agreed the hook-only negative is structurally weak. Keeping it as a tripwire on that module; `NPERoute.spec.tsx` already covers restore-without-attempt for NPE. A full report-selection restore spec (and the boot-vs-pending-NPE race) is a follow-up rather than expanding this PR.

src/components/report-selection/LocalFolderSelector.tsx
dcblundell: Agreed. Profiler upload now checks `response.data.status` when present (success payloads omit it) and records `missing_file` instead of a load, matching the performance handler.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Agreed. Dropped the non-200 `else` and the toast it invented. Failures record from `catch` via `notifyAndRecordRemoteFolderMountError`. Same for the unreachable non-OK branch on the local-fallback mount.

src/routes/NPE.tsx
dcblundell: Agreed it is a lot of machinery for one boolean. Leaving the attempt-id protocol: StrictMode and stale attempts are covered, and collapsing it into a hook is a larger refactor than this review pass. Will take it if we next touch the NPE load path.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Agreed `kind` duplicates `type` on the same element. Leaving the threading for this PR — folding `FolderTypes` into `ReportKind` is a wider UI change than the usage work. Follow-up if we touch this configurator again.

src/components/report-selection/RemoteSyncConfigurator.tsx
aidemsined: > [!NOTE]
> **Correction — you are right, and the strongest part of my finding was wrong.**

I checked `:460` and `:506` against the code rather than leaving it as a disagreement, and both are downstream of the gate you named:

- `getRemoteSyncFailureAction` returns `IGNORE_CANCEL` for `axios.isCancel(err)` (`getRemoteSyncFailureAction.ts:18`)
- `handleSyncFailure` returns on `IGNORE_CANCEL` before reaching the record at `:506`
- `:460` is inside `mountLocalFolderOnSyncFailure`, which is only reachable via `mountLocalFallback` from `handleSyncFailure` — so after that same early return

So neither site could record a cancel, and my "already inconsistent" framing was false. That was the line I used to escalate the finding from a duplication nit to a Definite, so it is worth retracting explicitly rather than letting it stand: the duplication was real and worth fixing, the inconsistency was not.

Thanks for pushing back instead of just applying it. `recordReportLoadFailure` plus the two `notifyAndRecord*` helpers is a better outcome than the guard sprinkling I was arguing against anyway.

One observation from verifying the rest, offered as an observation rather than a request. `tests/helpers/mockReportLoadUsage.ts` re-implements `recordReportLoadFailure`'s body in the double — which your docstring explains correctly (the original closes over the unmocked `recordReportLoadFailed`). The consequence is that the real helper's body never executes in the component specs: deleting the `axios.isCancel` guard from it fails exactly one test, the direct unit test in `reportLoadUsage.spec.ts`. That is defensible layering and the guard *is* pinned — but it is production logic living in a mock, so a future rule added to the real helper would silently not apply to any component spec. Worth a comment on the double, or asserting through the real helper on one surface.

Separately, and this is the useful part: I re-ran my original mutation against your fixed specs. Collapsing every classified reason to a literal `OTHER` now **fails a test**, where before it failed none. The privacy assertion also genuinely has the string in scope now. Both findings are properly closed.

tests/useRestoreInstance.spec.tsx
dcblundell: Addressed in `0e369573`. `ProtectedRoute` now keeps interactive report surfaces unavailable until instance restoration settles, including the report-list round trip, while still restoring if that list request fails. Integration coverage pins the pending surface, redirect ordering, zero usage events, and the query-failure fallback.

src/routes/NPE.tsx
dcblundell: Addressed in `0e369573`. Attempt IDs, source ownership, stale guards, and exactly-once recording now live in `useNpeLoadAttempt`; the windowed view receives one controller prop and its duplicate settled-ID ref is gone. Direct tests cover duplicate settlement and stale failures.

src/components/report-selection/RemoteSyncConfigurator.tsx
dcblundell: Addressed in `0e369573`. `RemoteFolderType` is now derived from the `ReportKind` enum members in `definitions/Reports.ts`; each action bundle carries that single `type`, which is reused by both `RemoteFolderSelector` and the recording helpers. The duplicate `kind` field and string-literal union are gone.

src/components/mlir/MlirFileResultsOverlay.tsx
dcblundell: Addressed in `0e369573`. View now disables the live button synchronously before the persistence await and tracks the in-flight state declaratively. The regression test double-clicks while persistence is pending and asserts one persistence call and one `report_loaded` event.

src/functions/reportLoadUsage.ts
dcblundell: Addressed in `0e369573`. NPE and MLIR validation-to-reason maps now live beside their enums, the NPE endpoint-specific classifier has its own feature module, and `reportLoadUsage.ts` no longer imports NPE vocabulary or exposes the one-off options type.

PR #1963, opened 2026-09-01
Record view-opened usage by route

### LINKED ISSUE #1832: Usage signals: record view_opened and reconcile the view enum with the real routes

dcblundell:
## Summary
- map application routes to the closed usage-view vocabulary
- record view opens centrally, including parameter changes and topology modal navigation
- share parameterised route and modal-state semantics to prevent drift
- add route, navigation, gating, and StrictMode coverage

## Testing
- `pnpm test`
- `pnpm lint`
- `pnpm lint:ts`
- `pnpm build`
- `uv run pytest backend/ttnn_visualizer/tests/test_usage_frontend_parity.py`
- `pnpm flask:lint`

Closes #1832

### DIFF
--- a/src/components/Layout.tsx
+++ b/src/components/Layout.tsx
@@ -18,6 +18,8 @@ import FeedbackButton from './FeedbackButton';
 import FileStatusOverlay from './FileStatusOverlay';
 import MlirFileResultsOverlay from './mlir/MlirFileResultsOverlay';
 import { initUsageRecording } from '../functions/recordUsage';
+import useRecordViewOpened from '../hooks/useRecordViewOpened';
+import { isModalOpen } from '../functions/modalRoute';
 
 const BounceIn = cssTransition({
     enter: `Toastify--animate Toastify__bounce-enter`,
@@ -29,12 +31,14 @@ const BounceIn = cssTransition({
 
 function Layout() {
     const location = useLocation();
-    const state = location.state as { background?: Location };
+    const isTopologyOpen = isModalOpen(location, ROUTES.CLUSTER);
 
     // Starts the usage flush lifecycle; it records nothing on its own. Here rather than at
     // module scope so importing the sender has no side effect, and so the listeners are
     // owned the way every other listener in this app is.
     useEffect(() => initUsageRecording(), []);
+
+    useRecordViewOpened();
 
     return (
         <>
@@ -58,7 +62,7 @@ function Layout() {
 
                 <main>
                     <ModalAwareOutlet />
-                    {location.pathname === ROUTES.CLUSTER && state?.background && <ClusterRenderer />}
+                    {isTopologyOpen && <ClusterRenderer />}
                 </main>
             </div>
 
--- a/src/definitions/Routes.ts
+++ b/src/definitions/Routes.ts
@@ -15,4 +15,13 @@ const ROUTES = Object.freeze({
     CLUSTER: '/cluster',
 });
 
+// Parameterised shapes are shared by React Router and usage-view matching. Keeping them
+// here means changing a route cannot silently leave instrumentation matching its old URL.
+export const ROUTE_PATTERNS = Object.freeze({
+    OPERATION_DETAILS: `${ROUTES.OPERATIONS}/:operationId`,
+    GRAPHTREE: `${ROUTES.GRAPHTREE}/:operationId?`,
+    NPE: `${ROUTES.NPE}/:filepath?`,
+    MLIR: `${ROUTES.MLIR}/:filepath?`,
+});
+
 export default ROUTES;
--- a/src/definitions/UsageEvent.ts
+++ b/src/definitions/UsageEvent.ts
@@ -62,6 +62,9 @@ export enum ReportLoadFailureReason {
  *
  * `styleguide` is excluded and stays excluded: counting a development surface would
  * pollute reach.
+ *
+ * `OPERATION_DETAILS` is owned by `view_opened`. A future `drilldown_opened` event must
+ * exclude it, or one navigation would be counted as two different actions.
  */
 export enum UsageView {
     REPORTS = 'reports',
--- a/src/definitions/UsageEvent.ts
+++ b/src/definitions/UsageEvent.ts
@@ -64,7 +64,9 @@ export enum ReportLoadFailureReason {
  * pollute reach.
  *
  * `OPERATION_DETAILS` is owned by `view_opened`. A future `drilldown_opened` event must
- * exclude it, or one navigation would be counted as two different actions.
+ * exclude it, or one navigation would be counted as two different actions. It counts
+ * once per operation viewed rather than once per visit to the surface, so its total is
+ * not comparable to the other nine and should be read per-session or deduplicated.
  */
 export enum UsageView {
     REPORTS = 'reports',
--- a/src/functions/modalRoute.ts
+++ b/src/functions/modalRoute.ts
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import type { Location } from 'react-router';
+
+interface ModalLocationState {
+    background?: Location;
+}
+
+export function getModalBackground(location: Location): Location | null {
+    const state = location.state as ModalLocationState | null;
+    return state?.background ?? null;
+}
+
+export function isModalOpen(location: Location, modalPath: string): boolean {
+    return location.pathname === modalPath && getModalBackground(location) !== null;
+}
+
+export function isReturningFromModal(previous: Location, current: Location, modalPath: string): boolean {
+    const background = getModalBackground(previous);
+
+    // The key identifies the exact history entry behind the modal. Comparing only
+    // pathnames would also suppress a real navigation to the same path with different
+    // search parameters or state.
+    return previous.pathname === modalPath && background?.key === current.key;
+}
--- a/src/functions/modalRoute.ts
+++ b/src/functions/modalRoute.ts
@@ -3,25 +3,34 @@
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
 import type { Location } from 'react-router';
+import { NAVIGATION_ITEMS } from '../definitions/NavigationItems';
 
-interface ModalLocationState {
+export interface ModalLocationState {
     background?: Location;
 }
 
-export function getModalBackground(location: Location): Location | null {
+export const MODAL_ROUTES: readonly string[] = Object.freeze(
+    NAVIGATION_ITEMS.filter((item) => item.isModal).map((item) => item.route),
+);
+
+export function getModalBackground(location: Pick<Location, 'state'>): Location | null {
     const state = location.state as ModalLocationState | null;
     return state?.background ?? null;
 }
 
-export function isModalOpen(location: Location, modalPath: string): boolean {
+export function isModalOpen(location: Pick<Location, 'pathname' | 'state'>, modalPath: string): boolean {
     return location.pathname === modalPath && getModalBackground(location) !== null;
 }
 
-export function isReturningFromModal(previous: Location, current: Location, modalPath: string): boolean {
+export function isReturningFromModal(previous: Location, current: Location): boolean {
     const background = getModalBackground(previous);
 
     // The key identifies the exact history entry behind the modal. Comparing only
     // pathnames would also suppress a real navigation to the same path with different
     // search parameters or state.
-    return previous.pathname === modalPath && background?.key === current.key;
+    return MODAL_ROUTES.includes(previous.pathname) && background?.key === current.key;
+}
+
+export function modalNavigationState(location: Location): { state: ModalLocationState } {
+    return { state: { background: location } };
 }
--- a/src/functions/viewUsage.ts
+++ b/src/functions/viewUsage.ts
@@ -2,18 +2,17 @@
 //
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
-import { matchPath } from 'react-router';
+import { type Location, matchPath } from 'react-router';
 import ROUTES, { ROUTE_PATTERNS } from '../definitions/Routes';
 import { UsageEvent, UsageView } from '../definitions/UsageEvent';
+import { isModalOpen } from './modalRoute';
 import recordUsage from './recordUsage';
 
 type RoutePath = (typeof ROUTES)[keyof typeof ROUTES];
 
-interface RouteUsageDefinition {
-    view: UsageView;
-    pattern?: string;
-    nestedView?: UsageView;
-}
+type RouteUsageDefinition =
+    | { view: UsageView; requiresModalBackground?: true }
+    | { view: UsageView; pattern: string; nestedView?: UsageView };
 
 /**
  * The single inventory of which application routes are countable views.
@@ -31,33 +30,56 @@ export const USAGE_VIEW_BY_ROUTE = Object.freeze({
     [ROUTES.TENSORS]: { view: UsageView.TENSORS },
     [ROUTES.BUFFERS]: { view: UsageView.BUFFERS },
     [ROUTES.STYLEGUIDE]: null,
-    [ROUTES.GRAPHTREE]: { view: UsageView.GRAPH, pattern: ROUTE_PATTERNS.GRAPHTREE, nestedView: UsageView.GRAPH },
+    [ROUTES.GRAPHTREE]: { view: UsageView.GRAPH, pattern: ROUTE_PATTERNS.GRAPHTREE },
     [ROUTES.PERFORMANCE]: { view: UsageView.PERFORMANCE },
-    [ROUTES.NPE]: { view: UsageView.NPE, pattern: ROUTE_PATTERNS.NPE, nestedView: UsageView.NPE },
-    [ROUTES.MLIR]: { view: UsageView.MLIR, pattern: ROUTE_PATTERNS.MLIR, nestedView: UsageView.MLIR },
-    [ROUTES.CLUSTER]: { view: UsageView.TOPOLOGY },
+    [ROUTES.NPE]: { view: UsageView.NPE, pattern: ROUTE_PATTERNS.NPE },
+    [ROUTES.MLIR]: { view: UsageView.MLIR, pattern: ROUTE_PATTERNS.MLIR },
+    [ROUTES.CLUSTER]: { view: UsageView.TOPOLOGY, requiresModalBackground: true },
 }) satisfies Readonly<Record<RoutePath, RouteUsageDefinition | null>>;
 
 const ROUTE_USAGE_ENTRIES = Object.entries(USAGE_VIEW_BY_ROUTE) as [RoutePath, RouteUsageDefinition | null][];
 
-export function getUsageView(pathname: string): UsageView | null {
+let lastSeenPathname: string | null = null;
+
+export function resetRememberedViewPathname(): void {
+    lastSeenPathname = null;
+}
+
+/**
+ * True when this pathname is a new view location. Survives a shell remount so a
+ * `ProtectedRoute` bounce cannot reopen the page it lands on.
+ */
+export function rememberViewPathname(pathname: string): boolean {
+    if (lastSeenPathname === pathname) {
+        return false;
+    }
+
+    lastSeenPathname = pathname;
+    return true;
+}
+
+export function getUsageView(location: Pick<Location, 'pathname' | 'state'>): UsageView | null {
     for (const [route, definition] of ROUTE_USAGE_ENTRIES) {
-        if (matchPath({ path: route, end: true }, pathname)) {
+        if (matchPath({ path: route, end: true }, location.pathname)) {
+            if (definition && 'requiresModalBackground' in definition && definition.requiresModalBackground) {
+                return isModalOpen(location, route) ? definition.view : null;
+            }
+
             return definition?.view ?? null;
         }
 
-        if (
-            definition?.nestedView &&
-            definition.pattern &&
-            matchPath(
+        if (definition && 'pattern' in definition) {
+            const matched = matchPath(
                 {
                     path: definition.pattern,
                     end: true,
                 },
-                pathname,
-            )
-        ) {
-            return definition.nestedView;
+                location.pathname,
+            );
+
+            if (matched) {
+                return definition.nestedView ?? definition.view;
+            }
         }
     }
 
--- a/src/functions/viewUsage.ts
+++ b/src/functions/viewUsage.ts
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { matchPath } from 'react-router';
+import ROUTES, { ROUTE_PATTERNS } from '../definitions/Routes';
+import { UsageEvent, UsageView } from '../definitions/UsageEvent';
+import recordUsage from './recordUsage';
+
+type RoutePath = (typeof ROUTES)[keyof typeof ROUTES];
+
+interface RouteUsageDefinition {
+    view: UsageView;
+    pattern?: string;
+    nestedView?: UsageView;
+}
+
+/**
+ * The single inventory of which application routes are countable views.
+ *
+ * `null` is an explicit exclusion rather than an omitted key, so adding a route
+ * without deciding how it should be counted fails type-checking.
+ */
+export const USAGE_VIEW_BY_ROUTE = Object.freeze({
+    [ROUTES.HOME]: { view: UsageView.REPORTS },
+    [ROUTES.OPERATIONS]: {
+        view: UsageView.OPERATIONS,
+        pattern: ROUTE_PATTERNS.OPERATION_DETAILS,
+        nestedView: UsageView.OPERATION_DETAILS,
+    },
+    [ROUTES.TENSORS]: { view: UsageView.TENSORS },
+    [ROUTES.BUFFERS]: { view: UsageView.BUFFERS },
+    [ROUTES.STYLEGUIDE]: null,
+    [ROUTES.GRAPHTREE]: { view: UsageView.GRAPH, pattern: ROUTE_PATTERNS.GRAPHTREE, nestedView: UsageView.GRAPH },
+    [ROUTES.PERFORMANCE]: { view: UsageView.PERFORMANCE },
+    [ROUTES.NPE]: { view: UsageView.NPE, pattern: ROUTE_PATTERNS.NPE, nestedView: UsageView.NPE },
+    [ROUTES.MLIR]: { view: UsageView.MLIR, pattern: ROUTE_PATTERNS.MLIR, nestedView: UsageView.MLIR },
+    [ROUTES.CLUSTER]: { view: UsageView.TOPOLOGY },
+}) satisfies Readonly<Record<RoutePath, RouteUsageDefinition | null>>;
+
+const ROUTE_USAGE_DEFINITIONS = Object.values(USAGE_VIEW_BY_ROUTE) as (RouteUsageDefinition | null)[];
+
+export function getUsageView(pathname: string): UsageView | null {
+    const exactDefinition = USAGE_VIEW_BY_ROUTE[pathname as RoutePath];
+
+    if (exactDefinition !== undefined) {
+        return exactDefinition?.view ?? null;
+    }
+
+    for (const definition of ROUTE_USAGE_DEFINITIONS) {
+        if (
+            definition?.nestedView &&
+            definition.pattern &&
+            matchPath(
+                {
+                    path: definition.pattern,
+                    end: true,
+                },
+                pathname,
+            )
+        ) {
+            return definition.nestedView;
+        }
+    }
+
+    return null;
+}
+
+export function recordViewOpened(view: UsageView): void {
+    recordUsage({ event: UsageEvent.VIEW_OPENED, details: { view } });
+}
--- a/src/functions/viewUsage.ts
+++ b/src/functions/viewUsage.ts
@@ -38,16 +38,14 @@ export const USAGE_VIEW_BY_ROUTE = Object.freeze({
     [ROUTES.CLUSTER]: { view: UsageView.TOPOLOGY },
 }) satisfies Readonly<Record<RoutePath, RouteUsageDefinition | null>>;
 
-const ROUTE_USAGE_DEFINITIONS = Object.values(USAGE_VIEW_BY_ROUTE) as (RouteUsageDefinition | null)[];
+const ROUTE_USAGE_ENTRIES = Object.entries(USAGE_VIEW_BY_ROUTE) as [RoutePath, RouteUsageDefinition | null][];
 
 export function getUsageView(pathname: string): UsageView | null {
-    const exactDefinition = USAGE_VIEW_BY_ROUTE[pathname as RoutePath];
+    for (const [route, definition] of ROUTE_USAGE_ENTRIES) {
+        if (matchPath({ path: route, end: true }, pathname)) {
+            return definition?.view ?? null;
+        }
 
-    if (exactDefinition !== undefined) {
-        return exactDefinition?.view ?? null;
-    }
-
-    for (const definition of ROUTE_USAGE_DEFINITIONS) {
         if (
             definition?.nestedView &&
             definition.pattern &&
--- a/src/hooks/useRecordViewOpened.ts
+++ b/src/hooks/useRecordViewOpened.ts
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { useEffect, useRef } from 'react';
+import { type Location, useLocation } from 'react-router';
+import ROUTES from '../definitions/Routes';
+import { isModalOpen, isReturningFromModal } from '../functions/modalRoute';
+import { getUsageView, recordViewOpened } from '../functions/viewUsage';
+
+/**
+ * Record countable route changes from the one shell shared by every view.
+ */
+export default function useRecordViewOpened(): void {
+    const location = useLocation();
+    const previousLocation = useRef<Location | null>(null);
+
+    useEffect(() => {
+        const previous = previousLocation.current;
+        previousLocation.current = location;
+
+        // React StrictMode repeats effect setup with the same location. Search/hash-only
+        // changes likewise do not open a different view.
+        if (previous?.pathname === location.pathname) {
+            return;
+        }
+
+        // Closing topology restores the history entry which remained mounted beneath it;
+        // that view was never reopened.
+        if (previous && isReturningFromModal(previous, location, ROUTES.CLUSTER)) {
+            return;
+        }
+
+        // `/cluster` without background state renders no topology overlay, so counting
+        // such a deep link would claim a view opened when the user received a blank route.
+        if (location.pathname === ROUTES.CLUSTER && !isModalOpen(location, ROUTES.CLUSTER)) {
+            return;
+        }
+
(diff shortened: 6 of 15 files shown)


src/definitions/Routes.ts
```diff
@@ -15,4 +15,13 @@ const ROUTES = Object.freeze({
     CLUSTER: '/cluster',
 });
 
+// Parameterised shapes are shared by React Router and usage-view matching. Keeping them
+// here means changing a route cannot silently leave instrumentation matching its old URL.
+export const ROUTE_PATTERNS = Object.freeze({
+    OPERATION_DETAILS: `${ROUTES.OPERATIONS}/:operationId`,
+    GRAPHTREE: `${ROUTES.GRAPHTREE}/:operationId?`,
```
aidemsined, line 22: > [!CAUTION]
> **Definite:** the optional-parameter half of these patterns is unpinned — dropping a `?` breaks three bare routes and no spec fails.

**Mutation:** change `'/npe/:filepath?'` to `'/npe/:filepath'` (or the same on `GRAPHTREE` / `MLIR`). Bare `/graphtree`, `/npe` and `/mlir` then match no route object and those three nav items render a blank page. The suite stays green.

Why nothing catches it: `tests/viewUsage.spec.ts:26-36` reaches those views through the **static** `matchPath({ path: route, end: true })` branch at `viewUsage.ts:45`, so the pattern is never consulted for them, and `${ROUTES.NPE}/trace.json` matches with or without the `?`. Meanwhile `grep -rn routeObjectList tests/` returns only `tests/routeRequirements.spec.ts:16`, which imports `RouteRequirements` and never the route array — so nothing in the suite renders or matches `routeObjectList` at all.

The comment two lines up is the reason this matters: "Keeping them here means changing a route cannot silently leave instrumentation matching its old URL." The consolidation is right, but only the instrumentation consumer is tested; the routing consumer that shares the constant isn't.

**Suggested action:** assert the routing side — for each `ROUTE_PATTERNS` entry, that `matchRoutes(routeObjectList, base)` and `matchRoutes(routeObjectList, base + '/x')` both resolve. That pins optionality and the shared-constant guarantee in one test.

src/functions/viewUsage.ts
```diff
@@ -0,0 +1,69 @@
...
+    nestedView?: UsageView;
+}
+
+/**
+ * The single inventory of which application routes are countable views.
+ *
+ * `null` is an explicit exclusion rather than an omitted key, so adding a route
+ * without deciding how it should be counted fails type-checking.
+ */
+export const USAGE_VIEW_BY_ROUTE = Object.freeze({
```
aidemsined, line 23: > [!CAUTION]
> **Definite:** route metadata is now hand-maintained in four places, and this repo already rejected that shape two lists over.

Adding route #11 requires four coordinated edits: a value in `Routes.ts`, an element entry in `routeObjectList.tsx`, a nav descriptor in `NavigationItems.ts`, and a usage decision here. Only the last is checked at all, and only at runtime by `viewUsage.spec.ts:14` — `routeObjectList` entries are free strings that nothing ties back to `ROUTES` or `ROUTE_PATTERNS`.

The precedent is in the file this PR edits: `routeObjectList.tsx:83-90` derives `RouteRequirements` from `NAVIGATION_ITEMS` with the comment *"Hand-maintaining a second copy drifts silently in both directions"*. So the question was live and answered here with a fifth independent list instead of a descriptor.

**Suggested action:** a single route-descriptor array — `{ path, pattern?, element, usageView: UsageView | null, nav?: {...} }` — living where `routeObjectList` already is, with `routeObjectList`, `RouteRequirements`, `USAGE_VIEW_BY_ROUTE` and the nav rail's `route` all `map`/`filter`ed out of it. That's the same technique `RouteRequirements` uses today. If that's too much for this PR, the minimum is to key `USAGE_VIEW_BY_ROUTE` off the same structure `routeObjectList` consumes, so the two cannot disagree about what routes exist.

To be clear about what *is* solid here: I tested the `satisfies` guarantee directly because it looked like it might not hold without `as const`, and it does — `Object.freeze`'s primitive-constrained overload preserves the literal types, the missing-key error fires, and `RoutePath` rejects an arbitrary string. The docblock's claim is accurate for a new `ROUTES` entry.

src/hooks/useRecordViewOpened.ts
```diff
@@ -0,0 +1,45 @@
...
+import ROUTES from '../definitions/Routes';
+import { isModalOpen, isReturningFromModal } from '../functions/modalRoute';
+import { getUsageView, recordViewOpened } from '../functions/viewUsage';
+
+/**
+ * Record countable route changes from the one shell shared by every view.
+ */
+export default function useRecordViewOpened(): void {
+    const location = useLocation();
+    const previousLocation = useRef<Location | null>(null);
```
aidemsined, line 15: > [!WARNING]
> **Possible:** the dedupe ref is discarded whenever `ProtectedRoute` bounces, so the landing view is recorded a second time.

`previousLocation` is a ref inside `Layout`, and `Layout` is a child of `ProtectedRoute` (`main.tsx:24-26`). When the active report a route needs is missing, `ProtectedRoute:34-56` returns `<Navigate to={ROUTES.HOME} replace />` **instead of** `children` — a different element type, so React unmounts `Layout` and the ref goes with it. The remount at `/` then records `view_opened: reports` again, for a view the user never left.

**Reachable sequence:** on `/performance`, go Home, switch to a report with no performance file, press browser Back. The guard correctly emits nothing for `/performance`, but the `Layout` remount emits a second `reports`. Back bypasses the nav rail's disabled state entirely. `RemoteSyncConfigurator:141,147` clears `active_report.performance_name` to `''` mid-session, which is exactly what `ProtectedRoute:46` reads.

Nothing covers this: there's no `ProtectedRoute` spec, and `useRecordViewOpened.spec.tsx` renders the hook directly under `MemoryRouter`, never through a gate that can unmount it.

**Suggested action:** hold the last-recorded pathname in module scope in `viewUsage.ts`, or mount the hook above `ProtectedRoute`, so a shell remount cannot re-open the view it lands on.

src/functions/viewUsage.ts
```diff
@@ -0,0 +1,69 @@
...
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import { matchPath } from 'react-router';
+import ROUTES, { ROUTE_PATTERNS } from '../definitions/Routes';
+import { UsageEvent, UsageView } from '../definitions/UsageEvent';
+import recordUsage from './recordUsage';
+
+type RoutePath = (typeof ROUTES)[keyof typeof ROUTES];
+
+interface RouteUsageDefinition {
```
aidemsined: > [!WARNING]
> **Possible:** `pattern` and `nestedView` are independently optional, so a half-filled definition silently stops counting a URL while passing every check.

`getUsageView` defends with `definition?.nestedView && definition.pattern`, so a definition carrying one without the other doesn't crash — it just never matches.

**Failure scenario:** someone adds `TENSOR_DETAILS: \`${ROUTES.TENSORS}/:tensorId\`` to `ROUTE_PATTERNS` plus a matching `routeObjectList` entry, but forgets to add `pattern`/`nestedView` to the `[ROUTES.TENSORS]` definition. `/tensors/5` renders the page and emits nothing. The `satisfies` passes (the `TENSORS` key exists) and `viewUsage.spec.ts:16` passes (the route key set is unchanged).

That's the drift the docblock above claims to prevent. "Adding a route without deciding how it should be counted fails type-checking" is true for a new `ROUTES` entry — I verified the `satisfies` really does catch that — but false for a new `ROUTE_PATTERNS` entry.

**Suggested action:** make the pair inseparable, `type RouteUsageDefinition = { view: UsageView } | { view: UsageView; pattern: string; nestedView: UsageView }`, and add a spec asserting `Object.values(ROUTE_PATTERNS)` is a subset of the definitions' `pattern` values. Related and cheap: `nestedView` equals `view` in 3 of the 4 parameterised entries, so defaulting it to `view` when a `pattern` is present would drop three redundant fields.

src/hooks/useRecordViewOpened.ts
```diff
@@ -0,0 +1,45 @@
...
+
+        // React StrictMode repeats effect setup with the same location. Search/hash-only
+        // changes likewise do not open a different view.
+        if (previous?.pathname === location.pathname) {
+            return;
+        }
+
+        // Closing topology restores the history entry which remained mounted beneath it;
+        // that view was never reopened.
+        if (previous && isReturningFromModal(previous, location, ROUTES.CLUSTER)) {
```
aidemsined: > [!WARNING]
> **Possible:** the modal handling hardcodes `ROUTES.CLUSTER` while `isModal` is a general capability, so a second modal route would re-count its background view on close.

`modalRoute.ts` is correctly parameterised on `modalPath`, but this is its only caller and it pins the value. Meanwhile `NavigationItems.ts:35` declares `isModal?: boolean` as a general flag and `useMainNavigationItems.ts:105` honours it generically.

**Failure scenario:** a second `isModal: true` item at `/foo`. Open it from `/tensors` → recorded. Close with `navigate(-1)` → guard B evaluates `previous.pathname === ROUTES.CLUSTER` as false, so it doesn't suppress, and `getUsageView('/tensors')` records `tensors` a second time. That breaks the one invariant #1832 states explicitly — "the back-navigation on close must not be counted as opening the background view a second time" — and nothing type-checks or tests against it.

**Suggested action:** derive the modal routes from the descriptor list — `NAVIGATION_ITEMS.filter((item) => item.isModal).map((item) => item.route)` — and check `previous.pathname` against that set. `routeObjectList.tsx:83-89` already derives `RouteRequirements` from `NAVIGATION_ITEMS` with a docblock making exactly this argument about hand-maintained second copies.

Worth noting the `key`-based comparison in `isReturningFromModal` is the right discriminator and I couldn't break it — `navigate(-1)` restores the same key while a fresh push to the same path gets a new one, and the spec at `useRecordViewOpened.spec.tsx:169` pins precisely that. It's only the route selection that's narrow.

src/components/Layout.tsx
```diff
@@ -29,13 +31,15 @@ const BounceIn = cssTransition({
 
 function Layout() {
     const location = useLocation();
-    const state = location.state as { background?: Location };
+    const isTopologyOpen = isModalOpen(location, ROUTES.CLUSTER);
```
aidemsined, line 34: > [!WARNING]
> **Possible:** this call site is the half the PR actually changed behaviourally, and it's the half with no coverage.

**Mutation:** `isModalOpen(location, ROUTES.CLUSTER)` → `location.pathname === ROUTES.CLUSTER`, dropping the background-state requirement the pre-PR line carried. Every spec passes. The regression is that a bare `/cluster` deep link paints the topology overlay over an empty outlet, since `routeObjectList.tsx:65-67` gives `/cluster` `element: null`.

Why nothing catches it: `Layout.spec.tsx:39` mocks `ClusterRenderer` as `() => null`, so the gate is unobservable, and `Layout.spec.tsx` is the only spec that renders `Layout`. The equivalent condition *inside the hook* is covered ("does not count a topology pathname which renders no overlay") — so it's specifically the render gate that's unpinned.

**Suggested action:** stub `ClusterRenderer` as an identifiable marker and assert it's absent for `/cluster` with no state and present with `state.background`.

Separately, the refactor itself is a genuine improvement worth noting: neither `Layout` nor `ModalAwareOutlet` imported `Location` from `react-router`, so the old inline casts were annotating `background` as the **DOM** `Location`. `modalRoute.ts:5` imports the right type, which fixes a latent typing bug alongside removing the third open-coded cast.

tests/useRecordViewOpened.spec.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    afterEach(() => {
+        cleanup();
+    });
+
+    it('records the initial view once under StrictMode', () => {
+        renderRecorder(ROUTES.OPERATIONS);
+
+        expect(recordViewOpened).toHaveBeenCalledTimes(1);
+        expect(recordViewOpened).toHaveBeenCalledWith(UsageView.OPERATIONS);
```
aidemsined, line 113: > [!WARNING]
> **Possible:** the null guard is uncovered at the hook layer, and the plausible mistake is type-valid.

**Mutation:** replace the `if (view !== null)` block with `recordViewOpened(getUsageView(location.pathname) ?? UsageView.REPORTS)`. That's a "sensible default" mistake someone could genuinely make, it type-checks, and **every assertion in this file still passes** — the harness has buttons for `/tensors`, `/operations/2`, `/graphtree/2`, `/cluster` and a query change, but none for `ROUTES.STYLEGUIDE` or an unknown path.

`viewUsage.spec.ts:49-54` covers `getUsageView` *returning* `null`, but nothing covers the hook *acting* on it. If it regressed, `/styleguide` and every 404 would land in `reports` — the one view both docblocks insist must stay clean, and the AC that says styleguide is excluded.

**Suggested action:** add two harness buttons (`ROUTES.STYLEGUIDE` and `/does-not-exist`) and assert `recordViewOpened` isn't called for either.

The rest of this spec is strong, for the record — I checked the mutants for all three guards and each is caught, and `:169` ("does not mistake a new background-path entry for closing the topology modal") pins the history-`key` comparison, which is a deliberately chosen non-obvious mutant. `Layout.spec.tsx:93` independently proving the StrictMode double-invocation is real in that harness is what makes the once-ness assertion meaningful rather than incidental.

src/functions/viewUsage.ts
```diff
@@ -0,0 +1,69 @@
...
+        nestedView: UsageView.OPERATION_DETAILS,
+    },
+    [ROUTES.TENSORS]: { view: UsageView.TENSORS },
+    [ROUTES.BUFFERS]: { view: UsageView.BUFFERS },
+    [ROUTES.STYLEGUIDE]: null,
+    [ROUTES.GRAPHTREE]: { view: UsageView.GRAPH, pattern: ROUTE_PATTERNS.GRAPHTREE, nestedView: UsageView.GRAPH },
+    [ROUTES.PERFORMANCE]: { view: UsageView.PERFORMANCE },
+    [ROUTES.NPE]: { view: UsageView.NPE, pattern: ROUTE_PATTERNS.NPE, nestedView: UsageView.NPE },
+    [ROUTES.MLIR]: { view: UsageView.MLIR, pattern: ROUTE_PATTERNS.MLIR, nestedView: UsageView.MLIR },
+    [ROUTES.CLUSTER]: { view: UsageView.TOPOLOGY },
```
aidemsined: > [!NOTE]
> **Possible:** this entry is only correct in conjunction with a guard that lives in the hook, and #1833 is about to be the second caller.

`getUsageView('/cluster')` returns `TOPOLOGY` unconditionally. The knowledge that a bare `/cluster` renders nothing — `element: null` and no `state.background` — lives only at `useRecordViewOpened.ts:36`. So the mapping that the module docblock calls "the single inventory" has one entry whose correctness is elsewhere.

`UsageEvent.ts:58-61` names `view_engaged` (#1833) as the next consumer of this same vocabulary. If it calls `getUsageView(location.pathname)` from its own hook, it counts every blank-route `/cluster` deep link as a topology view — the one view whose docstring says every navigation to it "is a deliberate act rather than something a redirect can produce".

**Suggested action:** put the requirement in the mapping — `[ROUTES.CLUSTER]: { view: UsageView.TOPOLOGY, requiresModalBackground: true }` — and have `getUsageView` take the `Location` so it can honour it. The hook is then left with only the dedupe logic, and the second consumer inherits the rule for free. Much cheaper now than after #1833 exists.

src/functions/modalRoute.ts
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import type { Location } from 'react-router';
+
+interface ModalLocationState {
```
aidemsined: > [!NOTE]
> **Possible:** this module centralised every *reader* of the modal location state but not the one *writer*, so it owns half its own invariant.

The three read sites were consolidated — `Layout.tsx:34`, `ModalAwareOutlet.tsx:11`, `useMainNavigationItems.ts:77` — which is the valuable part of the refactor. But `useMainNavigationItems.ts:105` still constructs `{ state: { background: location } }` as a bare object literal, and `ModalLocationState` is module-private here.

So nothing type-checks the producer against the shape the consumers parse. Renaming `background`, or adding a field the reader needs, changes this module and leaves the only writer silently emitting the old shape — and the readers all fail soft (`background?.key`, `?? null`), so it would present as "modal suppression stopped working" rather than a type error.

**Suggested action:** export the state type and a small `modalNavigationState(location): { state: ModalLocationState }` from here, and call it at `useMainNavigationItems.ts:105`. That makes this module the whole story for the convention rather than the read half.

Also worth noting while here: the `previous.pathname === modalPath` conjunct at :26 is unreachable today — `{ state: { background } }` is written in exactly one place, gated on `item.isModal`, and Topology is the only such item, so no location other than `/cluster` ever carries a background. It's harmless and arguably future-proofing, but a reader can't tell which.

src/definitions/UsageEvent.ts
```diff
@@ -62,6 +62,9 @@ export enum ReportLoadFailureReason {
  *
  * `styleguide` is excluded and stays excluded: counting a development surface would
  * pollute reach.
+ *
+ * `OPERATION_DETAILS` is owned by `view_opened`. A future `drilldown_opened` event must
```
aidemsined, line 66: > [!NOTE]
> **Possible:** this settles drilldown *ownership* but omits the magnitude caveat, in a docstring whose whole purpose is preventing misreads.

`operation_details` is the only per-item route in the app, and per #1832's AC a parameter change counts as a view open.

**Consequence:** a user arrowing through 40 operations with `OperationDetailsNavigation.tsx:58-68`'s ArrowLeft/ArrowRight handler produces 40 `view_opened{operation_details}` against 1 `view_opened{tensors}` for a comparable session on Tensors. Anyone answering #1819's Q4 by comparing per-view counts concludes operation details is an order of magnitude more popular than it is.

The same docstring spends six lines (`usage.py:258-263`) on `TOPOLOGY`'s denominator for exactly this class of reason, so the omission reads as out of character rather than out of scope.

**Suggested action:** one sentence, mirrored in `usage.py:272` — `operation_details` counts once per operation viewed rather than once per visit to the surface, so its total is not comparable to the other nine and should be read per-session or deduplicated.

aidemsined (commented):
The exactly-once machinery holds up. I went after it specifically and couldn't break it: StrictMode's double-invoked effect, a re-render, query-only and hash-only changes, a modal over a route, re-navigating to the same route, back/forward in both directions (including open→forward→back→back), and the initial mount all behave correctly, and `location.key` collisions are impossible because only the single initial entry uses `'default'`. `getUsageView` and React Router's `matchRoutes` agree on all 16 pathnames I probed against the installed react-router, including `/operations/`, `/TENSORS`, `/npe/dir/trace.json` and `/does-not-exist` — so no 404 can emit a bogus view. Privacy is clean: `getUsageView` discards `matchPath`'s params, so the pathname that carries report names and operation ids never escapes its parameter. Vocabulary and `details` keys match the backend exactly.

Commit 2 is a real correctness fix rather than churn — the exact-key lookup it replaced would have returned `null` for `/operations/` and `/TENSORS`, both of which React Router routes to a real page. I confirmed both.

The findings cluster on one theme: the shared constants and derived maps this PR introduces are pinned on the instrumentation side and unpinned on the routing side.

## Unanchored findings

- **Possible — route matching is now three implementations, and the fragile two survived.** `getUsageView` uses `matchPath`, but `useMainNavigationItems.ts:62`'s `isNestedMatch` still uses `candidate.includes(path)` with its own caveat ("no route in `ROUTES` is a substring of another") — an invariant the new machinery makes unnecessary. Two more hand-rolled matchers for the shape `ROUTE_PATTERNS.OPERATION_DETAILS` now owns are untouched: `FooterInfobar.tsx:90` (`/^\/operations\/\d+$/`, path literal inlined against `CONVENTIONS.md:687-689`) and `RangeSlider.tsx:68` (`pathname.includes(...)`). That undercuts the rationale at `Routes.ts:18-19` — changing the operations route still silently leaves two matchers on the old URL. Exporting one `matchesRoute(pathname, pathOrPattern)` and converting all three would make the constant's guarantee real; reasonable as a follow-up rather than here.
- **Possible — the inventory is data living in `functions/`.** `AGENTS.md:197` and `CONVENTIONS.md:655` put route maps in `definitions/` (with `Routes.ts` as the named example). `USAGE_VIEW_BY_ROUTE` is a frozen route-keyed map and `RouteUsageDefinition` a plain interface, both one directory from the two definition modules they're built from. The practical cost is that the module mixes inventory, matcher and network sender, so a parity test, a lint rule or #1833 wanting only the mapping also pulls in `recordUsage`. Moving the two data exports next to `Routes.ts`/`UsageEvent.ts` and keeping `getUsageView`/`recordViewOpened` here would split it cleanly.
- **Possible — static routes and their drill-downs are counted inconsistently under the same missing precondition.** `/operations` with no profiler report bounces to Home and emits nothing, while `/operations/42` and `/graphtree/42` are absent from `RouteRequirements`, render, and do emit. Reachable by bookmark, deep link, or Back after a report switch. Worth deciding once: either extend the guard to the parameterised patterns (they're shared constants now, so it's a one-line match), or state in `viewUsage.ts` that drill-downs count regardless of report state.

## General observations

- `tests/viewUsage.spec.ts:49` — the null-expecting rows pin `end: true` on the *static* match but not the nested one. Dropping `end: true` from the nested `matchPath` would make `/operations/42/extra` report `operation_details`, though `matchRoutes` renders nothing for it. Adding `/operations/42/extra` and `/npe/dir/trace.json` closes it — I verified both are `null` under react-router 7.18.2.
- `tests/recordUsage.spec.ts:486` — the only change to this file provides no coverage of this PR. It sits in `describe('recordUsage gating')` asserting `post`/`sendBeacon` weren't called, and passes identically with the original `REPORT_LOADED`; `VIEW_OPENED` was already used at 27 other call sites there. If it was meant to satisfy #1832's "nothing emitted when recording is disabled", it doesn't — that assertion is about the transport, not about `recordViewOpened`. Either revert the line or put the AC's assertion in `viewUsage.spec.ts` against the real `recordUsage`.
- `AGENTS.md:199` — the route-authoring checklist is now incomplete: a new route also needs a `USAGE_VIEW_BY_ROUTE` decision (enforced by `satisfies`, so at least it fails loudly) and, if parameterised, a `ROUTE_PATTERNS` entry (unenforced). The usage-recording section at :186 lists four non-inferable invariants for `recordUsage.ts`; this PR adds three more a reader can't infer — exactly-once via a ref rather than state, modal-close suppression by history `key`, bare-`/cluster` exclusion — and documents none.
- `CONVENTIONS.md:701-712` — the routing convention's code sample still shows `stripFirstSlash(\`${ROUTES.OPERATIONS}/:operationId\`)` inline and claims that keeps `ROUTES` the single source of truth, but `routeObjectList.tsx:33` now reads `ROUTE_PATTERNS.OPERATION_DETAILS`. It's a block a reader would copy.
- `viewUsage.ts:24` — `USAGE_VIEW_BY_ROUTE` maps to a *definition*, not a view; `AGENTS.md:214` asks for `valueByKey` named after the relation encoded, so `USAGE_DEFINITION_BY_ROUTE`.
- `useRecordViewOpened.ts:22` — the comment gives two of the guard's three effects. It also drops a deliberate re-navigation to the same pathname (a nav re-click, or the logo `<Link>` while already on `/`), which is a defensible product decision but shouldn't be something a reader derives.
- `useRecordViewOpened.ts:36` — `location.pathname === ROUTES.CLUSTER && !isModalOpen(location, ROUTES.CLUSTER)` re-tests the pathname `isModalOpen` already tests, giving a double negative that reads as redundant but is load-bearing. `getModalBackground(location) === null` says it without the trap.
- `useRecordViewOpened.ts:14` — default-exported hook; `CONVENTIONS.md:1403` prefers named exports for hooks, and the sibling this PR edits (`useMainNavigationItems`) is named-exported. Ten pre-existing hooks do default-export, so it's soft.
- `recordUsage.ts:238` — the teardown comment ("a no-op on an empty buffer — which is the StrictMode mount/unmount/mount case in dev") is falsified by this PR: hook order in `Layout` puts `initUsageRecording` before `useRecordViewOpened`, so StrictMode's destroy pass runs the cleanup with the initial `view_opened` already buffered. Dev-only, one immediate single-event POST per boot, no data loss — but the comment is now wrong.
- `useRecordViewOpened`'s location bookkeeping is private to a `void` hook, so #1833 must either duplicate the ref-and-compare loop (two notions of "previous location" in one shell) or be bolted inside. Extracting `useCountableView(): UsageView | null` would let both subscribe to one value.
- No coverage for back/forward between two *ordinary* routes — the harness's only `navigate(-1)` is labelled "Close topology", so "back from `/tensors` to `/operations` records `operations` again" is unasserted, and that's where guard B's key comparison is closest to misfiring. A hash-only case would also close the gap between the comment and the tests.

## Worth calling out as good

`tests/Layout.spec.tsx:72-82` is the right test at the right layer — it renders the real `Layout` with real `viewUsage`/`modalRoute` and only `recordUsage` mocked, then asserts one call with the exact wire payload. Deleting the hook from `Layout` fails it, and the sibling at :93 independently proves the StrictMode double-invocation is real in that harness, which is what makes the once-ness assertion exercise the dedupe rather than a harness that never double-invokes. And `useRecordViewOpened.spec.tsx:169` picks a genuinely non-obvious mutant — weakening `isReturningFromModal` to compare pathnames instead of history keys — and catches it.

Extracting `modalRoute.ts` also fixed a latent typing bug in passing: neither `Layout` nor `ModalAwareOutlet` imported `Location` from `react-router`, so the old inline casts were annotating `background` as the **DOM** `Location`.

aidemsined (changes requested):

src/definitions/Routes.ts
dcblundell: Addressed in 60722e09. `routePatterns.spec.ts` now exercises every shared pattern through `matchRoutes`, requiring optional patterns to resolve both the bare and parameterised paths.

src/functions/viewUsage.ts
dcblundell: Thanks. I kept the current separation in this PR: `ROUTES` remains exhaustive in the usage map, every `ROUTE_PATTERNS` value is now pinned to a usage definition, and routing behaviour is exercised with `matchRoutes`. Consolidating route elements, navigation metadata, requirements, and usage into one descriptor would be a broader routing/navigation refactor, so I am leaving that outside #1832.

src/hooks/useRecordViewOpened.ts
dcblundell: Addressed in 60722e09. The last-seen pathname is now module-scoped, so it survives a `Layout` remount caused by `ProtectedRoute`; the remount case has regression coverage.

src/functions/viewUsage.ts
dcblundell: Addressed in 60722e09. Parameterised definitions are now a discriminated union, `nestedView` defaults to `view`, and a spec requires every `ROUTE_PATTERNS` value to appear in the usage definitions.

src/hooks/useRecordViewOpened.ts
dcblundell: Addressed in 60722e09. Modal routes are derived from `NAVIGATION_ITEMS`, and modal-return suppression checks that shared set rather than hardcoding `ROUTES.CLUSTER`.

src/components/Layout.tsx
dcblundell: Addressed in 60722e09. `Layout.spec.tsx` now renders an identifiable cluster marker and checks both the bare `/cluster` deep link and the background-state overlay case.

tests/useRecordViewOpened.spec.tsx
dcblundell: Addressed in 60722e09. The hook spec now navigates to both `ROUTES.STYLEGUIDE` and an unknown route and asserts that neither records a view.

src/functions/viewUsage.ts
dcblundell: Addressed in 60722e09. The cluster definition now declares `requiresModalBackground`, and `getUsageView` accepts the location so every consumer inherits the blank-route exclusion.

dcblundell: Addressed in 60722e09. The cluster definition now declares `requiresModalBackground`, and `getUsageView` accepts the location so every consumer inherits the blank-route exclusion.

src/functions/modalRoute.ts
dcblundell: Addressed in 60722e09. `ModalLocationState` and `modalNavigationState(location)` are exported from `modalRoute.ts`, and `useMainNavigationItems` now uses that typed writer.

src/definitions/UsageEvent.ts
dcblundell: Addressed in 60722e09. The per-operation magnitude caveat is now documented in both the TypeScript and Python `UsageView` descriptions.

tests/viewUsage.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+describe('usage view route mapping', () => {
+    it('makes an explicit decision for every route', () => {
+        expect(new Set(Object.keys(USAGE_VIEW_BY_ROUTE))).toEqual(new Set(Object.values(ROUTES)));
+    });
+
+    it('uses every parameterised pattern in a route definition', () => {
+        const patterns = Object.values(USAGE_VIEW_BY_ROUTE)
+            .filter((definition): definition is { view: UsageView; pattern: string } =>
```
aidemsined: > [!CAUTION]
> **Definite:** this file has three TypeScript errors, and CI cannot see them.

`pnpm lint:ts` on this head:

```
tests/viewUsage.spec.ts(26,49): error TS2677: A type predicate's type must be assignable to its parameter's type.
  Type '{ view: UsageView; pattern: string; }' is not assignable to type
  '{ view: UsageView.REPORTS; } | { view: UsageView.OPERATIONS; pattern: "/operations/:operationId"; … } | … | null'.
tests/viewUsage.spec.ts(29,34): error TS18047: 'definition' is possibly 'null'.
tests/viewUsage.spec.ts(29,45): error TS2339: Property 'pattern' does not exist on type '…'.
```

`origin/dev` is clean — **0 errors** — so these arrive with this branch. They come from the fix for the `RouteUsageDefinition` finding: the new discriminated union is narrower than the predicate here asserts, so the guard can't be expressed against it as written.

**Why CI is green anyway:** `.github/workflows/lint-and-test.yml` runs `pnpm lint` (eslint, on changed files only) and `pnpm test`. Nothing in any workflow runs `tsc`, even though the repo has a `lint:ts` script and `tsconfig.json` includes `tests/**/*`. So all 14 checks pass with the type errors in place, and `pnpm test` passes too — vitest transpiles without typechecking.

**Suggested action:** narrow the predicate to the union member rather than a structural shape — something like `(d): d is Extract<RouteUsageDefinition, { pattern: string }> => d !== null && 'pattern' in d` — which also removes the `possibly 'null'` and the missing-property errors at :29.

Worth raising as its own issue regardless of this PR: a repo with a `lint:ts` script and no CI step that runs it will keep accumulating this silently. Happy to file that if you'd rather keep it off this thread.

aidemsined (commented):
Re-checked the fixes by re-running my own mutations against them rather than reading the diff, and both Definites are properly closed:

- **the `?` optionality** — dropping it from the `NPE` pattern now fails `keeps /npe/:filepath matching both its bare path and a parameterised child`. `routePatterns.spec.ts` going through `matchRoutes` is the right shape; it pins the routing consumer that was the actual gap.
- **the `ProtectedRoute` remount** — reverting `rememberViewPathname` to the old ref-only comparison now fails `does not re-record the same pathname after the recorder remounts`. Module scope is the right call, and the regression coverage is real rather than structural.

The rest of the round is good too — `requiresModalBackground` moving the blank-route rule into the mapping means #1833 inherits it, deriving modal routes from `NAVIGATION_ITEMS` removes the hardcoded route, and the exported `modalNavigationState` closes the reader/writer gap. Full suite is 2420 passing across 213 files.

I've flagged one thing above that arrived with the fix round and that CI structurally cannot catch. Agreed on leaving the single-descriptor consolidation outside #1832 — the scope argument is fair, and pinning every `ROUTE_PATTERNS` value to a usage definition covers the drift that mattered most.

tests/viewUsage.spec.ts
aidemsined: Filed the CI gap as #1969 so it is off this thread — `dev` is clean today, so it is prevention rather than cleanup, and the step would have caught these three before review.

The three errors here are still worth fixing in this PR though, since they arrive with it.

dcblundell: Addressed in d662f646 by replacing the invalid structural type predicate with a `flatMap` guard over the inferred union. `pnpm lint:ts`, ESLint, and the 22 `viewUsage` tests now pass.

PR #1972, opened 2026-09-02
Add TypeScript typechecking to PR CI

### LINKED ISSUE #1969: CI never runs tsc, so type errors merge with all checks green
## Problem

`package.json` defines `lint:ts` (`pnpm exec tsc --noEmit`) and `tsconfig.json` includes `src/**/*`, `tests/**/*` and the three config files — but **no CI workflow ever runs it**. Type errors merge silently with every check green.

Found while reviewing #1963, which is the live proof: its head has three real type errors and all 14 checks pass.

```
tests/viewUsage.spec.ts(26,49): error TS2677: A type predicate's type must be assignable to its parameter's type.
tests/viewUsage.spec.ts(29,34): error TS18047: 'definition' is possibly 'null'.
tests/viewUsage.spec.ts(29,45): error TS2339: Property 'pattern' does not exist on type '…'.
```

## Why nothing catches it

| CI step | what it actually does |
|---|---|
| `lint-and-test.yml` → "Lint changed files" | `pnpm lint` = eslint. Type-aware rules only; eslint does not report `tsc` diagnostics |
| `lint-and-test.yml` → "Run frontend tests" | `vitest run --environment jsdom` — transpiles via esbuild, never typechecks |
| `build-wheels.yml` → `pnpm build` | `vite build` — transpile only, no `tsc` step |

So the only thing that would surface a type error is a developer running `pnpm lint:ts` locally, or a pre-commit hook. `lint-staged` runs eslint and prettier on staged files, not `tsc`.

## Scope of the debt

`origin/dev` is currently **clean — 0 errors**. So this is prevention rather than cleanup; adding the step now costs nothing and would have failed #1963 before review.

## Suggested fix

Add a step to the frontend job in `lint-and-test.yml`:

```yaml
      - name: Typecheck
        run: pnpm lint:ts
        background: true
```

It belongs alongside the existing two rather than replacing either — eslint and `tsc` catch disjoint classes of problem. Whole-project rather than changed-files, because a type error is frequently in a file the change did not touch.

## Adjacent, worth deciding separately

The eslint step runs on **changed files only** (`git diff --name-only` filtered to `.ts`/`.tsx`/`.cjs`). That means a change in file A which breaks a lint rule in an unchanged file B is not caught either. Cheaper to keep as-is than a whole-repo eslint run on every PR, so this is a note rather than part of the ask — but it is the same shape of gap.

## Acceptance criteria

- [ ] A CI step runs `pnpm lint:ts` on every PR, and fails the check on a non-zero exit
- [ ] Introducing a deliberate type error in a spec file fails that check
- [ ] `dev` stays at zero errors (verify at the time of the fix — it is zero today)

dcblundell:
## Summary
- run the existing whole-project TypeScript typecheck in the frontend PR job
- execute it in parallel with ESLint and Vitest
- include a temporary type error for one CI run to verify the new gate

**Note**
Specifying input files when running `tsc` will cause it to ignore any `.tsconfig`. There are packages that get around this behaviour but the ts linting is quite fast so I haven't added it here.

Closes #1969

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -44,6 +44,10 @@ jobs:
         run: pnpm test
         background: true
 
+      - name: Typecheck
+        run: pnpm lint:ts
+        background: true
+
       - wait-all:
 
   backend-lint-and-test:
--- a/tests/viewUsage.spec.ts
+++ b/tests/viewUsage.spec.ts
@@ -8,6 +8,9 @@ import ROUTES, { ROUTE_PATTERNS } from '../src/definitions/Routes';
 import { UsageEvent, UsageView } from '../src/definitions/UsageEvent';
 import { USAGE_VIEW_BY_ROUTE, getUsageView, recordViewOpened } from '../src/functions/viewUsage';
 
+const DELIBERATE_TYPECHECK_FAILURE: string = 42;
+void DELIBERATE_TYPECHECK_FAILURE;
+
 const { recordUsage } = vi.hoisted(() => ({ recordUsage: vi.fn() }));
 
 vi.mock('../src/functions/recordUsage', () => ({ default: recordUsage }));
--- a/tests/viewUsage.spec.ts
+++ b/tests/viewUsage.spec.ts
@@ -8,9 +8,6 @@ import ROUTES, { ROUTE_PATTERNS } from '../src/definitions/Routes';
 import { UsageEvent, UsageView } from '../src/definitions/UsageEvent';
 import { USAGE_VIEW_BY_ROUTE, getUsageView, recordViewOpened } from '../src/functions/viewUsage';
 
-const DELIBERATE_TYPECHECK_FAILURE: string = 42;
-void DELIBERATE_TYPECHECK_FAILURE;
-
 const { recordUsage } = vi.hoisted(() => ({ recordUsage: vi.fn() }));
 
 vi.mock('../src/functions/recordUsage', () => ({ default: recordUsage }));


aidemsined (commented):
> [!NOTE]
> **Verified, no findings.** The gate does what #1969 asked for, including the part I expected to be the weak spot.

I filed #1969, so I checked the acceptance criteria rather than just reading the YAML.

**The gate genuinely rejects.** Run [33677381211](https://github.com/tenstorrent/ttnn-visualizer/actions/runs/33677381211) on `4adc849` — the deliberate-error commit — shows `Typecheck -> failure` **and** `Wait for all -> failure`, so the non-zero exit propagates to the job rather than being swallowed by `background: true`. The run after `Remove testing code` is green. That covers the "introducing a deliberate type error fails the check" criterion directly, and the propagation through `wait-all` was the specific thing I couldn't verify by inspection.

**Coverage is total.** The root `tsconfig.json` includes `vite.config.ts`, `vitest.config.ts`, `vitest.setup.ts`, `src/**/*` and `tests/**/*`, and declares no `references`, so all 670 tracked `.ts`/`.tsx` files fall inside this single `tsc --noEmit`. I went looking for a second project that would need its own step and there isn't one.

Your note about `tsc` ignoring `tsconfig.json` when given file arguments is correct, and whole-project is the right call independently of that: a type error is frequently in a file the PR never touched, so a changed-files typecheck would miss exactly the case that motivated the issue. The "quite fast" claim holds too — `pnpm lint:ts` is ~4.9s locally.

## General observations

- `tsconfig.node.json` exists but isn't referenced by the root config and has no `references` entry pointing at it, so nothing in this new step reads it. Pre-existing and out of scope here — noting it only so it isn't mistaken for extra coverage.

PR #1973, opened 2026-09-02
Type the config singleton factory

### LINKED ISSUE #1919: Annotate `Config.__new__` return type to unblock mypy in `database_migrations.py`
`backend/ttnn_visualizer/settings.py:717` declares `def __new__(cls):` with no return annotation, so the singleton's type is unknown to mypy. That surfaces as an `attr-defined` error at `backend/ttnn_visualizer/database_migrations.py:78`:

```python
uri = cast(DefaultConfig, Config()).SQLALCHEMY_DATABASE_URI
```

Fix is:

```python
def __new__(cls) -> "DefaultConfig":
```

which should also let the `cast(...)` at the call site go away.

---

dcblundell:
## Summary
- replace the `Config.__new__` factory with a typed callable singleton while preserving `Config()` and `Config._instance`
- remove the three `cast(DefaultConfig, Config())` workarounds
- cover environment-specific selection and singleton reuse, and remove the resolved conventions entry

The issue's proposed `__new__ -> DefaultConfig` annotation is rejected by mypy because `DefaultConfig` is not a `Config` subtype. The callable factory gives `Config()` its real static return type without a suppression.

## Testing
- `pnpm run flask:mypy`
- `pnpm run flask:lint`
- `pnpm run flask:test` (1,196 passed)

Fixes #1919


backend/ttnn_visualizer/settings.py
```diff
@@ -790,25 +790,25 @@ class ProductionConfig(DefaultConfig):
     TESTING = False
 
 
-class Config:
-    _instance = None
+class _ConfigFactory:
+    _instance: Optional[DefaultConfig] = None
```
> `_instance` is declared as a class attribute here, but `__call__` assigns `self._instance` — so the class slot stays `None` for the life of the process and the live cache is an instance attribute shadowing it.

Verified by execution: after one `Config()`, `_ConfigFactory.__dict__["_instance"]` is still `None`, `Config.__dict__` holds the config, and `_ConfigFactory()() is Config()` is `False`. Nothing constructs a second factory today, so this is latent rather than live, but it sets two traps the old class-attribute cache didn't have. A reset written the obvious way — `settings._ConfigFactory._instance = None`, or `monkeypatch.setattr(type(Config), "_instance", None)` in a test — passes `raising=True` and resets nothing, because the instance attribute keeps shadowing it. And anything that does construct a second `_ConfigFactory()` gets an independent cache, while `app.py:490`'s `assert Config._instance is None` reads `None` on the fresh factory and waves it through — the guard that exists to make a pre-CLI construction *loud* would be silent.

**Suggested action:** move the declaration to where the assignment lands — `def __init__(self) -> None: self._instance: Optional[DefaultConfig] = None` — so the per-instance nature of the cache is stated rather than implied. Confirmed compatible: mypy clean, 1196 tests pass with that change.

backend/ttnn_visualizer/settings.py
```diff
@@ -790,25 +790,25 @@ class ProductionConfig(DefaultConfig):
     TESTING = False
 
 
-class Config:
-    _instance = None
+class _ConfigFactory:
```
aidemsined, line 793: > [!WARNING]
> `_ConfigFactory` has no docstring and `Config = _ConfigFactory()` has no comment, so the reason for the shape lives only in the PR description.

This is now the least obvious construct in the module — a class-cased public name bound to an *instance*, invoked as if it were a constructor — and the file documents every comparable decision in place (`_AllowedOrigins`, `_UsageRecordingActive`, `_refuse_debug_under_server_mode`). The non-obvious half isn't in the issue either: mypy doesn't merely reject `__new__ -> DefaultConfig`, it also discards such an annotation for inference. I reproduced it — `reveal_type(Config())` comes back as `Config`, so all three casts would have survived even behind a suppression. That makes the callable factory the only shape that actually delivers what #1919 asked for, which is worth recording both here and on the issue when it closes.

**Suggested action:** add a short docstring on `_ConfigFactory` (or a comment on the `Config = _ConfigFactory()` binding) giving that rationale plus the consequence for callers: `Config` is a value now, so `isinstance(x, Config)` and `class Foo(Config)` no longer work. Both are caught statically, so the trap is guarded — it just isn't explained.

backend/ttnn_visualizer/tests/test_settings.py
```diff
@@ -1207,6 +1208,28 @@ def test_gunicorn_bind_is_recomputed_from_host_and_port(monkeypatch):
     assert config.GUNICORN_BIND == "127.0.0.1:9001"
 
 
+@pytest.mark.parametrize(
```
aidemsined, line 1211: > [!WARNING]
> Nothing pins the invariant `_config_after_cli_env` exists to protect: that the cache wins over a **later** `FLASK_ENV` change.

The test resets `_instance`, sets one `FLASK_ENV`, then asserts identity within that single value. I checked by mutation that this does catch caching being removed outright (all three params fail), and that the `monkeypatch` reset is genuinely exercised rather than decorative — the app fixtures populate the singleton first, so the `testing`/`production` rows would go red if the reset stopped working. What it doesn't catch is a cache that re-resolves *when the environment changed*, which is exactly the regression `_config_after_cli_env`'s docstring describes ("`--server` would yield `SERVER_MODE=False` with no error"). Under that mutation the whole suite stays green, including `test_the_cli_refuses_to_run_after_the_config_singleton_exists`.

**Suggested action:** a fourth case in this parametrised test — set `FLASK_ENV=production`, call `Config()`, `monkeypatch.setenv("FLASK_ENV", "development")`, then assert `Config() is config` and `isinstance(config, ProductionConfig)`. Three lines, and it closes the one behaviour of the factory that the CLI depends on and nothing currently observes.

aidemsined (commented):
Nothing here blocks merge. I re-ran the stated checks against the PR head: `mypy backend/ttnn_visualizer` → `Success: no issues found in 83 source files`, the full backend suite → `1196 passed` (matching the body exactly), `isort` + `black --check` clean, zero `cast` left anywhere in `backend/`, and no new suppression in any of the five touched files. Test isolation is genuinely preserved despite the class→instance move: `monkeypatch.setattr` snapshots a class attribute from `target.__dict__` but an instance attribute via `getattr`, and in every reachable case here both resolve to the same live value, so `undo` restores identical state — I confirmed that on the live object rather than reasoning about it. Three comments inline.

One thing worth carrying back to #1919 when it closes: the issue's own proposed fix wouldn't have worked, and not only for the reason the body gives. `reveal_type(Config())` under `def __new__(cls) -> "DefaultConfig":` comes back as `Config` — mypy discards a non-subtype `__new__` return for inference — so all three casts would have had to stay even with a suppression forcing the annotation through. The callable factory isn't a workaround for a mypy complaint; it's the only one of the two shapes that delivers the ticket's goal.

## General observations

- The new test lacks the leading "why" comment its three neighbours carry (`:1233`, `:1257`, `:1280`), in a file that otherwise treats it as mandatory.
- Two branches of `_determine_config` stay uncovered: the `"development"` default (every case calls `monkeypatch.setenv`, so the path a developer with nothing exported actually takes is never taken — `monkeypatch.delenv("FLASK_ENV", raising=False)`), and the `.lower()` normalisation. I mutation-checked the second: dropping `.lower()` leaves all 1196 tests green. A `("PRODUCTION", ProductionConfig)` row covers it.
- Related and pre-existing, but the `.lower()` gap makes it easier to hit: `create_app` compares `FLASK_ENV` case-sensitively at `app.py:133`/`:154`, so `FLASK_ENV=Production` already selects `ProductionConfig` in settings while skipping the production-only `robots.txt` and SPA catch-all routes. The two readers of one variable disagree on normalisation.
- `override_with_env_variables()` isn't covered by the new test either — deleting it leaves this test green and only fails two tests elsewhere. One line (`monkeypatch.setenv("PORT", "9101")`, then assert `GUNICORN_BIND`) would make the test self-sufficient. For what it's worth I did verify the override runs exactly once, not zero or twice, in both the old and new shape.
- Five sites now reach into the private `_instance` (`app.py:490` plus four test writes). A `reset()` / `is_initialised()` on the factory would turn the production assert into API and is immune to the shadowing in the first inline comment — though the tests may reasonably keep attribute access for `monkeypatch`'s automatic teardown.
- `CONVENTIONS.md:950` and `tests/fixture_settings.py:64` both still read "`Config` is a process singleton whose class attributes bind at import". That was already loose; it's now wrong about the named symbol, since `Config` has no config attributes at all — they're on `DefaultConfig`. This PR already edits `CONVENTIONS.md`.
- Since the #1919 known-inconsistency bullet is going away, a one-line replacement stating the resolved rule ("`Config` is a callable singleton instance, not a class; `Config()` returns the `DefaultConfig` subclass selected by `FLASK_ENV`") would leave the conventions net-richer rather than just shorter.
- The unrecognised-`FLASK_ENV` fall-through maps any typo, `"staging"`, or empty string to `DevelopmentConfig` — the most permissive posture. Neither deleted comment captured it and nothing does now; that's the sentence worth writing if a comment goes back into that region.
- `warn_redundant_casts = true` under `[tool.mypy]` would finish the job this PR started: I confirmed a re-added `cast(DefaultConfig, Config())` type-checks silently today, so the workaround can creep back as a no-op even though the return-type guarantee is CI-enforced.
- `_determine_config` touches neither `self` nor `cls`; the `@staticmethod` preserves only a `Config._determine_config` call shape nothing uses. Marginal.
- Both deleted comments were pure restatement, and removing them is endorsed by the repo's own convention example. No objection.
- Dropping the `super().__new__(cls)` allocation is a real improvement, not just a relocation: the old code published a bare `Config` for one statement, which is why the slot couldn't have been annotated honestly even in the old shape.

backend/ttnn_visualizer/settings.py
dcblundell: Fixed. Added a `_ConfigFactory` docstring explaining why `Config` is a callable instance and why its return type is environment-specific. The settings test suite passes: 165 passed.

backend/ttnn_visualizer/settings.py
dcblundell: Fixed. `_ConfigFactory` now initialises `self._instance` in `__init__`, making the per-factory cache explicit and ensuring resets and the CLI guard operate on the live attribute. The settings test suite passes: 165 passed.

backend/ttnn_visualizer/tests/test_settings.py
dcblundell: Fixed. Added `test_config_cache_wins_over_later_environment_changes`, which constructs the production config, changes `FLASK_ENV` to development, and verifies identity and type remain cached. The settings test suite passes: 165 passed.

PR #1975, opened 2026-09-03
Add optional Stylelint checking to PR CI

dcblundell:
This pull request improves the linting workflow and makes a small style adjustment. The main changes are the addition of SASS linting for changed `.scss` files in the GitHub Actions workflow and a minor CSS tweak to the `.monospace` class.

**CI/CD workflow improvements:**

* The GitHub Actions workflow `.github/workflows/lint-and-test.yml` now detects changed `.scss` files and runs `stylelint` on them, in addition to the existing linting for TypeScript/JavaScript files. This ensures SASS code quality checks are enforced for pull requests.

**Styling changes:**

* The `.monospace` CSS class in `src/scss/_common.scss` now includes `zoom: 1;`, which can help with layout consistency and browser compatibility.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -32,12 +32,19 @@ jobs:
         run: |
             git fetch origin ${{ github.base_ref }} --depth=1
             # Get changed files, filter for .ts, .tsx, .cjs, and .mjs, exclude deleted files, output as space-separated string
-            CHANGED_FILES=$(git diff --name-only --diff-filter=d origin/${{ github.base_ref }} | grep -E '\.(ts|tsx|cjs|mjs)$' | tr '\n' ' ' | xargs)
+            CHANGED_FILES=$(git diff --name-only --diff-filter=d origin/${{ github.base_ref }} | awk '/\.(ts|tsx|cjs|mjs)$/ { print }' | tr '\n' ' ' | xargs)
+            CHANGED_SCSS=$(git diff --name-only --diff-filter=d origin/${{ github.base_ref }} | awk '/\.scss$/ { print }' | tr '\n' ' ' | xargs)
             echo "changed=$CHANGED_FILES" >> $GITHUB_OUTPUT
+            echo "changed_scss=$CHANGED_SCSS" >> $GITHUB_OUTPUT
 
-      - name: Lint changed files
+      - name: Lint changed TypeScript/JavaScript
         if: ${{ steps.pr_changes.outputs.changed != '' }}
         run: pnpm lint ${{ steps.pr_changes.outputs.changed }} --report-unused-disable-directives --max-warnings 0 --no-warn-ignored
+        background: true
+
+      - name: Lint changed SASS
+        if: ${{ steps.pr_changes.outputs.changed_scss != '' }}
+        run: pnpm exec stylelint ${{ steps.pr_changes.outputs.changed_scss }}
         background: true
 
       - name: Run frontend tests
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -33,18 +33,16 @@ jobs:
             git fetch origin ${{ github.base_ref }} --depth=1
             # Get changed files, filter for .ts, .tsx, .cjs, and .mjs, exclude deleted files, output as space-separated string
             CHANGED_FILES=$(git diff --name-only --diff-filter=d origin/${{ github.base_ref }} | awk '/\.(ts|tsx|cjs|mjs)$/ { print }' | tr '\n' ' ' | xargs)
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-and-test.yml
```diff
@@ -32,14 +32,21 @@ jobs:
...
 
-      - name: Lint changed files
+      - name: Lint changed TypeScript/JavaScript
         if: ${{ steps.pr_changes.outputs.changed != '' }}
         run: pnpm lint ${{ steps.pr_changes.outputs.changed }} --report-unused-disable-directives --max-warnings 0 --no-warn-ignored
         background: true
 
+      - name: Lint changed SCSS
+        if: ${{ steps.pr_changes.outputs.changed_scss != '' }}
+        run: pnpm exec stylelint ${{ steps.pr_changes.outputs.changed_scss }}
```
aidemsined: > [!NOTE]
> **Possible** — this step is genuinely narrower than the TypeScript step it's modelled on, and the plumbing that makes it narrow costs more than it saves.

The `lint` script hardcodes `eslint .`, so `pnpm lint <changed files>` expands to `eslint . --ext ts,tsx … <changed files>` and eslint walks the whole repo regardless. I measured it on this branch: passing a single path still processes **675 files**. So for TypeScript the changed-files detection only decides *whether* the step runs, not what it covers.

`pnpm exec stylelint <changed>` has no equivalent glob, so this step really is changed-files-only. The two steps now read as symmetric but aren't — a `.scss` file that violates a rule without appearing in the diff is covered by neither step. And the saving is small: whole-src `pnpm lint:scss` is ~1.15s across 101 files, against ~0.74s for a single file.

**Suggested action:** use the existing `pnpm lint:scss` instead. That deletes the `CHANGED_SCSS` awk line, the `changed_scss` output, and the `if:` guard on this step, while making SCSS coverage match what the TypeScript step effectively already does.

.github/workflows/lint-and-test.yml
```diff
@@ -32,14 +32,21 @@ jobs:
...
+            echo "changed_scss=$CHANGED_SCSS" >> $GITHUB_OUTPUT
 
-      - name: Lint changed files
+      - name: Lint changed TypeScript/JavaScript
         if: ${{ steps.pr_changes.outputs.changed != '' }}
         run: pnpm lint ${{ steps.pr_changes.outputs.changed }} --report-unused-disable-directives --max-warnings 0 --no-warn-ignored
         background: true
 
+      - name: Lint changed SCSS
+        if: ${{ steps.pr_changes.outputs.changed_scss != '' }}
```
aidemsined, line 44: > [!NOTE]
> **Possible** — the failure path is never exercised here, unlike the equivalent gate in #1972.

This guard means the current head skips the step entirely: run [33759351270](https://github.com/tenstorrent/ttnn-visualizer/actions/runs/33759351270) reports `Lint changed SCSS -> skipped`, because `Remove debug style` left the PR with no `.scss` changes at all. The two earlier runs that did execute it (`15d203b`, `2c960ef`) used `.monospace { zoom: 1; }` — and `zoom: 1` passes stylelint cleanly on this branch, exit 0. So every green run so far demonstrates that the step *executes*, not that it *rejects*.

The gate itself is sound, for what it's worth. Against `stylelint@17.14.1` with this branch's config I get exit 2 for prettier indentation, long hex, single-colon `:before`, `-webkit-` prefixes, named colours, `!important`, and syntax errors. It just isn't demonstrated by any CI run on this PR.

**Suggested action:** do what you did on #1972 — push a commit with a real violation (`color: red` is enough), confirm the job goes red, then drop it.

aidemsined (commented):
Two non-blocking points inline. The step works — I verified against `stylelint@17.14.1` on this branch that it exits 2 for prettier indentation, long hex, single-colon `:before`, vendor prefixes, named colours, `!important` and syntax errors — so this is about scope and about proving it, not about correctness.

## General observations

- The description still describes a `.monospace { zoom: 1; }` tweak and "a minor CSS tweak to the `.monospace` class", but `Remove debug style` took that out — the final diff is workflow-only, one file. Worth trimming so a reviewer isn't hunting for a CSS change that isn't there.
- The `grep -E` → `awk` swap isn't mentioned in the description, and it's behaviour-neutral as things stand: this workflow sets no `shell:` and no `defaults.run`, so GitHub uses `bash -e` *without* `pipefail`, and `grep`'s exit 1 on no-match was already absorbed by the command substitution (`xargs` exits 0). No objection to keeping it — it removes a real hazard the moment anyone adds `shell: bash` — just flagging that it's unrelated to the stated purpose of the PR and reads as unexplained in the diff.

.github/workflows/lint-and-test.yml
dcblundell: Will do.

dcblundell: Simplified as you suggested.

.github/workflows/lint-and-test.yml
dcblundell: Already did that.

PR #1978, opened 2026-09-03
Include documentation for event logging

### LINKED ISSUE #1835: Usage signals: document what is recorded, where, and how to inspect, delete or disable it
Sub-issue of #1819.

Recording is on by default and the tool is public, so anything recorded is readable by anyone who reads the repo. #1819 treats that as a forcing function for keeping the recorded set defensible; the docs page is where that set becomes reviewable by someone who is not going to read the source.

## Scope

A new page under `docs/src/` — the docs are Sphinx, so add it to the toctree in `docs/index.rst` — covering:

- **The fixed path**, and plainly that wiping the app data directory does *not* remove it. This is the acknowledged cost of putting usage data outside `get_app_data_directory()`, and the mitigation is precisely that a fixed documented path is easier to inspect or delete than one the user would have to compute from their environment. That only holds if it is actually documented.
- **Every recorded field, event by event**, with its closed enum values.
- **The explicitly-not-recorded list** from #1819: no report, file, folder or directory names; no operation names, tensor shapes or kernel names; no model identifiers; no hostnames, usernames, IP addresses or SSH targets; no search or filter text; no stack traces or error bodies; no raw counts that could fingerprint a workload.
- **The off switch** — both the environment variable and the file at the usage path — and that recording is on by default. State it rather than leaving it implied.
- **A one-line delete command.**
- That `run_id` is generated fresh for each application launch, shared by its server workers, and never exported.
- That collection is out-of-band, and the tool has no knowledge of whether it is happening.
- That the hosted instance records nothing at all.
- Startup output that identifies why event logging is disabled and reports an unrecognised opt-out value beside the resulting disabled status.

Keep this clearly separate from the docs site's own PostHog analytics (caveat 19) — docs traffic and app usage are complementary but not comparable, and the page should not read as though one measures the other.

## Open question from #1819 to settle here

Is this page part of the definition of done for each new event? Recommend yes, and record it in `AGENTS.md` / `CONVENTIONS.md` alongside the answer to "who owns the event list, and what is the review gate for adding an event". Without a gate, the vanity-metric risk in caveat 7 is only a good intention.

## Out of scope

Startup output unrelated to event-logging status and invalid-opt-out guidance.

## Verification

`sphinx-build docs docs/output` per `building-the-docs.md`. New files need an SPDX header with the current year.

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1835.

This pull request clarifies and strengthens the documentation and implementation around event logging (formerly "usage recording"), especially the opt-out mechanism for disabling event logging, and introduces a new test to enforce parity between user-visible documentation and the backend event schema. The changes ensure that the behavior and documentation for event logging are consistent, explicit, and auditable, and that unrecognized values for the disabling environment variable are handled transparently. The most important changes are:

**Event logging terminology and documentation:**

* Replaces references to "usage recording" with "event logging" throughout documentation and code comments, and updates section headings and links accordingly for clarity and consistency. [[1]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L186-R190) [[2]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL23-R23) [[3]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL599-R609)
* Expands and clarifies the documentation for the event logging opt-out (`USAGE_RECORDING_DISABLED`), specifying that unset/`false`/`0` enables logging, while `true`/`1` and any unrecognized value disables it, with a startup warning that includes the value. [[1]](diffhunk://#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049cL23-R25) [[2]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L231-R231) [[3]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL1329-R1331)

**Backend implementation improvements:**

* Updates `_record_launch` in `app.py` to print a clear warning (including the unrecognized value) when event logging is disabled due to an unrecognized environment variable value, and ensures the opt-out logic is explicit and user-facing. [[1]](diffhunk://#diff-14674be812616fc4545d50f7661dafc8f85a2727687bda0471eda8b372f6af3eL433-R447) [[2]](diffhunk://#diff-14674be812616fc4545d50f7661dafc8f85a2727687bda0471eda8b372f6af3eL445-R458)
* Refactors imports and environment variable handling in `app.py` for clarity and to support the updated logging and warning behavior.

**Parity and test enforcement:**

* Adds a new test file, `test_event_logging_docs_parity.py`, which ensures that the user-facing event logging documentation (`docs/src/event-logging.md`) matches the backend event schema, including event types, field names, allowed values, and environment controls. This test enforces that documentation changes are reviewed alongside schema changes.

**Test fixture cleanup:**

* Removes unnecessary monkeypatching of a warning-tracking variable in the `usage_directory` test fixture, as warnings are now handled differently.

These changes collectively improve the reliability, transparency, and maintainability of event logging and its documentation, ensuring users and developers have a clear understanding of how to control and audit event logging behavior.


backend/ttnn_visualizer/app.py
```diff
@@ -430,9 +432,19 @@ def _record_launch(config):
     """
     # Exported unconditionally so every gunicorn worker this launch spawns shares one
     # identifier, even if recording is switched on part-way through the session.
-    os.environ[RUN_ID_ENV_VAR] = get_run_id()
+    os.environ[RUN_ID_ENV_VAR] = start_run()
 
     if not is_recording_enabled(config.SERVER_MODE):
+        print(f"📊 Recording usage has been DISABLED by the user")
```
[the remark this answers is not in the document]
dcblundell: Already addressed.

AGENTS.md
```diff
@@ -183,11 +183,11 @@ Applies on touch to views that draw data-proportional visuals (NPE chip cluster
...
 - Emit toasts via **`createToastNotification(message, fileName, ToastType.X)`** (`src/functions/createToastNotification.tsx`; `ToastType` lives in `src/definitions/ToastType.ts`); the same module exports **`createToast`** and **`dismissToast`** for a custom body or a toast you must dismiss later. Only that module may import `toast` from `react-toastify` — `no-restricted-imports` enforces it. [Signatures and the `<ToastContainer>` rule](./CONVENTIONS.md#emit-toasts-via-createtoastnotification).
 
-### Usage recording (frontend)
+### Event logging (frontend)
 
-`src/functions/recordUsage.ts` posts local usage events to `POST /api/usage`. Four invariants a reader cannot infer from the code alone — all detailed, with the caps table and the route's deliberate decorator stack, in [CONVENTIONS.md](./CONVENTIONS.md#usage-recording-frontend):
+`src/functions/recordUsage.ts` posts local usage events to `POST /api/usage`. Five invariants a reader cannot infer from the code alone — all detailed, with the caps table and the route's deliberate decorator stack, in [CONVENTIONS.md](./CONVENTIONS.md#event-logging-frontend):
```
aidemsined, line 188: > [!WARNING]
> **Definite** — the count no longer matches the list.

The sentence promises five invariants and four bullets follow. `dev` said "Four" with those same four; this PR raised the number and folded the new documentation-copy obligation into the first bullet rather than adding one, so the list length did not change.

**Suggested action:** either split the `docs/src/event-logging.md` copy out as its own fifth bullet — it is arguably a distinct obligation from the frontend enum copy, since a different test enforces it — or put the number back to four.

backend/ttnn_visualizer/tests/test_event_logging_docs_parity.py
```diff
@@ -0,0 +1,225 @@
...
+    source: str, pattern: str, description: str
+) -> Set[str]:
+    match = re.search(pattern, source)
+    if match is None:
+        raise AssertionError(f"No {description} found in {_EVENT_LOGGING_DOCS.name}")
+
+    return set(re.findall(r"`([^`]+)`", match.group(1)))
+
+
+def test_the_event_logging_docs_page_is_in_the_resources_toctree():
```
aidemsined, line 125: > [!NOTE]
> **Possible** — this asserts membership in any toctree, not the Resources one its name claims.

`^\s+src/\S+$` is matched against the whole of `index.rst`, so the assertion is satisfied by an entry under any of the three captions. I checked it rather than assuming: moving `src/event-logging` out of Resources and into the Get Started toctree leaves this test passing.

Worth tightening precisely because the thesis of this PR is that the documentation is held in place by tests. A test whose name claims more than its assertion delivers is the one gap this suite exists to close, and it is the sort of thing a future reader will trust without re-deriving.

**Suggested action:** scope the search to the Resources block — slice from `:caption: Resources` to the next `.. toctree::` or end of file, and look for the entry inside that slice.

aidemsined (commented):
I ran the parity suite against this head and sabotaged it three ways to check it actually bites: adding a `UsageEvent` member, bumping `SCHEMA_VERSION`, and deleting one documented field line each fail it. It is a real review gate rather than decoration, and deriving the expectations from `UsageEvent` / `CLIENT_EVENT_DETAIL_FIELDS` / `_DETAIL_FIELD_ENUMS` / `TRUE_VALUES` with `==` rather than a subset check is what makes it hold. That is the most valuable thing in this PR.

Two non-blocking points inline.

## General observations

- The warn-once machinery (`_warn_once`, `_warned_env_vars`) and its test are gone, and the unrecognised-value warning now surfaces only from `_record_launch`. I went looking for a path that would lose it: `[project.scripts]` maps `ttnn-visualizer` to `app:main`, `pnpm flask:start` runs `python -m ttnn_visualizer.app`, and `_record_launch` has exactly one production call site inside `main()` — so every launch still reports the misconfiguration, and the launcher output is a better place for it than a gunicorn worker's log. Flagging only so the removal is not misread as a regression later.
- `_is_recording_disabled_by_environment` reduces to `parsed is not False`, which preserves the previous truth table exactly: unset and `false`/`0` record, `true`/`1` and anything unrecognised disable. Worth keeping the docstring's account of *why* this one boolean diverges from `_coerce_env_value` — a deliberate asymmetry with no stated reason is the kind of thing that gets "tidied up" a year later.
- The vocabulary migration is half-applied on the user-facing page: the title and headings say event logging, while the prose still says "recording" ("keeps recording on", "switches recording off") and the paths stay under `~/.ttnn-visualizer/usage/`. That is coherent — the variable cannot be renamed without breaking existing opt-outs — but a reader meets both names on one page, so it may be worth a sentence saying the environment variable keeps its original name.
- `docs/index.rst` places the page last under Resources, after `troubleshooting`. Reasonable, though it is the page a privacy-minded user is most likely to go looking for; `support` / `event-logging` / `troubleshooting` would surface it slightly better.

AGENTS.md
dcblundell: Addressed in 7e49164a by splitting the documentation-copy obligation into its own fifth bullet.

backend/ttnn_visualizer/tests/test_event_logging_docs_parity.py
dcblundell: Addressed in 7e49164a. The test now selects the Resources toctree before checking for the event-logging entry.

PR #1979, opened 2026-09-03
Track events in hosted app

### LINKED ISSUE #1905: Usage signals: Remote host logging
Probably split by session cookie id and store under /data/app/events.log.

dcblundell:
This pull request updates the documentation and configuration to reflect a migration from "usage" event logging to a more robust and secure "event log" system, especially for hosted/server mode deployments. The changes clarify how event logging works in both local and server modes, introduce new invariants and security boundaries for hosted logs, and rename files, routes, and constants to match the new terminology and architecture. This ensures that event logging is more clearly documented, secure, and consistent across the codebase.

**Event Logging System Overhaul and Documentation Updates:**

_Event Logging Architecture and Security:_
* Migrated from `usage`-based event logging to a new `event log` system, updating all references to use `event_logging.py`, `event-log/events` endpoint, and new vocabulary/types (e.g., `EventLogEvent`, `MAX_EVENT_LOG_BATCH_EVENTS`). This includes renaming frontend files (e.g., `recordUsage.ts` → `recordEvent.ts`) and backend tests/constants for clarity and parity. [[1]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L188-R196) [[2]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL403-R403) [[3]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL601-R656)
* Documented that in server mode, event logs are partitioned per anonymous session, with identity derived from a signed Flask session cookie. The backend never exports or forwards events, and the log partition key is never user-controlled. [[1]](diffhunk://#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049cL18-R22) [[2]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL601-R656)
* Outlined new hosted admission controls: strong `SECRET_KEY` requirement, cross-worker log reservation caps, per-log and aggregate rate limits, and storage quotas to ensure bounded resource usage in hosted deployments. [[1]](diffhunk://#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049cL54-R58) [[2]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL601-R656)

_Frontend and API Usage:_
* Updated documentation to reflect that the only exceptions to using `axiosInstance` for HTTP requests are the socket connection and the event log unload beacon, which now posts to `/api/event-log/events` using a `Blob` with `application/json` content type. [[1]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L171-R171) [[2]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL403-R403)
* Clarified that the event logging system is opt-out via `USAGE_RECORDING_DISABLED`, and that the environment variable handling logic now resides in `event_logging.py`.

_Constants, Tests, and Parity:_
* Ensured all constants, tests, and documentation are kept in sync between backend and frontend (e.g., batch/event caps, event schema, and test parity files), with CI enforcing this contract. [[1]](diffhunk://#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fL601-R656) [[2]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L188-R196)

These changes collectively improve the reliability, security, and clarity of event logging, especially for hosted/server deployments, and ensure that all related documentation and configuration are up to date.

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1905.


backend/ttnn_visualizer/views.py
```diff
@@ -2820,6 +2822,11 @@ def record_usage_events():
     if not is_recording_enabled(current_app.config["SERVER_MODE"]):
         return Response(status=HTTPStatus.NO_CONTENT)
 
+    server_mode = is_flag_enabled(current_app.config["SERVER_MODE"])
+    event_log_id = ensure_event_log_id() if server_mode else None
```
smountenay-tt: Admission has a filesystem side effect before we know this is JSON or a valid event. In `SERVER_MODE`, a CORS-simple `text/plain` POST reaches this point without a preflight; I reproduced a 400 response that still minted a session ID and created a zero-byte `events.log`. A cookieless hostile page can consume 60 slots/minute and exhaust all 1,024 slots in about 17 minutes. Please reject non-JSON (and preferably validate the payload) before minting/admitting, with hosted malformed/plain-text tests asserting that neither a session ID nor a file is created.

backend/ttnn_visualizer/views.py
```diff
@@ -2820,6 +2822,11 @@ def record_usage_events():
...
+    if not admit_event_log_batch(server_mode, event_log_id):
```
smountenay-tt: `admit_event_log_batch()` can raise `PermissionError`/`OSError` while creating the root or opening `.quota.lock`, but this call is outside `record_events()`'s no-raise boundary. With a read-only event-log root I get a 500 plus traceback; every client flush can then repeat that and flood application logs. Please make admission best-effort/no-raise (including warning suppression) and preserve the silent 204 contract, with an endpoint test that injects a reservation filesystem failure.

backend/ttnn_visualizer/event_logging.py
```diff
@@ -657,25 +684,176 @@ def _format_line(
...
+            return False
+
+    return not _is_hosted_rate_limited(state)
+
+
+def _reserve_hosted_log(log_path: Path, state: _EventLogState) -> bool:
+    """Atomically reserve one bounded hosted log slot across server workers."""
+    global _hosted_quota_warning_logged
+
+    if state.reservation_checked:
```
smountenay-tt, line 800: `reservation_checked` assumes the file will continue to exist. If retention deletes `events.log` but leaves the session directory, `_open_log()` later recreates it through `O_CREAT` without taking the quota lock. Once another session has occupied the freed slot, the old session can therefore push the global count past `MAX_HOSTED_EVENT_LOGS`; I reproduced the production bound moving from 1,024 to 1,025. Please ensure a hosted open cannot create a missing file without re-reservation, and test file-only deletion while the quota is full.

backend/ttnn_visualizer/app.py
```diff
@@ -111,6 +115,23 @@ def _serialize_spa_js_config(js_config: dict) -> str:
...
+    if not is_flag_enabled(config.get("SERVER_MODE", False)):
+        return
+
+    secret_key = config.get("SECRET_KEY")
+    encoded = (
+        secret_key
+        if isinstance(secret_key, bytes)
+        else str(secret_key or "").encode("utf-8")
+    )
+    if secret_key == DEFAULT_SECRET_KEY or len(encoded) < MIN_HOSTED_SECRET_KEY_BYTES:
```
smountenay-tt, line 130: This makes `SECRET_KEY` an explicit hosted integrity boundary, but `main()`'s default `PRINT_ENV` loop still prints every value from `config.to_dict()`, including the full signing key. Hosted startup therefore writes the key to terminal/container logs, allowing anyone with log access to forge Flask sessions. Please redact sensitive settings from the environment dump (at minimum `SECRET_KEY`) and add a startup-output regression test.

backend/ttnn_visualizer/app.py
```diff
@@ -149,6 +170,7 @@ def create_app(settings_override=None):
     if settings_override:
         app.config.update(settings_override)
 
+    _validate_hosted_secret_key(app.config)
```
smountenay-tt, line 186: This validation only runs when Gunicorn imports `create_app()`. The launcher has already performed migrations, displayed the logging banner, potentially opened a browser, and—under `--daemon`—detached before a worker rejects the key. A daemon launch can consequently return success while every worker fails; the non-daemon path also ignores `subprocess.run()`'s return code. Please validate the hosted configuration in `main()` immediately after constructing `Config`, before side effects/daemonization, and propagate Gunicorn's nonzero exit status.

backend/ttnn_visualizer/event_logging.py
```diff
@@ -328,28 +341,73 @@ class UsageEventRejected(Exception):
...
+
+
+def ensure_event_log_id() -> str:
+    """Return the server-minted anonymous log identifier in the Flask session."""
+    from flask import session
+
+    event_log_id = session.get(EVENT_LOG_ID_SESSION_KEY)
+    if not _is_valid_event_log_id(event_log_id):
+        event_log_id = uuid.uuid4().hex
+        session[EVENT_LOG_ID_SESSION_KEY] = event_log_id
```
smountenay-tt, line 375: The first ID is minted inside the event POST, so two initial flushes sent before either response's `Set-Cookie` is applied both see an empty session, mint different IDs, and write separate logs. The competing responses then leave one log orphaned. `postEvents()` has no in-flight serialization, and a visibility/pagehide flush can race a scheduled request. Please establish the ID on the initial HTML response or serialize the first client write, and cover concurrent first requests.

smountenay-tt, line 375: This also adds another field to the client-side Flask session cookie, whose report lists are capped by count rather than encoded bytes. A valid cookie near Werkzeug's limit can exceed the ~4,093-byte header threshold after adding `event_log_id`; if the browser rejects that `Set-Cookie`, every flush mints another ID/log and can exhaust quota. Please reserve byte headroom or trim session data by serialized size, with a near-limit persistence test.

src/routes/NPE.tsx
```diff
@@ -61,7 +61,7 @@ const NPE = () => {
     const npeData = useMemo(() => demoData || loadedData || loadedTimeline, [demoData, loadedData, loadedTimeline]);
 
     // Demos are hosted-only. Keep their source wired for vocabulary parity, but
-    // recordUsage drops the event under SERVER_MODE. The non-windowed settle
+    // recordEvent drops the event under SERVER_MODE. The non-windowed settle
```
smountenay-tt: Non-blocking: this is now the opposite of the implementation—`recordEvent` intentionally records under `SERVER_MODE` when the published switch is active. Please update this rationale; `recordEvent.ts`'s opening description also still calls these local-only events.

smountenay-tt (changes requested):
The overall direction makes sense, but the hosted admission and session lifecycle currently leave several quota, availability, and session-integrity gaps. I reproduced the content-type reservation, read-only-storage 500, and collected-log quota bypass locally. The focused backend and frontend suites remain green, so these need new regression coverage.

backend/ttnn_visualizer/app.py
dcblundell: Addressed in `b5010921`. Hosted secret validation now runs immediately after CLI/environment configuration, before migrations, browser launch, daemonisation, or worker startup. Gunicorn's non-zero exit status is now propagated.

src/routes/NPE.tsx
dcblundell: Addressed in `b5010921`. Updated the NPE rationale and `recordEvent.ts` documentation to reflect that event logging can run in hosted mode when enabled; the stale local-only wording is removed.

backend/ttnn_visualizer/event_logging.py
dcblundell: Addressed in `b5010921`. Hosted pages establish the session log ID in the initial HTML response, and frontend POST flushes are serialised so concurrent initial batches cannot mint separate IDs. Added sender coverage for the queued first writes.

backend/ttnn_visualizer/views.py
dcblundell: Addressed in `b5010921`. Hosted admission now treats reservation `OSError` failures as best-effort drops, returns the existing silent 204 contract, and suppresses repeated warnings. Added endpoint coverage for repeated reservation failures.

backend/ttnn_visualizer/views.py
dcblundell: Addressed in `b5010921`. The endpoint now parses and validates the JSON envelope and events before minting a hosted session ID or admitting a log. Added regression coverage asserting malformed/plain-text requests create neither.

backend/ttnn_visualizer/event_logging.py
dcblundell: Addressed in `b5010921`. Before persisting the hosted log ID, the backend measures the signed session cookie and trims oldest report-history entries until it has reserved headroom. If the cookie cannot fit even after trimming, no ID/log is admitted. Added near-limit coverage.

backend/ttnn_visualizer/app.py
dcblundell: Addressed in `b5010921`. `SECRET_KEY` is now redacted from the environment dump, with a regression test covering startup configuration output.

backend/ttnn_visualizer/event_logging.py
dcblundell: Addressed in `b5010921`. Hosted appends no longer use `O_CREAT`; a missing log triggers re-reservation before recreation. Added coverage showing a deleted file cannot bypass a full hosted quota.

smountenay-tt (approved):

PR #1782, opened 2026-07-20
Move functions and definitions into more appropriate files

dcblundell:
Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1781.

PR #1981, opened 2026-09-04
Remove report sorting by link status

dcblundell:
Removes the sorting behaviour that places previously linked reports at the top and failed links at the end of the list.

The pickers now preserve the order supplied by the backend. Link-status badges remain visible but no longer change row order. The resulting long-list discoverability trade-off is tracked in #1986.

Multihost rank ordering is out of scope for this change.

Screenshot 2026-09-04 at 14 22 45

### DIFF
--- a/src/components/report-selection/LocalFolderPicker.tsx
+++ b/src/components/report-selection/LocalFolderPicker.tsx
@@ -2,17 +2,13 @@
 //
 // SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
 
-import { useMemo, useState } from 'react';
+import { useState } from 'react';
 import { Button, ButtonVariant, MenuItem, Position, Tooltip } from '@blueprintjs/core';
 import { ItemRenderer, Select } from '@blueprintjs/select';
 import { IconNames } from '@blueprintjs/icons';
 import { useInstance } from '../../hooks/useAPI';
 import 'styles/components/FolderPicker.scss';
-import {
-    getFolderLinkState,
-    shouldShowFolderLinkStatus,
-    sortByFolderLinkState,
-} from '../../functions/folderLinkStatus';
+import { getFolderLinkState, shouldShowFolderLinkStatus } from '../../functions/folderLinkStatus';
 import { ReportFolder } from '../../definitions/Reports';
 import getServerConfig from '../../functions/getServerConfig';
 import isDirectReportMode from '../../functions/isDirectReportMode';
@@ -77,18 +73,6 @@ const LocalFolderPicker = ({
     const isDeleteDisabled = loading;
     const showLinkStatus = shouldShowFolderLinkStatus(linkedIds, unlinkedIds);
 
-    // Linked first, unknown next, failed links last — preserve server order within each group.
-    const sortedItems = useMemo(
-        () =>
-            sortByFolderLinkState(
-                items ?? [],
-                (folder) => getReportId(folder.syncedName, folder.path),
-                linkedIds,
-                unlinkedIds,
-            ),
-        [items, linkedIds, unlinkedIds],
-    );
-
     const renderItem: ItemRenderer<ReportFolder> = (folder, { handleClick, handleFocus, modifiers, query }) => {
         if (!modifiers.matchesPredicate) {
             return null;
@@ -149,7 +133,7 @@ const LocalFolderPicker = ({
         <>
             <Select<ReportFolder>
                 className='folder-picker'
-                items={sortedItems}
+                items={items ?? []}
                 itemPredicate={(query, item) => !query || item.path.toLowerCase().includes(query.toLowerCase())}
                 itemRenderer={renderItem}
                 noResults={
--- a/src/components/report-selection/LocalFolderPicker.tsx
+++ b/src/components/report-selection/LocalFolderPicker.tsx
@@ -22,7 +22,10 @@ import FolderLinkStatusIcon from './FolderLinkStatusIcon';
 import SelectRowActions from './SelectRowActions';
 
 interface LocalFolderPickerProps {
-    /** `null` while the folder list query is still loading — the component renders disabled. */
+    /**
+     * `null` while the folder list query is still loading — the component renders disabled.
+     * Folders render in the supplied order; link-status badges do not reorder them.
+     */
     items: ReportFolder[] | null;
     value: string | null;
     handleSelect: (folder: ReportFolder) => void;
--- a/src/components/report-selection/RemoteFolderSelector.tsx
+++ b/src/components/report-selection/RemoteFolderSelector.tsx
@@ -5,13 +5,9 @@
 import { Button, MenuItem } from '@blueprintjs/core';
 import { IconName, IconNames } from '@blueprintjs/icons';
 import { type ItemPredicate, ItemRenderer, Select } from '@blueprintjs/select';
-import { type ReactNode, useMemo } from 'react';
+import { type ReactNode } from 'react';
 import 'styles/components/FolderPicker.scss';
-import {
-    getFolderLinkState,
-    shouldShowFolderLinkStatus,
-    sortByFolderLinkState,
-} from '../../functions/folderLinkStatus';
+import { getFolderLinkState, shouldShowFolderLinkStatus } from '../../functions/folderLinkStatus';
 import { RemoteConnection, RemoteFolder } from '../../model/RemoteConnection';
 import { TEST_IDS } from '../../definitions/TestIds';
 import { getReportId } from '../../functions/reportLinks';
@@ -114,16 +110,11 @@ const RemoteFolderSelector = ({
 
     const isDisabled = loading || remoteFolderList?.length === 0 || disabled;
 
-    const sortedFolderList = useMemo(
-        () => sortByFolderLinkState(remoteFolderList ?? [], getRemoteFolderId, linkedIds, unlinkedIds),
-        [remoteFolderList, linkedIds, unlinkedIds],
-    );
-
     return (
         <div className='form-container'>
             <Select
                 className='remote-select'
-                items={sortedFolderList}
+                items={remoteFolderList}
                 itemRenderer={remoteFolderRenderer({
                     type,
                     selectedFolder: remoteFolder,
@@ -147,7 +138,7 @@ const RemoteFolderSelector = ({
             >
                 <Button
                     icon={icon}
-                    endIcon={sortedFolderList.length > 0 ? IconNames.CARET_DOWN : undefined}
+                    endIcon={remoteFolderList.length > 0 ? IconNames.CARET_DOWN : undefined}
                     disabled={isDisabled}
                     loading={loading}
                     text={remoteFolder ? getRemoteFolderLabel(remoteFolder) : fallbackLabel}
--- a/src/components/report-selection/RemoteFolderSelector.tsx
+++ b/src/components/report-selection/RemoteFolderSelector.tsx
@@ -77,6 +77,7 @@ const remoteFolderRenderer =
 
 interface RemoteFolderSelectorProps {
     remoteFolder?: RemoteFolder;
+    /** Folders render in the supplied order; link-status badges do not reorder them. */
     remoteFolderList?: RemoteFolder[];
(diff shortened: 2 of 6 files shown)


src/components/report-selection/RemoteFolderSelector.tsx
```diff
@@ -114,16 +110,11 @@ const RemoteFolderSelector = ({
...
-        () => sortByFolderLinkState(remoteFolderList ?? [], getRemoteFolderId, linkedIds, unlinkedIds),
-        [remoteFolderList, linkedIds, unlinkedIds],
-    );
-
     return (
         <div className='form-container'>
             <Select
                 className='remote-select'
-                items={sortedFolderList}
+                items={remoteFolderList}
```
aidemsined, line 118: > [!WARNING]
> The ordering contract moves entirely off-screen and this diff deletes the only two places it was written down.

Order isn't an accident of the API — the backend sorts deliberately, most-recent-first: `views.py:1032` and `:1142` for local reports, `sftp_operations.py:1448`/`:1483` for SSH listings, `local_remote_reports.py:108` for the offline fallback. I traced the client side too: `mergeRemoteFolders` maps over the fresh list, react-query replaces `data` wholesale with no `select`, and `refetchOnWindowFocus` is off — so the order is stable across renders and refetches, and "the picker renders what it's given" is a defensible contract. That's the actual rationale for the PR, and after this merge it exists nowhere: the `// Linked first, unknown next, failed links last` comment and the `FOLDER_LINK_SORT_RANK` docblock both go, and nothing replaces them.

One consequence worth naming while you're here: for multihost, every rank of a launch is written within seconds of the others, so mtime-desc effectively scrambles rank order — and the new test bakes that in (`rankFolder` sets `lastModified: rank + 1`, and the assertion has `Rank 1` above `Rank 0`). That sits oddly next to how much `reportRank.ts` and the neighbouring tests invest in ranks being legible.

**Suggested action:** state the contract in the `items` / `remoteFolderList` prop docblocks (both files already document their props in exactly this style) — the picker renders the supplied order, the server returns newest first, don't re-sort client-side. If rank order matters for multihost, the fix is a secondary sort by `rank` in the listing (`get_remote_performance_folders`), not a `sort` prop on the picker, which would re-open the door this PR closes.

src/components/report-selection/LocalFolderPicker.tsx
```diff
@@ -149,7 +133,7 @@ const LocalFolderPicker = ({
         <>
             <Select<ReportFolder>
                 className='folder-picker'
-                items={sortedItems}
+                items={items ?? []}
```
aidemsined, line 139: > [!WARNING]
> The link badges keep their meaning but lose the only way of being found in a long list.

With ordering gone, link status is neither sortable nor searchable: the local listing is uncapped (`views.py:1034-1043` appends every valid directory), and the `itemPredicate` on this line matches path text only — same in the remote selector. A linked report at position 40 is invisible until the user scrolls to it, where before it was at the top.

I checked whether this reverses something that was asked for, and it doesn't. The sort arrived incidentally inside #1775's review rounds (`fe0b17f5` → `0a4ed14a`, later moved by #1782); #1775 closed #1767, which is about selectability without SSH and says nothing about ordering, and none of that PR's 19 threads asked for the sort. So this is a product call you're entitled to make — but with no issue referenced, a commit titled "Remove behaviour", and the explanatory comment deleted, nothing records that long-list discoverability was weighed.

**Suggested action:** say in the PR body whether that was considered, so the next person seeing "linked reports scattered through the list" doesn't just re-add the sort. Worth also claiming the incidental fix this earns: `RemoteSyncConfigurator.tsx:123` seeds the default performance selection from `remotePerformanceFolderList[0]` — index 0 of the *raw* server list — which previously disagreed with whatever the sorted picker showed first.

tests/remoteFolderSelector.spec.tsx
```diff
@@ -1667,6 +1672,33 @@ const renderPerformanceSelector = async (
...
+        expect.stringContaining(`Rank 1: ${TIMESTAMP}`),
+        expect.stringContaining(`Rank 0: ${TIMESTAMP}`),
+    ]);
+};
+
+it('preserves supplied order regardless of performance report link status', async () => {
+    await expectSuppliedOrder(ReportKind.PERFORMANCE);
+});
+
+it('preserves supplied order regardless of memory report link status', async () => {
```
aidemsined: > [!WARNING]
> This case executes the same code path as the `PERFORMANCE` one above it, so it adds no coverage.

`type` reaches only `formatRemoteFolderPath` and `filterFolders`. `formatRemoteFolderPath` short-circuits to `getRemoteFolderLabel(folder)` whenever `rank` is set (`RemoteFolderSelector.tsx:177-179`), and every `rankFolder(n)` fixture sets `rank`, so the `type`-dependent `paths` lookup never runs; `filterFolders` is unreachable with an empty query. The two tests differ by a value that provably can't change the outcome, and the naming compounds it — a "memory report" case routed through `renderPerformanceSelector`, against a `multihostConnection` whose `profilerPath` is `''`.

For the record, the order assertions themselves are sound: I re-inserted `sortByFolderLinkState` and re-ran the three specs, and exactly the three order tests failed and nothing else — so they're load-bearing, not transcriptions of current output. The local fixture being declared in the reverse of the old ranking is what makes that work.

**Suggested action:** either drop this test, or give it `rank: null` folders so `type` actually drives `formatRemoteFolderPath` and the case earns its place. At minimum fix the name so it doesn't claim a "memory report" path it never enters.

tests/remoteFolderSelector.spec.tsx
```diff
@@ -1650,6 +1650,9 @@ const renderPerformanceSelector = async (
     connectionList: RemoteConnection[],
     folderList: RemoteFolder[] = multihostFolders,
     onSelectFolder: (folder: RemoteFolder) => void = () => undefined,
+    type: RemoteFolderType = ReportKind.PERFORMANCE,
+    linkedIds?: Set<string>,
+    unlinkedIds?: Set<string>,
```
aidemsined: > [!WARNING]
> Six positional parameters, four of them optional, with two adjacent `Set<string>` args at the tail.

The only new caller has to pass a throwaway `() => undefined` purely to reach `type`, and the two Sets are the classic argument-order trap — swap them and the test still compiles and still passes for the wrong reason. An options object is unambiguously this repo's shape for test render helpers (`MlirServerDialog.spec.tsx:118`, `RemoteConnectionDialog.spec.tsx:162`, `OpGraphPanelPerfMetric.spec.tsx:36`, `OpGraphToolbarSwitches.spec.tsx:26`), and the component under test makes the same call for the same reason — `remoteFolderRenderer` takes a named options interface rather than seven positional args.

**Suggested action:** this file already solved the same problem two hundred lines up, with a generic `selectRemoteFolder(type, remotePath)` plus thin `selectProfilerFolder` / `selectPerformanceFolder` wrappers. Mirroring that — a generic `renderRemoteFolderSelector({ type, folderList, linkedIds, unlinkedIds })` with `renderPerformanceSelector` kept as a one-line wrapper — takes the options object *and* leaves the eight existing call sites untouched.

aidemsined (commented):
The removal itself is clean and nothing here blocks merge. All three removed symbols have zero references left anywhere in the tree (docs and `AGENTS.md`/`CONVENTIONS.md` included), no inline reimplementation of the ranking survives, the badges are untouched, and all three `FolderLinkState` members still earn their keep in `FolderLinkStatusIcon`. I ran the three touched specs against the PR head — 71 tests pass, `tsc --noEmit` and `eslint` clean — and then re-inserted `sortByFolderLinkState` and re-ran: exactly the three order tests failed and nothing else, so those assertions are load-bearing rather than a transcription of the new output. Four comments inline; the substantive one is that the ordering rationale ends up recorded nowhere.

## General observations

- `remoteFolderList?.length === 0` at `RemoteFolderSelector.tsx:111` is now the lone survivor of a nullish defence this diff removes everywhere else in the file — `:117` and `:141` both dereference bare. Not a crash path: the prop has no `null` in its type, the destructuring default covers `undefined`, and I walked every call site (both `RemoteSyncConfigurator` props are `useState<RemoteFolder[]>` values that are only ever reassigned to `[]` or `mergeRemoteFolders`' output; the three spec renders pass typed arrays; no `.js` specs; `tsc` clean). Dropping the `?.` is a one-character in-hunk cleanup. I'd leave the two pickers' *prop types* alone though — `items: ReportFolder[] | null` is a required prop with a documented loading state that `isDisabled` genuinely branches on, and the conventions bless both shapes.
- The delete-row test's switch to `folders[1]` is fine and I wouldn't ask for the old lookup back — that test passes no link sets, so the sort was already a no-op there, and the scoped `getByLabelText` still throws on a row/folder desync. The only loss is the deleted comment explaining why the indirection existed.
- Worth a comment at the `folders` fixture (`LocalFolderPicker.spec.tsx:37-41`) noting that its declared order is deliberately the reverse of the old ranking — that's what makes the flipped expectation a real assertion, and someone tidying the fixture alphabetically would quietly neuter it.
- Nothing asserts row↔badge pairing now that no reordering step sits between `items` and the renderer. The order tests check labels, the badge tests check that badges appear, but nothing checks that *this* row carries *that* badge. Structural and low-risk, but it's the one property the removal plausibly disturbs.
- `reportLinks.spec.ts` is left holding a single-test `shouldShowFolderLinkStatus` describe whose title ("hides badges when both sets are empty or null") under-describes its `true` assertion, and `folderLinkStatus.ts` coverage still lives in a spec named after a different module. A `tests/folderLinkStatus.spec.ts` is the obvious follow-up now that it's down to three assertions — the neighbouring `ReportLinkStatus`, `useReportLinkBadgeIds` and `useReportLinkMatch` specs each get their own file.
- `folderLinkStatus.ts` should stay its own module at 25 lines, for what it's worth — `src/functions/` is full of one- and two-export modules, and it's the behaviour half of `definitions/FolderLinkStatus.ts`. Merging it into `reportLinks.ts` would cross concerns (persisted link model vs per-row presentation) and folding it into the pickers' folder would make one component import from the other's.
- Pre-existing, but this change makes it the sole determinant of what the user sees: in `SERVER_MODE`, `views.py:1017`/`:1122` build the directory list via `list(set(...))` and `list.sort` is stable, so identical `st_mtime` entries retain hash-set order, which varies with `PYTHONHASHSEED` between server processes. Needs an exact mtime tie to bite — reports unpacked in the same tick, or copied with `-p`.
- Dropping the two `useMemo`s is a perf non-event in both directions, which I'd rather say than leave implied. They guarded a Set-lookup comparator over a dropdown list (tens of reports, a few hundred at worst); `items ?? []` returns `items` itself when non-null so identity is as stable as the memo's output was, and the fresh `[]` only happens in the `null` case where the picker is disabled anyway. Small net win, in fact: `items` identity no longer churns when `linkedIds`/`unlinkedIds` change while the popover is open, so Blueprint's keyboard-active row no longer resets when a link is recorded mid-selection.

tests/remoteFolderSelector.spec.tsx
dcblundell: Addressed in 7d2058cc. The redundant memory-report test was removed; the remaining order test covers the behaviour without claiming unused type-specific coverage.

src/components/report-selection/LocalFolderPicker.tsx
dcblundell: Addressed in the PR description and tracked in follow-up issue #1986. The PR now records that link-status sorting was intentionally removed; the picker preserves backend order and keeps the badges visible.

tests/remoteFolderSelector.spec.tsx
dcblundell: Addressed in 7d2058cc. The helper now takes a named options object, eliminating the positional argument and adjacent Set ordering traps.

src/components/report-selection/RemoteFolderSelector.tsx
dcblundell: Addressed in 7d2058cc. Both picker prop docs now state that folders render in supplied order and link-status badges do not reorder them. Multihost rank ordering is intentionally out of scope for this PR.

Issue #1810, opened 2026-07-29
Memory↔performance reports from the same run fail to link on multi-device captures


dcblundell:
Memory and performance reports captured from the same multi-device run are reported as "Failed to link" even though their device-operation sequences align exactly. The cause is `collapseMultideviceOPs` in `src/hooks/useAPI.tsx`, which discards most of the memory report's device-operation list whenever `devices.length > 1`.

Observed on `v0.95.1` (`dev` @ ce466d29). Reproduced with a remote-synced pair, but remote sync is incidental — nothing in the matching path depends on how the report was loaded.

## Evidence

Pair: `resnet50_jul28_1524` (memory) + `2026_07_28_15_27_43` (performance), synced from `wh-05`, 2 devices.

Extracting both sequences directly from the synced files — the memory report's `captured_graph` nodes filtered by `isDeviceOperation`, and the perf CSV's device rows:

| | count |
|---|---|
| Memory report flattened device-op list | 324 |
| Perf report rows (devices merged, the default) | 324 |
| Positional mismatches between the two | **0** |

The two lists are identical, element for element. The link should succeed.

## Root cause

`useReportLinkMatch` returns `UNLINKED` when `useGetDeviceOperationListPerf` yields an empty array. That hook aligns the two lists strictly by index:

```ts
// src/hooks/useAPI.tsx
const isValid = deviceOperations.every((deviceOperation, index) => {
    const perfData = data.report[index];
    if (perfData && perfData.raw_op_code === deviceOperation.name) { ... return true; }
    return false;
});
return isValid ? deviceOperations : [];
```

The `deviceOperations` side comes from `useGetDeviceOperationsList`, which runs the flattened list through `collapseMultideviceOPs(result, devices.length)`:

```ts
// src/hooks/useAPI.tsx
if (numDevices === 1) {
    return data;
}
// ...
if (!seen.has(key) && operationCountByKey.get(key) === numDevices) {
    result.push(item);
    seen.add(key);
}
```

This assumes a multi-device memory report records each device op **once per device**, so it keeps only `name-id` keys occurring exactly `numDevices` times. This report records each op once, so the filter drops 321 of 324 entries, leaving three `PadDeviceOperation` entries (ops 5, 104, 203).

[...]

PR #1985, opened 2026-09-08
Fix report linking when nested operations are in a different order

### LINKED ISSUE #1860: Memory - performance report linking fails when nested device ops have different graph and profiler order
## Summary

Memory and performance reports generated by the same process fail to link when the captured graph contains nested device operations.

TTNN Visualizer currently builds the memory-side matching sequence from `function_start` events and compares it positionally with profiler rows. For nested operations, however, the profiler row order can differ from the graph's `function_start` order.

This causes `matchDeviceOperationsToPerf()` to return an empty mapping even though both reports describe the same run and contain the same operations.

## Reproduction

I observed this while profiling a four-layer Qwen3.6 model that uses
`SparseMatmulDeviceOperation`.

The reports were generated in the same process. The relevant nested operation
is produced because `SparseMatmulDeviceOperation::create_output_tensors()`
initializes its output through `ttnn::zeros_like()`, which launches a nested
`UnaryDeviceOperation`.

The resulting graph lifecycle is:

```text
function_start SparseMatmulDeviceOperation
  function_start UnaryDeviceOperation
  function_end   UnaryDeviceOperation
function_end   SparseMatmulDeviceOperation
```

Comparing the reports from ttnn and profiler:

| Sequence | Count |
|---|---:|
| Memory-side device operations | 334 |
| Merged profiler rows | 334 |
| Positional mismatches using `function_start` order | 16 |
| Positional mismatches using `function_end` order | 0 |

All 16 mismatches came from eight occurrences of this nested `SparseMatmulDeviceOperation` / `UnaryDeviceOperation` pair.

The memory-side sequence is currently extracted from `function_start` events,
so this nested lifecycle produces:

```text
TTNN graph (`function_start` order):
SparseMatmulDeviceOperation
UnaryDeviceOperation
```

However, the corresponding profiler rows appear in the opposite order:

```
Profiler row order:
UnaryDeviceOperation
SparseMatmulDeviceOperation
```

The outer SparseMatmul graph event is recorded before
`create_output_tensors()` runs. During output creation, `ttnn::zeros_like()`
launches and profiles the nested Unary operation. The outer SparseMatmul
profiler event is emitted only afterward, when its own workload is enqueued.
As a result, the profiler order for this pair matches the graph's
function_end order:

```
TTNN graph (`function_end` order):
UnaryDeviceOperation
SparseMatmulDeviceOperation
```

Therefore, the existing `function_start`-based positional match rejects the reports, while the `function_end` sequence matches them without a mismatch.

This does not imply that the profiler directly reads or sorts by graph `function_end` events. The profiler event is emitted at a different point in the device-operation launch path. I have included the detailed cause in a comment below.

## Current behavior

`fetchOperations()` extracts device operation names from `function_start` events. `matchDeviceOperationsToPerf()` then requires every operation name to match the profiler row's `raw_op_code` at the same index.

A single mismatch causes the entire mapping to become empty. The performance table subsequently displays:

```text
Load a synced profiler report to see tensor details
```

even though both reports came from the same run.

## Suggested behavior

Preserve the existing `function_start` order as the primary matching candidate, but retry with a `function_end` candidate when the start-order match fails.

Conceptually:

```text
1. Extract the device-operation sequence using function_start events.
2. Try the existing raw/collapsed matching logic.
3. If it fails, extract the sequence using function_end events.
4. Retry the same raw/collapsed matching logic.
```

Graph rendering should continue to consume the original lifecycle stream. The `function_end` sequence should only be an additional report-linking candidate, so this does not change parent-first graph rendering.

I have implemented a prototype with a regression test on this [fork](https://github.com/kc-ml2/ttnn-visualizer/tree/fix-bug-function-start-end-mismatch) and **would be happy to open a PR if this approach looks reasonable**.

The prototype parameterizes device-operation extraction by node type and passes both `function_start` and `function_end` candidate sequences to the matcher. It preserves `function_start` as the preferred candidate and only uses `function_end` as a fallback.

The original reports are large and contain model-specific metadata, so I have not attached them publicly. I can provide them if needed.

dcblundell:
This pull request refactors the logic for matching device operations to performance data, improving support for nested and reordered device operations. It introduces a new matching strategy that first tries to match using the "function start" order and, if that fails but both orders contain the same operations, falls back to the "function end" order. The update also enhances test coverage for these scenarios and adapts the hooks and memoization to support the new matching logic.

**Device Operation Matching Enhancements:**

* Added `matchDeviceOperationOrdersToPerf` to prefer matching device operations in "function start" order, with a fallback to "function end" order when both contain the same operations. This improves alignment for nested or reordered operations.
* Updated hooks and memoization (`useGetDeviceOperationListPerf`, `getDeviceOperationOrderCandidates`, etc.) to use the new matching function and support both operation orders. [[1]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L34-R34) [[2]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L861-R898) [[3]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L906-R929) [[4]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L920-R957)

**API and Data Handling Updates:**

* Refactored how device operation names are extracted, now parameterized by node type (function start or end), and updated data flow in `fetchOperations` and related code. [[1]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L187-R203) [[2]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L244-R245)

**Testing Improvements:**

* Added comprehensive tests for the new matching logic, including fallback behavior, preference for start order, and rejection of mismatched cases.
* Extended integration tests to verify that nested operations are correctly matched and linked in both orders, ensuring public APIs remain consistent. [[1]](diffhunk://#diff-187f5177dcfca23ad809d6d54c3bcdde3d8d3fd871a2046aa574a94a949894f1R261-R278) [[2]](diffhunk://#diff-187f5177dcfca23ad809d6d54c3bcdde3d8d3fd871a2046aa574a94a949894f1R343-R379)

**Code Maintenance:**

* Updated imports and cache clearing to reflect the new function and memoization structure. [[1]](diffhunk://#diff-ddfd37f198b2396809dd223191aea15d18d1967abd65138c898cfc80556d0806L6-R10) [[2]](diffhunk://#diff-187f5177dcfca23ad809d6d54c3bcdde3d8d3fd871a2046aa574a94a949894f1R20) [[3]](diffhunk://#diff-28aac87d93df3e1b341a768b2c4850fe92744b5954a9afb7ef5962a6107ac604L906-R929)

These changes improve the robustness and flexibility of device operation to performance row matching, especially for complex nested cases.

Fixes https://github.com/tenstorrent/ttnn-visualizer/issues/1860.

### DIFF
--- a/src/functions/deviceOperationMatching.ts
+++ b/src/functions/deviceOperationMatching.ts
@@ -112,3 +112,58 @@ export const matchDeviceOperationsToPerf = (
 
     return alignToPerfRows(collapseMultideviceOperations(deviceOperations, numDevices), alignableRows);
 };
+
+const hasSameDeviceOperations = (
+    functionStartOperations: DeviceOperationMapping[],
+    functionEndOperations: DeviceOperationMapping[],
+): boolean => {
+    if (functionStartOperations.length !== functionEndOperations.length) {
+        return false;
+    }
+
+    const operationCountByKey = new Map<string, number>();
+
+    for (const { id, name } of functionStartOperations) {
+        const key = JSON.stringify([id, name]);
+        operationCountByKey.set(key, (operationCountByKey.get(key) ?? 0) + 1);
+    }
+
+    for (const { id, name } of functionEndOperations) {
+        const key = JSON.stringify([id, name]);
+        const remaining = operationCountByKey.get(key);
+
+        if (!remaining) {
+            return false;
+        }
+
+        operationCountByKey.set(key, remaining - 1);
+    }
+
+    return true;
+};
+
+/**
+ * @description Try the captured graph's parent-first function-start order,
+ * then its child-first function-end order when both contain the same device
+ * operations. Each order retains the raw-then-multidevice-collapse fallback.
+ */
+export const matchDeviceOperationOrdersToPerf = (
+    functionStartOperations: DeviceOperationMapping[],
+    functionEndOperations: DeviceOperationMapping[],
+    perfRows: PerfTableRow[],
+    numDevices: number,
+): DeviceOperationMapping[] => {
+    const functionStartMatch = matchDeviceOperationsToPerf(functionStartOperations, perfRows, numDevices);
+
+    if (functionStartMatch.length > 0) {
+        return functionStartMatch;
+    }
+
+    // An interrupted capture can omit function-end events. Since alignment
+    // tolerates trailing perf rows, only a complete reordering is safe to retry.
+    if (!hasSameDeviceOperations(functionStartOperations, functionEndOperations)) {
+        return [];
+    }
+
+    return matchDeviceOperationsToPerf(functionEndOperations, perfRows, numDevices);
+};
--- a/src/functions/deviceOperationMatching.ts
+++ b/src/functions/deviceOperationMatching.ts
@@ -3,7 +3,7 @@
 // SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 
 import { PerfTableRow } from '../model/PerfTable';
-import { DeviceOperationMapping } from '../model/DeviceOperationMapping';
+import { DeviceOperationMapping, DeviceOperationOrderCandidates } from '../model/DeviceOperationMapping';
 import { OpType } from '../definitions/Performance';
 
 /**
@@ -166,8 +166,7 @@ const hasSameDeviceOperations = (
  * cannot suppress the complete end-order match. See #1860.
  */
 export const matchDeviceOperationOrdersToPerf = (
-    functionStartOperations: DeviceOperationMapping[],
-    functionEndOperations: DeviceOperationMapping[],
+    { functionStartOperations, functionEndOperations }: DeviceOperationOrderCandidates,
     perfRows: PerfTableRow[],
     numDevices: number,
 ): DeviceOperationMapping[] => {
--- a/src/functions/deviceOperationMatching.ts
+++ b/src/functions/deviceOperationMatching.ts
@@ -83,6 +83,26 @@ const alignToPerfRows = (
 const alignableRowsOf = (perfRows: PerfTableRow[]): PerfTableRow[] =>
     perfRows.filter((perfRow) => perfRow.op_type !== OpType.SIGNPOST);
 
+const alignCollapsedToPerfRows = (
+    deviceOperations: DeviceOperationMapping[],
+    perfRows: PerfTableRow[],
+    numDevices: number,
+): DeviceOperationMapping[] => {
+    if (numDevices <= 1) {
+        return [];
+    }
+
+    const collapsedOperations = collapseMultideviceOperations(deviceOperations, numDevices);
+
+    // A valid per-device list contributes every operation exactly once per
+    // device. A smaller subset can prefix-match and falsely mark a report linked.
+    if (collapsedOperations.length * numDevices !== deviceOperations.length) {
+        return [];
+    }
+
+    return alignToPerfRows(collapsedOperations, perfRows);
+};
+
 /**
  * @description Match a memory report's device operations against a performance
  * report's rows, returning the mappings when the two sequences describe the same
@@ -104,13 +124,7 @@ export const matchDeviceOperationsToPerf = (
         return directMatch;
     }
 
-    // With nothing to collapse on, the fallback would repeat the pass that just
-    // failed.
-    if (numDevices <= 1) {
-        return [];
-    }
-
-    return alignToPerfRows(collapseMultideviceOperations(deviceOperations, numDevices), alignableRows);
+    return alignCollapsedToPerfRows(deviceOperations, alignableRows, numDevices);
 };
 
 const hasSameDeviceOperations = (
@@ -143,9 +157,13 @@ const hasSameDeviceOperations = (
 };
 
 /**
- * @description Try the captured graph's parent-first function-start order,
- * then its child-first function-end order when both contain the same device
- * operations. Each order retains the raw-then-multidevice-collapse fallback.
+ * TODO: remove once memory and performance reports carry a shared run id (#1800)
+ * @description Match nested device operations whose profiler events can be
+ * child-first because a child's workload is enqueued before its parent's. Start
+ * order remains preferred for reports whose profiler rows are parent-first;
+ * end order is the fallback when the same operations are child-first. Raw orders
+ * are tried before either multi-device collapse so a spurious collapsed prefix
+ * cannot suppress the complete end-order match. See #1860.
  */
 export const matchDeviceOperationOrdersToPerf = (
     functionStartOperations: DeviceOperationMapping[],
@@ -153,7 +171,8 @@ export const matchDeviceOperationOrdersToPerf = (
     perfRows: PerfTableRow[],
     numDevices: number,
 ): DeviceOperationMapping[] => {
-    const functionStartMatch = matchDeviceOperationsToPerf(functionStartOperations, perfRows, numDevices);
+    const alignableRows = alignableRowsOf(perfRows);
+    const functionStartMatch = alignToPerfRows(functionStartOperations, alignableRows);
 
     if (functionStartMatch.length > 0) {
         return functionStartMatch;
@@ -165,5 +184,17 @@ export const matchDeviceOperationOrdersToPerf = (
         return [];
     }
 
-    return matchDeviceOperationsToPerf(functionEndOperations, perfRows, numDevices);
+    const functionEndMatch = alignToPerfRows(functionEndOperations, alignableRows);
+
+    if (functionEndMatch.length > 0) {
+        return functionEndMatch;
+    }
+
+    const collapsedFunctionStartMatch = alignCollapsedToPerfRows(functionStartOperations, alignableRows, numDevices);
+
+    if (collapsedFunctionStartMatch.length > 0) {
+        return collapsedFunctionStartMatch;
+    }
+
+    return alignCollapsedToPerfRows(functionEndOperations, alignableRows, numDevices);
 };
--- a/src/hooks/useAPI.tsx
+++ b/src/hooks/useAPI.tsx
@@ -854,16 +854,16 @@ export const useGetDeviceOperationsListByOp = () => {
     }, [operations]);
 };
 
-// Memoised across call sites, not per hook invocation: both derived values are
-// read by a handful of hooks and by one component instance per virtualised row,
-// and `useMemo` would run the flatMap and the O(rows) match once for each. The
-// inputs are React Query results, so their identity is shared by every caller in
-// a render pass. Callers must not mutate the results — they share them now.
 interface DeviceOperationOrderCandidates {
     functionStartOperations: DeviceOperationMapping[];
     functionEndOperations: DeviceOperationMapping[];
 }
 
+// Memoised across call sites, not per hook invocation:
+// `useGetDeviceOperationListPerf` is read by several hooks and by one component
+// instance per virtualised row, so `useMemo` would rebuild both order candidates
+// and rerun the O(rows) match for each. The inputs are shared React Query results.
+// Callers must not mutate the derived values — they share them now.
 const getDeviceOperationOrderCandidates = memoiseLatest(
     (operations?: OperationDescription[]): DeviceOperationOrderCandidates => {
         const functionStartOperations: DeviceOperationMapping[] = [];
@@ -930,17 +930,6 @@ export const clearReportCaches = (queryClient: QueryClient) => {
     getDeviceOperationListPerf.reset();
     getOpToPerfIds.reset();
     getDeviceOperationListPerfByOpId.reset();
-};
-
-/**
- * @description Every device operation in the memory report, flattened in report
- * order. Multi-device collapsing happens at match time, not here, because only
- * the performance report reveals which shape this report has (#1810).
- */
-export const useGetDeviceOperationsList = (): DeviceOperationMapping[] => {
-    const { data: operations } = useOperationsList();
-
-    return getDeviceOperationOrderCandidates(operations).functionStartOperations;
 };
 
 export const useGetDeviceOperationListPerf = () => {
--- a/src/hooks/useAPI.tsx
+++ b/src/hooks/useAPI.tsx
@@ -30,7 +30,7 @@ import parseMemoryConfig, { memoryConfigPattern } from '../functions/parseMemory
 import { MemoryConfig } from '../model/MemoryConfig';
 import getServerConfig from '../functions/getServerConfig';
 import { PerfTableRow } from '../model/PerfTable';
-import { DeviceOperationMapping } from '../model/DeviceOperationMapping';
+import { DeviceOperationMapping, DeviceOperationOrderCandidates } from '../model/DeviceOperationMapping';
 import { matchDeviceOperationOrdersToPerf } from '../functions/deviceOperationMatching';
 import memoiseLatest from '../functions/memoiseLatest';
 import {
@@ -854,11 +854,6 @@ export const useGetDeviceOperationsListByOp = () => {
     }, [operations]);
 };
 
-interface DeviceOperationOrderCandidates {
-    functionStartOperations: DeviceOperationMapping[];
-    functionEndOperations: DeviceOperationMapping[];
-}
-
 // Memoised across call sites, not per hook invocation:
 // `useGetDeviceOperationListPerf` is read by several hooks and by one component
 // instance per virtualised row, so `useMemo` would rebuild both order candidates
@@ -934,13 +929,12 @@ export const clearReportCaches = (queryClient: QueryClient) => {
 
 export const useGetDeviceOperationListPerf = () => {
     const { data: operations } = useOperationsList();
-    const { functionStartOperations, functionEndOperations } = getDeviceOperationOrderCandidates(operations);
+    const deviceOperationOrderCandidates = getDeviceOperationOrderCandidates(operations);
     const { data: devices } = useDevices();
     const { data } = useLinkedPerformanceReport();
 
     return getDeviceOperationListPerf(
-        functionStartOperations,
-        functionEndOperations,
+        deviceOperationOrderCandidates,
         (data ?? EMPTY_PERF_RETURN).report,
         devices?.length ?? 0,
     );
--- a/src/hooks/useAPI.tsx
+++ b/src/hooks/useAPI.tsx
@@ -31,7 +31,7 @@ import { MemoryConfig } from '../model/MemoryConfig';
 import getServerConfig from '../functions/getServerConfig';
 import { PerfTableRow } from '../model/PerfTable';
 import { DeviceOperationMapping } from '../model/DeviceOperationMapping';
-import { matchDeviceOperationsToPerf } from '../functions/deviceOperationMatching';
+import { matchDeviceOperationOrdersToPerf } from '../functions/deviceOperationMatching';
 import memoiseLatest from '../functions/memoiseLatest';
 import {
     PerformanceReportParams,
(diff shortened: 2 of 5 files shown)


src/functions/deviceOperationMatching.ts
```diff
@@ -112,3 +112,58 @@ export const matchDeviceOperationsToPerf = (
...
+ */
+export const matchDeviceOperationOrdersToPerf = (
+    functionStartOperations: DeviceOperationMapping[],
+    functionEndOperations: DeviceOperationMapping[],
+    perfRows: PerfTableRow[],
+    numDevices: number,
+): DeviceOperationMapping[] => {
+    const functionStartMatch = matchDeviceOperationsToPerf(functionStartOperations, perfRows, numDevices);
+
+    if (functionStartMatch.length > 0) {
```
aidemsined, line 176: > [!CAUTION]
> On a multi-device report the *collapsed* start pass can win with a tiny spurious match, and this gate then suppresses the end-order retry — so the fix silently never fires.

`matchDeviceOperationsToPerf` already carries its own raw→collapsed fallback, so `functionStartMatch` is non-empty if *either* start pass succeeded. `collapseMultideviceOperations` keeps only keys seen exactly `numDevices` times, so the collapsed list can be a small subset of the real sequence — and `alignToPerfRows` tolerates trailing perf rows, so a subset that happens to match the first rows is a "success".

Reproduced:

```
start = [Pad(1), Pad(1), Matmul(2), Outer(3), Inner(3)]
end   = [Pad(1), Pad(1), Matmul(2), Inner(3), Outer(3)]
perf  = [Pad, Pad, Matmul, Inner, Outer]

collapse(start, 2) → ['Pad']        // only Pad-1 appears exactly twice
numDevices: 2      → 1 entry  (Pad/1)
numDevices: 1      → 5 entries (correct, via the end-order retry)
```

`useReportLinkMatch` reports LINKED on `length > 0`, so the user sees a linked report with one of five device operations mapped. Not a regression — `dev` returns the same single entry — but it means this fix may not reach the multi-device reports it was written for, and #1860's model is a four-layer Qwen. The repeated-name-within-one-operation shape is already attested by the suite's own case at `tests/deviceOperationMatching.spec.ts:38` ("op 5 genuinely repeats `Pad` twice — which the collapse heuristic mistakes for per-device duplication").

**Suggested action:** try both **raw** orders before either **collapsed** order — raw-start, raw-end, collapsed-start, collapsed-end — rather than nesting the existing raw/collapsed fallback inside each order. Preferring the longest match would also work. Either way it wants the `numDevices > 1` test below.

src/functions/deviceOperationMatching.ts
```diff
@@ -112,3 +112,58 @@ export const matchDeviceOperationsToPerf = (
...
+        return functionStartMatch;
+    }
+
+    // An interrupted capture can omit function-end events. Since alignment
+    // tolerates trailing perf rows, only a complete reordering is safe to retry.
+    if (!hasSameDeviceOperations(functionStartOperations, functionEndOperations)) {
+        return [];
+    }
+
+    return matchDeviceOperationsToPerf(functionEndOperations, perfRows, numDevices);
```
aidemsined: > [!WARNING]
> The collapsed half of this retry can invent a partial link where `dev` correctly returned nothing.

The collapsed end order is a same-length permutation of the collapsed start order, so it can prefix-align on its own even when the raw end order cannot. Reproduced:

```
numDevices: 2
start = [Outer(1), Inner(1), Outer(1), Inner(1), Matmul(2)]
end   = [Inner(1), Outer(1), Inner(1), Outer(1), Matmul(2)]
perf  = [Inner, Outer, Matmul]

collapse(start, 2) → ['Outer', 'Inner']   // fails
collapse(end, 2)   → ['Inner', 'Outer']   // aligns on rows 0-1
pre-PR  → 0 entries  (UNLINKED)
post-PR → 2 entries, both op 1; op 2's Matmul silently dropped
```

Worth being precise about severity: this widens an existing hazard rather than creating a new class, because any successful collapsed match is a subset by construction — partial links were already reachable through the collapsed *start* pass. What's new is a second opportunity for one, and it converts a loud UNLINKED into a quiet partial link.

Note the pass reordering suggested above does **not** fix this one: with raw-first ordering this case still ends on collapsed-end and still returns two entries.

**Suggested action:** require a collapsed match to consume every alignable row before accepting it, or gate the end-order retry to the raw pass only. The second is the smaller change and keeps this PR's blast radius to the shape #1860 describes.

tests/deviceOperationMatching.spec.ts
```diff
@@ -192,6 +196,64 @@ describe('matchDeviceOperationsToPerf', () => {
     });
 });
 
+describe('matchDeviceOperationOrdersToPerf', () => {
```
aidemsined, line 204: > [!WARNING]
> Every case in this block passes `numDevices: 1`, which is exactly the quadrant the two findings above live in.

`matchDeviceOperationsToPerf` short-circuits at `numDevices <= 1`, so the collapse path never runs in any of these five cases, and the new hook test overrides its enclosing block's two-device fixtures with `[{ device_id: 0 }]` — so the whole suite exercises only start-raw → guard → end-raw. The pass sequence that actually ships for a multi-device report is four passes deep and completely uncovered.

For what it's worth the four-pass path *does* work on the shape it was designed for — I checked: a per-device-duplicated nested pair against merged perf rows resolves correctly on pass four. So this is a coverage gap rather than a second bug, but it is the gap that would have caught both findings above.

**Suggested action:** two cases with `numDevices: 2` — the working shape (per-device duplication plus a nested reorder, resolving on the collapsed end pass) and the suppression shape from the `Pad` reproduction above. `duplicatedPerDevice` at `:34` and `NUM_DEVICES`/`unmergedRows` in `useLinkedPerformanceReport.spec.tsx:229` already exist to build them from.

src/hooks/useAPI.tsx
```diff
@@ -858,21 +859,43 @@ export const useGetDeviceOperationsListByOp = () => {
 // and `useMemo` would run the flatMap and the O(rows) match once for each. The
 // inputs are React Query results, so their identity is shared by every caller in
 // a render pass. Callers must not mutate the results — they share them now.
-const getDeviceOperationsList = memoiseLatest((operations?: OperationDescription[]): DeviceOperationMapping[] => {
-    if (!operations) {
-        return [];
-    }
+interface DeviceOperationOrderCandidates {
```
> Inserting this interface here detached the memo comment above from what it documents — and that comment is now stale as well.

Two separate problems in the same few lines. The comment at `:857` ("Memoised across call sites, not per hook invocation…") documents `getDeviceOperationOrderCandidates` at `:867`, but this interface now sits between them, so on first read it documents the interface. And the comment says "`useMemo` would run the flatMap and the O(rows) match once for each" — the implementation below is two nested `for` loops; the flatMap is gone.

`CONVENTIONS.md:104` puts this in the repo's own blocker class: "Stale comments are bugs… Reviewers will flag stale comments as blockers."

**Suggested action:** move the comment below this interface so it sits immediately above the memo, and drop the flatMap clause. While there, "both derived values are read by a handful of hooks" is now only half true — `functionEndOperations` has exactly one reader.

src/hooks/useAPI.tsx
```diff
@@ -917,15 +940,21 @@ export const clearReportCaches = (queryClient: QueryClient) => {
 export const useGetDeviceOperationsList = (): DeviceOperationMapping[] => {
     const { data: operations } = useOperationsList();
 
-    return getDeviceOperationsList(operations);
+    return getDeviceOperationOrderCandidates(operations).functionStartOperations;
```
> This hook has no production callers left after the change, so the "public APIs remain consistent" guarantee in the PR body is guarding an export nothing reaches.

On `dev` its only consumer was `useGetDeviceOperationListPerf`. This diff rewires that hook to call `getDeviceOperationOrderCandidates(operations)` directly, so a repo-wide grep at this head finds only the declaration and the new test — nothing in `src/` renders it. I also checked by mutation: making it return a copy (`[...functionStartOperations]`) fails no test, which confirms nothing downstream depends on its identity either.

**Suggested action:** either delete it along with its docstring and the assertion at `tests/useLinkedPerformanceReport.spec.tsx:376`, or keep it and say in a comment why the export is retained. Worth noting its neighbour `useGetDeviceOperationsListByOp` is also unreferenced — that one is pre-existing, not yours, but the two together suggest this corner wants a sweep.

src/functions/deviceOperationMatching.ts
```diff
@@ -112,3 +112,58 @@ export const matchDeviceOperationsToPerf = (
...
+ * @description Try the captured graph's parent-first function-start order,
+ * then its child-first function-end order when both contain the same device
+ * operations. Each order retains the raw-then-multidevice-collapse fallback.
+ */
+export const matchDeviceOperationOrdersToPerf = (
+    functionStartOperations: DeviceOperationMapping[],
+    functionEndOperations: DeviceOperationMapping[],
+    perfRows: PerfTableRow[],
+    numDevices: number,
+): DeviceOperationMapping[] => {
```
aidemsined, line 172: > [!WARNING]
> Two adjacent parameters of the same type, where swapping them type-checks and silently inverts the preference the whole fix rests on.

The only thing keeping the call correct is the argument names at the single call site, which destructures the candidates object and then re-flattens it into two positional arguments. The repo's usual answer at this arity is a params object (`buildChartIndexEntries`, `buildLateDeallocationReports`), but a fresh object literal per render would defeat `memoiseLatest`'s reference key — which is presumably why it wasn't done here.

`DeviceOperationOrderCandidates` solves both at once: it is already reference-stable, because it comes out of the cache of one. Passing it drops the signature to three arguments with no same-type adjacency, shrinks the memo key, and makes "these two orders came from one pass over the same operations" structural rather than incidental. As written, nothing stops a future caller pairing a start list from the shared memo with an end list from a per-caller `useMemo` — which is simultaneously the swap trap and the memo-degradation trap `memoiseLatest`'s own docstring warns about.

**Suggested action:** take `candidates: DeviceOperationOrderCandidates` as the first parameter, and move the interface next to `DeviceOperationMapping` in `src/model/` once it crosses modules.

src/functions/deviceOperationMatching.ts
```diff
@@ -112,3 +112,58 @@ export const matchDeviceOperationsToPerf = (
...
+    }
+
+    return true;
+};
+
+/**
+ * @description Try the captured graph's parent-first function-start order,
+ * then its child-first function-end order when both contain the same device
+ * operations. Each order retains the raw-then-multidevice-collapse fallback.
+ */
```
aidemsined, line 167: > [!WARNING]
> This restates the four lines below it and leaves out the one thing a maintainer can't recover from the code.

Every other docstring in this file explains a *why* and cites its issue: `collapseMultideviceOperations` explains that two report shapes exist and nothing distinguishes them (#1810, #1800), `alignableRowsOf` explains that signposts are their own op type so `hideHostOps` misses them (#1943), `alignToPerfRows` explains why validation precedes allocation. This one names no issue and no mechanism.

The missing why is the interesting part: the profiler emits the outer op's event when *its own* workload is enqueued, so a nested child lands first in the perf stream while the graph records it second. Which means the end order is arguably the semantically correct sequence for nested ops, and start-first is backwards compatibility rather than correctness — worth saying out loud, because as written the preference reads as a correctness claim.

I checked whether it should therefore be flipped, and it should not: `segformer_decoder` in `demo-reports/` has 41 device operations with 8 positional differences between its start and end orders, and its perf order agrees with **start**. Preferring the end order would break a report already in the repo. That is worth a line in the docstring too, since it is the evidence that settles the ordering.

**Suggested action:** state the mechanism, cite #1860, and note that the fallback stack is now three layers deep (signpost filter #1943, raw→collapsed #1810, start→end here) — ideally with the same `TODO: remove once … shared run id (#1800)` marker `collapseMultideviceOperations` already carries, since a shared run id retires this layer too.

src/hooks/useAPI.tsx
```diff
@@ -858,21 +859,43 @@ export const useGetDeviceOperationsListByOp = () => {
...
+            for (const name of operation.deviceOperationNameList) {
+                functionStartOperations.push({
+                    name,
+                    id: operation.id,
+                    operationName: operation.name,
+                });
+            }
 
-const getDeviceOperationListPerf = memoiseLatest(matchDeviceOperationsToPerf);
+            for (const name of getDeviceOperationNameList(operation, NodeType.function_end)) {
```
aidemsined, line 880: > [!NOTE]
> The end order is derived here at render time while its twin is precomputed at fetch time — worth converging, but **not** on perf grounds.

`deviceOperationNameList` is built once inside `fetchOperations` (`:245`) and cached on the operation; this re-scans every operation's full `device_operations` frame stream inside a memo that runs during the render pass after the query settles. `getDeviceOperationNameList` is already parameterised, `OperationDescription` already carries one derived order, and `fetchOperations` already walks the same array for `processedConnections` — so a `deviceOperationEndNameList` sibling puts both orders on one testable boundary and reduces this helper to two flat maps.

I measured the cost before raising it, because it looks like a perf issue and isn't: **0.39 ms** for the largest shipped report (720 operations / 12,960 graph nodes), 2.3 ms at 3,000 operations / 600k nodes, and it runs once per operations-query identity rather than per render. So treat this as placement and symmetry only — if you'd rather leave it, that's defensible and I wouldn't push.

**Suggested action:** optional. If you do move it, it also answers "who pays for building the list `useGetDeviceOperationsList` never uses" rather than leaving it open.

aidemsined (commented):
The design is right and it matches what #1860 prescribed — start order preferred, end order retried only behind a guard, graph rendering left on the parent-first stream. The guard is load-bearing rather than decorative: deleting it fails all three rejection cases, and the length check is the mechanism — without it a truncated end list aligns against a perf prefix and hands back a silently partial link. The suite (220 files, 2505 tests), `tsc` and `eslint` are all clean, and the unchanged path really is unchanged: a report matching on start order runs the identical single call the old code made and returns before the guard.

Two things are worth resolving before merge, both in the same place, and both only reachable when `numDevices > 1` — which is the one quadrant with no test coverage at all. In short: because the raw→collapsed fallback is nested *inside* each order rather than sitting outside both, a spurious collapsed-start match can suppress the correct end-order retry, and the new collapsed-end pass can produce a partial link where `dev` returned nothing. Reproductions are in the inline comments.

Some positives worth recording, since they took measuring and would otherwise be re-litigated later:

- **Keep start-order first.** `segformer_decoder` in `demo-reports/` has 41 device operations with 8 positional differences between its start and end orders, and its perf order agrees with **start** — so preferring the end order, which the root-cause analysis might suggest, would break a report already in the repo.
- **A non-empty *raw* start match cannot be silently wrong.** Reordering is confined to one operation's contiguous block; permuting equal-named entries yields identical mappings and permuting differently-named ones changes the name sequence, which the positional check rejects. The hazard is specific to the collapsed pass.
- **Identity and memoisation hold.** Both array arguments come from the single memoised candidates object keyed on the shared query result; mutating either to a copy fails the existing identity test, so `useOpPerfRowScores` and `useTopNAnnotations` stay memo-stable. All four `memoiseLatest` caches are still reset by `clearReportCaches` after the rename.
- **The fixtures match the wire.** `NodeType` is a string enum, so `node_type: 'function_start'` is the real payload value, and all three shipped demo reports have exactly equal `function_start`/`function_end` counts (204/204, 284/284, 2146/2146) with no `function_end` node missing `params.name` — the symmetry the fallback rests on is real, not assumed.

## General observations

- `CONVENTIONS.md:529-556` documents this exact call site as `memoiseLatest(matchDeviceOperationsToPerf)` with the three-argument hook body. It's now stale, and the repo's maintenance contract makes updating it part of the change — the four-argument call is also the most interesting live example of the "every argument must come from a source all consumers share" constraint that section is about.
- `hasSameDeviceOperations` has no docstring, even though its purpose is subtle enough that the call site needs a comment to explain it. The truncation constraint should travel with the function rather than living at `:162`.
- Two key encodings for one conceptual `(id, name)` key in one file: `JSON.stringify([id, name])` here, `` `${name}-${id}` `` in `collapseMultideviceOperations`. The JSON form is the safer of the two (a name containing `-` can't collide), so the inconsistency is the wrong way round. The house style elsewhere is a named `*Key` helper — `endpointKey`, `getRunKey`, `cbDeviceKey`.
- The guard compares the **raw** lists, so an asymmetric per-device duplication (starts duplicated per device, ends recorded once) refuses a retry that would have been correct. Not attested in any shipped report, so I'd only note it in the comment rather than handle it.
- `hasSameDeviceOperations([], [])` returns `true`, so an empty end list runs a second alignment pass that `alignToPerfRows` rejects immediately. Harmless; an early length bail reads more plainly.
- The new hook test sits inside `describe('report matching under a filtered performance tab')`, whose fixtures assume `NUM_DEVICES = 2`, and overrides that block's entire axios mock including forcing a single device. Its own `describe` would be clearer about what it's setting up.
- No hook-level case covers the common direction — a nested capture whose perf report is in *start* order — which is the regression case for this change rather than the fix case.
- `clearReportCaches` has no test anywhere, and this PR renames one of its four reset targets. All four are still reset, but a future rename that missed one would leak the previous report's rows for the page lifetime with nothing failing.
- The commit message is "Code review round", which describes a process rather than the change and doesn't reference #1860. A squash merge will take the PR title, so this only costs the commit-level record.

src/hooks/useAPI.tsx
dcblundell: Removed in 04a7585f, together with the test-only import and public-list assertion. The integration test now exercises the production matching hook directly.

src/functions/deviceOperationMatching.ts
dcblundell: Fixed in 04a7585f. Matching now tries raw start and raw end before either collapsed order, so a spurious collapsed start prefix cannot suppress a complete end-order match. Added the Pad/Matmul/Outer/Inner reproduction as a regression test.

tests/deviceOperationMatching.spec.ts
dcblundell: Added in 04a7585f. Coverage now includes a valid two-device nested sequence that resolves on collapsed end order, the collapsed-start suppression reproduction, and rejection of a collapsed end prefix that omits a non-duplicated operation.

src/functions/deviceOperationMatching.ts
dcblundell: Fixed in 04a7585f. Collapsed matching is now accepted only when the collapsed entries account for the entire raw memory-side sequence (`collapsed.length * numDevices === raw.length`), preserving trailing perf-row tolerance while rejecting subsets that omit non-duplicated operations. Added the reported partial end-order case.

src/hooks/useAPI.tsx
dcblundell: Fixed in 04a7585f. The comment now sits directly above the memoised derivation and describes rebuilding both order candidates plus rerunning the match; the stale `flatMap` and inaccurate reader wording are gone.

src/functions/deviceOperationMatching.ts
dcblundell: Updated in 04a7585f. The docstring now explains child-first profiler enqueue ordering, why start order remains preferred for compatibility, why raw orders precede collapsed orders, cites #1860, and carries the shared-run-id #1800 TODO.

src/hooks/useAPI.tsx
dcblundell: I’m leaving this as-is. The measured scan cost is negligible and runs once per operations-query identity; keeping both candidates in one memoised derivation also preserves their shared identity without adding an internal-only field to the API-shaped `OperationDescription`. Removing the unused public hook in 04a7585f also makes the ownership clearer: the end-order list is built solely for report matching.

src/functions/deviceOperationMatching.ts
dcblundell: Fixed in 95c809a7. `DeviceOperationOrderCandidates` now lives beside `DeviceOperationMapping` in `src/model/`, the matcher accepts that stable object as one parameter, and `useGetDeviceOperationListPerf` passes the memoised candidate object through without destructuring it into swappable arrays.

Issue #1991, opened 2026-09-08
Three loose ends in the op-graph fold state


aidemsined:
Three small behaviour questions found reviewing `feature/1976-op-graph-layer-semantics`, each one a decision rather than a defect, so none was taken during remediation.

**The layout cache limit predates weight fans.** `LAYOUT_CACHE_LIMIT` was cut from 16 to 8 when block folding made every detected instance an independently toggleable key dimension, and its comment says so. Fans multiply that dimension by roughly an order of magnitude — `bge_m3` has 293 `from_torch` ops in fans of about six, so ~48 independently toggleable fans, `sentence_bert` ~25 — against an 8-entry LRU of fully laid-out graphs. A user exploring more than eight fans pays a full Dagre pass for a state visited two clicks earlier, and on an 8k-op report that pass is seconds. Worth re-deriving the limit now the option space has changed shape.

**Opening the first fan mints a maximal cache key for an identical graph.** Folding or unrolling one instance from the `null` default has to materialise every sibling as expanded first (`new Set(null)` is empty, and deleting from that would fold the whole graph). So the first fan click turns the key from `none` into a sorted join of every grouping instance id, describing a graph identical to the one already cached under `none` — one redundant layout per session, plus a long key.

**The toolbar shows the outgoing detector's count for a frame after a grouping switch.** `detectedBlocks` is not cleared by the switch and the buttons only disable after the 200 ms spinner delay, so REPEATS(89) → LAYERS briefly reads "Layers (89)" — and Unroll/Fold render and then vanish if the new detector finds nothing. The count exists so the two strategies are comparable by switching, which is exactly the moment it is briefly wrong.

Issue #1990, opened 2026-09-08
Layer detection finds nothing in a report whose op names use `::`


aidemsined:
Role detection matches on the leaf of an op name, and `leafNameOf` derives it with `name.slice(name.lastIndexOf('.') + 1)` — a `.` split only. Given `ttnn::transformer::scaled_dot_product_attention` that returns the whole string, which matches no anchor and no delimiter, so the report partitions into nothing and the Layers button reads "no layers detected" with nothing to say why.

That both spellings reach this code is already recorded in the codebase: `DEALLOCATE_OP_NAME_LIST` lists `ttnn.deallocate` *and* `ttnn::deallocate` precisely because reports arrive with either, and the same filter feeds the sequence the detector runs on. The role tables themselves were partly read off the C++ names (`ttnn::operations::reduction::moe` is where the `moe` anchor comes from), so the vocabulary is right and only the split is wrong.

The fix is one line — `name.split(/[.:]+/).pop()` — but it is not verifiable against anything currently checked in: all four fixtures under `tests/fixtures/opRoles/` are dot-spelled, so the change would ship unexercised. Wants a `::`-spelled capture as a fifth fixture, which also pins that the anchor tables really are namespace-agnostic rather than accidentally dot-shaped.

Found while reviewing `feature/1976-op-graph-layer-semantics`.

Issue #1989, opened 2026-09-08
A layer span can still swallow a graph of 32 ops or fewer


aidemsined:
`MIN_LAYER_SPAN_ALLOWANCE = 32` exists so the 25% fraction does not reject a small capture's real layer, but it re-admits the exact case the bound was added to prevent: `spanLeaves.length > Math.max(32, N * 0.25)` is false for a 32-op graph whose only span is the whole graph, so the entire model folds into one box. Above 32 the fraction takes over immediately, so the window is narrow — but it is not only tiny reports, because the op-range slider filters `operationList` before detection runs, so any user narrowing the range to a few dozen ops reaches it through a first-class control.

The obvious fix — a span covering the whole graph is never a layer, whatever the size — is not a drive-by. I tried it: it fails a dozen pinned classification tests, whose chains are single spans by construction (`[embedding, gelu, layer_norm]` and friends). So the contract needs deciding rather than patching: either the guard applies only above some graph size, or those specs move to two-span fixtures and the detector's contract becomes "a layer is a proper part of a graph".

Measured bounds from the four captures, for whoever picks a value: the largest span that must be admitted is 30 ops (`bge_m3`'s embedding, unfiltered — the 26 in the current comment is the hide-deallocate figure, and detection sees the raw sequence when that filter is off), and the smallest that must be rejected is 119 (`test_ttnn_moe`). The admissible floor band is therefore 31–118, and 32 sits at the very bottom of it with two ops of margin. The fraction is much better placed: largest admitted 6.6%, smallest rejected 41%, so anything in 0.07–0.40 behaves identically and 0.25 is near the middle.

Found while reviewing `feature/1976-op-graph-layer-semantics`.

Issue #1988, opened 2026-09-08
A weight fan's id still moves when a grouping fold merges two fans


aidemsined:
`fanIdOf` keys a fan on its first member's operation id, which fixed the original defect — the consumer's *rendered* id was moving under a grouping fold. The narrower version of that defect survives, because fan *membership* is fold-dependent.

Two facts combine. `isClaimed` reports only the grouping blocks that are currently folded, and #1977 opens them unrolled, so a source op inside an unrolled layer is unclaimed and free to join a fan. And "the same rendered node" is what groups a fan's members, so folding a layer merges fans that were previously separate: sources 2 and 3 feeding op 4 plus source 1 feeding op 5 are `weights:2` and a singleton while ops 4 and 5 render separately, and become one `weights:1 [1,2,3]` once both consumers fold into the same block node.

So a fan the user unrolled as `weights:2` re-folds itself on the next grouping fold, and `weights:2` stays behind in a set nothing prunes — the same user-visible failure the re-key was meant to end, reached by a different route. The module comment claimed members are "never inside a grouping block", which is false: every `bge_m3` attention span contains six `from_torch` sources. That comment has been corrected on the branch; the mechanism has not.

Not observed on the four captures under review, because no fan there straddles a span boundary in the way that triggers it — this is a mechanism, not a reported bug.

Options: key a fan on its sorted member list rather than its first member (stable under merges, but the id changes when membership legitimately changes, which is arguably correct); or prune ids that name no current instance on each rebuild, which would also clean up the general staleness of the expansion set.

Found while reviewing `feature/1976-op-graph-layer-semantics`.

PR #1992, opened 2026-09-08
Operations graph: group by layer semantics read from op roles

### LINKED ISSUE #1976: Operations graph: classify layer semantics from op roles, not repetition alone
## Context

#1583 groups operations by **repetition** and #1953 names them from **stack ancestry**. Neither asks what a group *does*. A third signal is available and, in most reports, is the strongest of the three: **ttnn op names already assert their own role**, so an attention block or a norm can be identified outright rather than inferred.

This is not a replacement for either. They answer different questions and compose:

- **Semantics** say *what* — "an attention block".
- **Ancestry** (#1953) says *which* — `encoder.layer.3`.
- **Repetition** (#1583) says *how many*.

Semantics also reach the unique regions — embeddings, output heads, stage transitions — that repetition structurally cannot, which is the gap #1953 opens with.

## Evidence: the vocabulary is already semantic

`bge_m3_aug19_1645`, 943 operations:

| Op | Count |
|---|---|
| `ttnn.transformer.scaled_dot_product_attention` | 24 |
| `ttnn.experimental.nlp_create_qkv_heads` | 24 |
| `ttnn.experimental.nlp_concat_heads` | 24 |
| `ttnn.gelu` | 24 |
| `ttnn.layer_norm` | 49 |
| `ttnn.linear` | 96 |

A contiguous window reads as the textbook block with nothing inferred:

```
340  nlp_create_qkv_heads          <- QKV split
348  expand                        <- mask broadcast
349  scaled_dot_product_attention  <- attention
353  nlp_concat_heads              <- head merge
354  linear                        <- output projection
356  add                           <- residual
359  layer_norm                    <- post-attention norm
```

### Counts corroborate multiplicity for free

`linear` 96 = 24 x 4 (QKV, attn-out, MLP-up, MLP-down); `layer_norm` 49 = 24 x 2 + 1. Independently, `sentence_bert_aug19_1726`: `linear` 48 = 12 x 4, `matmul` 24 = 12 x 2, `layer_norm` 25 = 12 x 2 + 1.

"Hypothesise N layers, check the histogram divides cleanly" is a cheap high-confidence check that needs no detector, and it is exactly the corroboration #1953's confidence tiers ask for.

### Topology delimits where naming is absent

`resnet50_aug06_1521` has no attention-style vocabulary — `quasar.conv2d` x 159 and `quasar.add_` x 48. The residual add is the delimiter. Partitioning convs by add position:

```
convs per block: {3: 36 blocks, 4: 9, 5: 3}
```

36 blocks of exactly 3 convs is the canonical bottleneck (1x1 -> 3x3 -> 1x1); the 4s and 5s are projection-shortcut variants at stage transitions. That partition falls out of op type plus topology alone.

## Constraints the reports impose

**Fusion changes the signature for identical semantics.** `bge_m3` spells attention as one fused `scaled_dot_product_attention`. `sentence_bert` spells the same concept as `experimental.split_query_key_value_and_split_heads` + `transformer.attention_softmax_` + two `matmul`. The anchor table needs both spellings, and that set grows every time the backend fuses something.

**Op names are arch-namespaced.** ResNet is `ttnn.experimental.quasar.*` throughout. Matching must be prefix-tolerant rather than exact-string.

**Known negative case: `test_ttnn_moe_aug26_2217`.** Its vocabulary is inert — `squeeze` x 53, `as_tensor` x 29, four `matmul`, no `topk`, no softmax, no expert ops. There is nothing semantic to anchor on. This report must be named in the tests as the case this detector does **not** cover, so nobody claims coverage it does not have; ancestry (#1953) is the only signal there.

## Approach

1. **Anchor table** mapping role -> op patterns, prefix-tolerant, with fused and unfused spellings of the same role as alternates of one entry.
2. **Grow a candidate group** around each anchor, bounded by the delimiters that already work: residual `add`, `layer_norm`, or the next anchor of the same role.
3. **Corroborate** with count arithmetic and, where available, #1583's topology match. Emit a confidence rather than an assertion.
4. **Reconcile** with #1583's blocks and #1953's ancestry groups so one region cannot carry competing identities — the same open question #1953 records.

## Why this is worth doing before #1953

#1953 is gated on two hard design decisions: which tree level counts as "the layer" (the same ResNet gives 25 groups at 15-op granularity in `jul28` and 61 at 6-op in `aug06`), and how a tier-1 group reconciles with a #1583 block.

An op-role classifier has neither problem — no deny-list for dispatch shims, no path truncation, no level selection, because **the anchor op is the level**. It is cheaper, higher-confidence, and it gives #1953 an independent signal to resolve its granularity question against.

## Acceptance criteria

- [ ] Anchor table covers attention (fused and unfused), MLP/feed-forward, normalisation, embedding and convolutional-residual roles.
- [ ] `bge_m3` yields 24 attention blocks and `sentence_bert` 12, from the anchors alone, with no repetition input.
- [ ] `resnet50_aug06` yields the residual-delimited partition, with the 3-conv bottleneck as the dominant shape.
- [ ] Matching is prefix-tolerant, verified against the `ttnn.experimental.quasar.*` names.
- [ ] ~~Count-arithmetic corroboration raises or lowers confidence and is asserted on at least two reports.~~ **Amended — see below.**
- [ ] `test_ttnn_moe_aug26_2217` is covered by a test asserting the detector reports **no** semantic groups rather than guessing.
- [ ] A region covered by both this and a #1583 block resolves to one identity.

## Amendments made while implementing

**Corroboration is a validation technique, not a runtime confidence signal.** The original criterion asked for count arithmetic to raise or lower a group's reported confidence. It is not implemented that way and deliberately so: dividing an op histogram by a group count needs a per-role *signature op* — `linear` for attention, which is four per block in `bge_m3` and two in a report that fuses its projections — and any such table is a guess that would move labels in production. What shipped instead is span-length uniformity, which needs no table and is the same evidence in a stronger form: 24 attention spans of exactly 25 ops, 12 feed-forward spans of exactly 5, 48 conv blocks whose 3-conv shape dominates. A partition landing on arbitrary boundaries produces a spread, so uniformity is what says the structure is real, and it is asserted on three reports rather than two. Count arithmetic did do its job — it is how the anchor tables were checked against the captures — and it lives in the spec as `countsCorroborate`, off the production path. Raised by Copilot and @dcblundell on #1992.

**Two roles beyond the original list are intentional.** The criteria name attention, feed-forward, normalisation, embedding and convolutional-residual. The detector also reports **expert routing** (MOE) and **positional encoding**, both added after checking the hand-written table against ttnn's own taxonomy: `ttnn::operations::reduction::moe` and the `ttnn.experimental.deepseek_prefill.*` routing ops are first-class, and `rotary_embedding` ships as a family. Leaving them out would have meant an MoE report finding nothing where its own op names say what it is. Normalisation, by contrast, is not reported as a role — it is what the partition *cuts on*, which is the more useful thing to do with it.

**Repeats and layers stay mutually exclusive per build**, which is how the last criterion is met: a region cannot carry two identities because only one detector runs. Reconciling them into a single combined view remains #1953.

## Related

- #1583 - repeated-subgraph collapse (shipped 0.101.0)
- #1953 - layer inference from stack ancestry
- #1192 - identifying groups of operations

### LINKED ISSUE #1977: Operations graph: repeats should open unrolled, not folded by default
## Problem

#1583 shipped repeat detection that **folds every detected instance on first layout**. Detection describes the graph; it should not decide how the graph opens. A report currently arrives already restructured, so the first thing a user sees is not the graph they captured, and understanding it starts with undoing a decision nobody asked for.

Folding stays available — it is just a choice now rather than a default.

## Change

`expandedBlockIds` becomes tri-state in `OpGraphBuildOptions`:

| Value | Meaning |
|---|---|
| absent | nothing has been folded yet, so the graph renders unrolled |
| `[]` | a deliberate fold-all |
| `[...ids]` | those instances unrolled, the rest folded |

The default cannot be expressed as an id list, because the instance ids only exist once detection has run inside the build — which is why the original design folded instead. Absence is the only way to say "no decision yet" without a second layout pass on every mount.

Detections are still reported when nothing is folded, so the toolbar's Repeats row and the Fold button remain available.

## Notes

- Toggling `Hide deallocate ops` re-runs detection, so it drops the fold decision and returns to unrolled, matching the report-change reset.
- Folding one instance from the unrolled default has to materialise its siblings as expanded first: `new Set(null)` is empty, and deleting from that folds the whole graph.

## Follow-up left open

Whether the default should eventually be a user preference, or a heuristic (for example, fold automatically above some node count where the unrolled graph is unreadable), is deliberately not decided here.

## Acceptance criteria

- [ ] A report with detected repeats opens unrolled.
- [ ] The Repeats row and Fold are offered on that unrolled first layout.
- [ ] Fold-all, unroll-all, per-instance fold and unroll all still work.
- [ ] Folding one instance leaves its siblings unrolled.
- [ ] Mount still costs one layout, with no second pass to apply the default.

### LINKED ISSUE #1980: Operations graph: collapse weight-load fans into one node
## Problem

Weight loading dominates the node count and none of it is model structure. In
`sentence_bert_aug19_1726`, `ttnn.to_device` is **151 of 406 operations** — 37% of the
graph is getting parameters onto the device. `bge_m3_aug19_1645` is the same shape with a
different op: `ttnn.from_torch` **× 293**.

Rendered, they arrive as fans of roughly six nodes pointing at each detected layer, which
is most of what is on screen and none of what anyone came to look at.

## Rule

Collapse a **weight fan**: a set of *source* operations — no inputs from within the graph —
whose **every** consumer resolves to the same rendered node. Draw one node for the set and
a single unlabelled edge to that consumer.

Defined topologically rather than by op name. `sentence_bert` uses `to_device`,
`bge_m3` uses `from_torch`, and a name list would have missed one of the two reports we
already have. ttnn adds spellings faster than we notice.

Not restricted to detected layers. The rule holds for any single rendered consumer, so it
keeps working in Repeats mode and on reports where nothing is detected — and a six-fan
into a plain `ttnn.linear` is just as noisy as one into a layer.

### The safety condition is mandatory

If any member's consumers differ, the set is **not** a fan and must not collapse.
Otherwise the graph claims a weight belongs to one layer while it feeds two.

Measured: all **149 of 149** `to_device` operations with edges in `sentence_bert` have
exactly one consumer, so the condition holds everywhere here. It still has to be enforced
rather than assumed — the first model that shares a parameter tensor would otherwise get a
graph that lies. A test covers the shared-consumer case explicitly.

### No contiguity requirement

Verified that per-block fans happen to be contiguous in `sentence_bert` (the feed-forward
block at ops 183-186 draws loads 12-17, the attention block at 187-200 draws 18-23), but
the rule does not depend on it: the collapsed node is emitted at its first member and the
rest are skipped, exactly as a folded block already is.

## Approach

Fans join `collapsedInstanceByOpId` in the builder as pseudo-instances. That is the whole
integration: `renderedNodeIdOf` then maps members to the fan, and the existing edge path
already suppresses labels and dedupes parallel edges across a collapsed boundary, so the
single unlabelled edge falls out with no new code. Node emission sums member duration and
memory the same way, which gives the aggregate for free.

They are deliberately **not** added to `graph.blocks`, so the grouping count badge keeps
meaning "layers/repeats detected" and Fold/Unroll-all keep acting on grouping alone.

Because "same rendered node" depends on the current fold state, this resolves at build
time rather than in a pre-pass detector like #1583 and #1976.

## Toggle

A `Collapse weight loads` switch, **on by default**, independent of the grouping mode.

On-by-default matches `Hide deallocate ops`, which is already on by default in the same
toolbar: both hide plumbing rather than restructure the model. This is not in tension with
#1977 — that was about semantic grouping deciding how a report opens, where the objection
was that detection should not restructure the model before anyone has looked at it. Hiding
plumbing does not restructure the model; it removes what is not the model.

## Acceptance criteria

- [ ] A set of source ops whose every consumer is one rendered node collapses to one node.
- [ ] A single **unlabelled** edge joins the collapsed node to its consumer, whatever the
      member count.
- [ ] A source op whose consumers differ is left alone — asserted by a test that would
      otherwise misattribute a shared weight.
- [ ] The collapsed node carries the member count and summed bytes; parameter size is the
      one useful thing those nodes held.
- [ ] Works with grouping set to Repeats, to Layers, and with nothing detected.
- [ ] Off restores every member node and its labelled edge.
- [ ] Matching is topological; asserted against both `to_device` and `from_torch` without
      naming either.
- [ ] The grouping count badge and Fold/Unroll-all are unaffected.

## Related

- #1976 — layer grouping, whose spans these fans point at
- #1583 — repeated-subgraph collapse, whose node and edge machinery this reuses
- #1977 — repeats open unrolled, for the contrast on defaults

### LINKED ISSUE #1982: Operations graph: distinguish repeat, layer and weight blocks by colour
## Problem

Repeats (#1583), layers (#1976) and weight fans (#1980) all render as the same node
type, so the only way to tell which detector found a block is to read its label. With
weight fans on by default and layer grouping a click away, three kinds can be on screen
at once and they are indistinguishable at graph zoom, where the labels are unreadable.

## Approach

The kind is carried on the block instance rather than parsed back out of `patternId`.
A prefix check works today and silently mislabels the next detector someone adds; it is
also the discriminator a panel or a test should reason about instead of a string.

One `--graph-block-*` token pair per kind, per the repo's colour rule that each semantic
role gets its own token even where two share a value.

| Kind | Colour | Why |
|---|---|---|
| Repeats | blue (unchanged) | nothing moves for a graph that already shipped |
| Layers | ~~purple~~ **teal** | purple sits next to blue on the wheel and at these tints the two were near-indistinguishable on the dark canvas; cyan separates from indigo at a glance |
| Weight fans | ~~grey~~ **sand** | a fan is plumbing and should recede — but grey read as *disabled*; a warm hue separates from both cool ones by temperature, which survives a low zoom |

Green and yellow are deliberately unused: `--graph-input-node`, `--graph-output-node` and
`--graph-focused-node` already own them, and a block in either would read as an I/O
highlight.

### Two things worth knowing for anyone touching this later

**The class is set on the built node, not in the restyle memo.** It is static, and that
memo runs on every selection and filter keystroke — putting it there would defeat the
byte-for-byte node identity `OperationGraphRebuildDiscipline` asserts across a drag frame.

**That memo replaces `className` outright**, so it now seeds from whatever the node
already carries. Seeded empty, the colour disappeared the moment a block was selected —
visible only while selected, which is the hardest kind of styling bug to notice. Covered
by a test.

Selection is ordered last in the stylesheet so it outranks the kind: which block is
selected matters more than which detector found it.

## Acceptance criteria

- [ ] A repeat, a layer and a weight fan carry different classes and different colours.
- [ ] The kind is on the node data as well as the class, so it can be reasoned about
      without parsing a string.
- [ ] The kind colour survives selection, filtering and the perf overlay.
- [ ] Selection styling outranks the kind colour.
- [ ] No hex values outside `_base.scss`.
- [ ] The expander pill on a block tracks its kind.

## Related

- #1583 repeats, #1976 layers, #1980 weight fans — the three kinds being distinguished

---

**Updated 2026-09-04:** purple and grey were tried and rejected once rendered. Teal and sand shipped; the table above records both.

aidemsined:
image

image

## Summary

The operations graph could only group by **repetition** (#1583), which says how many but never what — and structurally cannot see a region that occurs once, so embeddings and output heads were invisible to it. ttnn op names already assert their own role, so this adds a second grouping mode that reads them: a span of operations is identified as attention, feed-forward, expert routing, positional encoding, an embedding or a residual conv block, and folds under the same machinery repeats already use.

Three things ship alongside it because the feature is unusable without them. Grouping no longer folds itself on first render — detection describes the graph, it does not decide how to show it. Weight-load fans collapse into one node, because 151 of `sentence_bert`'s 406 operations are `ttnn.to_device` and 293 of `bge_m3`'s are `ttnn.from_torch`: rendered, they were most of what was on screen and none of what anyone came to look at. And the three kinds of block are now told apart by colour rather than by reading their labels.

Closes #1976
Closes #1977
Closes #1980
Closes #1982

Refs #1583, #1953

### How a layer is identified

Partition the execution order on delimiters, then label each span by the anchors it holds — rather than growing outward from an anchor, which needs a stop rule per role and produces overlapping spans wherever two anchors sit close together.

- **Delimiters** are normalisation ops where the capture has any, residual adds otherwise. Inclusive of the delimiter, so a trailing norm belongs to the sub-block it terminates. `resnet50` has no norm at all, and there the residual add *is* the boundary: 159 convs partitioned on 48 adds gives 36 spans of exactly three, the canonical bottleneck.
- **Anchors** come in two strengths. A naming anchor means the op says what the span is (`scaled_dot_product_attention`, `nlp_create_qkv_heads`, `moe_grouped_topk`, `embedding`); a supporting anchor merely suggests it (an activation, a convolution) and is reported at lower confidence. Direct evidence outranks circumstantial, so an `embedding` is never renamed by a `gelu` sharing its span; between two supporting anchors the convolution wins, since it names a shape where the activation only says one happened. A span with no anchor at all is not a layer — it is the weight-loading run between two of them, and is dropped rather than guessed at.
- Anchors match on the **leaf** of the op name, so `ttnn.experimental.quasar.conv2d` and `ttnn.transformer.scaled_dot_product_attention` both match. Families are matched on a leading segment where metal ships a set of variants (`nlp_create_qkv_heads_{decode,vit,falcon7b,segformer}`), since an exact table silently misses whole report classes.
- A span past a size bound is reported as no grouping rather than as one useless block. A span that large does not mean the layer is big — it means the partition failed, and folding it replaces the model with a single box.

Two spellings of one role sit in the same tables, because the same semantic layer arrives differently per report: `bge_m3` emits a fused `scaled_dot_product_attention`, `sentence_bert` spells the identical concept as a split + softmax + two matmuls. `sentence_bert` also fuses its feed-forward GELU into the matmul's `program_config`, so that activation is read out of the argument and classified as though it were an op leaf — otherwise its 12 feed-forward blocks carry no anchor and are dropped, while `bge_m3`'s 24 explicit `ttnn.gelu` ops detect fine. Fusion is the direction of travel, so the explicit spelling is the lucky case rather than the representative one.

Measured on four real captures: `bge_m3` 49 groups (24 attention + 24 feed-forward + the embedding no repeat scan can reach), `sentence_bert` 24, `resnet50` 48. Every span of a given role comes out the same length, which is the corroboration that the partition found real structure rather than landing on arbitrary boundaries.

### Frontend

- **`opGraphOpRoles.ts`** (new) — the detector. Anchor and delimiter tables checked against ttnn's own taxonomy, plus the partition. Softmax is *subtracted* from the delimiter set rather than merely documented as excluded: ttnn files it under `operations/normalization/`, so the next person widening the list from that directory would otherwise pull it in and halve every attention block.
- **`opGraphLayerBlocks.ts`** (new) — adapts a role group to `RepeatBlockInstance`, so layer grouping reuses the block nodes, expanders, edge remapping, buried filter matches and block panel #1583 already built instead of adding a second block type to render the same shape.
- **`opGraphWeightFans.ts`** (new) — collapses a fan of weight-loading sources into one node. Matched topologically, not by op name: `sentence_bert` and `bge_m3` already disagree about the name, so a list would have missed one of them. A source is absorbed only when *every* one of its consumers resolves to the same rendered node — a weight shared between two layers belongs to neither, and collapsing it into one would have the graph assert it is that layer's.
- **`opGraphFusedActivation.ts`** (new) — reads an activation out of a matmul `program_config`. Parsed at the mapping boundary and carried as the bare name, since the full argument payload is hundreds of KiB per report and crosses to the worker.
- **`opGraphTypes.ts`** — `OpGraphBlockKind` and `OpGraphGrouping`. Kind is carried on the instance rather than parsed back out of `patternId`, because a prefix check would silently mislabel the next detector added. `expandedBlockIds` becomes a tri-state: absent means nobody has folded anything yet, an empty array is a deliberate fold-all.
- **`opGraphBuilder.ts`** — the weight-fan pass, and the kind class on the built node (static, so it is not recomputed by the restyle memo on every selection and filter keystroke). The candidate-edge pass is taken as an option when a caller already holds it, so the walk the worker runs for detection is not repeated per layout.
- **`opGraphBlockDetectors.ts`** — which detector each mode runs, as one record over the enum. The builder and the worker both need it, and a ternary in each would route a future third mode to repeats silently.
- **`opGraphLayoutWorker.ts`** — detection cached per grouping as well as per filter, and the build options forwarded by spread with a cache key keyed off `keyof OpGraphBuildOptions`, so a new option cannot be silently dropped or left out of the key.
- **`OperationGraphReactFlow.tsx`** — the fold state. A grouping block is unrolled unless folded (#1977); a weight fan is folded unless unrolled, since its switch is already on. Unroll-all and Fold-all each act on the detector's own blocks and leave the other kind's ids alone, because both kinds share one expansion set.
- **`OpGraphToolbar.tsx`** — the Grouping row, with the active mode carrying its block count so the two strategies are comparable by switching rather than by folding each and counting nodes. Always rendered, including when a detector finds nothing, which says which one came up empty instead of hiding the control.
- **SCSS** — one colour token pair per detector (blue / teal / sand). Green and yellow are deliberately unused: the I/O and focus highlights own them, and a block in either would read as one of those instead. The toolbar controls take the same tokens through classes we own, which have to outrank Blueprint's *dark-theme* button rules — those carry a class more than the light-theme ones, and the app is dark by default.

- Repeats and layers are **mutually exclusive per build**, so a region cannot carry two competing identities. Reconciling them is the open question in #1953.
- In Repeats mode nothing collapses into a weight fan, because repeat detection already claims the loader pairs.
- A weight fan is not yet a first-class block: selecting one shows its first member's detail rather than the fan's, and once unrolled it has no fold affordance. #1987
- A fan's id can move when a grouping fold merges two fans. #1988
- On a graph of 32 operations or fewer, one span can still cover the whole graph — reachable by narrowing the op-range slider. #1989
- A report whose op names use `::` rather than `.` finds no layers. #1990
- A CNN that emits its norms as ops partitions on those norms rather than on residual adds, because the delimiter strategy is chosen from any norm in the graph. #1993

## Test plan

- [x] Frontend: `pnpm vitest run` — 224 files, 2627 tests
- [x] `tsc --noEmit`, `eslint --max-warnings 0`, `stylelint` clean
- [x] No backend changes
- [x] Detector pinned against four real captures as fixtures (`bge_m3`, `sentence_bert`, `resnet50`, `test_ttnn_moe`) — group counts, span uniformity, per-role confidence and the anchor tables row by row
- [x] Manual, against `sentence_bert_aug19_1726` in a dev build: Layers reports **24**, matching the fixture measurement exactly, and folds the named spans; 28 weight fans collapse alongside them; the palette was checked by reading the computed style rather than by eye — layer blocks and the active Layers control are both `rgb(155, 219, 240)` on `rgb(62, 183, 222)`, fans and the switch indicator both sand, and selection still outranks the kind colour on a selected block

## Follow-ups

- #1987 — weight fans are second-class in the block model
- #1988 — a fan's id moves when a grouping fold merges two fans
- #1989 — a layer span can still swallow a graph of 32 ops or fewer
- #1990 — `::`-spelled op names find no layers
- #1991 — three loose ends in the fold state
- #1993 — a CNN with explicit norms partitions on norms instead of residual adds


tests/OpGraphToolbarSwitches.spec.tsx
```diff
@@ -215,6 +231,89 @@ describe('switch tooltips', () => {
...
+        renderToolbar({ status: PerfOverlayStatus.READY, hasBlocks: false, grouping: OpGraphGrouping.LAYERS });
+
+        expect(screen.getByText('no layers detected')).toBeInTheDocument();
+        expect(screen.queryByText('no repeats detected')).toBeNull();
+    });
+});
+
+describe('weight-load control', () => {
+    it('carries the swatch class that ties it to the fan colour', () => {
+        renderToolbar({ status: PerfOverlayStatus.READY });
```
dcblundell, line 289: **Should fix:** Test the weight-load switch behaviour

This covers the swatch class and ordering, but not that toggling invokes `onCollapseWeightLoadsChange` with the new value. Add enabled/disabled interaction coverage, and ideally assert that the value reaches `runBuild`.

src/components/operation-graph/opGraphBuilder.ts
```diff
@@ -102,11 +123,18 @@ export function buildOpGraph(
     const operationById = new Map<number, OpGraphSourceOperation>(
         keptOperations.map((operation) => [operation.id, operation]),
     );
-    const detectedBlocks = providedBlocks ?? detectRepeatBlocks(keptOperations);
-    const expandedBlocks = new Set<string>(expandedBlockIds);
+    const detectedBlocks =
```
dcblundell: **Should fix:** Avoid collecting candidate edges twice

`collectCandidateEdges(operations)` already runs before detection, and the detector path consumes the same candidate set. The build then traverses the full edge set again, adding O(edges) work to every uncached layout, including slider-driven builds. Reuse the collected candidates or pass them through the detection/build boundary.

tests/OperationGraphRebuildDiscipline.spec.tsx
```diff
@@ -553,7 +554,9 @@ describe('OperationGraphReactFlow rebuild triggers', () => {
         expect(runBuild).toHaveBeenLastCalledWith({
             hideDeallocate: false,
             deviceSubgraphs: [],
-            expandedBlockIds: [],
+            expandedBlockIds: undefined,
```
dcblundell, line 557: **Should fix:** Assert stale graph cleanup on report swaps

The critical-path report-swap tests around `OperationGraphRebuildDiscipline.spec.tsx:~999–1010` only verify that the checkbox turns off. Also assert that stale node/edge classes, annotations, and the summary are cleared after both profiler and performance report changes.

src/components/operation-graph/opGraphOpRoles.ts
dcblundell, line 1: **Consider / nit:** Follow-up and specification notes

- Selecting a collapsed weight fan still shows the first member’s detail/perf data rather than an aggregate; this is documented as #1987.
- Detector dispatch is duplicated between the worker and builder, so a shared helper could prevent drift.
- Count-arithmetic corroboration is test-only and remains deferred per the amended #1976 requirement.
- The detector also adds MOE and positional-encoding roles beyond the original criteria; amend the specification if those roles are intentional.

tests/OpGraphToolbarSwitches.spec.tsx
dcblundell, line 289: **Should fix:** Cover the all-collapsed disabled branch

The direction test covers `areAllBlocksExpanded`, but not the independent `areAllBlocksCollapsed` branch. Add a case asserting Fold is disabled and Unroll remains enabled when all blocks are collapsed.

src/scss/components/OpGraphToolbar.scss
```diff
@@ -51,3 +51,46 @@
...
+    font-size: 12px;
+    font-style: italic;
+    white-space: nowrap;
+}
+
+// The grouping controls wear the colour of the blocks they produce, so the toolbar and
+// the canvas read as one feature rather than two. Borders always, fill only when
+// active: an inactive button that is already filled cannot show that it is selected.
+// #1982
+.op-graph-grouping-repeat.bp6-button {
```
dcblundell: **Should fix:** Keep Blueprint overrides out of component-local SCSS

These new selectors directly target `.bp6-button`, `.bp6-active`, `.bp6-control`, and `.bp6-control-indicator`, contrary to the repository convention for component styles. Add app-owned state/element classes, or move the necessary global Blueprint override to `_blueprintjs.scss`.

dcblundell (commented):
Posted the five Should fix comments and one consolidated Consider / nit comment from the combined review.

src/components/operation-graph/OperationGraphReactFlow.tsx
```diff
@@ -716,25 +798,51 @@ const OperationGraphInner = ({
...
+            // and the difference only appeared after a separate Fold. Dropping the
+            // outgoing mode's ids folds whatever this mode detects and stops the set
+            // naming blocks that no longer exist; weight-fan ids are not mode-specific,
+            // so they stay. #1976
+            setExpandedBlockIds((previous) => withoutGroupingBlocks(previous, detectedBlockIds));
+            setExpandedOperationIds((previous) => withoutBlockMembers(previous, detectedBlocks));
+            setRevealedNodeIds(null);
+        },
+        [detectedBlocks, detectedBlockIds],
+    );
```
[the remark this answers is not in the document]
aidemsined: Good catch, and it reproduces — the test I added fails without the fix: from the unrolled default `withoutGroupingBlocks(null, …)` returns an empty set, so the active button folded the graph.

Fixed by returning early when `next === grouping`, in `handleGroupingChange` rather than on the button. One guard at the state owner covers both anchors of this finding, and it keeps the button ordinary — a disabled or inert active button in a `ButtonGroup` reads as "unavailable" rather than "already selected", and `aria-pressed` is what conveys the state.

`0213f22d`, with `does nothing when the grouping already in use is clicked`.

src/components/operation-graph/opGraphOpRoles.ts
```diff
@@ -0,0 +1,349 @@
...
+ * `embedding, gelu, layer_norm` read as feed-forward and threw the embedding away.
+ */
+const ROLE_PRIORITY: readonly OpSemanticRole[] = [
+    OpSemanticRole.ATTENTION,
+    OpSemanticRole.MOE,
+    OpSemanticRole.FEED_FORWARD,
+    OpSemanticRole.POSITIONAL_ENCODING,
+    OpSemanticRole.EMBEDDING,
+    OpSemanticRole.CONV_RESIDUAL,
+];
```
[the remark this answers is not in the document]
aidemsined: Confirmed by reproduction rather than reading: `conv2d → relu → add` came back as `feedForward/medium/relu`. `resnet50` fuses its activations (159 `conv2d`, 48 `add_`, no activation op), so no fixture covered it.

Taken the first of your two options — convolution now outranks activation in the role order. A convolution names a shape; an activation only says one happened, and every conv block contains one, so the specific evidence has to win. The delimiter tie-break would work too, but it would only fire on the residual-add strategy, and this needs to hold in a norm-delimited graph as well.

Fixtures are unchanged, and there is a test for the case none of them reach. `0213f22d`.

tests/opGraphOpRoles.spec.ts
```diff
@@ -0,0 +1,519 @@
...
+    type OpRoleGroup,
+    type OpRoleSourceOperation,
+    OpSemanticRole,
+    detectOpRoleGroups,
+} from '../src/components/operation-graph/opGraphOpRoles';
+import { DEALLOCATE_OP_NAME_LIST } from '../src/definitions/Deallocate';
+import bgeM3 from './fixtures/opRoles/bge_m3.json';
+import moe from './fixtures/opRoles/moe.json';
+import resnet50 from './fixtures/opRoles/resnet50.json';
+import sentenceBert from './fixtures/opRoles/sentence_bert.json';
```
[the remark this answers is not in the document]
aidemsined: Right, and my own architecture pass flagged the same thing — I skipped it and should not have. `CONVENTIONS.md:990` and `AGENTS.md:126` both put JSON fixtures in `tests/data/`.

Moved to `tests/data/opRoles/` with both importing specs updated. `0213f22d`.

src/components/operation-graph/opGraphOpRoles.ts
```diff
@@ -0,0 +1,349 @@
...
+/**
+ * Normalisation when the capture has any, residual adds otherwise. Chosen from the
+ * whole graph rather than per span: a graph that mixes both would otherwise partition
+ * inconsistently down its length.
+ */
+const delimitersFor = (leaves: readonly string[]): ReadonlySet<string> => {
+    if (leaves.some((leaf) => NORMALISATION_DELIMITERS.has(leaf))) {
+        return NORMALISATION_DELIMITERS;
+    }
+    return RESIDUAL_DELIMITERS;
```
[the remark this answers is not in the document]
aidemsined: Reproduced: `conv2d, batch_norm, relu, conv2d, batch_norm, add` partitions into `[1,2]` and `[3,4,5]`, cut at each norm, where the intended shape is one residual block. So the residual-conv grouping does only work where norms are fused or absent, which is why `resnet50_aug06` works — it has no norm op at all.

Not fixing it here, and the reason is that I cannot validate a fix: every fixture in the repo is either a transformer with `layer_norm` or a CNN with no norms, so any change to the strategy would ship unexercised. Filed as #1993 with the reproduction and the two candidate shapes (choose the set from the anchors present, or partition on residual adds *within* a norm-delimited span whose anchor is a convolution), blocked on getting a `batch_norm`-emitting capture in as a fixture.

Worth noting the role-priority fix in this PR improves the symptom: those fragments now name themselves after the convolution instead of the activation, so a batch-norm CNN mislabels nothing even while it over-partitions.

src/components/operation-graph/opGraphOpRoles.ts
```diff
@@ -0,0 +1,349 @@
...
+    for (const leaf of anchorLeaves) {
+        const naming = NAMING_ANCHORS.get(leaf) ?? familyRoleOf(leaf);
+        const role = naming ?? SUPPORTING_ANCHORS.get(leaf);
+        // First anchor per role wins, so the span is named by the op a reader meets
+        // first in execution order rather than by whichever table was consulted last.
+        if (role !== undefined && !found.has(role)) {
+            found.set(role, {
+                role,
+                confidence: naming === undefined ? OpRoleConfidence.MEDIUM : OpRoleConfidence.HIGH,
+                anchorName: leaf,
```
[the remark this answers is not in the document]
aidemsined: Fair, and it is the criterion I am least comfortable with. It is not implemented as confidence input, so I have amended the criterion on #1976 rather than leave the claim standing — with the reasoning, not just a strikethrough.

Short version: a runtime divisibility check needs a per-role *signature op*, and that table is a guess. `linear` is four per attention block in `bge_m3` and two in a report that fuses its projections, so the same arithmetic that corroborates one report contradicts another. What shipped instead is span-length uniformity — 24 attention spans of exactly 25 ops, 12 feed-forward of exactly 5, and `resnet50`'s 3-conv shape dominant — which needs no table, is the same evidence in a stronger form, and is asserted on three reports rather than the two asked for. A malformed one-off span shows up as a length outlier, which is the case you are worried about.

Count arithmetic did its job as a *validation* technique: it is how the anchor tables were checked against the captures. It stays in the spec as `countsCorroborate`, off the production path.

src/components/operation-graph/opGraphBuilder.ts
aidemsined: You are right, and `getKeptOperations` already had the parameter for exactly this reason — `buildOpGraph` was the one caller still re-collecting.

Threaded through as a worker-only option, so the pass the worker already runs once per source for detection is the one the build uses. Two things worth noting: the exhaustive cache key added last round immediately made me justify it (adding the option was a compile error until I said whether it belongs in the key — it does not, it is a function of the operations, which the source version already identifies), and the test proves the parameter is *honoured* rather than merely accepted, by supplying a set that omits a real edge and asserting the ops it would have connected are absent.

`0213f22d`.

tests/OpGraphToolbarSwitches.spec.tsx
aidemsined: Correct — they are independent props, not one tri-state, so the all-expanded test says nothing about this branch. Added `disables the other direction when everything is already folded`, asserting Fold disabled and Unroll enabled. `0213f22d`.

aidemsined: Added three: the switch reports `false` from on and `true` from off, and one asserting it is *not* perf-gated like the two switches beside it in the row — fans exist whether or not a performance report is loaded, and that is the property most likely to be broken by someone tidying that row.

I did not add the "value reaches `runBuild`" assertion, because it already exists at the view level: `OperationGraphRebuildDiscipline` covers `collapseWeightLoads` reaching the build options and the layout worker spec covers it surviving the message hop. Duplicating it in the toolbar spec would test the wiring twice and the toolbar not at all — happy to add it if you would rather have it local.

`0213f22d`.

tests/OperationGraphRebuildDiscipline.spec.tsx
aidemsined: Added to both report-kind tests: no node carries `op-graph-node-critical-path`, no edge carries `op-graph-edge-critical-path`, and the `Critical path` annotation is gone. They pass.

One caveat you should have, because I checked rather than assuming: I could not make them load-bearing. I mutated the scope guard to `criticalPathScope !== null` — dropping the report comparison, which is the regression these assertions are meant to catch — and both tests still passed, so after a swap the drawing ends up empty by some route other than the guard. The `criticalPathSwitch().checked` assertion that was already there is what actually catches it. I have not chased down which of the path's inputs goes empty; that is #1613 territory and it is not what this PR changed.

So: the assertions are documentation-grade rather than regression-grade. Kept them because they state the intent and cost nothing, but I would rather you knew that than read three green assertions as coverage. Happy to reshape the harness so a stale path has data to draw, if you want the real thing — that is its own piece of work.

`0213f22d`.

src/scss/components/OpGraphToolbar.scss
aidemsined: Taken, and following it up found a bug — the buttons were not getting their colours at all.

Chasing the convention led me to check the computed style in a running app instead of reasoning about the cascade, and the active control was `rgba(143, 153, 168, 0.3)` with a `rgba(255, 255, 255, 0.4)` border: Blueprint's own outlined-active grey, not the block palette. The reason is that I had counted the light-theme selectors. The rules that actually apply are the dark ones, and they carry an extra class:

- `.bp6-dark .bp6-button.bp6-outlined` — 3 classes, sets the border
- `.bp6-dark .bp6-button.bp6-outlined.bp6-active` — 4 classes, sets the fill

Both my `.op-graph-grouping-repeat.bp6-button` (2) and `…​.bp6-active` (3) lost to those, so this shipped with Blueprint's grey on a dark default theme and nothing on the canvas matched the control that produced it. Your comment is the reason I looked.

Now: the palette is keyed off classes we own — the component emits `op-graph-grouping-active` alongside Blueprint's state class, and the toolbar row scopes the swatch, giving 3 classes for the border and 4 for the fill, which match the theme rules and win on order. Verified in the browser rather than on paper: the active control computes to `rgb(155, 219, 240)` on `rgb(62, 183, 222)`, which is exactly what its layer blocks are drawn in.

One `.bp6-*` selector stays, and I would like to keep it: the switch indicator. It is Blueprint's own element with no hook of ours, which is the documented exception at CONVENTIONS.md:257 — and I kept it local rather than moving it to `_blueprintjs.scss`, because a fan-coloured indicator there would reach every switch in the app, which is the cross-component bleed the rule exists to prevent. Local precedent for reaching `.bp6-control` in a component sheet: `CircularBufferPressureModal.scss:64` and `BufferSummaryControls.scss:43`, both for this same Blueprint switch margin. Say the word if you would rather it went global anyway.

`0213f22d`.

src/components/operation-graph/opGraphOpRoles.ts
aidemsined: Taking these in order.

**Fan panel** — yes, #1987, and it is the one I most want to do next. Filed with the four behaviours that fall out of fans being kept out of `graph.blocks`.

**Detector dispatch** — done. One `BLOCK_DETECTOR` record over `OpGraphGrouping`, consumed by both the builder and the worker. The reason to prefer the record over a shared function is exhaustiveness: the two ternaries would each have routed a third mode to repeats silently, and #1953 anticipates exactly such a mode, so this is now a compile error at both sites instead.

**Count corroboration** — amended on #1976 rather than left as an unmet criterion, with the reasoning: a runtime divisibility check needs a per-role signature op, and `linear` is four per attention block in `bge_m3` but two where the projections are fused, so the arithmetic that corroborates one report contradicts the next. Span-length uniformity is what shipped and is asserted on three reports.

**MOE and positional encoding** — intentional, and now recorded on #1976. Both came from checking the hand-written table against ttnn: `ttnn::operations::reduction::moe` and the `deepseek_prefill.*` routing ops are first-class, and `rotary_embedding` ships as a family. Omitting them would mean an MoE report finding nothing when its own op names say what it is. Worth noting the inverse too — normalisation is in the criteria as a role but is deliberately *not* reported as one; it is what the partition cuts on, which is the more useful thing to do with it.

src/components/operation-graph/OpGraphToolbar.tsx
```diff
@@ -232,27 +292,63 @@ const OpGraphToolbar = memo(
...
+                            key={value}
+                            content={description}
+                            position={PopoverPosition.BOTTOM}
+                        >
+                            <Button
+                                className={swatchClass}
+                                variant={ButtonVariant.OUTLINED}
+                                active={grouping === value}
+                                disabled={isDisabled}
+                                onClick={() => onGroupingChange(value)}
```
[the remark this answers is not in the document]
aidemsined: Fixed at the handler rather than on the button — see the reply on `OperationGraphReactFlow.tsx` for the reasoning and the test. Keeping the active button clickable-but-inert: in a `ButtonGroup`, a disabled active button reads as "unavailable" rather than "already selected", and `aria-pressed` already carries the state. `0213f22d`.

tests/OperationGraphRebuildDiscipline.spec.tsx
dcblundell: Thanks for checking this rather than overstating the coverage. I agree the assertions are useful as intent documentation, but the mutation result means they do not independently prove stale graph cleanup. I’m happy to leave this as a follow-up rather than block the PR, provided the test comments make that limitation clear.

PR #1994, opened 2026-09-09
Upgrade tt-perf-report to 1.3.0 and include new subdevice columns in the performance table

### LINKED ISSUE #1940: Surface subdevice fields from tt-perf-report in the performance report

dcblundell:
**New Sub Device and Available Cores columns**

Screenshot 2026-09-09 at 11 16 45

Closes https://github.com/tenstorrent/ttnn-visualizer/issues/1940.

This pull request introduces support for new performance report columns, specifically `available_cores`, `sub_device_id`, and `dram_sharded`, and ensures these fields are handled throughout the backend and frontend. It updates the CSV parsing logic to be more robust to column order, synchronizes with the latest `tt-perf-report` version, and enhances the display and coloring logic for these new fields in the UI. Comprehensive tests are also added to verify these changes.

**Backend: CSV Parsing and Data Model Updates**
- Refactored `OpsPerformanceReportQueries` to use a `REPORT_COLUMN_HEADERS` mapping, allowing for flexible matching of CSV columns by name rather than relying on their order. Added logic to resolve column indexes and parse rows accordingly. (`csv_queries.py`) [[1]](diffhunk://#diff-aaa21ce9241cb2231a1ccef9c82a1c3a13a090dfe1fa9768290673c11ef9891eL541-R568) [[2]](diffhunk://#diff-aaa21ce9241cb2231a1ccef9c82a1c3a13a090dfe1fa9768290673c11ef9891eR628-R654) [[3]](diffhunk://#diff-aaa21ce9241cb2231a1ccef9c82a1c3a13a090dfe1fa9768290673c11ef9891eR837-L820)
- Added support for new columns: `available_cores`, `sub_device_id`, and `dram_sharded` in both the backend data model and contract tests, ensuring these fields are parsed and included in reports. (`csv_queries.py`, `test_perf_report_contract.py`) [[1]](diffhunk://#diff-aaa21ce9241cb2231a1ccef9c82a1c3a13a090dfe1fa9768290673c11ef9891eL541-R568) [[2]](diffhunk://#diff-9b8a18573555b088ba855bae1deebb6e1df47795866cd144b970424d33680ca3R78-R79) [[3]](diffhunk://#diff-9b8a18573555b088ba855bae1deebb6e1df47795866cd144b970424d33680ca3R132-R146) [[4]](diffhunk://#diff-9b8a18573555b088ba855bae1deebb6e1df47795866cd144b970424d33680ca3R342-R343)

**Backend: Dependency and Version Synchronization**
- Updated the `tt-perf-report` dependency from version `1.2.8` to `1.3.0` in `pyproject.toml` and synchronized the expected version in the contract tests. (`pyproject.toml`, `test_perf_report_contract.py`) [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L24-R24) [[2]](diffhunk://#diff-9b8a18573555b088ba855bae1deebb6e1df47795866cd144b970424d33680ca3L45-R45)

**Frontend: Table and Row Model Enhancements**
- Extended the performance table definitions and row types to include `available_cores`, `sub_device_id`, and `dram_sharded`, ensuring these fields are available for display, filtering, and comparison. (`PerfTable.ts`, `PerfTableRow`, `TypedPerfTableRow`) [[1]](diffhunk://#diff-33118f62a39080b18535301836fb375ea2105cce68a7808eeb1cf30fdae18e10R20-R24) [[2]](diffhunk://#diff-33118f62a39080b18535301836fb375ea2105cce68a7808eeb1cf30fdae18e10R71) [[3]](diffhunk://#diff-33118f62a39080b18535301836fb375ea2105cce68a7808eeb1cf30fdae18e10R93) [[4]](diffhunk://#diff-33118f62a39080b18535301836fb375ea2105cce68a7808eeb1cf30fdae18e10R117-R118)
- Updated the columns configuration for the performance table to display the new fields, and adjusted the footer and insertion logic accordingly. (`PerfTable.ts`, `PerfTableColumns.ts`) [[1]](diffhunk://#diff-6bd984ea8e7c5d63ee0dd216213d6578251024c43e7b87887a7f2b8544835d61R75) [[2]](diffhunk://#diff-6bd984ea8e7c5d63ee0dd216213d6578251024c43e7b87887a7f2b8544835d61R84) [[3]](diffhunk://#diff-6bd984ea8e7c5d63ee0dd216213d6578251024c43e7b87887a7f2b8544835d61L105-R119) [[4]](diffhunk://#diff-534501e4d91e089fd13809a1ee547e1cd3e758d7fd6c32d17378fffcb4b4113cL15-R15)

**Frontend: Coloring and Data Enrichment Logic**
- Enhanced the core coloring logic to consider `available_cores` and `dram_sharded`, providing more accurate visual feedback in the UI. Updated data enrichment to parse and normalize the new fields. (`perfFunctions.tsx`, `enrichPerfRowData.ts`, `normalisePerformanceData.ts`) [[1]](diffhunk://#diff-06811d0e2ee912b54c0d90b02cd8926d71cbf822029522c63197b919c69c0c9aL358-R358) [[2]](diffhunk://#diff-06811d0e2ee912b54c0d90b02cd8926d71cbf822029522c63197b919c69c0c9aL433-R454) [[3]](diffhunk://#diff-3bd520441fdbdc637e0e308c432bccf31e2601a2ef1bc5b20a2f2bfd90e9339bR70) [[4]](diffhunk://#diff-6c0d370bcf499513fc4241e0af417451c46ea36061e7f4e85042c9ebf49bf189L29-R29)

**Frontend: Component Integration**
- Updated components such as `OperationListPerfData` to pass the new fields to the coloring logic and display them as appropriate. (`OperationListPerfData.tsx`)

These changes improve the robustness and expressiveness of performance reporting and visualization, ensuring new device-level metrics are accurately handled and presented throughout the stack.


backend/ttnn_visualizer/tests/test_perf_report_contract.py
```diff
@@ -42,7 +42,7 @@
 
 # Keep in lockstep with the tt-perf-report pin in pyproject.toml. Bumping the pin without
 # bumping this constant fails the version assertion below, forcing a deliberate parity review.
-EXPECTED_TT_PERF_REPORT_VERSION = "1.2.8"
+EXPECTED_TT_PERF_REPORT_VERSION = "1.3.0"
```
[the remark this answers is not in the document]
dcblundell: Updated.

src/functions/perfFunctions.tsx
```diff
@@ -437,10 +437,6 @@ export const getCoreColour = (value: string | string[] | boolean | number): Cell
         if (cores < 10) {
             return CellColour.Red;
         }
```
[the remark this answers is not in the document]
dcblundell: Fixed in commit aff86231. Core colouring now uses available_cores and dram_sharded, including the operation-list call site, with parity scenarios covering the new behaviour.

src/definitions/PerfTable.ts
```diff
@@ -80,6 +80,7 @@ export enum ColumnKeys {
     OP = 'op',
     HighDispatch = 'high_dispatch',
     GlobalCallCount = 'global_call_count',
+    SubDevice = 'sub_device_id',
```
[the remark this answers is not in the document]
dcblundell: Fixed in commit aff86231. Renamed the enum member to SUB_DEVICE and updated its references.

src/model/PerfTable.ts
```diff
@@ -17,6 +17,7 @@ export interface PerfTableRow {
     op_code: string;
     raw_op_code: string;
     device: string;
+    sub_device_id?: string | null;
```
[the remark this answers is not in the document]
dcblundell: Fixed in commit bc6c9b08. available_cores is now present in the raw/typed row model, visible and comparison columns, with rendering coverage.

backend/ttnn_visualizer/tests/test_perf_report_contract.py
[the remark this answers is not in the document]
dcblundell: Fixed. The parity contract now includes full-budget, small sub-device-budget, and DRAM-sharded core scenarios, and passes both fields into the generated tt-perf-report cells.

src/functions/perfFunctions.tsx
```diff
@@ -430,15 +430,28 @@ export const getKernelRiscColour = (riscUs: number, deviceKernelUs: number | nul
...
+        if (isDramSharded) {
+            return CellColour.Green;
+        }
+
+        if (cores < 10 && usesLittleOfBudget) {
             return CellColour.Red;
         }
 
-        if (cores === 64) {
+        if (parsedAvailableCores != null && cores === parsedAvailableCores) {
```
aidemsined: > [!CAUTION]
> This green now contradicts the Flags cell two columns to its left on the same row, because the core heuristic and the utilisation charts still divide by the whole chip.

`computePerfHeuristicFlags.ts:72` is still `cores / maxCores < UNDERUTILISED_CORES_RATIO`, with `maxCores` the architecture-wide count from `resolveMaxCores`, and its detail string is literally ``Cores: ${row.cores} / ${maxCores}``. `getCoreUtilization.ts:18` still multiplies by `maxCores / coreCount`. Neither reads `available_cores`.

So an above-min-impact op given a 12-core subdevice and using all 12 of it gets, on one row:

- a **green** Cores cell from this line (`12 === available_cores`) — correct per 1.3.0
- a **warning** flag reading `Underutilised cores — Cores: 12 / 64` (12/64 = 0.19 < 0.25)
- a charted utilisation overstated by `maxCores / available_cores` ≈ 5.3×, feeding `PerfCoreCountUtilizationChart`, `PerfKernelDurationUtilizationChart` and `PerfOperationKernelUtilizationChart`, and suppressing `LOW_UTILISATION`

That false positive is exactly what 1.3.0's `uses_little_of_budget` was added to remove — its comment says "an op given a 12-core dispatch subdevice and using 6 of them is not reported as underutilizing the device". This PR imports the fix for the colour and leaves the flag and the charts on the old denominator.

Neither file is in this diff, so nothing here is regressed — but the PR is what creates the disagreement, and `available_cores` is now on every row.

**Suggested action:** `row.available_cores ?? maxCores` as the denominator in `isUnderutilisedCores` and `getCoreUtilization`, and use the same value in the flag's detail string so it reads `12 / 12` rather than `12 / 64`.

backend/ttnn_visualizer/tests/test_perf_report.py
```diff
@@ -183,6 +183,101 @@ def _fake_generate_perf_report(*args, **kwargs):
...
+                "ttnn_visualizer.csv_queries.perf_report.generate_perf_report",
+                side_effect=_fake_generate_perf_report,
+            ),
+        ):
+            report = OpsPerformanceReportQueries.generate_report(instance)["report"]
+
+        self.assertEqual(report[0]["raw_op_code"], "Matmul")
+        self.assertEqual(report[0]["advice"], ["useful advice"])
+        self.assertEqual(report[0]["sub_device_id"], "subdevice-7")
+        self.assertEqual(report[0]["available_cores"], "108")
```
aidemsined: > [!CAUTION]
> Four assertions pin a 26-entry mapping, so any single mislabelled display name passes the whole suite — which is the exact failure the refactor was meant to make impossible.

Two independent mutations confirm it. Changing `"cores": ("Cores", "cores")` to `("DRAM", "cores")` — which would render DRAM GB/s under the Cores heading, and colour it — passes **1251/1251** backend tests. Mislabelling five entries at once (swapping the two `Output Subblock` names, swapping the two `Input N Datatype` names, and pointing `dram` at `"DRAM %"`) passes all **9** tests in this module. Two of those survive because the snake_case aliases match first; this test survives because it never looks at those fields.

The consequence is a whole column of numbers under the wrong heading, with a green suite and nothing that looks wrong on screen. That is strictly worse than the positional bug this refactor fixed, because a positional error shifts everything visibly.

The fix is nearly free here — the test already writes the full rotated header row, and every fixture value is distinct:

**Suggested action:** assert the whole parsed row against an expected dict rather than four fields. For something durable, two options with existing precedent in this repo: drive tt-perf-report's real CSV writer over a synthetic row in `test_perf_report_contract.py` (which already imports the installed package) and assert `REPORT_COLUMN_HEADERS`' display names against the emitted header row — that catches renames on the next upgrade; and/or add a column parity test modelled on `test_event_logging_frontend_parity.py`, asserting the mapping's keys against `ColumnKeys` in `src/definitions/PerfTable.ts`.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -623,6 +625,33 @@ class OpsPerformanceReportQueries:
...
+                    for name in header_names
+                    if name in header_index_by_name
+                ),
+                None,
+            )
+
+        return column_indexes
+
+    @classmethod
+    def _parse_report_row(cls, column_indexes, row):
```
aidemsined, line 645: > [!CAUTION]
> 1.3.0 also added CSV formula escaping, which this parser doesn't undo — so some cells now reach the UI with a stray leading apostrophe.

I checked the 1.3.0 wheel directly rather than inferring it. `_CSV_FORMULA_PREFIXES = ("=", "+", "-", "@", "\t", "\r")` at `perf_report.py:528`, and `escape_csv_formula` prefixes a literal `'` to any **string** cell starting with one of them. It is applied on the CSV write path to every column (`:2751`) and to Advice (`:2757`). Absent from 1.2.8, so it arrives with this bump.

Probed the function's behaviour:

```
'=> BF16'        -> "'=> BF16"
'-3'             -> "'-3"
'@x'             -> "'@x"
'HiFi4 => BF16'  -> unchanged
42, None         -> unchanged
```

`math_fidelity` is built as `math_fidelity += f" => {short_name(output_datatype)}"` (`:1280`), so an op with no recorded fidelity but an output datatype yields `"=> BF16"` → the Math Fidelity column renders `'=> BF16`.

Worth noting upstream's own docstring names the other case: the escaping exists because `--csv` carries "op codes, and **a subdevice id that was not a whole number**" — i.e. it can land on the column this PR adds.

**Suggested action:** strip one leading `'` from string cells in `_parse_report_row`. Cheap and total; restricting it to specific text columns would need revisiting every time upstream escapes another one.

src/definitions/PerfTable.ts
```diff
@@ -102,16 +104,19 @@ export const Columns: ColumnDefinition[] = [
...
     },
     { name: 'Flags', key: ColumnKeys.Flags, footerSpan: 0 },
     { name: 'Device', key: ColumnKeys.Device, footerSpan: 0 },
+    { name: 'Sub Device', key: ColumnKeys.SubDevice, footerSpan: 0 },
     { name: 'Type', key: ColumnKeys.BufferType, sortable: true, filterable: true, footerSpan: 0 },
     { name: 'Layout', key: ColumnKeys.Layout, sortable: true, filterable: true },
     { name: 'Device Time', key: ColumnKeys.DeviceTime, unit: 'µs', decimals: 0, sortable: true },
     { name: 'Op-to-Op Gap', key: ColumnKeys.OpToOpGap, colour: 'red', unit: 'µs', decimals: 0, sortable: true },
     { name: 'Cores', key: ColumnKeys.Cores, colour: 'green', sortable: true },
+    { name: 'Available Cores', key: ColumnKeys.AvailableCores, sortable: true },
```
aidemsined: > [!WARNING]
> Both columns are unconditional and visible by default, where 1.3.0 deliberately gates them — and neither is filterable, though the PR body says they are.

Upstream shows these only when they carry information: `show_sub_device_ids = sub_device_count > 0` and `show_available_cores = show_sub_device_ids or len(reported_budgets) > 1` (`perf_report.py:2669`), precisely because on every other run one is blank and the other is a constant. The repo's own fixture CSV has neither raw column, so for that capture — and every pre-subdevice capture — Sub Device is empty on all rows and Available Cores reads `64` on all rows, permanently, with `hiddenPerfTableColumnsAtom` defaulting to `[]`.

`getEligiblePerfColumns` already has exactly this mechanism (`hasOpIds`, `hasL1PressureData`, `hiliteHighDispatch`, `hasNpe`), so this is two more flags rather than new machinery. It also removes the cost: two always-on columns are ~8% more cells on a table that isn't virtualised — `PerfTable.tsx:327` maps every row, and the repo's `react-virtuoso` usage is confined to the operation, tensor and buffer lists — plus ~41 bytes per row on the wire, which is ~1.2 MiB on a 30k-row report.

Separately, the body claims the new fields are "available for display, filtering, and comparison". Display and comparison are both there, but neither line carries `filterable: true` — and filtering by subdevice is the first thing anyone will reach for on a partitioned run.

**Suggested action:** gate both on the same data signal 1.3.0 uses, and add `filterable: true` to at least Sub Device.

backend/ttnn_visualizer/tests/test_perf_report_contract.py
```diff
@@ -127,6 +129,21 @@
     {"id": "cores-full", "column": "cores", "row": {"cores": 64}},
```
aidemsined, line 129: > [!WARNING]
> The regenerated golden can't distinguish the new budget-aware green from the hardcoded `cores === 64` it replaced.

Every scenario carries `available_cores: 64` except the two small-budget cases, and none has `cores === available_cores` with a non-64 budget. So `cores-full` (64/64) is green under both the new rule and the old one, `cores-mid` (32/64) is white under both, and the two 12-budget cases are decided by the red branch rather than the green one.

Demonstrated: replacing `perfFunctions.tsx:454` with `if (cores === 64)` passes **all 2504 frontend tests, including this 41-scenario golden**. On Blackhole P150 — 110 worker cores in 1.3.0's arch registry — or on any subdevice budget, a full-grid op would silently lose its green and nothing would fail.

Worth saying the harness is otherwise good: it's generated from the real installed library, CI runs it under `--frozen`, and inverting `isDramSharded` *is* caught (5 failures). This is one uncovered axis, not an inert fixture.

**Suggested action:** one scenario with `{"cores": 110, "available_cores": 110}` (or any non-64 pair) pins the rule that actually shipped. While there: `_build_op_data` supplies exactly the 14 keys `color_row` reads today, but upstream reads them via `op_data.get(key, Cell(default))`, so a future colour input that nobody adds here records upstream's *default-input* behaviour as if it were real — which is how the first commit shipped a golden asserting `cores-full → white`. Asserting that the provided key set matches the accessed one closes that permanently.

src/functions/perfFunctions.tsx
```diff
@@ -430,15 +430,28 @@ export const getKernelRiscColour = (riscUs: number, deviceKernelUs: number | nul
...
-export const getCoreColour = (value: string | string[] | boolean | number): CellColour => {
+export const getCoreColour = (
+    value: string | string[] | boolean | number,
+    availableCores: string | number | null | undefined = null,
+    dramSharded: boolean | string | null | undefined = null,
+): CellColour => {
     const cores = (typeof value === 'string' ? parseInt(value, 10) : value) as number;
+    const parsedAvailableCores = typeof availableCores === 'string' ? parseInt(availableCores, 10) : availableCores;
+    const isDramSharded =
+        dramSharded === true || (typeof dramSharded === 'string' && ['true', '1'].includes(dramSharded.toLowerCase()));
```
aidemsined: > [!WARNING]
> Second truthiness vocabulary for the same field, and the branch production actually takes is untested.

`enrichPerfRowData.ts:70` accepts only `'true'`; this accepts `'true'` and `'1'`. `OperationListPerfData.tsx` passes a raw `PerfTableRow` (all strings) while `getCellColour` passes an already-typed row, so a `"1"` cell would colour the operation list green and the perf table white. The repo's convention names `isFlagEnabled` as the single frontend owner of that vocabulary precisely because a second copy drifts in both directions.

The string branches are also outside the chain of trust: the golden drives `getCellColour` with a `TypedPerfTableRow` built from JSON, so `dram_sharded` arrives as a boolean and `available_cores` as a number. Inverting the enrich parse to `!== 'true'` passes all 2504 tests. I confirmed the CSV literal is `True`/`False` (Python `bool` through `DictWriter`), so the current code is right — nothing pins it.

One latent trap in the same block: `available_cores === ''` gives `parseInt('') === NaN`, and `NaN == null` is false, so `usesLittleOfBudget` is false and both colour rules disable — where Python's `None` yields red. `Number.isNaN(parsed) ? null : parsed` makes the two paths identical.

**Suggested action:** normalise once at the boundary — coerce through `isFlagEnabled` in `enrichRowData`, have the operation-list path consume a typed row, and let `getCoreColour` take one input shape so the golden exercises the code the app runs.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -538,32 +538,34 @@ def get_all_folders(directory: str) -> List[str]:
...
-        "dram_sharded",
-        "input_0_memory",
-        "inner_dim_block_size",
-        "output_subblock_h",
-        "output_subblock_w",
-        "global_call_count",
-        "advice",
-        "raw_op_code",
-    ]
+    REPORT_COLUMN_HEADERS = {
```
aidemsined, line 541: > [!WARNING]
> The second element of every tuple is the dict key repeated, and it exists only to keep one legacy test fixture green.

All 26 entries are `"x": ("Display Name", "x")` — I checked each one, the alternate equals the key every time. And no producer emits snake_case headers: both 1.2.8 and 1.3.0 write the CSV through `csv.DictWriter(f, fieldnames=all_headers)` with the display-name list, and the real output header row is `ID,Total %,Bound,...`.

Its only consumer is `REPORT_HEADER` at `test_perf_report.py:30`, which dates from when parsing was positional and header content was irrelevant. Stripping all 26 aliases fails exactly two tests — `test_raw_csv_kernel_durations_are_passed_through` and `test_signpost_row_reaches_the_report_with_its_op_type` — and nothing else.

So the production parser now accepts a schema that cannot occur, and two of the three report-parsing tests exercise headers the parser will never see. That also doubles the surface a typo can hide in, which is the same exposure as the assertion gap I've flagged on the test file.

**Suggested action:** reduce the mapping to `{field: "Display Name"}` and fix the two fixtures to use real header names. The resolver simplifies with it — the `next((...), None)` generator becomes a single lookup.

aidemsined (commented):
The column refactor is the right call and it is correct: I resolved all 26 keys against 1.3.0's real emitted header order and every field lands on its own column, and the refactor was necessary rather than optional — 1.3.0 inserts the two new columns *before* `Advice`/`Raw OP Code`, so the old positional list would have read `advice` from `Sub Device ID`. Backwards compatibility is safe by construction, not by luck: 1.3.0 writes a fixed header set regardless of input, and the report CSV is always regenerated by the pinned library rather than read from disk. Probes with a missing new column, a missing old column, a missing `ID`, extra unknown columns and short rows all degraded gracefully — key absent, row skipped, or a clean 422 — never shifting a neighbouring column and never escaping as a 500. `pyproject.toml`, `uv.lock` and the contract pin all agree on 1.3.0, and CI runs the contract under `--frozen`, so the pin is genuinely gated.

Three things I'd want addressed before merge, all inline. The one I'd start with is not in this diff at all: the core heuristic and the utilisation charts still divide by the whole chip, so the green this PR adds to a full-budget subdevice op now sits on the same row as an "Underutilised cores — Cores: 12 / 64" warning.

## Unanchored findings

- **1.3.0's partitioned-run totals warning is dropped, while the UI shows exactly the figures it names.** On the same `show_available_cores` signal that promotes these columns, upstream prints that totals assume sequential execution — summed device time, Total %, the op-to-op gap totals, the DRAM roofline. `PerfTable.tsx:477` computes `deviceTimeSum` and `opToOpGapSum` and hardcodes Total % as `'100%'` with no caveat, and the stacked table and charts consume the same sums. Two concurrent 100 µs ops on disjoint subdevices are footered as 200 µs and each shown at half its true share — and this PR is what adds the Sub Device column that tells the user the run was partitioned. `PerfMultiDeviceNotice.tsx` is the existing precedent for surfacing this class of caveat. (`PerfTable.tsx` isn't in the diff, hence no inline anchor.)
- **HiFi3 is missing from the frontend fidelity logic while tt-perf-report handles it.** `definitions/MathFidelity.ts` declares only HiFi4/HiFi2/LoFi and `evaluateFidelity` has no HiFi3 branch. Measured side by side: `BFLOAT8_B × BFLOAT8_B ⇒ BFLOAT16 @ HiFi3` gives TS `unknown` (white cell) against Python `too_high` (red); `BF16 × BF16 ⇒ BF16 @ HiFi3` gives white against cyan. Pre-existing, but the 13 fidelity scenarios contain no HiFi3 case, which is exactly why this bump's regenerated golden shows zero fidelity diffs — and a version bump is the "deliberate parity review" the contract's own docstring invokes.

## General observations

- `footerSpan: 5` is inert — `getFooterColumns` unconditionally overwrites it with `1 + trailingSkippedCount`, and setting it to `99` passes all 2504 tests. This PR bumped 4→5 and rewrapped a "keep in sync with `getFooterColumns`" comment for a value nothing reads; dropping the field would have made those three edits zero.
- `csv_queries.py:843` re-tests `id_index is None` inside the row loop, so an unresolvable ID header makes every row raise, get swallowed by the `except` at `:887`, and `continue` — the endpoint returns `{"report": []}` and the UI shows "No data to display" with no indication that parsing failed, while the log takes one full 26-element row repr per row. `id_index` can't change between rows; hoisting the check and raising `DataFormatError` once is both correct and cheaper.
- `OP_ID_INSERTION_POINT = 1` and `L1_PRESSURE_INSERTION_POINT = 2` are still literals immediately above the `HIGH_DISPATCH_INSERTION_POINT` this PR converted to a `findIndex`. The conversion was the right instinct — inserting Sub Device is what broke the literal — but the file now teaches two conventions, and the next inserted column breaks the two left behind, in the same silent way.
- The new header test landed inside `TestPerfReportKernelDurationSchemaCompatibility`, whose docstring scopes the module to per-RISC kernel durations (#1518). Column resolution wants its own module — which would also stop one file carrying two hand-written, mutually inconsistent 26-name header lists.
- `comparisonKeys` is a parallel list to `Columns`, so each new column is declared twice on the frontend while the other seven attributes live on `ColumnDefinition`. A `comparable?: boolean` retires that site permanently.
- `PLACEHOLDER` in `normalisePerformanceData.ts` didn't gain `sub_device_id` / `available_cores`, though its sibling `signpostRowDefaults` did — in the same commit that edited `PLACEHOLDER.dram_sharded`. Both fields are optional on `PerfTableRow`, so TypeScript can't catch the divergence and the two objects will keep drifting.
- Header matching is exact and case-sensitive, duplicates resolve to the last occurrence, and a whitespace-padded or BOM-prefixed header row silently unresolves every column. Fine while tt-perf-report is the only producer; a `.strip()` on header names is free insurance.
- `dram_sharded` is typed `boolean | null` but `enrichRowData` always yields a boolean, so "the report had no DRAM Sharded column" is indistinguishable from "this op isn't DRAM-sharded".
- `cores < 10` and `parsedAvailableCores / 2` are bare literals where this file already names `KERNEL_CRITICAL_PATH_SHARE`, `['true', '1']` is reallocated per call, and upstream's ten-line rationale for why red needs *both* halves didn't come across — without it `usesLittleOfBudget` reads as arbitrary. (`usesLittleOfBudget` is also computed before the `isDramSharded` early return that discards it.)
- Three of the eight commits exist only to repair or reverse earlier ones in the same PR, and `aff86231` reverses `d8646523` — which had deleted the full-grid green outright rather than reinterpreting it. Worth squashing, or the history reads as if green-at-full-grid was removed on purpose.
- The tt-perf-report upgrade procedure lives only in `test_perf_report_contract.py`'s docstring, and this PR shows it's incomplete: it says to bump the pin, regenerate the golden and reconcile the TS colour tests, but nothing about re-checking the column set, extending the colour-input builders when the *inputs* change, or the frontend row/column/comparison/footer edits that follow. This is the PR that discovered those steps, so it's the natural place to write them down.

backend/ttnn_visualizer/csv_queries.py
dcblundell: Fixed. REPORT_COLUMN_HEADERS now contains only the emitted display header for each field, the resolver performs a direct lookup, and parser fixtures use real display headers.

backend/ttnn_visualizer/tests/test_perf_report.py
dcblundell: Fixed. The parser test now asserts every mapped report field rather than only a small subset, so display-name/key mismatches are caught.

src/functions/perfFunctions.tsx
dcblundell: Fixed. Boolean parsing now uses isFlagEnabled, numeric budgets normalise invalid and empty values to null, and both table and operation-list paths use the same accepted vocabulary.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -538,32 +538,34 @@ def get_all_folders(directory: str) -> List[str]:
...
+        "input_0_datatype": ("Input 0 Datatype", "input_0_datatype"),
+        "input_1_datatype": ("Input 1 Datatype", "input_1_datatype"),
+        "dram_sharded": ("DRAM Sharded", "dram_sharded"),
+        "input_0_memory": ("Input 0 Memory", "input_0_memory"),
+        "inner_dim_block_size": ("Inner Dim Block Size", "inner_dim_block_size"),
+        "output_subblock_h": ("Output Subblock H", "output_subblock_h"),
+        "output_subblock_w": ("Output Subblock W", "output_subblock_w"),
+        "global_call_count": ("Global Call Count", "global_call_count"),
+        "sub_device_id": ("Sub Device ID", "sub_device_id"),
+        "available_cores": ("Available Cores", "available_cores"),
```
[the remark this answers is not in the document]
dcblundell: Fixed. The parser now passes through DEVICE FW START/END CYCLE and METAL TRACE ID / METAL TRACE REPLAY SESSION ID, with frontend row fields and table columns. Focused backend and frontend tests pass.

src/definitions/PerfTable.ts
dcblundell: Fixed. Sub Device and Available Cores are now gated by the data present in active and comparison rows, and Sub Device is filterable.

src/model/PerfTable.ts
```diff
@@ -17,9 +17,11 @@ export interface PerfTableRow {
...
     device_time: string;
     op_to_op_gap: string;
     cores: string;
+    available_cores?: string | null;
```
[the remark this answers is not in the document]
dcblundell: Fixed. available_cores is now parsed to number | null before reaching TypedPerfTableRow, so sorting compares numeric budgets. The table model and focused tests were updated.

backend/ttnn_visualizer/tests/test_perf_report_contract.py
dcblundell: Fixed. The colour contract now includes a non-default full-budget scenario (110/110), and the regenerated golden passes against tt-perf-report 1.3.0.

src/functions/perfFunctions.tsx
dcblundell: Fixed. Core heuristic flags and utilisation now use row.available_cores ?? maxCores, and the detail text reports the same denominator. This keeps the colour, flag, and charts consistent for subdevice budgets.

backend/ttnn_visualizer/csv_queries.py
dcblundell: Fixed. CSV parsing now removes tt-perf-report's formula-safety apostrophe only when it prefixes a recognised formula character, preserving ordinary leading apostrophes. A regression test covers both paths.

src/definitions/PerfTable.ts
```diff
@@ -72,6 +72,7 @@ export enum ColumnKeys {
     Layout = 'layout',
     OpToOpGap = 'op_to_op_gap',
     Cores = 'cores',
+    AvailableCores = 'available_cores',
```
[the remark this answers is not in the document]
dcblundell: Fixed. The new enum members are now AVAILABLE_CORES and SUB_DEVICE, with all references updated.

PR #1996, opened 2026-09-09
Expose perf triage to an agent over MCP

### LINKED ISSUE #1995: Expose perf-triage tools to an agent over MCP (TT-NN bring-up, first slice)
An agent doing TT-NN model bring-up asks a small number of questions repeatedly: which ops dominate, where did the time go inside them, and did my change help. It can answer none of them today without being handed a whole report, which is the wrong shape — a perf report is 10⁴–10⁵ rows × ~30 fields, and an agent's context is a tighter budget than a browser's heap.

This is the first slice of that: a stdio MCP server over the query classes the backend already has, with every tool returning aggregates or bounded slices and never the table.

## Tools

- `load_report(profiler_path, performance_path?)` → a handle plus what the report actually contains, so an agent knows which questions are answerable before it asks.
- `top_ops(handle, by, limit)` → bounded top-N from the perf report.
- `zone_timings(handle, limit)` → per-zone, per-RISC aggregate from `profile_log_device.csv`. This is the one genuinely new capability; the file is read today only by `query_by_timer_id` / `query_zone_statistics`, both of which need you to already know what you're looking for.
- `diff_reports(a, b, by, limit)` → op-level deltas between two reports, which is the shape of the actual iteration loop (change something, rerun, compare).

## Two things it must not do

**Serve view-filtered rows as if they were the report.** `get_performance_results_report` defaults `hide_host_ops=true` and `merge_devices=true` and can narrow to a signpost range. Hand an agent that and it will reason confidently about a partial set — the #1883 lesson, where link status was polluted by view filters and six downstream features silently dropped out. The MCP calls the query classes with explicit canonical arguments and reports the projection it used in every response.

**Amended after review on #1996:** one filter is kept deliberately — `merge_devices` stays true, because unmerged rows report the same operation once per device, so a merged row is the unit an operation is reported in rather than a subset of the report. It is named in the response and in `docs/src/agent-tools.md` rather than left to be inferred. Separately, the partitioned-run caveat below is implemented but dormant until #1994 lands: `sub_device_id` is not parsed on `dev` and the pin is tt-perf-report 1.2.8, so no generated row carries the field yet.

**Report unsound totals silently.** On a partitioned run the summed device time, Total % and op-to-op gap totals all assume sequential execution; tt-perf-report prints a warning about this that we do not surface (raised on #1994). Any tool returning a total says so when the report has more than one sub-device.

## Not in this slice

The Blaze join. It needs `blaze_viz.json` exported (nobody owns that yet) *and* kernel-side zone instrumentation carrying op names. I surveyed every device log available locally: five of six carry only the six default `BRISC/NCRISC/TRISC × FW/KERNEL` zones, and the one exception carries protocol-level names (`ALL-GATHER-READ`, `SYNC-ZONE-SENDER`) rather than op names. So the join key does not exist today and will not appear on its own — worth confirming appetite with Marko or Joseph Chu before designing anything on top of it.

Transport is hand-rolled JSON-RPC over stdio to keep this dependency-free while it is a spike; swapping in the official `mcp` SDK is a contained follow-up once the tools prove useful.

aidemsined:
## Summary

An agent doing TT-NN bring-up asks a report the same few questions a person does: which operations dominate, where the time went inside them, and whether a change helped. It could answer none of them without being handed the whole report, which is the wrong shape — a performance report is 10⁴–10⁵ rows of ~35 fields, and an agent's context is a tighter budget than the browser heap that already forced server-side windowing on the NPE view.

This adds a stdio MCP server over the report readers the backend already has, with four tools: `load_report`, `top_ops`, `zone_timings`, `diff_reports`.

Closes #1995

### No new dependency

The tools read the query classes directly rather than the HTTP API. `Instance` turns out to be a plain model of two paths, so nothing here needs Flask, a database session or an `instanceId` — a report is registered once by path and addressed afterwards by handle. Two handles stay live at once, because `diff_reports` needs both and a cache of one could not hold them.

The transport is newline-delimited JSON-RPC in one file rather than the official `mcp` SDK, which keeps a resolver tree and a licence scan off a spike. Swapping the SDK in touches only `server.py` — nothing above it imports from there.

### Backend

- **`agent/handles.py`** — path → handle registry, and the inventory `load_report` returns. It reports what is *answerable* rather than leaving an agent to discover it one failed call at a time: the report kinds are independent, so a performance-only capture has no operation graph and no tensor data.
- **`agent/tools.py`** — the four tools, `CANONICAL_PROJECTION`, and the caveat logic.
- **`agent/server.py`** — the stdio transport. Every byte on stdout is protocol; logs go to stderr, and `tt_perf_report`'s own printing is captured because it would otherwise land mid-frame and corrupt the stream.
- **`csv_queries.py`** — `query_zone_summary`, the one new capability. The two existing zone queries both need you to already know the zone (one takes a name, the other a timer id), so nothing answered "what is in this capture" — the first question anyone asks of a log they did not produce. It pairs each `ZONE_START` with the `ZONE_END` closing it on the same core, RISC and run, holding open starts in a dict rather than the rows: a pair split across chunks still matches, and peak memory tracks concurrently-open zones rather than the ~288 MB file. Plus `read_capture_metadata`, since `offset=1` skips the preamble line that carries the clock.

### Three properties, because a wrong answer looks like a right one

An agent's caller has nothing rendered beside the number to catch an omission, so these are enforced rather than documented:

- **Bounded results.** Aggregates or slices capped at 100 — never the table.
- **A stated projection.** The performance route defaults `hide_host_ops` and `merge_devices` to true and can narrow to a signpost range. Those are display choices, and #1883 is what they cost when they leak — a link status polluted by view filters silently dropped six downstream features. The tools read with an explicit `CANONICAL_PROJECTION` and return it beside the rows. Device merging is the one filter kept, deliberately, and the response says so.
- **Caveats attached to the number.** On a partitioned run, summed device time and total percentages assume sequential execution; tt-perf-report warns on the terminal, which a tool response does not have. Any total now carries that, naming the sub-devices. `zone_timings` carries its own: cycles are summed across cores, so they are occupancy rather than elapsed time.

### Docs

`docs/src/agent-tools.md` (in the User Guide toctree), the convention in `CONVENTIONS.md` § Agent-facing tools with the one-liner and pointers in `AGENTS.md`, and a developer-only `agent/README.md` that points at the page rather than repeating it.

- **No Blaze join.** It needs `blaze_viz.json` exported *and* kernel-side zones carrying op names. I surveyed every device log available locally: five of six carry only the default firmware/kernel zones, and the one exception carries protocol-level names (`ALL-GATHER-READ`, `SYNC-ZONE-SENDER`) rather than op names. The join key does not exist today and will not appear on its own — worth confirming appetite before designing on top of it.
- `zone_timings` describes RISC-level phases, not model operations, on any capture whose kernels were not instrumented — which is most of them.
- Durations need the device log's `type` column to pair starts with ends. It is deliberately not required by the reader, so a capture without it reports occurrence counts and `null` durations rather than a guess.
- No SQLite-backed tools yet (operations, tensors, buffers), so the graph and memory questions are still unanswerable.

## Test plan

- [x] Backend: `pytest` — 1277 passed (26 new in `test_agent_tools.py`)
- [x] `mypy`, `black`, `isort` clean; SPDX flags nothing tracked
- [x] `pnpm run build:docs` — sphinx builds the page, the toctree links it, and the `AGENTS.md` → `CONVENTIONS.md` anchor resolves
- [x] No frontend change
- [x] Driven end to end over stdio against real captures in `reports/`: `load_report` reports `[top_ops, zone_timings]` answerable and `operations` not; `zone_timings` returns BRISC-FW at 19.95 s occupancy over 1840 occurrences on 80 cores, with the all-gather protocol zones firing 80k–116k times; `top_ops` puts AllGatherAsync top at 187 µs over 30 ops; `diff_reports` gives all-zero deltas for a report against itself and correct add/remove with counts for two different ones; an unknown handle comes back as tool content rather than a protocol error

Two bugs only surfaced by running it, both now covered: `itertuples` renames any column that is not a valid identifier, so `"zone name"` arrived as `_1` and every lookup silently missed (first run returned zero rows); and cycles ÷ MHz is already microseconds, where I had divided by 1000 twice.

## Follow-ups

- Point a real agent at two reports from an actual iteration and see whether `diff_reports` changes how the loop feels — the cheap test of whether to build more.
- Swap the hand-rolled transport for the official `mcp` SDK once the tool set settles.
- SQLite-backed tools for the operation and tensor questions.


backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+    """Run the report with stdout captured.
+
+    `tt_perf_report` prints its progress and the temp paths it writes to. On a
+    stdio transport that lands in the middle of a JSON-RPC frame and corrupts
+    the stream, so it is captured here rather than at the transport -- a tool
+    called in-process should not write to stdout either.
+    """
+    captured = io.StringIO()
+    with contextlib.redirect_stdout(captured):
+        report = OpsPerformanceReportQueries.generate_report(
```
dcblundell, line 112: **Should fix:** Avoid regenerating the full report for every question

Each `top_ops` call reparses and regenerates the complete performance report for the same loaded handle. Repeated agent queries can therefore repeat large I/O, CPU, and temporary allocations. Cache suitable canonical data or derived aggregates per handle, while preserving the ability to keep two reports live for `diff_reports`.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+    # Largest movement in either direction: a regression matters as much as a win.
+    changes.sort(
+        key=lambda change: abs(_as_number(change["delta"]) or 0.0), reverse=True
+    )
+
+    return {
+        "metric": by,
+        "before": handle_a,
+        "after": handle_b,
+        "delta_total": round(
```
dcblundell: **Must fix:** Attach the partition caveat to `diff_reports`

`delta_total` is a total, but this response has no partitioned-run caveat. In addition, generated report rows do not preserve `sub_device_id`, so `_partitioned_caveat()` cannot reliably detect real partitioned reports. Preserve that field and apply the caveat to every relevant total, including diffs.

backend/ttnn_visualizer/agent/server.py
```diff
@@ -0,0 +1,245 @@
...
+        try:
+            message = json.loads(line)
+        except json.JSONDecodeError as error:
+            stdout.write(
+                json.dumps(_error(None, -32700, f"parse error: {error}")) + "\n"
+            )
+            stdout.flush()
+            continue
+
+        response = handle_message(message, table)
```
dcblundell, line 279: **Must fix:** Validate JSON-RPC message shape before dispatching

`json.loads()` accepts arrays, scalars, and `null`, but `handle_message()` assumes a mapping and calls `.get()`. A malformed client message can terminate the stdio server instead of returning `-32600`; validate the request object and cap the maximum input line size.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+# (`perf_report.py:480`). It is inert on this path today -- passing no signpost
+# range means `ignore_signposts`, which returns the frame before the strip can
+# run -- but a projection that exists to not filter should not declare a filter
+# it would honour the moment a tool took a range. Rows carry `op_type`, so an
+# agent can tell a marker from an operation.
+#
+# `merge_devices` is the one filter kept, and deliberately: unmerged rows report
+# the same op once per device, so a merged row is the unit an op is reported in
+# rather than a subset of the report. Every response repeats this projection.
+CANONICAL_PROJECTION: Dict[str, object] = {
```
dcblundell, line 47: **Must fix:** Do not present a view-filtered report as canonical

`merge_devices=True` retains the performance view's merging filter, and `generate_report()` still applies `DEFAULT_MIN_PERCENTAGE = 0.5`, dropping low-percentage operations. This contradicts the promise that agents receive the report rather than a partial view. Pass an explicitly unfiltered canonical threshold/projection, or amend #1995 and the new documentation to justify the intentional merge.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -411,6 +412,159 @@ def get_all_entries(
...
+            "zone name",
+            "RISC processor type",
+            "core_x",
+            "core_y",
+            "run host ID",
+            "time[cycles since reset]",
+        ] + (["type"] if has_entry_type else [])
+
+        try:
+            chunks = pd.read_csv(
```
dcblundell, line 530: **Should fix:** Bound zone-summary parsing state

This reads every CSV column even though the aggregation uses only a small subset, and unmatched `ZONE_START` entries remain in `open_starts` until EOF. A large or malformed log can therefore use avoidable CPU/memory and grow state with the number of unmatched rows. Restrict the parsed columns and define bounded recovery for unmatched starts.

backend/ttnn_visualizer/agent/handles.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    available: List[str] = []
+    missing: List[str] = []
+
+    def record(name: str, present: bool) -> None:
+        (available if present else missing).append(name)
+
+    perf_csv = None
+    if instance.performance_path:
+        perf_csv = next(
```
dcblundell: **Should fix:** Use one report-file selection policy

Inventory picks the lexicographically first `ops_perf_results*.csv`, while the execution path uses a different selection policy. `load_report` can therefore advertise one CSV while `top_ops` reads another. Reuse a shared canonical file-selection helper, and keep the advertised capabilities aligned with tools actually registered.

dcblundell (commented):
Posted six line-linked findings: three Must fix and three Should fix items covering malformed JSON-RPC input, incomplete canonical projections, partition caveats, report selection, repeated report generation, and zone-parser resource usage.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -411,6 +412,159 @@ def get_all_entries(
...
+                        continue
+                    risc = str(values[1]).strip()
+                    cycles = values[5]
+                    entry_type = str(values[6]).strip() if has_entry_type else ""
+
+                    zone_key = (zone, risc)
+                    if zone_key not in cores_seen:
+                        zone_keys.append(zone_key)
+                    core = (values[2], values[3])
+                    cores_seen[zone_key].add(core)
```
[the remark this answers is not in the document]
aidemsined: Right, and worse than the comment suggests — both captures in `reports/` are multi-slot (8 and 32), so this was wrong on every log I tested against.

Distinct-core counts were understated by that factor: `report_with_zones` reported 18 cores where there are 80, and the MoE capture 130 where there are 3840. Fixed by making a core `(PCIe slot, core_x, core_y)` in both the pair key and the count, with a two-slot fixture pinning it.

One correction to the mechanism you describe: the durations came out *unchanged*. The log is ordered per device, so each start is closed before another device reuses the coordinates — the collision existed but never fired. That is an accident of file layout rather than a guarantee, which is exactly why the identity should not have depended on it.

`2f27d025`.

backend/ttnn_visualizer/agent/server.py
```diff
@@ -0,0 +1,245 @@
...
+def serve(
+    stdin: TextIO, stdout: TextIO, registry: Optional[ReportRegistry] = None
+) -> None:
+    table = _tool_table(registry or ReportRegistry())
+    for line in stdin:
+        line = line.strip()
+        if not line:
+            continue
+        try:
+            message = json.loads(line)
```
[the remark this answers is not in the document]
aidemsined: Confirmed and fixed. `[]`, `null` and `7` all reached `.get`, raised out of `serve`, and ended the session on one bad frame.

Non-object frames and non-string methods now return -32600, non-object `params`/`arguments` return -32602, and a line over 1 MB is refused rather than accumulated. One subtlety worth noting for the review: the `params` check has to run *before* the `or {}` default, because `[]` is falsy and would otherwise be coerced to an empty object and answered as an unknown tool rather than refused.

Driven end to end afterwards — six malformed frames followed by a valid `tools/list`, which still returns four tools. Parametrised tests cover each shape. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+    sub-device; the visualizer does not surface it (raised on #1994). An agent
+    summing device time across subdevices that ran concurrently would overstate
+    the total and understate every op's share, so any tool reporting a total says
+    so here.
+    """
+    sub_devices = {
+        str(row.get("sub_device_id", "")).strip()
+        for row in rows
+        if str(row.get("sub_device_id", "")).strip()
+    }
```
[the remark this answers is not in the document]
aidemsined: Correct, and the honest answer is that it cannot fire on this branch: `sub_device_id` is not in `REPORT_COLUMN_HEADERS` on `dev` and the pin is still tt-perf-report 1.2.8, so only the mocked rows carry the field. I verified rather than assumed — `grep -c sub_device_id csv_queries.py` is 0 here.

I have not made this PR depend on #1994, and I would rather not: the mechanism is three lines at each of two call sites, it starts reporting the moment the column arrives with no change here, and gating an agent surface on a dependency-bump PR trades a dormant caveat for a blocked branch. What was wrong was claiming it works today, so that is now stated where a reader will look — a paragraph in the docs limitations naming #1994, and a note on `_partitioned_caveat` itself.

If you would rather it not ship dormant at all, say so and I will pull the caveat out and add it in a follow-up once #1994 lands. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+    ops = [
+        {
+            "id": row.get("id"),
+            "op_code": row.get("op_code") or row.get("raw_op_code"),
+            by: _rounded(_as_number(row.get(field))),
+            "cores": _as_number(row.get("cores")),
+            "device": row.get("device"),
+            "sub_device_id": row.get("sub_device_id") or None,
+            "bound": row.get("bound") or None,
+        }
```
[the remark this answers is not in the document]
aidemsined: Good catch — the projection justifies keeping signpost rows on the grounds that `op_type` distinguishes them, and then the response dropped `op_type`. The rationale and the output disagreed.

`op_type` is now in each returned row, with a test that a signpost row is identifiable in the response. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+        for row in ranked
+    ]
+
+    result: Dict[str, object] = {
+        "handle": handle,
+        "metric": by,
+        "op_count": len(rows),
+        "returned": len(ops),
+        f"{by}_total": round(total, 3),
```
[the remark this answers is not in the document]
aidemsined: Agreed, and taken as written: totals are now emitted only for `device_time`, `op_to_op_gap` and `total_percent`.

`dram` is GB/s, `flops` is TFLOPS and `cores` is a per-operation allocation — summing any of them across a report produces exactly the dimensionally invalid, authoritative-looking number you describe. An `ADDITIVE_METRICS` set gates the key, so ranking by DRAM still works and simply comes back without a total. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+        zones.append(
+            {
+                **entry,
+                # A megahertz clock ticks once per microsecond, so cycles over
+                # MHz is already microseconds.
+                "total_us": (
+                    round(total_cycles / clock_mhz, 3)
+                    if total_cycles and clock_mhz
+                    else None
+                ),
```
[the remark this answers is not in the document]
aidemsined: Right — `total_cycles and clock_mhz` made a measured zero indistinguishable from "this capture cannot say". Now `total_cycles is not None`, with a fixture whose zone opens and closes on the same cycle asserting `0` and `0.0` rather than `null`. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,315 @@
...
+        "before": handle_a,
+        "after": handle_b,
+        "delta_total": round(
+            sum(entry["total"] for entry in after.values())
+            - sum(entry["total"] for entry in before.values()),
+            3,
+        ),
+        "returned": min(len(changes), _bounded(limit)),
+        "changes": changes[: _bounded(limit)],
+        "projection": dict(CANONICAL_PROJECTION),
```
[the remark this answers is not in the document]
aidemsined: Taken. Both reports' sub-device sets are retained through the aggregation and the caveat is attached to `diff_reports` too — it is the response most likely to be acted on, since it answers "did my change help", and a delta of two unsound totals is unsound the same way.

`delta_total` is also now gated on the metric being additive, for the same reason as the `top_ops` total. `2f27d025`.

backend/ttnn_visualizer/agent/handles.py
```diff
@@ -0,0 +1,153 @@
...
+            iter(sorted(Path(instance.performance_path).glob("ops_perf_results*.csv"))),
+            None,
+        )
```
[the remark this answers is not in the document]
aidemsined: Agreed — the inventory globbed and took the lexicographically first match while every tool reads through `get_local_ops_perf_file_path`, which picks the newest. In a directory holding two captures `load_report` could name one and `top_ops` analyse the other.

It now calls that same selector, so the advertised file is the consumed file by construction rather than by two policies agreeing. `2f27d025`.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -411,6 +412,159 @@ def get_all_entries(
...
+        # occurrence counts and no durations rather than a guess.
+        has_entry_type = "type" in (
+            self.runner.df.columns if self.runner.df is not None else []
+        )
+
+        columns = [
+            "zone name",
+            "RISC processor type",
+            "core_x",
+            "core_y",
```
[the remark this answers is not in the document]
aidemsined: The gap is real, but I have not fixed it the way you suggest, and I think the suggested fix would be a regression.

`REQUIRED_DEVICE_LOG_COLUMNS` gates *every* device-log route, and its own comment records why it is deliberately short: "which columns exist varies by producer and gating on them rejects captures that work fine — the segformer demo reports have no `meta data`, the smoke fixture has no `run ID`". Adding four columns there to protect one new query would let it reject captures that other routes serve today, and #1941 is the ticket for having done a version of that before.

So the columns this query indexes are validated inside `query_zone_summary` instead: a capture missing one now gets a `DataFormatError` naming it — surfacing as a clean refusal rather than a `KeyError` mid-parse — and the other routes are untouched. Test drops `core_y` from the header and asserts the message names it.

Happy to widen the shared contract instead if you disagree, but it wants its own change with the demo reports checked against it. `2f27d025`.

backend/ttnn_visualizer/agent/handles.py
```diff
@@ -0,0 +1,153 @@
...
+    profiler_path = instance.profiler_path
+    record(
+        "zone_timings",
+        performance_path is not None
+        and Path(performance_path, DeviceLogProfilerQueries.DEVICE_LOG_FILE).is_file(),
+    )
+    record(
+        "operations",
+        profiler_path is not None and Path(profiler_path, PROFILER_DB_FILE).is_file(),
+    )
```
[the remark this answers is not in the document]
aidemsined: Correct, and it was a contradiction with the PR text: `operations` is not a registered tool, so an agent reading `answerable` had a name it could not call.

Taken the second of your options. `answerable`/`unanswerable` now contain tool names only, and data the report holds without a tool is reported separately as `data_present_without_tools: ["operations_database"]`. The test asserts the union of both tool lists is a subset of what `_tool_table` actually registers, so the two cannot drift again. `2f27d025`.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -411,6 +412,159 @@ def get_all_entries(
...
+            reverse=True,
+        )
+        # Reported rather than dropped: a capture stopped mid-zone leaves ends
+        # without starts, and a silent aggregate would read as a complete one.
+        if unpaired_ends:
+            logger.info(
+                "%s: %d zone ends had no matching start",
+                self.DEVICE_LOG_FILE,
+                unpaired_ends,
+            )
```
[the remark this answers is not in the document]
aidemsined: You are right, and the comment had it backwards: truncation leaves entries in `open_starts`, not `unpaired_ends`, so the condition ignored exactly the case it described.

Both directions are counted now and returned from `query_zone_summary` rather than logged — stderr is invisible to the caller, as you say. `zone_timings` turns a non-zero count into a caveat on the response saying how many starts and ends failed to pair and that the totals cover only what did. Test asserts `{"unmatched_starts": 1, "unmatched_ends": 1}` on a fixture built to leave one of each. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
aidemsined: Two halves, and they land differently — I checked both.

**The min-percentage half does not hold.** `DEFAULT_MIN_PERCENTAGE` reaches `color_row`, which mutes a cell colour and returns the row (`perf_report.py:1206`); the caller is `rows = [color_row(...) for op_data in rows]`, so every row survives. Generated at 0, 0.5 and 50 the report is **30 rows each time**, and a row at `total_percent = 0.00016%` survives all three:

```
min_percentage=  0.0: rows=30  lowest total_percent=0.000164…
min_percentage=  0.5: rows=30  lowest total_percent=0.000164…
min_percentage= 50.0: rows=30  lowest total_percent=0.000164…
```

Colour never crosses into the CSV we read, so nothing is dropped. Not changed.

**The merge half is fair**, and it was already the one thing I wanted stated rather than defaulted — but it was stated in the wrong place. `docs/src/agent-tools.md` now calls it out explicitly as the one filter applied and why (unmerged rows report one op once per device, so merged is the unit an op is reported in rather than a subset), and `CONVENTIONS.md` § Agent-facing tools carries the same. I have also amended #1995 so the issue does not read as though nothing is filtered.

If you would rather the tools default to `merge_devices=False` and let an agent merge, I will take that — it is a one-line default and a docs change, and I do not feel strongly. `2f27d025`.

backend/ttnn_visualizer/agent/server.py
aidemsined: Fixed, both parts. Non-object frames and non-string methods return -32600, non-object `params`/`arguments` return -32602, and a line over 1 MB is refused rather than buffered.

Worth flagging one subtlety the fix needed: the `params` check has to precede the `or {}` default, because `[]` is falsy and would otherwise be silently coerced to an empty object and answered as an unknown tool instead of refused. Parametrised tests cover `[]`, `null`, `7` and `"text"`, and an end-to-end run feeds six malformed frames followed by a valid `tools/list` that still answers. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
aidemsined: Taken on both counts. Both reports' sub-device sets are now retained through the op-code aggregation and the caveat is attached to `diff_reports`, with `delta_total` gated on the metric being additive.

On the second half — `sub_device_id` genuinely is not preserved, and the reason is upstream of this PR: the field is not in `REPORT_COLUMN_HEADERS` on `dev` and the pin is tt-perf-report 1.2.8, so no generated row carries it regardless of what this code does. `_partitioned_caveat` therefore returns `None` for every real report here and only fires on the mocked rows. I have not made this branch depend on #1994; instead the dormancy is now stated in the docs limitations and on the function, and it starts working when #1994 lands with no change here. Say the word if you would rather it came out entirely until then. `2f27d025`.

backend/ttnn_visualizer/agent/tools.py
aidemsined: Agreed and fixed — four questions of one handle meant four CSV re-parses, four trips through tt-perf-report and twelve temp files.

Cached per handle, but on the `ReportRegistry` rather than in a module global, which matters for the reason you name: `diff_reports` needs two reports live, so a cache of one could not serve it, and a registry-scoped dict keeps both while making the lifetime match the handles. It also fixed a test-isolation bug I had introduced with the first attempt — a module global leaked a patched report between tests, which a fresh registry cannot do by construction. Test asserts three `top_ops` calls generate once. `2f27d025`.

backend/ttnn_visualizer/csv_queries.py
aidemsined: Both taken, one fully and one partly.

`pd.read_csv` now takes `usecols`, so it parses the eight columns the aggregate reads instead of all fifteen.

On bounded recovery for unmatched starts: I have made them *visible* rather than bounded — counted and returned, and surfaced as a caveat on `zone_timings` saying the totals cover only what paired. I stopped short of capping the dict because any cap has to decide which starts to discard, and discarding silently is how a partial total comes to look complete; the count at least tells the caller. The bound in practice is concurrently-open zones, which is tens on a well-formed capture and unbounded only on a malformed one — if you want a hard cap with an explicit "state truncated" flag I will add it, but I would rather that were a deliberate shape than a number I picked. `2f27d025`.

backend/ttnn_visualizer/agent/handles.py
aidemsined: Agreed — the inventory globbed and took the first match alphabetically while the tools read through `get_local_ops_perf_file_path`, which picks the newest, so two captures in one directory could have `load_report` naming one and `top_ops` reading the other.

It calls that selector now, so there is one policy rather than two that happen to agree. The capability half is fixed too: `answerable` contains registered tool names only, and a test asserts the union of the tool lists is a subset of what `_tool_table` registers. `2f27d025`.

backend/ttnn_visualizer/agent/server.py
```diff
@@ -0,0 +1,278 @@
...
+        return _result(
+            request_id,
+            {
+                "content": [
+                    {"type": "text", "text": json.dumps(payload, indent=2, default=str)}
+                ]
+            },
+        )
+
+    return _error(request_id, -32601, f"method not found: {method}")
```
smountenay-tt, line 254: **Release blocker:** implement the MCP `ping` request before the generic `-32601` fallback. The advertised 2025-06-18 protocol requires a receiver to answer `ping` promptly with `result: {}`; a compliant client may treat this response as a failed health check and terminate/restart the server. This makes the new entry point protocol-incompatible with clients that use the standard liveness mechanism.

backend/ttnn_visualizer/agent/tools.py
```diff
@@ -0,0 +1,364 @@
...
+            lambda: {"total": 0.0, "count": 0.0}
+        )
+        for row in rows:
+            value = _as_number(row.get(field))
+            if value is None:
+                continue
+            op_code = str(row.get("op_code") or row.get("raw_op_code") or "").strip()
+            if not op_code:
+                continue
+            totals[op_code]["total"] += value
```
smountenay-tt, line 312: **Release blocker:** this still sums non-additive metrics inside each op-code group. The new `ADDITIVE_METRICS` guard only removes the report-wide `delta_total`; for `by=flops`, `dram`, or `cores`, each returned `*_before`, `*_after`, and `delta` remains a sum of rates/allocations. For example, one 10-TFLOPS Matmul versus two 8-TFLOPS Matmuls is reported as 10 → 16 (`+6`) even though every invocation regressed. Use a meaningful aggregation for these metrics or restrict `diff_reports` to additive metrics.

backend/ttnn_visualizer/csv_queries.py
```diff
@@ -411,6 +413,204 @@ def get_all_entries(
...
+
+                    zone_key = (zone, risc)
+                    if zone_key not in cores_seen:
+                        zone_keys.append(zone_key)
+                    # Slot included: coordinates repeat on every device.
+                    core = (values[2], values[3], values[4])
+                    cores_seen[zone_key].add(core)
+
+                    if not has_entry_type:
+                        occurrences[zone_key] += 1
```
smountenay-tt, line 560: **Release blocker:** pre-rename logs already distinguish boundaries through `zone phase=begin/end` (the existing `PRE_RENAME_*` fixtures pin that format), but this fallback ignores that column and counts both rows as occurrences. One zone invocation is therefore reported as two occurrences, while durations are discarded. Since `zone_timings` presents these as aggregate facts, consume `zone phase` when `type` is absent (or explicitly refuse that format) rather than returning a materially false count.

smountenay-tt (commented):
Release-blocking findings on the current head:

backend/ttnn_visualizer/csv_queries.py
aidemsined: This is the one I am most glad you caught — it was reporting a false number, not just an incomplete one.

Confirmed against `PRE_RENAME_HEADER`: `zone phase` with `begin`/`end`. My fallback keyed only on `type`, so on such a capture a begin and its end each incremented `occurrences` — one invocation reported as two — and the duration between them was discarded. `zone_timings` then presented that as an aggregate fact.

Both spellings are read now (`type` → ZONE_START/ZONE_END, `zone phase` → begin/end), resolved once per capture rather than per row. A capture with neither column still degrades to occurrence counts, which is the honest case. Test builds a pre-rename log and asserts one occurrence and a 700-cycle duration where the old code gave two and none.

`b8c31af1`.

backend/ttnn_visualizer/agent/tools.py
aidemsined: Correct, and my earlier fix was half a fix — I gated the report-wide `delta_total` and left every per-group number a sum of rates. Your example is the right one: 10 → 16 as an apparent gain when both invocations regressed.

Taken the second option, refusing rather than re-aggregating. A diff is per-op-code sums on both sides, so for a rate there is no number left to return that means anything, and choosing between a mean, a max and per-invocation is a product decision I would rather not make silently. `diff_reports` now raises for `flops`, `dram` and `cores`, naming the three that work and pointing at `top_ops` for ranking. The tool schema advertises only the additive three, so a client sees the constraint before it calls.

`top_ops` by those metrics is unaffected — it ranks and simply omits a total.

`b8c31af1`.

backend/ttnn_visualizer/agent/server.py
aidemsined: Agreed, and it is a real incompatibility rather than a missing nicety — a client using the standard liveness check would read `-32601` as a dead server and restart it.

`ping` now returns `result: {}`, placed before the unknown-method fallback. Verified over stdio: `{"jsonrpc":"2.0","id":1,"method":"ping"}` → `{"result":{}}`.

`b8c31af1`.

backend/ttnn_visualizer/csv_queries.py
aidemsined: Correction: the commit is `166aafd0`, not the sha I quoted above — I wrote it before committing.

backend/ttnn_visualizer/agent/tools.py
aidemsined: Correction: the commit is `166aafd0`, not the sha I quoted above — I wrote it before committing.

backend/ttnn_visualizer/agent/server.py
aidemsined: Correction: the commit is `166aafd0`, not the sha I quoted above — I wrote it before committing.

backend/ttnn_visualizer/agent/tools.py
dcblundell: The response logic is a good partial fix, but the caveat remains dormant on this branch because generated rows still lack `sub_device_id`. Please keep this thread open until #1994 lands, or explicitly track the dependency before treating the partition warning as complete.

backend/ttnn_visualizer/csv_queries.py
dcblundell: `usecols` addresses the avoidable parsing cost, and exposing unmatched counts improves honesty. The `open_starts` map is still unbounded for malformed/truncated input, though, so the resource-bound concern remains. Please keep this open unless the intended design explicitly accepts unbounded state for trusted local stdio inputs.

aidemsined: Fair — "visible" is not "bounded", and the concern was about the bound. Now bounded, in `7506cc63`.

`MAX_OPEN_ZONE_STARTS = 200_000`. The sizing: a well-formed capture can only hold one open start per `(slot, core, RISC, nesting level)` at a time, so the widest local shape — 32 slots × 130 cores × 5 RISCs — tops out around 21k. A malformed one can open a zone on every row and close none, which on that same file is 724k. Ten times the legitimate worst case caps the map at roughly 40 MB and still cannot be reached by a real log.

Past the ceiling, starts are counted rather than kept, and the count comes back as `dropped_starts` in the `pairing` dict and is appended to `zone_timings`' caveat — worded as "a well-formed capture does not reach this, treat this log as malformed", because unlike unmatched starts it is not a normal truncation. A bounded read must not be indistinguishable from a complete one.

Two tests in `TestOpenZoneCeiling` monkeypatch the ceiling down to 2 and 1: the first asserts 5 never-closed starts give `unmatched_starts=2`, `dropped_starts=3`; the second asserts the caveat text reaches the tool response.

Leaving this open as you asked — the design does not accept unbounded state, so if the sizing argument has a hole in it I would rather hear it here.

backend/ttnn_visualizer/agent/tools.py
aidemsined: Agreed on both counts — the caveat is dormant, and "we'll remember when #1994 lands" is not tracking. `7506cc63` makes the dependency a test.

`TestPartitionCaveatDependency` asserts the dormancy directly: it reads the column table off `OpsPerformanceReportQueries` and asserts `sub_device_id` is *not* in it. Today that passes. The day #1994 parses the column, it fails, and the failure message says what to do — verify `top_ops` and `diff_reports` against a real partitioned report, drop the dormancy note from `docs/src/agent-tools.md`, and delete the test.

It reads the attribute via `getattr(..., "REPORT_COLUMN_HEADERS", ...REPORT_COLUMNS)` because #1994 both renames the table and adds the field, so keying on either name alone would make the tripwire fire (or not) for the wrong reason. I checked both directions: it passes on this branch, and injecting `sub_device_id` into `REPORT_COLUMNS` makes it fail with the intended message.

Keeping the thread open as you asked. On the underlying question — whether the dormant caveat should ship at all before #1994 — I lean yes: the code path is right and tested against mocked rows, and shipping it means #1994 lands complete rather than needing a follow-up. But it is your call, and if you would rather it came out until the column exists, say so and I will pull it.

backend/ttnn_visualizer/tests/test_agent_tools.py
```diff
@@ -0,0 +1,750 @@
...
+    def test_the_caveat_is_dormant_until_the_column_is_parsed(self):
+        from ttnn_visualizer.csv_queries import OpsPerformanceReportQueries
+
+        # `REPORT_COLUMNS` on `dev`; #1994 renames it to `REPORT_COLUMN_HEADERS`
+        # and makes it a mapping, so both shapes are read rather than assuming
+        # either — that rename is part of the change this watches for.
+        columns = getattr(
+            OpsPerformanceReportQueries,
+            "REPORT_COLUMN_HEADERS",
+            OpsPerformanceReportQueries.REPORT_COLUMNS,
```
smountenay-tt: **Release blocker:** #1994 already merged into `dev`, so this tripwire is firing in the PR's projected merge and Backend CI is red (1316 passed, 1 failed). The immediate `AttributeError` occurs because Python evaluates the `getattr` default eagerly and `REPORT_COLUMNS` no longer exists; after fixing that, the assertion must still fail because `REPORT_COLUMN_HEADERS` now contains `sub_device_id`. Please integrate current `dev`, remove the obsolete dormancy test/notes, and verify the now-live partition caveat against the landed parser.

smountenay-tt (commented):
One release blocker remains on the current head:

backend/ttnn_visualizer/tests/test_agent_tools.py
aidemsined: Right on all three, and the `getattr` mistake is worth naming: Python evaluates the default eagerly, so `OpsPerformanceReportQueries.REPORT_COLUMNS` was resolved *before* `getattr` could return the attribute that does exist. Locally that read as a pass; in the projected merge it raised. The construct I reached for to survive the rename is the one that guaranteed I could not see it.

Fixed in `38fee6cd`, by deleting the test rather than repairing it — the column landed, so the tripwire has done its job.

- Merged current `dev` (`47a6530c`), synced the venv to the pinned tt-perf-report 1.3.0.
- Removed the dormancy claims from `docs/src/agent-tools.md`, `_partitioned_caveat`'s docstring, the mocked test's docstring, and the PR body.
- Verified the caveat against the landed parser instead of asserting it works. `TestPartitionCaveatEndToEnd` builds captures from the smoke fixture carrying a real `SUB DEVICE ID` column and runs them through the real report generator: rows come back with `sub_device_id` `{"0", "1"}`, `top_ops` and `diff_reports` caveat the partitioned capture, and the single-sub-device capture stays uncaveated.

That last one is the coverage that was actually missing. The mocked tests patch `_generate_canonical_report`, which is exactly the parser link #1994 changed — a rename either side would have left them green and every real report uncaveated. Mutation check: deleting the `sub_device_id` entry from `REPORT_COLUMN_HEADERS` kills three of the four new tests and leaves all the mocked ones passing.

1320 backend tests pass locally against merged `dev`.

backend/ttnn_visualizer/agent/tools.py
aidemsined: Update, and it resolves the open question rather than deferring it: #1994 had already merged into `dev` when I posted the tripwire — Scott caught it on the red Backend CI. So the dependency you asked me to track has landed, and the caveat is live.

In `38fee6cd`: merged current `dev`, synced to tt-perf-report 1.3.0, deleted the tripwire, and removed the dormancy notes from the docs, the docstring and the PR body.

The caveat is now verified against the landed parser rather than only against mocked rows — `TestPartitionCaveatEndToEnd` builds captures carrying a real `SUB DEVICE ID` column, runs them through the real report generator, and asserts `top_ops` and `diff_reports` caveat a two-sub-device capture while a single-sub-device one stays clean. Removing `sub_device_id` from `REPORT_COLUMN_HEADERS` kills three of those four and leaves every mocked test green, which is precisely the link the mocks could not see.

So the "should the dormant caveat ship" question is moot — it is not dormant. Happy for you to close this thread whenever you have looked at the new tests.

backend/ttnn_visualizer/tests/test_agent_tools.py
aidemsined: Follow-up, because "verify against the landed parser" deserved better evidence than a fixture I built the column for. I found a genuinely partitioned capture locally — an MoE run on a `MeshShape([8; 4])`, 808 raw rows — and ran the tools against it unmodified:

```
parsed sub_device_id: ['', '0', '1']
rows after canonical projection: 33
caveat: This report spans 2 sub-devices (0, 1). Ops on different sub-devices
        can run concurrently, so summed device time, total percentages and
        op-to-op gaps overstate elapsed time. …
device_time_total: 16335.306
```

Two things worth noting from it. 616 of the 808 rows carry a blank `SUB DEVICE ID`, and the caveat says "2 sub-devices", not 3 — blank is excluded, which matches tt-perf-report's own `count_sub_devices` treating an empty value as "no subdevice reported" rather than as a distinct one. And a real 16.3 ms total is exactly the number the caveat exists for: without it an agent reads that as elapsed time for a run whose two sub-devices overlapped.

The capture is not checked in, so the committed tests still synthesise the column onto the smoke fixture. This was the manual check behind them.

smountenay-tt (approved):

PR #2000, opened 2026-09-10
Compare against the operation range's lower bound in the Graph view

### LINKED ISSUE #1999: Graph view ignores the range's lower bound, and renders nothing when operation ids start at 0
The Graph view filters the operation list with

```js
operationList?.filter((op) => selectedOperationRange[0] && op.id <= selectedOperationRange[1])
```

(`GraphView.tsx:39`). The range *start* is used as a truthiness guard rather than compared against, so the lower bound is never applied. `OperationList.tsx:72` spells the same filter correctly — `op.id >= selectedOperationRange[0] && op.id <= selectedOperationRange[1]` — which is what makes this a dropped `op.id >=` rather than a deliberate choice.

Two consequences, and the second is what surfaced it.

**The lower handle of the Range control does nothing in the Graph view.** On `bge_m3` with the range set to [500, 600]:

```
old predicate kept: 600 ops, ids 1 -> 600
new predicate kept: 101 ops, ids 500 -> 600
```

Every report is affected. It is invisible because the upper bound *is* applied, so the graph does visibly narrow when you drag the right-hand handle.

**A report whose operation ids start at 0 renders nothing at all.** `selectedOperationRange[0]` is then `0`, the predicate is falsy for every op, the filtered list is empty, and `GraphView` treats `length === 0` as still-loading — so the view sits on its spinner forever with no error. Reproduced against a 1184-op capture with ids 0-1183 (`graph_report/visualizer_db`, a device-op-level graph capture); the operations API returns 200 with all 1184 rows, so the data is present and the loss is entirely client-side.

Whether that capture is a valid report is a separate question, but the failure mode is worth closing regardless: an empty filter result is indistinguishable from a pending fetch, so any future cause of an empty list produces the same silent spinner.

The same "an operation id of 0 is legitimate but falsy" hazard exists in three other places, not fixed here:

- `OperationDetailsComponent.tsx:107` — `lateDeallocationsByOperation.get(operation?.id || -1)` looks up key `-1` for op 0.
- `RangeSlider.tsx:289` and `:304` — `parseInt(value, 10) || opMin || …` discards a typed `0`.
- `RangeSlider.tsx:163-171` — `const updatedMin = matchMin || (…)`, on the perf-to-op sync path, where `matchMin` is an `opId`. Same hazard, found during review of #2000.

Separately, and adjacent: those same range inputs write the atom with no clamp and no `min <= max` enforcement, so an inverted range is typable. Every view now handles it honestly — `GraphView` draws an empty graph as of #2000, the lists show nothing — but `clampSelectionToRange` in `functions/perfRangeSelection.ts` already does this job for the perf side and should do it here too.

(The `parseInt(...) || defaultNumberCores` uses in `DeviceOperationsFullRender` and `OperationDetails` are correct — a zero core count is not meaningful.)

aidemsined:
## Summary

The Graph view filtered the operation list on `selectedOperationRange[0] && op.id <= selectedOperationRange[1]`, using the range's *start* as a truthiness guard rather than comparing against it. The lower bound was therefore never applied, and a report whose operation ids start at `0` filtered to nothing and never rendered. `OperationList.tsx:72` spells the same filter with both comparisons, which is what identifies this as a dropped `op.id >=`. This was never a regression: `git log -L36,44:src/routes/GraphView.tsx` traces the predicate to `3153d506` (Feb 2025), so the lower bound has never been applied.

Closes #1999

### Frontend

- `src/functions/filterByOperationRange.ts` — new. The predicate, shared. It existed verbatim in three places and this bug was the copy that drifted, so a fourth should not be possible. `TensorList` keeps its own — it tests producer and consumer arrays, not an id.
- `src/routes/GraphView.tsx` — compares `op.id` against both ends, and no longer reads an empty list as a pending fetch.
- `src/components/OperationList.tsx`, `src/hooks/useAPI.tsx` — call the shared helper.
- `tests/GraphViewOperationRange.spec.tsx` — new. Pins both bounds, the 0-based case, an empty range, and the unfiltered pass-through.

## Behaviour

Two symptoms, one cause.

**The Range control's lower handle moved nothing in this view.** On `bge_m3` with the range set to `[500, 600]`:

| | ops kept | ids |
| --- | --- | --- |
| before | 600 | 1 – 600 |
| after | 101 | 500 – 600 |

Every report is affected. It stayed hidden because the *upper* bound was applied, so the graph does visibly narrow when the right-hand handle moves.

**A report with 0-based operation ids rendered nothing.** `selectedOperationRange[0]` is then `0`, the guard is falsy for every op, and the empty filtered list reads as still-loading — so the view sits on its spinner indefinitely with the operations already fetched (`/api/operations` returns 200 with every row). Reproduced against a 1184-op device-op-level capture with ids 0–1183, which now renders.

## The empty-list dead end

Comparing both bounds made a previously unreachable state reachable, so it is closed here rather than deferred. The footer's min input writes the range atom unclamped, so a min above the max selects no operations — and `GraphView` read an empty list as a pending fetch, giving a spinner that never resolved with the data already fetched. While the lower bound was ignored, that input could not produce it.

The list is now rendered whenever the fetch is done, so a range selecting nothing draws an empty graph. `OperationGraphReactFlow` only maps and iterates `operationList`, with no non-empty assumption.

The range inputs are still unclamped, so `min > max` remains typable — honest now in every view, but worth clamping (`clampSelectionToRange` already does this for the perf side). Listed on #1999 with the remaining instances of the "an operation id of `0` is legitimate but falsy" hazard: `OperationDetailsComponent.tsx:107`, `RangeSlider.tsx:163`, `:289` and `:304`.

## Test plan

- [x] Frontend: `pnpm vitest run` — 2646 passing, 225 files
- [x] The two new range cases fail against the previous predicate and pass with it fixed
- [x] `tsc --noEmit`, `eslint`, `prettier --check` clean on touched files
- [x] Manual: load a capture whose operation ids start at 0, open Graph — renders, with repeat detection working; then drag the Range lower handle on a normal report and confirm the graph narrows from the left
- [x] Manual: type a min above the max in the footer — an empty graph, not a spinner (this reproduced the spinner before the empty-list change)

### DIFF
--- a/src/routes/GraphView.tsx
+++ b/src/routes/GraphView.tsx
@@ -10,6 +10,7 @@ import { useAtomValue } from 'jotai';
 import { useGetDeviceOperationListPerf, useLinkedPerformanceReport, useOperationsList } from '../hooks/useAPI';
 import OperationGraph from '../components/operation-graph/OperationGraphReactFlow';
 import LoadingSpinner from '../components/LoadingSpinner';
+import { filterByOperationRange } from '../functions/filterByOperationRange';
 import useClearSelectedBuffer from '../hooks/useClearSelectedBuffer';
 import { activePerformanceReportFolderNameAtom, selectedOperationRangeAtom } from '../store/app';
 import { PerfOverlaySource } from '../functions/perfOverlay';
@@ -35,10 +36,8 @@ const GraphView = () => {
 
     const filteredOperationList = useMemo(
         () =>
-            selectedOperationRange
-                ? operationList?.filter(
-                      (op) => op.id >= selectedOperationRange[0] && op.id <= selectedOperationRange[1],
-                  )
+            selectedOperationRange && operationList
+                ? filterByOperationRange(operationList, selectedOperationRange)
                 : operationList,
         [operationList, selectedOperationRange],
     );
@@ -81,7 +80,9 @@ const GraphView = () => {
         <div className='data-padding'>
             <Helmet title='GraphTree' />
 
-            {isLoading || filteredOperationList === undefined || filteredOperationList.length === 0 ? (
+            {/* An empty list is a range that selected nothing, not a pending fetch: reading
+                it as loading left a spinner that never resolved. #1999 */}
+            {isLoading || filteredOperationList === undefined ? (
                 <div className='graph-tree-loader'>
                     <LoadingSpinner />
                 </div>
--- a/src/routes/GraphView.tsx
+++ b/src/routes/GraphView.tsx
@@ -36,7 +36,9 @@ const GraphView = () => {
     const filteredOperationList = useMemo(
         () =>
             selectedOperationRange
-                ? operationList?.filter((op) => selectedOperationRange[0] && op.id <= selectedOperationRange[1])
+                ? operationList?.filter(
+                      (op) => op.id >= selectedOperationRange[0] && op.id <= selectedOperationRange[1],
+                  )
                 : operationList,
         [operationList, selectedOperationRange],
     );
--- a/tests/GraphViewOperationRange.spec.tsx
+++ b/tests/GraphViewOperationRange.spec.tsx
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+import '@testing-library/jest-dom/vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { cleanup, render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router';
+import { Provider, createStore } from 'jotai';
+import type { NumberRange } from '@blueprintjs/core';
+
+import type { OperationDescription } from '../src/model/APIData';
+
+// The range filter is route-level and never reaches the DOM, so the graph is stubbed
+// to record the list the route actually handed it.
+const { mockUseOperationsList, mockUseLinkedPerformanceReport, mockUseMatchedPerfOps, graphProps } = vi.hoisted(() => ({
+    mockUseOperationsList: vi.fn(),
+    mockUseLinkedPerformanceReport: vi.fn(),
+    mockUseMatchedPerfOps: vi.fn(),
+    graphProps: [] as { operationList: OperationDescription[] }[],
+}));
+
+vi.mock('react-helmet-async', () => ({ Helmet: () => null }));
+vi.mock('../src/hooks/useClearSelectedBuffer', () => ({ default: () => {} }));
+vi.mock('../src/hooks/useAPI', () => ({
+    useOperationsList: () => mockUseOperationsList(),
+    useLinkedPerformanceReport: () => mockUseLinkedPerformanceReport(),
+    useGetDeviceOperationListPerf: () => mockUseMatchedPerfOps(),
+}));
+vi.mock('../src/components/operation-graph/OperationGraphReactFlow', () => ({
+    default: (props: { operationList: OperationDescription[] }) => {
+        graphProps.push(props);
+        return <div data-testid='op-graph' />;
+    },
+}));
+
+/* eslint-disable import/first */
+import GraphView from '../src/routes/GraphView';
+import { selectedOperationRangeAtom } from '../src/store/app';
+/* eslint-enable import/first */
+
+const operationsFrom = (firstId: number, count: number) =>
+    Array.from({ length: count }, (_, index) => ({
+        id: firstId + index,
+        name: 'ttnn.matmul',
+    })) as unknown as OperationDescription[];
+
+const renderRoute = (operations: OperationDescription[], range: NumberRange | null) => {
+    const store = createStore();
+    store.set(selectedOperationRangeAtom, range);
+    mockUseOperationsList.mockReturnValue({ data: operations, isLoading: false });
+    mockUseLinkedPerformanceReport.mockReturnValue({ data: undefined });
+    mockUseMatchedPerfOps.mockReturnValue([]);
+
+    render(
+        <Provider store={store}>
+            <MemoryRouter>
+                <GraphView />
+            </MemoryRouter>
+        </Provider>,
+    );
+
+    return graphProps[graphProps.length - 1];
+};
+
+afterEach(() => {
+    cleanup();
+    graphProps.length = 0;
+    vi.clearAllMocks();
+});
+
+describe('GraphView operation range', () => {
+    it('honours the lower bound as well as the upper one', () => {
+        // The start used to be read as a truthiness guard rather than compared against,
+        // so the left-hand handle of the Range control moved nothing. #1999
+        const graph = renderRoute(operationsFrom(1, 1000), [500, 600]);
+
+        expect(graph.operationList).toHaveLength(101);
+        expect(graph.operationList[0].id).toBe(500);
+        expect(graph.operationList[graph.operationList.length - 1].id).toBe(600);
+    });
+
+    it('renders a report whose operation ids start at 0', () => {
+        // A start of 0 made the old guard falsy for every op, so the list came back empty
+        // and the route read that as still-loading: a spinner that never resolved. #1999
+        const graph = renderRoute(operationsFrom(0, 1184), [0, 1183]);
+
+        expect(screen.getByTestId('op-graph')).toBeInTheDocument();
+        expect(graph.operationList).toHaveLength(1184);
+    });
+
+    it('passes the whole list through when no range is selected', () => {
+        const graph = renderRoute(operationsFrom(1, 10), null);
+
+        expect(graph.operationList).toHaveLength(10);
+    });
+});
--- a/tests/GraphViewOperationRange.spec.tsx
+++ b/tests/GraphViewOperationRange.spec.tsx
@@ -10,6 +10,8 @@ import { Provider, createStore } from 'jotai';
 import type { NumberRange } from '@blueprintjs/core';
 
 import type { OperationDescription } from '../src/model/APIData';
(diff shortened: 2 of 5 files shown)


src/routes/GraphView.tsx
```diff
@@ -36,7 +36,9 @@ const GraphView = () => {
     const filteredOperationList = useMemo(
         () =>
             selectedOperationRange
-                ? operationList?.filter((op) => selectedOperationRange[0] && op.id <= selectedOperationRange[1])
+                ? operationList?.filter(
+                      (op) => op.id >= selectedOperationRange[0] && op.id <= selectedOperationRange[1],
```
smountenay-tt: **The deferred "empty list reads as loading" limitation is now reachable through the very control this fix repairs.**

Deferring the empty state is fair — it does need a design decision. But the *reachability* changed underneath that deferral, and I think that is worth resolving here rather than on the follow-up.

`RangeSlider.tsx:288`/`:303` write `setSelectedOperationRange` straight from the footer text inputs, with no clamp and no `min <= max` enforcement:

```ts
parseInt(value, 10) || opMin || selectedOperationRange[0],
selectedOperationRange[1],
```

Type a min above the current max and this predicate now keeps nothing, and `GraphView.tsx:84` reads `length === 0` as loading — a spinner that never resolves, with the data already fetched. **Before this change the lower bound was ignored, so that input could not empty the list.** The dead end is new.

Worth noting the slider *handles* can't do this: `merge_rank_filter`'s docstring (`backend/ttnn_visualizer/queries.py:265`) records that ids restart at 1 per rank, so ids are contiguous within a rank and any handle selection contains at least one id. It is specifically the unclamped text inputs.

Either fix closes it:
- clamp the inputs (there is already `clampSelectionToRange` in `functions/perfRangeSelection.ts` doing exactly this for the perf side), or
- render the graph when `!isLoading && filteredOperationList.length === 0`, so an empty range is honest rather than a spinner.

Traced in code, not driven through the UI — worth a quick manual confirm before deciding.

src/routes/GraphView.tsx
```diff
@@ -36,7 +36,9 @@ const GraphView = () => {
     const filteredOperationList = useMemo(
         () =>
             selectedOperationRange
-                ? operationList?.filter((op) => selectedOperationRange[0] && op.id <= selectedOperationRange[1])
+                ? operationList?.filter(
```
smountenay-tt: Optional, but it is the root-cause fix rather than the instance fix: this predicate now exists verbatim in three places — here, `OperationList.tsx:72`, and `useBuffers` at `useAPI.tsx:1139` (plus a variant in `TensorList.tsx:79-81`).

This bug *is* the copy that diverged. A `filterByOperationRange(list, range)` helper in `src/functions/` would make the fourth copy impossible, and all three call sites collapse to a one-liner.

tests/GraphViewOperationRange.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+    default: (props: { operationList: OperationDescription[] }) => {
+        graphProps.push(props);
+        return <div data-testid='op-graph' />;
+    },
+}));
+
+/* eslint-disable import/first */
+import GraphView from '../src/routes/GraphView';
+import { selectedOperationRangeAtom } from '../src/store/app';
+/* eslint-enable import/first */
```
smountenay-tt: This suppression is both unnecessary and undocumented — AGENTS.md ("Lint discipline") requires an explanatory comment on the same or preceding line for every suppression.

It can just go: vitest hoists `vi.mock` above the imports, so these two do not need to sit below the mocks. I verified by moving them up with the rest of the imports — `eslint` clean, all 3 tests still pass.

tests/GraphViewOperationRange.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    render(
+        <Provider store={store}>
+            <MemoryRouter>
+                <GraphView />
+            </MemoryRouter>
+        </Provider>,
+    );
+
+    return graphProps[graphProps.length - 1];
```
smountenay-tt, line 60: Nit: this is `undefined` whenever the route renders the spinner instead of the graph, so a future regression surfaces as `TypeError: Cannot read properties of undefined (reading 'operationList')` rather than a readable assertion — which is a shame, since the spinner path is exactly what the second test exists to catch.

An explicit `expect(graph).toBeDefined()` here, or throwing a named error, would keep the failure legible.

smountenay-tt (commented):
## Review

Fix is correct and matches both siblings (`OperationList.tsx:72`, `useAPI.tsx:1139`). Verification below, then two things worth a decision and some minors inline.

### Verified locally at 3826bbb

| Claim | Result |
|---|---|
| Full suite | 225 files, 2645 tests pass — matches the PR |
| New tests fail against the old predicate | Reverted the predicate: 2 failed / 1 passed |
| `tsc --noEmit`, `eslint`, `prettier --check` on touched files | Clean |
| SPDX year on the new file | 2026, correct |

One addition for the description: this was **never a regression**. `git log -L36,44:src/routes/GraphView.tsx` shows the predicate has been wrong since `3153d506` introduced it — the lower bound has never been applied.

### A fourth instance of the same 0-is-falsy hazard, not listed on #1999

`RangeSlider.tsx:163-171` (off-diff, so no inline anchor):

```ts
const updatedMin = matchMin || (…fallback…);   // matchMin is an opId
```

Same "an operation id of 0 is legitimate but falsy" hazard, on the perf→op sync path. #1999 lists only `:289`/`:304`. Given this PR's whole point is making 0-based reports render, it belongs on that issue alongside the other three.

### Verdict

Good to merge from my side. The only thing I'd want a decision on first is the inline note on the range predicate: the deferral of the empty-list-reads-as-loading limitation is reasonable, but this change moves it from theoretical to user-reachable, so it deserves either a one-line guard or an explicit "we accept that" on #1999.

smountenay-tt (approved):

src/routes/GraphView.tsx
aidemsined: You're right, and this is the most useful thing anyone has said on either of these PRs: the deferral was fine but the *reachability* changed underneath it, and I didn't notice because I only reasoned about the predicate.

Confirmed through the UI, as you suggested — not just traced. On the 0-based capture (range `[0, 1183]`) I typed `2000` into the footer min input: the graph went straight back to the endless spinner, data already fetched. Before this PR that input could not do it, because the lower bound was discarded. So the dead end is new, and it's mine.

Fixed in `94248a4f` by the second option — the list is rendered whenever the fetch is done, so a range that selects nothing draws an empty graph instead of a spinner. Re-ran the same repro: min `2000` / max `1183` now renders the empty canvas with its toolbar. I checked `OperationGraphReactFlow` only maps and iterates `operationList` with no non-empty assumption, so this is safe rather than merely quiet.

I took that over clamping because it closes the dead end for *every* cause of an empty list, not just this input — which was the actual complaint in the deferred limitation. Clamping is still worth doing on its own merits, and `clampSelectionToRange` is right there, but as a UX fix rather than a dead-end fix: an unclamped min > max is wrong for `OperationList` and `useBuffers` too, they just show an honest empty list rather than lying. I've put it on #1999 with the other `RangeSlider` items instead of widening this PR.

New test covers it: `renders an empty graph when the range selects nothing` with `[2000, 50]`.

src/routes/GraphView.tsx
aidemsined: Taken — done in `94248a4f`. `filterByOperationRange` in `src/functions/`, with `GraphView`, `OperationList:71` and `useBuffers` all calling it.

Your framing is what decided it: this bug *is* the copy that drifted, so leaving three copies in place would be fixing the instance and preserving the mechanism. That is the same argument that landed on the sibling PR #2001 an hour ago — the `::` leaf bug there turned out to be one module's leaf rule drifting from a helper in the same directory that already had it right, and the fix was to unify rather than patch. Two independent instances of the same failure mode in one afternoon is a decent argument for taking the root-cause option both times.

`TensorList:79-81` stays as it is: it tests whether any producer or consumer falls in the range, over two arrays, which is a different predicate rather than a fourth copy of this one. The helper is typed `<T extends { id: number }>` so it does not tempt anyone to force it there.

The doc comment on the helper names the hazard rather than the symptom — "both bounds are compared, never tested for truthiness; an operation id of 0 is legitimate" — so the next person reading it sees why the shape matters.

tests/GraphViewOperationRange.spec.tsx
aidemsined: Fixed in `94248a4f`, and worth recording that your "it can just go" needed the second half of your sentence to actually work.

Dropping the suppression alone did not: `import/first` still fired, because the two imports were still sitting below the `vi.mock` calls. It is the *move* that fixes it, exactly as you described — the imports are now up with the rest and there is no suppression. `eslint` clean, all four tests still pass, and the mocks still bind (they must, or the route would reach the real `useAPI` and the graph stub would never record props).

I had copied the pattern from `GraphViewPerfReportLoaded.spec.tsx`, which carries the same suppression, without checking whether it was load-bearing there or just inherited. AGENTS.md is right that an undocumented suppression is the problem — had it carried a reason I would have had something to check it against.

tests/GraphViewOperationRange.spec.tsx
aidemsined: Agreed, and the irony was the point — the one test whose whole job is catching the spinner would have reported it as `TypeError: Cannot read properties of undefined (reading 'operationList')`.

`94248a4f` adds an `expectGraphRendered` helper that asserts the stub is present and returns it narrowed, so all four cases read through it. A spinner regression now fails as `expected undefined to be defined` on the named case rather than as a property read.

It earns its keep immediately: the new empty-range test asserts `operationList` has length 0, which is indistinguishable from "the graph never rendered" without that check.

PR #2001, opened 2026-09-10
Read op-name leaves through one spelling-agnostic helper

### LINKED ISSUE #1990: Layer detection finds nothing in a report whose op names use `::`

aidemsined:
## Summary

Role detection derived an op name's leaf with a dot-only split, so `ttnn::transformer::scaled_dot_product_attention` came back whole, matched no anchor and no delimiter, and a namespaced report partitioned into nothing — the Layers button read "no layers detected" with nothing to say why. Both spellings reach this code, as `DEALLOCATE_OP_NAME_LIST` already records.

Closes #1990

### Frontend

- `src/components/operation-graph/opGraphOpNames.ts` — new. One spelling-agnostic `shortOperationName`, carrying the two-spellings fact and the caveat that matching a *whole* name is a different question.
- `src/components/operation-graph/opGraphOpRoles.ts` — reads leaves through it instead of its own dot-only split.
- `src/components/operation-graph/opGraphRepeatBlocks.ts` — its private `shortOperationName` was already the correct rule; it now imports the shared one rather than keeping a second copy.
- `tests/opGraphOpRoles.spec.ts` — a re-spell property over the four existing captures, a minimal positive case on the ticket's own exemplar name, and two tests for the new capture.
- `tests/data/opRoles/cpp_namespaced.json` — new fixture, a real device-op-level capture (1184 ops, 860 `::`-named).

`isDeallocate` deliberately still matches the full lowercased name: `Tensor::deallocate` is a different op that shares this leaf.

## Why one helper rather than a fixed split

`opGraphRepeatBlocks` already answered "what is this op name's leaf, either spelling" correctly, in the same directory. Fixing only the role detector would have left three spellings of one rule with three different delimiter sets — and one module's leaf rule drifting from its sibling's is precisely what this bug is an instance of. `stripEnum` in `src/functions/formatting.ts` is a byte-identical implementation of a *different* concept (dtype/enum labels) and is intentionally left alone.

## How it is verified

The property is the load-bearing test: re-spelling each of the four real captures to `::` must produce **identical** groups — same boundaries, roles, confidences and anchor names, not merely the same count. That holds only because no entry in any anchor, family or delimiter table contains a separator, so the property pins that the tables really are namespace-agnostic rather than accidentally dot-shaped.

The fifth capture pins the opposite direction: real C++ device-op names (`Tensor::pad`, `tt::tt_metal::create_device_tensor`, 60 × `Conv2dDeviceOperation`) must claim nothing. That assertion runs over a 32-op window rather than the whole file, because the whole-graph span is rejected on size before the anchor tables are ever read — over the full capture the assertion would pass no matter what the tables contained.

## Known limitations (out of scope)

No local capture has a `::`-spelled *anchor* op — a sweep of 131 report databases found zero. Every `::` name available is device- or tensor-level (`ttnn::deallocate`, `Tensor::cpu`, `tt::tt_metal::*`), so the ticket's exemplar spelling is exercised by the property test and the synthetic positive case rather than by a captured transformer report.

Separately, `src/functions/filterOperations.ts` treats the *presence* of `::` as the discriminator for "not a ttnn python op", which is the opposite reading of the same characters from the one here. Both are correct for their purpose, and nothing in `CONVENTIONS.md` records that op names arrive in two spellings — worth an entry, not made here.

## Test plan

- [x] Frontend: `pnpm vitest run` — 2649 passing, 224 files
- [x] Reverting the split fails 4 of the new cases; widening an anchor to catch `Conv2dDeviceOperation` fails the device-op window test
- [x] `tsc --noEmit`, `eslint`, `prettier --check` clean on touched files
- [x] Manual: opened a `::`-spelled capture in the Graph view and confirmed repeat detection runs and the Layers control behaves

### DIFF
--- a/src/components/operation-graph/opGraphOpNames.ts
+++ b/src/components/operation-graph/opGraphOpNames.ts
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+/**
+ * The leaf of an op name, whichever separator the report spelled it with.
+ *
+ * Reports arrive in both spellings — `DEALLOCATE_OP_NAME_LIST` enumerates
+ * `ttnn.deallocate` and `ttnn::deallocate` for exactly that reason — so a rule that
+ * knows only one of them silently stops matching on half the corpus. #1990
+ *
+ * Not to be confused with matching a *whole* name: `isDeallocate` compares the full
+ * name because `Tensor::deallocate` is a different op that shares this leaf.
+ */
+export const shortOperationName = (name: string): string => {
+    const trimmed = name.trim();
+    if (!trimmed) {
+        return '';
+    }
+    const parts = trimmed.split(/::|\./);
+    return parts[parts.length - 1] || trimmed;
+};
--- a/src/components/operation-graph/opGraphOpRoles.ts
+++ b/src/components/operation-graph/opGraphOpRoles.ts
@@ -11,6 +11,8 @@
  * #1976
  */
 
+import { shortOperationName } from './opGraphOpNames';
+
 /** Roles a span can be identified as. Ordered by classification priority. */
 export enum OpSemanticRole {
     ATTENTION = 'attention',
@@ -53,13 +55,6 @@ export interface OpRoleSourceOperation {
      */
     fusedActivation?: string;
 }
-
-/**
- * Anchors match on the name's leaf: ResNet arrives as
- * `ttnn.experimental.quasar.conv2d`, so an exact-string table misses whole
- * architectures.
- */
-const leafNameOf = (name: string): string => name.slice(name.lastIndexOf('.') + 1);
 
 /**
  * A span this large does not mean the layer is big — it means the partition failed,
@@ -306,7 +301,9 @@ const classifySpan = (anchorLeaves: readonly string[]): SpanClassification | nul
  * delimiters are already unambiguous and give each op exactly one owner.
  */
 export const detectOpRoleGroups = (operations: readonly OpRoleSourceOperation[]): OpRoleGroup[] => {
-    const leaves = operations.map((operation) => leafNameOf(operation.name));
+    // Anchors match the leaf: ResNet arrives as `ttnn.experimental.quasar.conv2d`,
+    // so an exact-string table on the full name misses whole architectures.
+    const leaves = operations.map((operation) => shortOperationName(operation.name));
     const fusedActivations = operations.map((operation) => operation.fusedActivation);
     const delimiters = delimitersFor(leaves);
 
--- a/src/components/operation-graph/opGraphRepeatBlocks.ts
+++ b/src/components/operation-graph/opGraphRepeatBlocks.ts
@@ -4,6 +4,7 @@
 
 /* eslint-disable no-continue -- window scan skips consumed spans instead of nesting four levels */
 
+import { shortOperationName } from './opGraphOpNames';
 import type { OpGraphSourceOperation, RepeatBlockInstance } from './opGraphTypes';
 import { OpGraphBlockKind } from './opGraphTypes';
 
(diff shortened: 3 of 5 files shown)


smountenay-tt (commented):
Frontend-only change (`src/components/operation-graph/**`, `tests/**`) with no API contract touched, so this is a bugs/security/release-blockers pass rather than a design one. The fix is correct, the consolidation onto `opGraphRepeatBlocks`' existing rule is the right direction, and I found nothing blocking.

I checked the two things that could have made this a data-correctness problem rather than a fix, since neither is covered by the tests as written:

**Does the new leaf rule change what existing reports show?** No. Over every op name in all four dot-spelled captures, the old `leafNameOf` (`slice(lastIndexOf('.') + 1)`) and `shortOperationName` produce an identical leaf for every single name — zero divergences. The added `trim()` and the `|| trimmed` fallback only differ on trailing-separator or whitespace-padded names, and no real capture has one. So this is a pure fix, not a fix plus a silent behaviour change on reports that were working.

**Is `cpp_namespaced` inert everywhere, or just in the chosen window?** Inert everywhere. It has no `layer_norm`-family leaf and no `add`/`add_` at all, so `delimitersFor` falls back to residual adds, finds none, and the capture is one span. I ran `detectOpRoleGroups` over every sliding 32-op window (1,153 of them) and over 64/128/256/512-op slices: zero groups anywhere. So `slice(149, 181)` isn't concealing over-claiming elsewhere in the file, and the "widening an anchor to catch `Conv2dDeviceOperation` would claim conv blocks here" guard is doing real work.

Spot-checks on the claims in the description all hold: `isDeallocate` does compare the full lowercased name against `DEALLOCATE_OP_NAME_LIST`, so `Tensor::deallocate` stays excluded; `fusedActivationOf` reads `UnaryOpType::` out of the program config and was already separator-independent, so the fused path needed nothing; and `stripEnum` really is the same implementation of a different concept — leaving it alone is right.

CI locally on the branch: `tsc --noEmit`, `eslint`, `prettier --check`, `check-spdx.mjs`, and the full `vitest run` (2,649 tests, 224 files) are all clean. The fixture needs no SPDX header — `check-spdx.mjs` only inspects `.json` whose path contains `package.json` — and at 106K it's in line with the existing captures.

**Optional, non-blocking:** `CLASSIFIABLE_WINDOW = slice(149, 181)` is load-bearing at exactly 32 ops — `isSpanTooLarge(32, 32)` is `32 > max(32, 8)`, i.e. false, so the span reaches classification, but at 33 it's rejected on size and the test would pass no matter what the anchor tables held. The conv-presence `expect` guards the other half of the setup; an `expect(CLASSIFIABLE_WINDOW).toHaveLength(32)` (or deriving the slice from `MIN_LAYER_SPAN_ALLOWANCE`) would stop it going vacuous if the fixture ever shifts. Same class of problem the second test already calls out about itself.

The `filterOperations.ts` observation in the description is worth the `CONVENTIONS.md` entry you suggest — two spellings of an op name, read as opposite discriminators in two places, is exactly the drift this PR is collapsing.

smountenay-tt (approved):

smountenay-tt (approved):
Approving — findings are in my review above; the one test note is optional and not blocking.

Issue #2002, opened 2026-09-10
Make the hosted SECRET_KEY check entropy-aware, and stretch weak keys with a KDF


smountenay-tt:
## Context

`v0.102.0` added a startup check refusing `SERVER_MODE` unless `SECRET_KEY` is
non-default and meets a minimum length (`_validate_hosted_secret_key`,
introduced in b5010921 "Harden hosted event logging"). The requirement shipped
without a migration path for deployments provisioned before it existed, so it
was not satisfiable at upgrade time and hosted startup failed.

The immediate unblock (#2003) lowered `MIN_HOSTED_SECRET_KEY_BYTES`. That is
deliberately a stopgap: it needed no change outside this repo and preserved
existing session cookies, but it leaves the check weaker than it should be.
This issue tracks doing it properly.

## Two problems with the check as written

1. **It measures UTF-8 byte length, not entropy.** `"x" * 32` passed the old
   floor; a 16-byte CSPRNG key did not. Two emoji satisfy an 8-byte floor while
   seven ASCII characters do not, and `"password"`, `"90909909"` and eight
   spaces all pass it. So it is a floor against unset or placeholder values
   rather than a strength check, at any value.
2. **It only inspects `SECRET_KEY`.** If `SECRET_KEY_FALLBACKS` is ever wired
   up (Flask 3.1.3 is pinned, so it is available), a fallback list containing a
   weak key would pass validation while still verifying forged cookies.

## Why the boundary matters

The hosted deployment has no authentication, and Flask cookies are
self-contained HMAC tokens, so cookie integrity rests entirely on the signing
key. The cookie scopes which reports appear in a caller's listing
(`views.py:1004-1021`, built from `session["instances"]` and
`session["profiler_paths"]`) and mints the event-log partition key
(`event_logging.py:375`).

**Scoping this correctly matters, so that the KDF work is not mis-specified.**
The cookie is *not* the only gate on hosted report access. `with_instance`
(`decorators.py:29-64`) takes `instanceId` from the query string and never
checks that the session owns it — it only appends — and `/operations`
(`views.py:434`), `GET /instance` (`views.py:2497`) and `PUT /instance`
(`views.py:2507`) carry it with no `@local_only`. A caller sending no cookie at
all can therefore create an instance and point it at a report. What a *forged*
cookie additionally yields is the report names in the caller's own listing and a
chosen event-log partition id.

That weakens the case for treating the length floor as load-bearing, and it is
[...]

PR #2003, opened 2026-09-10
Lower the hosted SECRET_KEY floor from 32 to 8 bytes

smountenay-tt:
## Why

`v0.102.0` added a startup check refusing `SERVER_MODE` unless `SECRET_KEY` is
non-default and meets a minimum length (`_validate_hosted_secret_key`, from
b5010921 "Harden hosted event logging").

The requirement was introduced without a migration path for deployments
provisioned before it existed, so it was not satisfiable at upgrade time and
hosted startup failed. That was an oversight in how the change shipped, not a
deliberate configuration choice.

`SECRET_KEY` is provisioned outside this repo, so raising an existing
deployment to meet the old floor requires a cross-team change, and rotating
additionally invalidates every existing session cookie — users lose their
uploaded-report lists. Lowering the floor unblocks the upgrade without either.

## What this is not

This is a stopgap, not a considered security position, and it is recorded as
such at the constant and in every doc that states the number. **#2002 tracks
the real fix.**

The approach recommended there needs no change to any deployment's
configuration: derive the signing key from whatever the operator supplies via a
slow KDF. That adds no entropy, but it makes each brute-force candidate orders
of magnitude more expensive to test, so a shorter key stops being cheap to
attack while the deployment keeps the key it already has.

#2002 also covers two pre-existing weaknesses in the check itself:

- It measures **UTF-8 byte length, not entropy**. `"x" * 32` passed the old
  floor while a 16-byte CSPRNG key would not, and two emoji satisfy an 8-byte
  floor while seven ASCII characters do not.
- It only inspects `SECRET_KEY`, never `SECRET_KEY_FALLBACKS`. Flask 3.1.3 is
  pinned, so if fallbacks are ever wired up, a fallback list containing a weak
  key would pass validation while still verifying forged cookies.

Operators should supply a long random value regardless of the floor; the
guidance and a generator snippet are in `.env.sample`.

The floor is **not** what makes the signed cookie an integrity boundary across
workers and restarts — stability and non-defaultness are. An earlier revision of
this PR's `.env.sample` text conflated the two and also described the cookie as
the sole gate on the hosted report listing; both were wrong and are corrected in
`a502923c`. `with_instance` reads `instanceId` from the query string without
checking the session, and `/operations`, `GET /instance` and `PUT /instance`
carry it with no `@local_only`, so the cookie gates less than that text claimed.

## Changes

| File | Change |
|---|---|
| `backend/ttnn_visualizer/settings.py` | `MIN_HOSTED_SECRET_KEY_BYTES` 32 → 8, with a comment recording why and pointing at #2002 |
| `docs/src/event-logging.md` | Required — `test_event_logging_docs_parity.py:237` pins the literal byte count, so CI fails without it |
| `CONVENTIONS.md` | Prose stated the old value |
| `AGENTS.md` | No longer claims startup "requires a strong `SECRET_KEY`", which a length floor does not establish |
| `.env.sample` | Updated, plus a `token_urlsafe(48)` snippet so the floor is not read as the recommendation |
| `tests/test_settings.py` | New boundary test |

The new test exists because every existing case derives from the constant
(`"x" * MIN_HOSTED_SECRET_KEY_BYTES`), so **none of them failed when the floor
changed**. The boundary is now pinned explicitly, making a future change a
deliberate edit.

Still refused: the development default (`90909`), empty values, `None`, and
anything below the floor.

## Verification

- Black, isort, mypy clean; full backend suite **1322 passed**
- Verified `create_app` boots in hosted posture with `SERVER_MODE=True` and
  `SESSION_COOKIE_SECURE=True`, and that the development default, empty values
  and under-floor values are all still refused

Committed with `--no-verify`: the pre-commit hook deadlocked in pnpm 11.22.0's
SQLite store index (`pnpm run flask:format` triggers a dependency-verification
`pnpm install` first, which wedged with hundreds of open fds on
`store/v11/index.db`, no network activity and no CPU). `flask:format`'s actual
commands — `uv run isort` and `uv run black` — were run directly instead and
produced zero changes.

## Follow-ups tracked in #2002

- A hosted-posture boot check in CI. `smoke-tests.yml` starts the wheel with no
  env, so `SERVER_MODE` is false and the hosted boot path is never exercised
  end to end — this class of failure can ship again today.
- A deploy-time preflight validating the target environment against the
  release's requirements before restarting the service.
- Flagging new hosted startup requirements as breaking changes in release notes.


backend/ttnn_visualizer/tests/test_settings.py
```diff
@@ -90,6 +90,35 @@ def test_server_mode_accepts_a_strong_secret_key():
     )
 
 
+@pytest.mark.parametrize("length, accepted", [(7, False), (8, True)])
+def test_server_mode_secret_key_length_boundary(length, accepted):
+    """Pins the floor with literals rather than deriving it from the constant.
```
aidemsined, line 95: > [!CAUTION]
> This test cannot fail for the reason it was added, and its docstring says otherwise.

Both parameters derive from the constant, so the test asserts that the floor behaves like a floor — true for every possible value of it. I mutated `MIN_HOSTED_SECRET_KEY_BYTES` and ran the suite: `test_server_mode_secret_key_length_boundary` passes at 1, 4, 8, 16, 24 and 32 alike. It fails in exactly the way it accuses the pre-existing cases of failing.

So the safeguard the PR describes does not exist. After this merges, nothing fails if someone drops the floor to 6 or restores it to 32. What actually constrains the value is the pre-existing `"short"` literals at `test_settings.py:77` (floor ≥ 6 only) and `test_event_logging_docs_parity.py:237`, which pins docs/code *parity* rather than the number — it is derived from the constant too, so it surfaces a change as a docs diff but never as a failure.

The docstring at `:97-101` ("Pins the floor itself, which the cases above only track relative to the constant") is therefore inaccurate on arrival, which `CONVENTIONS.md:105` treats as a blocker. The `(MIN, True)` case is also a duplicate of `test_server_mode_accepts_a_strong_secret_key` at `:85-91`.

**Suggested action:** express the boundary in literals — `[(7, False), (8, True)]` — and add `assert MIN_HOSTED_SECRET_KEY_BYTES == 8`. Alternatively drop the test and remove the claim from the PR body and commit message; what it uniquely covers today is the comparison's inclusivity, which is worth one sentence rather than the stated purpose.

.env.sample
```diff
@@ -52,9 +52,12 @@
 # GUNICORN_TIMEOUT=60
 # GUNICORN_WORKER_CLASS=gevent
 # GUNICORN_APP_MODULE=ttnn_visualizer.app:create_app()
-# When SERVER_MODE=true, SECRET_KEY must be stable, non-default, and at least 32 bytes
+# When SERVER_MODE=true, SECRET_KEY must be stable, non-default, and at least 8 bytes
 # so signed session cookies remain an integrity boundary across workers and restarts.
-# An empty value or the development fallback (90909) is refused.
+# An empty value or the development fallback (90909) is refused. The 8-byte minimum is
+# only a floor against an obviously-short key: the cookie is the sole gate on the hosted
+# report listing, so use a long random value, e.g.
```
aidemsined: > [!CAUTION]
> This paragraph now asserts and denies the same thing, and the sentence added to fix the overclaim contains a second one.

The retained clause reads "at least 8 bytes **so** signed session cookies remain an integrity boundary across workers and restarts", and three lines later the new text says the minimum is "only a floor against an obviously-short key". Both cannot hold. What makes the cookie a boundary across workers and restarts is that the key is *stable and non-default* — the length floor contributes nothing to that property. The PR corrects precisely this species of overclaim in `AGENTS.md` while leaving a stronger one intact in a file it edits in the same hunk.

The new sentence also states a threat model that does not match the code: "the cookie is the sole gate on the hosted report listing". It is not. `with_instance` (`decorators.py:29-64`) reads `instanceId` straight from the query string and never checks the session — it only appends to it — and `GET /instance` (`views.py:2497`), `PUT /instance` (`views.py:2507`) and `GET /operations` (`views.py:434`) carry it with no `@local_only`. A hosted client sending no cookie at all can create an instance, point `active_report.profiler_name` at any report directory name and read that report. What a *forged* cookie adds is the report names in the attacker's own listing plus a chosen event-log partition id.

That cuts in favour of merging — the floor is weakening a boundary that gates less than the docs claim — but it should not be written down wrongly, and #2002 will mis-scope the KDF work if it inherits this framing.

**Suggested action:** split the clause so *stable and non-default* carries the integrity-boundary rationale and the byte floor stands alone as an explicitly weak check, and drop or correct "sole gate". The `token_urlsafe(48)` snippet is a good addition and should stay.

CONVENTIONS.md
```diff
@@ -613,7 +613,7 @@ The event list exists to answer Q1–Q5 in #1819, not to accumulate counters. Ev
...
+One hosted session gets one directory; the identifier is not copied into logfmt and the collector must not export directory names or per-user series. Minting it does not make a session permanent, although the existing upload flow can later make the whole cookie permanent for Flask's default 31-day lifetime. `create_app` refuses `SERVER_MODE` unless `SECRET_KEY` is non-default and at least 8 bytes — a floor against an obviously-short key, not a strength check, and a stopgap pending #2002. The writer caps hosted logs at 1,024, serialises cross-worker reservations under `.quota.lock`, limits new logs to 60 per minute across workers, and limits each worker to 120 batches per log per minute. The 10 MiB per-file cap therefore has a bounded aggregate worst case; deployment retention reclaims slots after collection, while edge controls absorb rates above the worker-aware application limits.
```
aidemsined: > [!CAUTION]
> Four copies of the number and five of the caveat, and only one of them is machine-checked.

After this PR the byte count appears in `settings.py:224`, `docs/src/event-logging.md:29`, here, and `.env.sample:55`; the "a floor, not a strength check / stopgap pending #2002" claim appears in all four plus `AGENTS.md:200`. `test_event_logging_docs_parity.py:237` reads only `docs/src/event-logging.md`. No test reads `CONVENTIONS.md`, and the `.env.sample` assertions in `test_settings.py` cover boolean lines and variable names, not byte counts.

So when #2002 lands, CI fails on `docs/` alone and these two go silently stale. That is the exact failure mode this file documents for `parse_bool` at `CONVENTIONS.md:1334-1340` — "a second copy of the token list drifts, and both drift directions fail silently". And the evidence that it already happened is in this diff: `AGENTS.md`'s "requires a strong `SECRET_KEY`" was wrong at 32 too, and it survived because it stated no number for the parity test to catch.

**Suggested action:** de-number this sentence and `.env.sample` the way the PR de-numbers `AGENTS.md`, pointing at `docs/src/event-logging.md` as the canonical statement (`AGENTS.md` already carries a dozen `CONVENTIONS.md#anchor` cross-links, so reference-rather-than-restate is the established shape here). Extending the parity assertion to both files would also work.

backend/ttnn_visualizer/settings.py
```diff
@@ -214,7 +214,14 @@ def __get__(self, instance: object, owner: type) -> bool:
...
 DEFAULT_SECRET_KEY = "90909"
-MIN_HOSTED_SECRET_KEY_BYTES = 32
+# A floor against an obviously-short key, not a strength check: it measures string
+# length rather than entropy. Lowered from 32 because the previous value was introduced
+# without a migration path for existing deployments, whose keys are provisioned outside
+# this repo and cannot be rotated without a cross-team change and a user-visible cookie
+# reset. Deliberately a stopgap: #2002 replaces the length test with a slow-KDF
+# derivation, which makes a short key expensive to brute-force without any deployment
+# having to change the key it already has.
+MIN_HOSTED_SECRET_KEY_BYTES = 8
```
aidemsined, line 221: > [!WARNING]
> The justification for 8 rests on a number the PR never states: how long the deployed key actually is.

The whole argument is that the deployed key cannot be rotated, which is convincing. But without its length a reviewer cannot tell whether 8 was forced or merely convenient — and nothing in the tree obstructs a higher floor. I mutated the constant and ran the suite: at 16 and at 24 the only failure is `test_event_logging_docs_parity.py:237`, i.e. the same one-line docs edit this PR already makes, and `fixture_settings.py:86`'s key is 38 bytes so no fixture breaks either.

The difference is not cosmetic. The check counts bytes of the supplied string, not entropy, so 8 admits `"password"` — and `"        "` and `"90909909"`, both of which I verified pass. Flask signs the session with `URLSafeTimedSerializer` over a fixed `"cookie-session"` salt and itsdangerous defaults to SHA-1 key derivation, so one captured cookie makes a short low-entropy key a commodity offline GPU workload. If the deployed key is 16 or 20 bytes, then 16 unblocks the upgrade identically and leaves a materially better floor in the tree for however long #2002 takes.

**Suggested action:** state the deployed key's length in the PR or on #2002 and set the floor to the highest value that clears it. One question to the deploying team, and it may change the number.

Related, and worth a sentence rather than a change: #2002's acceptance criteria contain no length floor at all, so this stopgap institutionalises an arbitrary threshold across six coupled artifacts that all have to be unwound later. Refusing only the development default and empty values while *warning* below a recommended length would have unblocked on identical terms — no deployment change, no cookie reset. Not asking for that rewrite here, but the PR justifies the value without justifying the check.

backend/ttnn_visualizer/settings.py
```diff
@@ -214,7 +214,14 @@ def __get__(self, instance: object, owner: type) -> bool:
...
 _DEFAULT_SSH_PORT = 22
 DEFAULT_SECRET_KEY = "90909"
-MIN_HOSTED_SECRET_KEY_BYTES = 32
+# A floor against an obviously-short key, not a strength check: it measures string
+# length rather than entropy. Lowered from 32 because the previous value was introduced
+# without a migration path for existing deployments, whose keys are provisioned outside
+# this repo and cannot be rotated without a cross-team change and a user-visible cookie
+# reset. Deliberately a stopgap: #2002 replaces the length test with a slow-KDF
+# derivation, which makes a short key expensive to brute-force without any deployment
+# having to change the key it already has.
```
> The comment misdescribes the check it annotates, and it is long enough to go stale in three ways.

"It measures string length rather than entropy" is not what the code does: `_validate_hosted_secret_key` encodes to UTF-8 and measures bytes (`app.py:125-129`). I verified two emoji — eight bytes, two characters — are accepted at the floor while `"kkkkkkk"` (seven characters) is refused. The constant's name gets this right; the new comment is the thing that is wrong. "byte length rather than entropy" fixes it. Worth correcting in the PR body and #2002 too, which both say "string length, not entropy".

On length: three of the four facts are genuinely non-obvious and earn their place — floor-not-strength, the rotation constraint, and #2002 superseding it. But the deployment history in `:218-221` is already recorded in the commit message, the PR body and #2002's Context section, and the same framing is now restated in four docs. And "#2002 replaces the length test with a slow-KDF derivation" goes stale the day #2002 lands, since its acceptance criteria keep no floor at all.

**Suggested action:** trim to roughly "Measures byte length, not entropy — a floor against an obviously-short key, not a strength check. Stopgap, see #2002." and let the ticket hold the deployment rationale. Worth considering whether the rationale belongs on `_validate_hosted_secret_key` in `app.py:120`, which has no docstring at all: a reader there sees the floor and the error message with no hint it is a stopgap, and `CONVENTIONS.md:625` states the repo's own rule for this ("the batch cap lives in `event_logging.py`, beside the atomicity guarantee it bounds").

aidemsined (commented):
The tradeoff is argued honestly and the direction is defensible: the floor at 32 bought less than it looked like it did (`"x" * 32` always passed), the ticket is filed, and rewriting `AGENTS.md` to stop claiming a "strong" key is exactly the kind of honesty a posture reduction should force. Most of what follows is about the claims around the change rather than the change.

Two things I verified rather than took on trust. The `--no-verify` bypass cost nothing — in a fresh checkout at `524d16a2`, `black --check` (89 files unchanged), `isort --check-only` and `mypy` (89 sources) are all clean. And every "still refused" claim holds at floor 8: `None`, `""`, `"90909"`, `b"90909"`, `"short"`, `"k" * 7`, `0` and `False`.

## Unanchored findings

- **The hosted boot path is still unexercised in CI**, so this class of failure can ship again the day after this merges. `.github/workflows/smoke-tests.yml` sets only `BASE_PATH` and `NPM_CONFIG_PRODUCTION` and starts the wheel bare, so `SERVER_MODE` is false and `_validate_hosted_secret_key` returns at `app.py:121` without ever entering the hosted branch — the PR's own follow-up note is accurate. Deferring the deploy-time preflight to #2002 is fair, but one step exporting `SERVER_MODE=true` with a generated key and asserting the wheel serves would close the exact hole that caused this incident, in the PR that exists because of it. The ingredients are already there: `fixture_settings.py:85-86` boots hosted in tests.

## General observations

- `test_server_mode_accepts_a_strong_secret_key` (`test_settings.py:85`) now asserts that `"x" * 8` is "a strong secret key" — the same overclaim this PR removes from `AGENTS.md`, one file away. `..._accepts_a_secret_key_at_the_floor` says what it checks.
- The development-default check compares the raw value rather than `encoded` (`app.py:130`), so `b"90909"` is refused only because 5 < 8. That margin just went from 27 bytes to 3; comparing `encoded == DEFAULT_SECRET_KEY.encode()` makes it independent of the floor.
- `"        "` (eight spaces) and `"90909909"` both pass. "A floor against an obviously-short key" is fair, but eight spaces is not short — a `.strip()` before measuring closes the most embarrassing case.
- `fixture_settings.py:86` still reads `"test-secret-key-with-at-least-32-bytes"`. The value is 38 bytes so nothing breaks, but it is the last place in the tree naming the old floor.
- The stated reason `SECRET_KEY_FALLBACKS` is dormant is not quite right. Flask 3.1.3 *already* extends the verification key list with fallbacks (`flask/sessions.py:309-312`), so the plumbing exists now rather than "if fallbacks are ever wired up". It is dormant because nothing sets the key and it is not a `DefaultConfig` attribute, so `override_with_env_variables` cannot pick it up either. No code change needed — the sentence in the PR and #2002 just overstates the condition. Incidentally `pyproject.toml` declares `Flask>=3.1.0,<4.0`; only `uv.lock` pins 3.1.3.
- `is_flag_enabled("yes")` is `False` while `local_only` (`decorators.py:165`) and the hosted listing branch (`views.py:1004`) read raw truthiness of the same config value. Not reachable from the environment — `SERVER_MODE` is in `_STRICT_BOOLEANS`, so `SERVER_MODE=yes` is fatal at import — and only reachable via `settings_override`. Pre-existing; noted only because this PR makes the key check the load-bearing gate.
- `AGENTS.md:200`'s "bounded at three layers" now counts a nominal 8-byte floor as a peer of two genuinely enforced layers. "Two enforced layers plus a nominal key floor" carries the same information without the arithmetic doing rhetorical work.
- For #2002's threat model: the value that actually reaches hosted report data is `instanceId`, and `src/libs/axiosInstance.ts:36-40` builds it from four concatenated `Math.random().toString(36)` slices rather than a CSPRNG. Not a finding against this PR, but if #2002 is writing down what gates hosted reports, that is the value doing it.
- Residual blast radius is genuinely bounded and worth recording: the endpoints carrying SSH and local-filesystem material are `@local_only` (`decorators.py:160`) and 403 under `SERVER_MODE`.

backend/ttnn_visualizer/tests/test_settings.py
```diff
@@ -90,6 +90,25 @@ def test_server_mode_accepts_a_strong_secret_key():
     )
 
 
+@pytest.mark.parametrize(
+    "length, accepted",
+    [(MIN_HOSTED_SECRET_KEY_BYTES - 1, False), (MIN_HOSTED_SECRET_KEY_BYTES, True)],
```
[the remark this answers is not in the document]
smountenay-tt: Fixed in `ac95a0c4` — literal `[(7, False), (8, True)]` plus `assert MIN_HOSTED_SECRET_KEY_BYTES == 8`.

Verified the way you'd want: setting the constant to 4 now fails two tests, where previously it failed none.

backend/ttnn_visualizer/tests/test_settings.py
smountenay-tt: Correct on every point, including that the docstring claimed exactly the property the test lacked. Fixed in `ac95a0c4`, taking your first option: literal `[(7, False), (8, True)]` plus `assert MIN_HOSTED_SECRET_KEY_BYTES == 8`.

I confirmed the failure the way you did — mutating the constant to 4 previously produced zero failures and now produces two. Your point that `test_event_logging_docs_parity.py:237` surfaces a change as a docs diff but never as a failure is also right, and it's why the assertion on the constant is there rather than relying on parity.

On the duplicate: `(8, True)` does overlap `test_server_mode_accepts_a_strong_secret_key`. I kept it, because a boundary test that only asserts the refusing side stops reading as a boundary and invites someone to "simplify" the remaining case back to a derived value. Happy to drop it if you'd rather have the deduplication.

backend/ttnn_visualizer/settings.py
smountenay-tt: Right, and the constant's name was the giveaway. Fixed in `a502923c`.

Confirmed before changing it: `"🔑🔑"` is two characters and eight bytes and is accepted, while `"kkkkkkk"` is refused. The comment now says UTF-8 bytes, and I corrected "string length" in the PR body and #2002 as well.

Trimmed as suggested, and took the `app.py` suggestion — `_validate_hosted_secret_key` now carries the rationale, since a reader arriving at the floor and its error message had no hint it was a stopgap. That follows the rule this repo states for the batch cap at `CONVENTIONS.md:625`. The deployment history is gone from the comment; the commit message and #2002 hold it.

.env.sample
smountenay-tt: Both wrong, both fixed in `a502923c`.

The contradiction is exactly as you describe — the retained "at least N bytes **so** cookies remain an integrity boundary" clause asserted what the sentence three lines below denied. Split so that *stable and non-default* carries the integrity-boundary rationale and the byte floor stands alone as an explicitly weak check. The `token_urlsafe(48)` snippet stays.

"Sole gate" was my error and I've removed rather than reworded it. I verified your reading: `/operations` (`views.py:434`), `GET /instance` (`views.py:2497`) and `PUT /instance` (`views.py:2507`) all carry `@with_instance` with no `@local_only`, and `with_instance` only appends to `session["instances"]` without checking ownership.

Your point about #2002 inheriting the framing was the important one — I've rewritten its "why the boundary matters" section to state what the cookie does and does not gate, and added the `instanceId` weakness as tracked work there rather than leaving it implied.

CONVENTIONS.md
smountenay-tt: Agreed, and the `AGENTS.md` evidence is the part that convinced me — it was wrong at the old value too and survived precisely because it stated no number.

Fixed in `a502923c` using your first option: `CONVENTIONS.md` and `.env.sample` now point at `docs/src/event-logging.md` as the canonical statement, and say so, rather than restating the count. That leaves one numbered copy in prose and one in code, both covered by the parity test, so #2002 will fail CI on the file that actually needs editing instead of leaving two behind.

I went with de-numbering over extending the parity assertion because it also removes the caveat duplication, which was five copies to the number's four.

backend/ttnn_visualizer/settings.py
smountenay-tt: You're right that the PR justifies the value without justifying the check, and I've acted on the second half of your comment: #2002's acceptance criteria now include deciding explicitly whether a floor survives the KDF work, and considering your alternative shape — refuse only the development default and empty values, warn below a recommended length. You're correct that it would have unblocked this on identical terms, and it's the better design. I'd rather land that deliberately on #2002 than rewrite the stopgap mid-review.

I'm not going to state the deployed key's length here. This repository is public, and that is a live detail about a deployment whose weakness we're documenting in the same breath — it belongs in a private channel, not an open PR. Happy to confirm it to the deploying team directly.

What I can say: the value was determined by the deployment constraint rather than chosen for convenience, so a higher floor would not have unblocked the upgrade. Your in-tree experiment is accurate but doesn't settle it — the binding constraint is outside the tree, which is why only the docs parity test moves when you mutate the constant. If that's not verifiable enough for a reviewer, the honest fix is your warn-instead-of-refuse shape, which removes the need to defend any particular number.

Your entropy point stands regardless and I've stopped implying otherwise: `"password"`, `"90909909"` and eight spaces all pass this floor — I verified all three — and the docstring now added to `_validate_hosted_secret_key` in `a502923c` says the check catches unset or placeholder keys and nothing more, so it can't be read as establishing strength. The itsdangerous detail (fixed `"cookie-session"` salt, SHA-1 default key derivation) is a good argument for prioritising #2002 over tuning this number, and I've noted it there.

aidemsined (approved):

Issue #2006, opened 2026-09-11
Hosted SECRET_KEY check: default comparison misses the bytes form, and whitespace passes the floor


aidemsined:
Two residual gaps in `_validate_hosted_secret_key` (`backend/ttnn_visualizer/app.py:138-146`), noticed while reviewing #2003 and left out of it deliberately — they were low severity against that PR's purpose, and they survive #2002 either way, since neither is fixed by replacing the length test with a KDF.

**The development-default check compares the raw value, not the encoded one.** `app.py:143` reads `secret_key == DEFAULT_SECRET_KEY`, while the length test just above operates on `encoded`. So a `bytes` key never matches the default branch and is caught only if it is short. Verified against the released code:

```
b"90909"      -> refused      (only because 5 < 8)
b"909090909"  -> ACCEPTED     (the development default, padded)
```

That second case is the point: an obviously-placeholder key boots a hosted deployment. Comparing `encoded == DEFAULT_SECRET_KEY.encode()` makes the default check independent of the floor, which matters more now that the two are 3 bytes apart rather than 27.

**Whitespace passes the floor.** Also verified:

```
"        "  (8 spaces) -> ACCEPTED
"\t\t\t\t\t\t\t\t" (8 tabs)   -> ACCEPTED
```

The floor is documented as catching "an obviously-short key", which is fair, but eight spaces is not short and is not a key. A `.strip()` before measuring closes it.

Neither is reachable by accident from a real deployment — both need someone to set a deliberately silly value — so this is about the check saying what it means rather than a live exposure.

**Also, trivially:** `backend/ttnn_visualizer/tests/fixture_settings.py:86` still reads `"test-secret-key-with-at-least-32-bytes"`. The value is 38 bytes so nothing breaks, but it is the last place in the tree naming the old floor.

Issue #2004, opened 2026-09-10
Prevent new startup requirements from breaking existing deployments (v0.102.0 retrospective)


smountenay-tt:
Retrospective on the `v0.102.0` hosted startup failure. The purpose is to fix
the class of failure, not the individual setting — #2003 unblocked the symptom
and #2002 tracks the specific check.

## What happened

`v0.102.0` introduced a new hard startup requirement on operator-supplied
configuration: `create_app` began refusing to start under `SERVER_MODE` unless
`SECRET_KEY` met new conditions (`_validate_hosted_secret_key`, from b5010921).

Deployments provisioned before that requirement existed did not meet it, so
every gunicorn worker raised at import, the arbiter halted with
`HaltServer: 'Worker failed to boot.'`, and systemd restart-looped ~27 times
before the failure was noticed. There was no state in which the service ran
degraded — it simply never came up.

## Why the tests did not catch it

Worth stating plainly, because the intuitive answer is wrong: **this was not a
coverage gap, and no test could have caught it.**

The check was well covered before release. `tests/test_settings.py` had five
direct tests, including one asserting `create_app` raises this exact
`RuntimeError` in hosted posture. `test_event_logging_docs_parity.py` pinned
the error message against the docs. And the whole backend suite already runs in
hosted posture — `fixture_settings.base_test_settings` sets `SERVER_MODE: True`
for every fixture app.

Every one of those tests passed, before and after the change, because they
supply their own config. The validated value comes from the deployment's
environment, which lives outside this repository. **CI has nothing to validate
against, so the failing input is invisible to it by construction.**

That is the actual lesson: our test suite verifies that validators behave
correctly given an input, and cannot verify that any real deployment supplies a
conforming one. Any change that makes a previously-acceptable operator
configuration fatal is therefore untestable here, and needs a different control.

Two secondary observations from the same incident:

- **Tests that derive from a constant cannot detect a change to that constant.**
  Every case was written as `"x" * MIN_HOSTED_SECRET_KEY_BYTES`, so when #2003
  changed the constant, none of them failed. Thresholds that carry a policy
  decision want at least one test that pins the literal value, so changing it
  is a deliberate edit rather than a silent follow.
- **The requirement was documented, in three files, with rationale.** So this
  was not an unnoticed change slipping through review. What review did not do
  was ask whether any *existing* deployment satisfied the new condition. That
  question is the missing step, not more documentation.

### 1. CI never boots the app in hosted posture

[...]

PR #2009, opened 2026-09-14
Judge the hosted SECRET_KEY default and floor on the same value

smountenay-tt:
Judge the hosted SECRET_KEY default and floor on the same value

## Why

Follow-up to #2003, closing the two residual gaps recorded in #2006. Both are in
`_validate_hosted_secret_key` (`backend/ttnn_visualizer/app.py`), and neither is
fixed by #2002 replacing the length test with a KDF.

**The development-default check compared the raw value, not the trimmed one.** The
comparison read `secret_key == DEFAULT_SECRET_KEY` while the length test just above
operated on `encoded`, so the default branch was unreachable for a `bytes` key —
`b"90909"` was refused for being five bytes, not for being the default. Comparing
the same value both clauses measure makes the default check independent of the
floor, which matters now that the two are 3 bytes apart rather than 27.

**Whitespace passed the floor.** Eight spaces and eight tabs were both accepted by
a check documented as catching "an obviously-short key". Whitespace of that length
is not short, and is not a key. The floor now measures the trimmed length.

Neither was reachable by accident from a real deployment — both need someone to set
a deliberately silly value — so this is about the check saying what it means rather
than a live exposure.

## The floor is unchanged

`MIN_HOSTED_SECRET_KEY_BYTES` is untouched at 8, and `settings.py` is not in this
diff. Over a 4,072-case corpus spanning both key types, every ASCII and non-ASCII
whitespace character, and 200 randomly generated 8-character keys drawn from the
full printable-ASCII alphabet, **no input goes from refused to accepted** — the
change is monotonically stricter. The 7/8 boundary test still pins the floor exactly
where it was.

**The newly-refused class is every key whose trimmed length falls below the floor,
at any raw length** — not, as an earlier revision of this section claimed, only an
8-byte key whose first or last character is a space or tab. 1,361 of the 1,529
newly-refused cases in the corpus are longer than 8 raw bytes (`"  abcdefg  "` at
11, `" " * 20`), and `\n`, `\r`, `\x0b` and `\x0c` trim too, so `"abcdefg\r"` and
`"\x0babcdefg"` are refused as well. Every affected key has a sub-floor core, which
is why the compatibility conclusion survives the correction.

That class is now asserted rather than only described: `"xxxxxxx "` and `" xxxxxxx"`
are the two cases that distinguish `strip` from `lstrip` and `rstrip`.

## Trimming decides admission, not the key

The trim never touches what Flask signs. `" key12345 "` is admitted on its 8-byte
core and signed with all 11, so it and `"key12345"` stay two different signing keys
whose cookies do not interchange. The asymmetry is deliberate — normalising at the
settings boundary would rotate the key for every deployment carrying stray
whitespace — but the refusal message has to not promise otherwise. An operator who
reads "padding is ignored", tidies a trailing newline out of their `.env` and
restarts drops every session and every user's uploaded-report list.

So the message and the user reference both say **"at least 8 bytes excluding
surrounding whitespace"**, the docstring states that the configured value is what
signs, `.env.sample` warns the operator who is about to quote a padded value, and
`test_a_padded_key_is_admitted_on_the_trim_but_signs_with_its_whitespace` pins it.

**An earlier revision of this section claimed `.env` parsing strips the padding. It
does not.** `dotenv_values` preserves whitespace inside quotes, so
`SECRET_KEY="secret  "` yields 8 raw bytes and 6 trimmed; only the unquoted form is
stripped. And the primary path is `os.getenv` at `settings.py:534`, which bypasses
dotenv entirely — a Docker `-e`, a Helm value, or `kubectl create secret
--from-file` (which keeps the trailing newline) all arrive unstripped. That is the
reason the `.env.sample` note and the wording above matter rather than being
belt-and-braces.

## Which whitespace trims

The `str` is trimmed before encoding, so everything `str.isspace` accepts goes —
including the non-breaking and ideographic spaces that reached the floor as
multi-byte UTF-8 and were accepted before. `"\xa0" * 8`, `"　" * 8` and
`"\xa090909\xa0"` are all refused now, the last on both clauses at once.

A `bytes` or `bytearray` key, reachable only through `settings_override`, loses
exactly the six bytes in `b" \t\n\r\x0b\x0c"`, because arbitrary bytes cannot be
decoded to find the rest. The docstring frames that as a consequence of the
ordering rather than a property of trimming, and gives the exact byte set — `\x1c`
through `\x1f` are ASCII and `str.isspace` calls them whitespace, but `bytes.strip`
does not remove them, so "ASCII whitespace only" was imprecise.

**One review case does not hold: a byte-order mark is not whitespace to Python.**
`'﻿'.isspace()` is `False`, so `str.strip()` does not remove it either, and
`"﻿" * 8` and `"﻿90909"` remain accepted on both paths. Recorded in the
docstring so the next reader does not have to retest it.

`bytearray(b"90909")` is also fixed: it is not `bytes`, so it stringified to
`"bytearray(b'90909')"` — 20 bytes — and was accepted, with Flask signing
successfully. `isinstance(secret_key, (bytes, bytearray))` closes it.

## Behaviour

| `SECRET_KEY` | before | after |
|---|---|---|
| `b"90909"` | refused — for its length | refused — as the default |
| `bytearray(b"90909")` | **accepted** | refused |
| `"        "` (8 spaces) | **accepted** | refused |
| `"\t" * 8` | **accepted** | refused |
| `b" " * 8` | **accepted** | refused |
| `"\xa0" * 8` (NBSP) | **accepted** | refused |
| `"　" * 8` (ideographic) | **accepted** | refused |
| `"\xa090909\xa0"` | **accepted** | refused |
| `"xxxxxxx "` / `" xxxxxxx"` | **accepted** | refused |
| `"  " + "x" * 8 + "\n"` | accepted | accepted |
| `"﻿" * 8` (BOM) | accepted | accepted — BOM is not whitespace |
| `(b"\xa0" * 8)` | accepted | accepted — `bytes` path |
| any 8-byte non-whitespace key | accepted | accepted |
| 7-byte key | refused | refused |

Still accepted, and out of scope: `"909090909"` — the issue's "development default,
padded". It is a different value from the default, and refusing it would mean prefix
or substring matching on a five-character placeholder, which would also refuse
random keys that happen to contain `90909`. It clears the floor on its own length in
either form.

**#2006 should not close silently on this PR.** Its first gap named `b"909090909"`,
which is still accepted; deferring the general low-entropy-placeholder problem to
#2002 is the right call, but the report needs to stay tracked there rather than
disappear. Also unchanged and still #2002's: the check measures UTF-8 bytes rather
than entropy, and inspects only `SECRET_KEY`, never `SECRET_KEY_FALLBACKS`.

## Changes

| File | Change |
|---|---|
| `backend/ttnn_visualizer/app.py` | Both clauses judge one trimmed value; `str` trimmed before encoding; `bytearray` accepted as a byte type; message says "excluding surrounding whitespace"; docstring records the placement rationale, the admission-only trim, and the exact byte sets |
| `backend/ttnn_visualizer/tests/test_settings.py` | Refusal cases for the `bytes` and `bytearray` defaults, non-ASCII whitespace, and one-sided padding over a sub-floor core; a floor-patched test isolating the default clause; a test that a padded key is admitted without being normalised; a header stating the literal-vs-derived rule |
| `backend/ttnn_visualizer/tests/test_event_logging_docs_parity.py` | One phrase constant shared by the docs and the refusal message, asserted against both |
| `backend/ttnn_visualizer/tests/fixture_settings.py` | Placeholder promoted to `HOSTED_TEST_SECRET_KEY`, composed from `DEFAULT_SECRET_KEY` and the floor so both properties it needs are named |
| `docs/src/event-logging.md` | States the exclusion and that padding is not removed |
| `.env.sample` | Warns the operator about to quote a padded value |

### Why the docstring carries the placement rationale

`settings.py:219-220` points readers at this function, so that paragraph is where
someone lands, and the architectural answer was unrecorded. It is timing:
`settings.py`'s own cross-key hosted guard, `_refuse_debug_under_server_mode`, runs
inside `Config.__init__`, before `create_app` applies `settings_override` — which is
why `fixture_settings.py:69-73` has to pin `DEBUG` by hand. A settings-side
secret-key validator would inherit that defect, and
`test_create_app_refuses_the_development_secret_in_server_mode`, which reaches this
function *through* `settings_override`, would stop exercising the real path. Without
that sentence the split reads as an accident and the next person tidying `app.py`
moves the rule into `settings.py` and silently weakens it.

`CONVENTIONS.md:616` and `AGENTS.md:200` correctly need no change, because both
defer to `docs/src/event-logging.md` as the canonical copy — the indirection from
#2003 doing its job, noted here so its absence from the diff does not read as an
omission.

### Not done

The validator still accepts keys Flask cannot sign with: `12345678` (int),
`1234567.5`, `["abcdefghij"]` and `object()` clear the floor on their `repr` and
then raise `TypeError: key: expected bytes or bytearray` at the first session write
— a request, not a boot. A type guard would keep that failure at startup, which is
what a boot-time gate is for, but it is a new refusal class rather than one of the
two gaps #2006 reported, and it belongs with #2002's rework of what this function
measures. Filed there rather than widened into this diff.

## Verification

- 11 mutants checked against the new tests, **all dead**: the default clause
  reverted, deleted, and pointed at a bogus literal; `strip` → `lstrip` and
  → `rstrip`; the `str` and `bytes` trims each removed; `bytearray` dropped from the
  `isinstance`; `<` → `<=`; and the trimming clause deleted from the error message
  and from the docs. Before this commit, all but two of those left the suite green.
- 4,072-case corpus: zero inputs go from refused to accepted
- `isort`, `black`, `mypy` clean over `backend/ttnn_visualizer` (89 files)
- Full backend suite: **1322 passed**, 4 skipped


backend/ttnn_visualizer/tests/test_settings.py
```diff
@@ -74,7 +74,22 @@
...
 @pytest.mark.parametrize(
     "secret_key",
-    [None, "", DEFAULT_SECRET_KEY, "short", b"short"],
+    [
+        None,
+        "",
+        DEFAULT_SECRET_KEY,
+        # The ``bytes`` default used to reach only the floor, so it was refused for
+        # being five bytes rather than for being the default. See #2006.
+        DEFAULT_SECRET_KEY.encode("utf-8"),
```
aidemsined, line 83: > [!CAUTION]
> This case cannot distinguish the fix it is attached to, and the comment above it claims that it does.

Reverting `app.py:149` to `secret_key == DEFAULT_SECRET_KEY` leaves the **entire suite green** — 1316 tests, and 198 for `test_settings.py` + the docs-parity spec. So does deleting the clause outright, and so does comparing against a bogus literal. The only mutants that die are the `.strip()` removal and `<` → `<=`.

The arithmetic is why: `DEFAULT_SECRET_KEY` is five bytes and the floor is eight, so any value equal to the encoded default already fails the floor. `b"90909"` was refused before this change and is refused after, and nothing observable changes — only which branch fires. The comment at `:81-83` says the bytes default "used to reach only the floor, so it was refused for being five bytes rather than for being the default", but with the floor at 8 that is still true after the change, so the case pins the outcome rather than the mechanism.

That matters beyond tidiness: the clause is real future-proofing — it decouples the default check from the floor, which is exactly what protects you if #2002 moves or removes the floor — but as shipped it is latent *and* unverified, so a later floor change silently re-opens #2006. That is the failure mode #2004 was filed about.

**Suggested action:** isolate the clause. Monkeypatch `MIN_HOSTED_SECRET_KEY_BYTES` below `len(DEFAULT_SECRET_KEY)` and assert both `"90909"` and `b"90909"` are still refused — that dies on the revert. Then either correct the comment here or let the new test carry the claim.

backend/ttnn_visualizer/app.py
```diff
@@ -140,10 +145,14 @@ def _validate_hosted_secret_key(config: Mapping[str, Any]) -> None:
...
-    if secret_key == DEFAULT_SECRET_KEY or len(encoded) < MIN_HOSTED_SECRET_KEY_BYTES:
+    if (
+        encoded == DEFAULT_SECRET_KEY.encode("utf-8")
+        or len(encoded.strip()) < MIN_HOSTED_SECRET_KEY_BYTES
+    ):
         raise RuntimeError(
             "SERVER_MODE requires SECRET_KEY to contain at least "
-            f"{MIN_HOSTED_SECRET_KEY_BYTES} bytes and not use the development default"
+            f"{MIN_HOSTED_SECRET_KEY_BYTES} bytes once surrounding whitespace is "
+            "trimmed, and not use the development default"
```
aidemsined, line 155: > [!CAUTION]
> "once surrounding whitespace is trimmed" reads as *the app trims your key*, and it does not.

Flask keeps signing with the untrimmed `app.config["SECRET_KEY"]`. I verified `"   abcdefgh"` is judged at 8 bytes and signed with all 11 — so `"key12345"` and `" key12345 "` are two different signing keys whose cookies do not interchange.

The failure mode is the one #2003 reorganised itself to avoid. An operator refused by this message reads it as "padding is ignored", tidies the trailing newline out of their `.env`, and rotates the signing key — dropping every session and every user's uploaded-report list. The message that told them padding did not matter is the thing that cost them the sessions.

The asymmetry itself is right, and I'd keep it: the signed key always has at least as many bytes as the check credited it with, so the floor never over-credits real key material, and normalising at the settings boundary would rotate keys for every deployment with stray whitespace — the same harm.

**Suggested action:** attribute the trimming to the check rather than to the key, in both copies. Something like "at least 8 bytes excluding surrounding whitespace" here and at `docs/src/event-logging.md:29`, plus a clause in the new docstring paragraph saying the trim decides admission only and the configured value is what signs, whitespace included.

backend/ttnn_visualizer/app.py
```diff
@@ -140,10 +145,14 @@ def _validate_hosted_secret_key(config: Mapping[str, Any]) -> None:
         if isinstance(secret_key, bytes)
         else str(secret_key or "").encode("utf-8")
     )
-    if secret_key == DEFAULT_SECRET_KEY or len(encoded) < MIN_HOSTED_SECRET_KEY_BYTES:
+    if (
+        encoded == DEFAULT_SECRET_KEY.encode("utf-8")
+        or len(encoded.strip()) < MIN_HOSTED_SECRET_KEY_BYTES
```
aidemsined, line 150: > [!WARNING]
> Trimming after encoding leaves the development default itself accepted when it is wrapped in non-breaking spaces.

`bytes.strip()` removes only `b" \t\n\r\x0b\x0c"`, so every non-ASCII whitespace run survives and counts toward the floor. Verified accepted, with Flask signing successfully in each case:

```
"\xa0" * 8      (NBSP)              accepted
"　" * 8    (ideographic space) accepted
" " * 8                        accepted
"﻿" * 8    (BOM)               accepted
"\xa090909\xa0"  (the default, NBSP-wrapped)   accepted
"﻿90909"                                  accepted
```

By the docstring's own rationale — "eight spaces are not a short key but are not a key at all" — eight non-breaking spaces fail the same test. And the last two are a plausible copy-paste of the placeholder out of rendered documentation.

This is not inherent to trimming, which is how the new docstring frames it. `str.strip()` *does* remove all of them, and `os.getenv` always returns `str`, so trimming before `.encode()` on the non-`bytes` path refuses all five — and collapses `"\xa090909\xa0"` onto the default, catching it on both clauses at once.

**Suggested action:** strip on the `str` path before encoding, and reword the docstring so the remaining gap reads as a consequence of ordering rather than a property of trimming. While there: `\x1c`–`\x1f` are ASCII and `str.isspace()` calls them whitespace, but `bytes.strip()` does not remove them, so "ASCII whitespace only" is imprecise — the exact set is the six bytes in `b" \t\n\r\x0b\x0c"`.

backend/ttnn_visualizer/tests/test_settings.py
```diff
@@ -74,7 +74,22 @@
...
+        "short",
+        b"short",
+        # Long enough to clear the floor unstripped, which is why these are written
+        # against the constant: the cases stay all-whitespace-over-the-floor whatever
+        # the floor becomes.
+        " " * MIN_HOSTED_SECRET_KEY_BYTES,
+        "\t" * MIN_HOSTED_SECRET_KEY_BYTES,
+        b" " * MIN_HOSTED_SECRET_KEY_BYTES,
```
aidemsined, line 91: > [!WARNING]
> The trim is only pinned as two-sided by accident — the one behaviour change this PR deliberately makes has no test.

`.strip()` → `.lstrip()` passes 198 tests. `.strip()` → `.rstrip()` also passes 198. These refusal cases are whitespace-only runs, which either one-sided strip kills equally, and the acceptance case at `:108-119` has an eight-byte core, so it clears the floor under `lstrip` (9) and `rstrip` (10) alike.

The two cases that discriminate are exactly the "deliberate exception" your behaviour table documents: a seven-byte core with trailing whitespace (`"xxxxxxx "`) must be refused, and the leading twin (`" xxxxxxx"`) for the other direction. Right now the documented exception is the only newly-refused class and nothing asserts it.

**Suggested action:** add both. They also make the exception discoverable from the tests rather than only from the PR body, which matters once the PR body is out of view.

Worth saying that the split of idioms in this file is otherwise right, and I'd leave it: deriving these from the constant keeps them all-whitespace-over-the-floor whatever the floor becomes, while the boundary test at `:122` correctly uses bare literals because its invariant *is* the number. Two different jobs, two correct choices. The one thing missing is a line stating the rule in general form, so someone adding a case has a basis for choosing — the floor's position is pinned with literals in exactly one test, everything else derives.

backend/ttnn_visualizer/app.py
```diff
@@ -127,6 +127,11 @@ def _validate_hosted_secret_key(config: Mapping[str, Any]) -> None:
     catches an unset or placeholder key, and nothing more. Do not read it as
     establishing key strength.
 
+    Both tests run on the encoded form so that a ``bytes`` key is judged the same as
+    the equivalent ``str``, and the floor measures what survives ``bytes.strip()``,
+    because eight spaces are not a short key but are not a key at all. Trimming covers
+    ASCII whitespace only; a run of non-breaking spaces still counts toward the floor.
```
aidemsined, line 133: > [!CAUTION]
> This docstring is now the canonical home of the rule, and it still does not answer the question the split invites: why is the rule here and not beside the constants?

`settings.py:219-220` explicitly points readers at this function ("the rationale lives with the validator that applies it"), so this paragraph is where someone lands — and there is a good architectural answer that goes unrecorded.

The answer is timing. `settings.py`'s own cross-key hosted guard, `_refuse_debug_under_server_mode`, runs inside `Config.__init__`, which happens *before* `create_app` applies `settings_override`. `fixture_settings.py:69-73` and `CONVENTIONS.md:958` both document what that costs: `DEBUG` has to be pinned in the fixture precisely because the settings-side guard has already run. A settings-side secret-key validator would inherit the same defect, and `test_create_app_refuses_the_development_secret_in_server_mode` — which reaches this function *through* `settings_override` — would stop exercising the real path.

So the placement is correct and deliberate, and nothing in the tree says so. Without that sentence the split reads as an accident, and the next person tidying `app.py` moves the rule into `settings.py` and silently weakens it.

**Suggested action:** one sentence here saying the check must run after `settings_override` is applied, which is why it is not a `Config` method. The `bytes.strip()` and non-breaking-space sentences are the strongest lines in the paragraph and should stay; the opener ("Both tests run on the encoded form…") is the one I'd trade for it, since it narrates the two lines immediately below and advertises a parity benefit no current input exercises.

backend/ttnn_visualizer/tests/fixture_settings.py
```diff
@@ -83,7 +83,7 @@ def base_test_settings(tmpdir: str, **overrides: Any) -> Dict[str, Any]:
         "TESTING": True,
         "SQLALCHEMY_DATABASE_URI": f"sqlite:///{Path(tmpdir) / 'app.db'}",
         "SERVER_MODE": True,
-        "SECRET_KEY": "test-secret-key-with-at-least-32-bytes",
+        "SECRET_KEY": "test-secret-key-that-is-not-the-development-default",
```
aidemsined, line 86: > [!WARNING]
> This trades one name-encoded requirement for another, and the fixture needs two properties while the name states one.

The value has to be non-default **and** clear the floor after trimming. The new name covers the first only, so a future floor above 50 bytes breaks every fixture-backed test with no hint in the value about why.

Its own neighbours in this dict argue the other way: `:96-99` says "The shipped default rather than a literal, so it follows a change to it", and `:104-105` says "Derived from `DefaultConfig` rather than written as a literal so it follows a change to the shipped defaults". `CONVENTIONS.md:768` says the same for literals shared between test setup and production code.

And the old value is the proof of the failure mode — #2006 called it the last place in the tree naming the superseded 32-byte floor, and it went stale through #2003 with nothing failing.

**Suggested action:** promote it to a module-level constant composed from `DEFAULT_SECRET_KEY` and `MIN_HOSTED_SECRET_KEY_BYTES`, naming both properties once.

docs/src/event-logging.md
```diff
@@ -26,7 +26,7 @@ An installation running in `SERVER_MODE` stores one log per anonymous browser se
...
+`SERVER_MODE` refuses to start unless `SECRET_KEY` is non-default and at least 8 bytes once surrounding whitespace is trimmed. That is a floor against an obviously-short key rather than a strength check, so operators should still supply a long random value. The application accepts at most 1,024 hosted event logs, at most 60 new event logs per minute across workers, and at most 120 batches per event log per minute in each worker. Combined with the 10 MiB per-file cap, the log-count limit bounds aggregate event-file storage to approximately 10 GiB. Hosted deployments must still mount writable persistent storage at `/data/usage`, remove collected session logs to reclaim quota, and apply edge-level request controls appropriate to their worker count and expected traffic.
```
aidemsined, line 29: > [!WARNING]
> The trimming semantics now exist as two hand-maintained prose copies with nothing coupling them to the code.

`test_event_logging_docs_parity.py:237` asserts only `f"at least {MIN_HOSTED_SECRET_KEY_BYTES} bytes" in source`. Your insertion lands after that substring so the assertion still passes — I confirmed, 15/15 — but it means deleting "once surrounding whitespace is trimmed" from this line, or dropping `.strip()` from the validator, leaves the suite green. The same clause is duplicated in the `RuntimeError` string.

The repo already has the shape for this: `CONVENTIONS.md:604` and `:608` establish that docs and frontend copies of security-relevant vocabulary get pinned by parity tests.

**Suggested action:** extend the existing assertion to cover the trimming clause in both this file and the error message, or say in the docstring that the duplication is accepted and unpinned. Worth noting `CONVENTIONS.md:616` and `AGENTS.md:200` both correctly need no change here, because they defer to this file as canonical — that indirection from #2003 is working, and saying so in the PR body would stop the absence reading as an omission.

aidemsined (commented):
Both gaps are real fixes and the whitespace half is well made — reverting `.strip()` kills three tests in both `str` and `bytes` form, and the acceptance case for a padded key is a good defensive addition that stops the trim growing into "strip, then also refuse padded keys". The reasoning for leaving `"909090909"` alone is right, too: prefix or substring matching on a five-character placeholder would refuse random keys containing `90909`, and the general low-entropy-placeholder problem is #2002's remit.

I verified the compatibility claim rather than taking it: 200 seeded 8-character keys over `ascii_letters + digits + punctuation` in both `str` and `bytes` form, plus `!@#$%^&*`, `aB3$xK9!` and `~|+=<>?/` — zero refusals. Across a wider 2,817-case corpus, **no input went from refused to accepted**, so the change is monotonically stricter. The trust boundary is intact: validation still runs at `app.py:208`, ahead of `middleware`, `register_blueprint` and `extensions`, and `SENSITIVE_CONFIG_KEYS` still redacts the key in `_print_environment`. `black`, `isort` and `mypy` clean; 1316 backend tests pass.

Two things in the body want correcting before merge.

**The newly-refused class is broader than "an 8-byte key whose first or last character is a space or tab".** It is every key whose stripped length falls below the floor, at any length: 510 of the 546 newly-refused inputs I found are longer than 8 bytes (`"  abcdefg  "` at 11, `" " * 20`), and 364 have neither first nor last byte a space or tab — `\n`, `\r`, `\x0b` and `\x0c` all trim too, so `"abcdefg\r"` and `"\x0babcdefg"` are newly refused. The conclusion survives the correction, since every affected key has a sub-floor core; the characterisation does not.

**"`.env` parsing strips them" does not hold.** `dotenv_values` preserves whitespace inside quotes — `SECRET_KEY="secret  "` yields `'secret  '`, 8 raw bytes and 6 stripped — so that value boots today and is refused after this change. Only the unquoted form is stripped. And the primary path is `os.getenv` at `settings.py:534`, which bypasses dotenv entirely: a Docker `-e`, a Helm value, or `kubectl create secret --from-file` (which keeps the trailing newline) all arrive unstripped.

## Unanchored findings

- **Don't close #2006 on this PR without a note.** #2006 named `b"909090909"` — the development default padded — as the point of its first gap, and that value is still accepted. Deferring it to #2002 is the right call, but if #2006 closes silently the reported exposure stops being tracked anywhere.

## General observations

- `bytearray(b"90909")` is not `bytes`, so it stringifies to `"bytearray(b'90909')"` (20 bytes) and is accepted — Flask then signs with it successfully. The #2006 gap in a different container, only reachable via `settings_override`, so low severity; `isinstance(secret_key, (bytes, bytearray))` closes it and matches the PR's stated goal of judging both tests on the same value.
- The validator accepts keys Flask cannot sign with. `12345678` (int), `1234567.5`, `["abcdefghij"]` and `object()` all clear the floor on their `repr` and then raise `TypeError: key: expected bytes or bytearray` at the first session write — a request, not a boot. A boot-time gate exists to avoid exactly that; a type guard would keep the failure at startup.
- `"empty"` in the summary line at `app.py:121` no longer covers the refused set, since the paragraph below argues that eight spaces are neither empty nor short.
- The `or ""` collapse means the "empty" half of the promise is carried by truthiness rather than by either clause in the `if`: `0`, `False`, `[]` and `0.0` all become `""` and are refused, while `True` becomes `"True"` and is refused only by the floor. Right answers, incidental mechanism.
- `.env.sample` is the file an operator with a whitespace-wrapped value reads first, and it says nothing about padding not counting toward the floor.
- Worth a line in #2002: once the KDF lands, derivation is consumed by both settings and app and `SECRET_KEY_FALLBACKS` comes into scope, at which point this stops being a startup assertion and wants its own module. Not worth extracting today for one 15-line function with two callers.

backend/ttnn_visualizer/tests/test_settings.py
smountenay-tt: Right, and the mutation result is the proof — reverting `app.py:149`, deleting the clause and pointing it at `b"zzzzz"` all left the suite green. Isolated it as suggested: `test_the_development_default_is_refused_independently_of_the_floor` patches `MIN_HOSTED_SECRET_KEY_BYTES` down to 1 and asserts `"90909"`, `b"90909"` and `"  90909 "` are still refused. All three mutants now die.

Patched on `ttnn_visualizer.app` rather than `ttnn_visualizer.settings`, since the validator imports the name directly — noted in the docstring so the next person does not lose ten minutes to it.

The misleading comment at `:81-83` is gone; the new test carries the claim, and its docstring says why the clause is worth keeping latent (if #2002 moves or removes the floor, it is the only thing still refusing the default).

backend/ttnn_visualizer/app.py
smountenay-tt: Agreed, and this was the most important one — the message was describing a normalisation the code does not do. Taken your wording in both copies: "at least 8 bytes excluding surrounding whitespace", in the `RuntimeError` and at `docs/src/event-logging.md:29`.

Added the clause you asked for to the docstring, and went a bit further since the harm is session loss rather than confusion:

- the docs sentence now says padding is not counted *and* not removed, and that tidying it out of an accepted key rotates the signing key
- `.env.sample` warns the operator who is about to quote a padded value, which is where they are standing when they make the mistake
- `test_a_padded_key_is_admitted_on_the_trim_but_signs_with_its_whitespace` asserts the config value is unchanged after validation, so a later "strip it at the boundary while we are here" fails

Kept the asymmetry itself for your reason, recorded in the docstring.

backend/ttnn_visualizer/app.py
smountenay-tt: Done — trimming the `str` before `.encode()`, so `"\xa0" * 8`, `"　" * 8` and `"\xa090909\xa0"` are all refused now, the last on both clauses at once as you predicted. Docstring reframed: which whitespace trims is a consequence of the ordering, not a property of trimming. Took the exact byte set too, `b" \t\n\r\x0b\x0c"` rather than "ASCII whitespace only" — the `\x1c`–`\x1f` mismatch is a real trap.

**One case does not hold: the BOM.** `'﻿'.isspace()` is `False`, so `str.strip()` does not remove it either — `"﻿" * 8` stays 24 bytes and `"﻿90909"` stays 8, and both are still accepted after the change. I checked before writing the docstring because the sentence would otherwise have claimed something false. So four of the six cases are fixed, and the docstring now states that a run of U+FEFF counts toward the floor whatever the type, rather than leaving it to be rediscovered.

Also took the `bytearray` fix from your general observations, since it is the same gap in a different container and one `isinstance` change: `bytearray(b"90909")` was stringifying to `"bytearray(b'90909')"` and clearing the floor at 20 bytes. Dropping `bytearray` from the tuple now fails a test.

backend/ttnn_visualizer/tests/test_settings.py
smountenay-tt: Confirmed both — `lstrip` and `rstrip` each passed the whole file. Added the two cases: `f"{'x' * (MIN_HOSTED_SECRET_KEY_BYTES - 1)} "` and its leading twin. Both mutants die now.

Also added the general-form line you asked for, as a header above the secret-key cases: a case whose invariant *is* the floor's position writes the number as a bare literal, which is `test_server_mode_secret_key_length_boundary` and nothing else; every other case derives from the constant because its invariant is a relationship to the floor that has to survive the floor moving. That gives someone adding a case a basis for choosing instead of pattern-matching on whichever neighbour they read first.

backend/ttnn_visualizer/app.py
smountenay-tt: This is the one I would not have written without the prompt, and the timing argument is right — I had treated the placement as settled without recording why. Added it as the second paragraph: the check has to see `settings_override`, `_refuse_debug_under_server_mode` runs inside `Config.__init__` and is the cautionary example, `fixture_settings.py` pinning `DEBUG` by hand is what that costs, and `test_create_app_refuses_the_development_secret_in_server_mode` is what a settings-side validator would stop exercising.

Traded the opener for it as you suggested. It was narrating the two lines below it and advertising a parity benefit nothing exercised, so nothing was lost.

backend/ttnn_visualizer/tests/fixture_settings.py
smountenay-tt: Fair — I swapped one name-encoded requirement for another. Promoted to a module-level `HOSTED_TEST_SECRET_KEY` composed from both constants:

```python
HOSTED_TEST_SECRET_KEY = f"test-secret-key-that-is-not-{DEFAULT_SECRET_KEY}".ljust(
    MIN_HOSTED_SECRET_KEY_BYTES, "x"
)
```

The prefix keeps it non-default if the default changes, and the pad keeps it over the floor if the floor rises past the prefix — so both properties are named once and neither is in the name. Matches what `:96-99` and `:104-105` already do for their neighbours, and `CONVENTIONS.md:768`.

docs/src/event-logging.md
smountenay-tt: Right, and the substring assertion was the reason — my insertion landed after `at least 8 bytes`, so it passed regardless. Extended it as suggested, through one constant:

```python
_HOSTED_SECRET_KEY_FLOOR_PHRASE = (
    f"at least {MIN_HOSTED_SECRET_KEY_BYTES} bytes excluding surrounding whitespace"
)
```

`test_the_event_logging_docs_name_the_fixed_paths_and_environment_control` now asserts that phrase rather than the bare number, and a new `test_the_hosted_secret_key_floor_reads_the_same_in_the_docs_and_the_refusal` raises the real `RuntimeError` and asserts the phrase appears in both it and the docs. Deleting the clause from either copy now fails — verified both directions.

Added the `CONVENTIONS.md:616` / `AGENTS.md:200` note to the body too: both correctly need no change because they defer to this file, and saying so stops the absence reading as an omission.

smountenay-tt:
Thanks — the mutation testing was the useful part of this review. Nine of the eleven mutants I checked after your comments were green before this push, including three on the default clause and both one-sided strips, so the "pins the outcome rather than the mechanism" framing was exactly right.

**All seven inline comments addressed**, replied to individually. Summary of what changed in the code rather than the prose:

- the `str` is trimmed before encoding, so non-ASCII whitespace is refused on every operator-facing path
- `bytearray` is accepted as a byte type (taken from your general observations)
- the default clause is isolated behind a patched floor, the one-sided-padding cases are asserted, and the floor's wording is pinned through one phrase shared by the docs and the refusal message
- the docstring records the placement rationale and that the trim decides admission only

**11 mutants, all dead now**; 4,072-case corpus with zero inputs going from refused to accepted; 1322 backend tests pass.

Three things to flag rather than bury in the replies:

**Your BOM case does not hold.** `'﻿'.isspace()` is `False`, so `str.strip()` does not remove it — `"﻿" * 8` and `"﻿90909"` are still accepted after the change. Four of the six cases in that comment are fixed; the docstring now states the BOM exception explicitly rather than leaving it to be retested.

**Both body corrections are in, and you were right that the characterisation was the weaker claim.** 1,361 of the 1,529 newly-refused cases in my corpus are longer than 8 raw bytes and only 142 have a space or tab at either end, so the class is "every key whose trimmed length falls below the floor, at any raw length". The `.env` paragraph was worse than imprecise — `dotenv_values` preserving whitespace inside quotes means the quoted form boots today and is refused after this change, and `os.getenv` bypasses dotenv anyway. That is now the stated reason the `.env.sample` warning earns its place.

**The type guard is deferred, not dropped.** `12345678`, `1234567.5` and `["abcdefghij"]` clearing the floor on their `repr` and failing at the first session write is a real defect in a boot-time gate, but it is a new refusal class rather than one of the two gaps #2006 reported, so it goes with #2002's rework of what this function measures. Called out under "Not done" in the body so it is not silently lost — same reason as the #2006 note you asked for, which is also in.

aidemsined:
Re-reviewed at `05d4b2b5`. Everything you claim checks out, including the part where I was wrong. One residual, and it is the same class as the gap you just closed — on the other path.

## You're right about the BOM, and I was wrong

```
>>> '﻿'.isspace()
False
>>> ('﻿' * 8).strip()
'﻿﻿﻿﻿﻿﻿﻿﻿'
```

Four of my six cases held (NBSP and ideographic space are `isspace()` and do go); BOM and zero-width space are not whitespace to Python at all. Confirmed against the branch — `"﻿" * 8` and `"﻿90909"` are still accepted. Stating the exception in the docstring is the right resolution, better than a test that would just re-encode the surprise.

## Verified on the branch

- **Trim before encode works.** Eight NBSP and eight ideographic spaces are now refused on the `str` path; eight ASCII spaces and `"  12345 "` likewise.
- **`bytearray` is handled, and it tightened things.** `bytearray(b"short")` used to be *accepted* — it fell to the `str()` branch and its `repr` is 19 bytes. It is 5 now.
- **The default clause is testable, and moving it to the trimmed value fixed something extra**: `" 90909 "` is refused now and was not before, since the old comparison was against the raw value.
- **No input got looser.** My own corpus, built independently of yours — 3,175 values across `str`/`bytes`/`bytearray`, ten whitespace characters, one-sided and two-sided padding, plus the non-string types — gives **0 refused → accepted** and 642 accepted → refused. That is the direction that matters and it holds.
- **10 of the 11 mutants I ran are dead**, including trim-after-encode, both one-sided `str` strips, the floor off-by-one, each clause of the `or` individually, and dropping `bytearray` from the isinstance tuple.

## The one survivor — bytes path, asymmetric padding

`bytes(secret_key).strip()` → `.lstrip()` **and** → `.rstrip()` both survive the whole backend suite, not just `test_settings.py`.

The cause is the same one you fixed for `str`: the only whitespace case on the byte path is `b" " * MIN_HOSTED_SECRET_KEY_BYTES` (`test_settings.py:99`), which is symmetric — every one-sided strip reduces it to `b""` and refuses it anyway, so nothing discriminates. The summary says both one-sided strips are now asserted; that is true of the `str` path only.

Two cases that do discriminate, 5 content bytes plus 3 padding so the raw length clears the floor while the trimmed length does not:

```python
b"12345   "   # strip() -> 5 bytes, refused; lstrip() -> 8 bytes, accepted
b"   12345"   # strip() -> 5 bytes, refused; rstrip() -> 8 bytes, accepted
```

Minor, since the byte path is only reachable through `settings_override` rather than from an operator's environment — but it is the one remaining place where the tests pin the outcome instead of the mechanism.

## Observation, no action

`DEFAULT_SECRET_KEY` is `"90909"` — 5 bytes against a floor of 8 — so the default clause can never be the reason a hosted start is refused unless the floor drops below 6 or the default gets longer. The patched-floor test is what makes it meaningful, and keeping it is right; worth a line in the docstring that the clause is load-bearing only if one of those two moves, otherwise the next reader may reasonably wonder why it is there.

The deferred type guard reads well under **Not done** — `12345678` clearing the floor on its `repr` and failing at the first session write is exactly the kind of thing that gets lost, and #2002 is the right home for it.

Nothing here blocks. Fix the byte-path pair if you want the mutation score clean; I would merge it either way.

smountenay-tt:
Fixed at `06ef6e8b` — you were right, and my summary comment overstated it. "Both one-sided strips are now asserted" was true of the `str` path only; the byte path had exactly the symmetric-case blind spot you describe, and `b" " * MIN_HOSTED_SECRET_KEY_BYTES` dying to all three strips is why it read as covered.

Added the asymmetric twins, derived from the constant to match the idiom rule in the header rather than written as the 5+3 literals:

```python
b"x" * (MIN_HOSTED_SECRET_KEY_BYTES - 1) + b" ",
b" " + b"x" * (MIN_HOSTED_SECRET_KEY_BYTES - 1),
```

**11 mutants, 0 survivors** — and I reran the whole set against the full backend suite this time rather than `test_settings.py` plus the docs spec, since that is how you caught this one. Both one-sided strips on both paths, both trims removed, both clauses of the `or` individually, the floor off-by-one, the raw-value revert, and `bytearray` dropped from the tuple. 1324 passed.

Took the "no action" observation too, since it costs a sentence and the question will recur: the docstring now says the default clause is load-bearing only if the floor drops below six or `DEFAULT_SECRET_KEY` grows, that nothing observable rests on it at 5-against-8 today, and that it stays for #2002 with the patched-floor test keeping it honest. Better there than rediscovered as apparent dead code.

Your `bytearray(b"short")` finding is a good one I had not checked — it was accepted before on a 19-byte `repr` and is refused at 5 now. Same for `" 90909 "` falling out of moving the comparison to the trimmed value. Both are strictly-tighter side effects rather than anything I aimed at, and both are in the corpus result.

Nothing outstanding from my side. Ready for approval whenever you are.

PR #2011, opened 2026-09-15
Add a launch config for the dev servers

aidemsined:
Adds `.claude/launch.json` so an editor or agent can start the dev servers by name instead of rediscovering the commands and ports each time.

The two entries mirror the `package.json` scripts already used for local development:

| Entry | Command | Port |
| --- | --- | --- |
| `backend` | `uv run python -m ttnn_visualizer.app` (`flask:start`) | 8000 |
| `frontend` | `pnpm run dev` (`vite`) | 5173 |

8000 is the port `vite.config.ts` already proxies `/api` to, so the pair works together without further configuration.

Config only — no application code is touched and nothing runs in CI.


smountenay-tt (commented):
Config-only, 17 lines, no application code — and the factual claims in the body check out: the two entries do mirror `flask:start` and `dev`, `PORT` defaults to `8000` (`settings.py:605`, and `GUNICORN_BIND` is what the app actually binds), Vite's default is `5173`, and `vite.config.ts:59` does proxy `/api` there. SPDX doesn't apply (`scripts/check-spdx.mjs:96` only checks `.json` paths containing `package.json`), so nothing here breaks CI.

My problem is with the premise rather than the contents: I can't find anything that reads `.claude/launch.json`, which means the file doesn't currently do what the description says it does.

## Blocker — name the consumer

- The two keys this file leans on, `runtimeExecutable` and `runtimeArgs`, appear **nowhere** in the Claude Code 2.1.272 binary. The string `launch.json` does appear, but only inside what reads as a list of filenames, with no schema handling near it. The things Claude Code actually reads under `.claude/` are `settings.json`, `settings.local.json`, `agents/`, `commands/` and `skills/`.
- VS Code *does* understand `runtimeExecutable`/`runtimeArgs` — but it reads `.vscode/launch.json`, not `.claude/`, its schema is `"version": "0.2.0"` with a required `type` and `request` per configuration, and its `port` is the debug adapter's attach port, not the app's listening port. So these entries wouldn't launch under VS Code either, in this directory or any other.
- Nothing in the repo references the path.
- `"version": "0.0.1"` matches no published schema I can find.

Which tool is meant to consume this, and is the format documented anywhere? If there's a spec, link it in the body and I'll withdraw this — but if it's a personal editor/agent setup, it belongs in a local, untracked file rather than in everyone's checkout.

## Should fix

**`.gitignore` has no `.claude` rule, and this would be the first tracked file under that directory.** The repo already solves exactly this shape for editors at `.gitignore:27-28` (`.vscode/*` plus `!.vscode/extensions.json`). Without a mirror, every contributor's `.claude/settings.local.json` — machine-local permission rules and absolute paths — and their `.claude/worktrees/` sit uncovered beside a tracked file, one `git add .claude` away from being committed. If this lands:

```
.claude/*
!.claude/launch.json
```

Worth noting that line 27-28 pair is itself a deliberate decision not to ship `.vscode/launch.json`. Adding the same thing one directory over is a reversal of that call, so it's worth making it explicitly rather than by side effect.

**The frontend entry will start with the wrong Node.** `package.json:17` pins `engines.node: 24.15.0`, `.nvmrc` is `v24.15.0`, and AGENTS.md:156 tells you to run `nvm use` from the repo root. A launcher that spawns `pnpm run dev` directly inherits whatever Node its parent process has, which for an editor or agent is usually the system default — pnpm then fails before Vite ever starts. Either the entry handles the version or the file states that the shell must already have `nvm use` applied.

## Nits

- The backend entry doesn't match the documented dev pairing. AGENTS.md:58 says frontend dev is `pnpm dev` **and `pnpm flask:start-debug`** in parallel; `flask:start` drops the `DEBUG=true` that gives you backend logs while you're working on the UI.
- This is a third copy of the dev commands, after AGENTS.md:37/58-59 and `docs/src/running-from-source.md`, with no test or lint pinning it to `package.json`. A script or port rename drifts here silently — and a launcher that's quietly wrong is worse than no launcher.
- Hardcoding `5173` is fragile once something does read it: Vite falls back to 5174+ when the port is taken (`docs/src/running-from-source.md:42` calls this out), so a consumer waiting on 5173 would hang on a machine that already has a dev server up.

aidemsined:
Every factual claim here checks out, and three of the findings were right about real problems. Fixed in 7b924abe. The blocker has an answer, though it's not in the binary you grepped.

## The consumer

**The Claude Code desktop app's browser/preview tooling**, not the CLI. That's why the strings aren't in 2.1.272 — different surface. Its tool refuses with the schema spelled out verbatim when the file is missing:

```
No .claude/launch.json found. Create .../.claude/launch.json with this format:
{ "version": "0.0.1", "configurations": [ { "name": ..., "runtimeExecutable": ...,
  "runtimeArgs": [...], "port": ... } ] }
```

which is where `"version": "0.0.1"` and the key names come from. I used it in this session to start these servers and drive the app in the pane, and I re-ran it after the changes below — the `frontend` entry brings Vite up on 5173.

You're right that I can't point you at a published spec, and right that VS Code's `.vscode/launch.json` is a different format entirely (`0.2.0`, `type`/`request`, debug-adapter port). So the file is app-specific, and your underlying question survives the answer — see the end.

## `.gitignore` — right, and the sharpest catch here

No `.claude` rule, and this would be the first tracked file under it, so everyone's `settings.local.json` (machine-local permission rules, absolute paths) and `worktrees/` sat uncovered beside it. Added, mirroring the pair you pointed at:

```
.claude/*
!.claude/launch.json
```

## Node — right, and my first fix was wrong

I tried `bash -lc` on the theory that a login shell defines `nvm`. Measured on this machine it doesn't: the shell here is zsh, bash's profile never loads nvm, and pnpm still refused at v16.20.0. It now sources `$NVM_DIR/nvm.sh` explicitly and falls through to whatever Node is present when nvm is absent — which is what would have happened anyway. Verified by starting it: Vite on 5173 with pnpm under v24.15.0 from a v16 parent.

## `flask:start-debug` — taken

Both entries go through `bash -c` now, which is also what lets the backend set `DEBUG=true` without pulling Node into its path.

## The drift nit — taken, because you're right that a quietly wrong launcher is worse than none

`tests/claudeLaunchConfig.spec.ts` ties the backend port to the port `vite.config.ts` proxies `/api` to, the commands to the `package.json` scripts they stand in for, and the node alignment to its own presence. Four drift mutations all fail it, including renaming the proxy port out from under the config.

## The 5173 nit stands

Vite falls back to 5174+ when the port is taken, so a consumer waiting on the declared port can hang. That's a property of a format that requires one port per entry, not something this file can fix. Worth knowing rather than worked around.

## Your real question

> if it's a personal editor/agent setup, it belongs in a local, untracked file

That's the part I can't settle, and you've framed it correctly: `.gitignore:27-28` is a deliberate decision *not* to ship `.vscode/launch.json`, and tracking this is that call reversed one directory over. It's a team convention question rather than a technical one, so I'd rather it were decided than defaulted. Happy either way — if the answer is "keep it local", the `.gitignore` rule and the drift test are still worth keeping, and I'll drop the file itself and close this.

smountenay-tt (commented):
Thanks — that's a real answer to the consumer question and three real fixes. I checked the new work rather than taking it on trust, and the test in particular holds up: 5/5 pass, `tsc --noEmit` is clean, and the four drift mutations you describe each fail exactly one test on my machine (vite proxy port → 8001; launch backend port → 9999; `nvm use` removed; the module renamed in `flask:start-debug`). The `.gitignore` pair behaves — `launch.json` stays tracked, `settings.local.json` is ignored by `.gitignore:33` — and `DEBUG=true exec uv run …` does export `DEBUG` under `bash -c`, which isn't obvious for a prefix assignment on a special builtin.

One new thing, from the Node fix itself.

## Should fix — the node fix now hides its own failure

`nvm use >/dev/null 2>&1` discards stderr, and `nvm use` resolves `.nvmrc` from the **current working directory**, not from the location of the config that invokes it. Measured from a directory without `.nvmrc`:

```
$ bash -c '... nvm use >/dev/null 2>&1; node --version'
v20.11.1          # silent
$ bash -c '... nvm use; node --version'
No .nvmrc file found
v20.11.1
```

So if the launcher's cwd is ever not the repo root, the chain says nothing, falls through to the parent's Node, and `pnpm run dev` fails with pnpm's own error about the engine rather than the actual cause. That's precisely the failure the entry exists to prevent, now without the message that would explain it.

It evidently works on your machine, so the launcher's cwd is the repo root today — but that's an assumption about an undocumented format, and it costs one redirect to stop betting on it:

```
nvm use >/dev/null || echo 'launch.json: nvm use failed; pnpm will run on' "$(node -v)" >&2
```

Same fall-through, stdout still quiet, and the bad case becomes legible instead of looking like a pnpm bug.

## Nit — the body is now stale

> Config only — no application code is touched and nothing runs in CI.

`tests/claudeLaunchConfig.spec.ts` is picked up by `pnpm test`, which `.github/workflows/lint-and-test.yml:49` runs on every PR (`pnpm lint:ts` at :53 covers it too). That's a fine change — the drift test is worth its CI seconds — but it's the sentence a later reviewer would trust, so it should say so.

## On the consumer, for the record

I can't verify a desktop-app surface from here, so I'll take the reproduction as given. Worth stating plainly for whoever reads this later: the format is private, undocumented and unversioned, `"version": "0.0.1"` is the only hint of a contract, and nothing validates the file against it. Your drift test pins the file to this repo's scripts and ports, which is the half that's in our control — it can't catch the consumer changing the schema underneath. Not something to fix, just not a guarantee to lean on.

## The tracking question

Leaving that where you put it — it's a maintainer call, not mine, and I've flagged it for a decision rather than letting it default. Your read is the right one either way: if the answer is "keep it local", the `.gitignore` allowlist is worth keeping on its own merits, since it also stops anyone's private agent config under `.claude/` from being committed by an absent-minded `git add`.

smountenay-tt (approved):

Issue #2013, opened 2026-09-15
Agent tools: shared read-only open, tool descriptors, and response-shape consistency


aidemsined:
Raised in review of #2014 (issue #2012). The correctness findings were fixed there; these are the structural ones, which change shape rather than behaviour.

## `Instance.profiler_path` means two different things

It is the path to `db.sqlite` everywhere in the application — `instances.py:585` concatenates it, `queries.py:104` opens it, `sftp_operations.py:274` reads it — and the *directory* inside `agent/handles.py`. The agent surface therefore resolves the file itself, and the same `db.sqlite` join plus `mode=ro` URI now appears in both `agent/operations.py::_profiler_db` and `agent/handles.py::_has_operations_table`, the second of which also hand-rolls `_check_table_exists`.

An `Instance` built by the agent layer is a different animal from every other `Instance`, and handing one to another consumer misbehaves silently. Either resolve to the file in `ReportRegistry.add` and keep the directory as its own field, or give `DatabaseQueries` an `open_read_only(directory)` classmethod both call sites use. Read-only is a property `LocalQueryRunner` cannot provide — it connects read-write with `isolation_level=None` — so the agent layer opening its own connection is right; doing it twice is not.

## `server._tool_table` is hand-written schema for eight tools

Each tool's caveat text now lives in three places — the schema description, the response `note`/`caveat`, and the function docstring — with nothing keeping them in step. Adding a tool means edits in three files, and there is no shared `_HANDLE_SCHEMA`, so `handle` appears eight times as a bare `{"type": "string"}` with no description.

Have each tool module export a descriptor list (`name`, `description`, `schema`, `fn`) and let `server.py` merge them, leaving it pure transport. The README's containment claim is currently one-directional: nothing imports *from* `server.py`, but `server.py` knows every tool's signature through `lambda arguments: …(registry, **arguments)`. This is what makes the planned MCP SDK swap mechanical rather than a rewrite.

## Response-shape drift between the CSV and database halves

An agent calling both meets two conventions:

- Every `tools.py` response carries `"projection": dict(CANONICAL_PROJECTION)`, which is the rule `CONVENTIONS.md` names. The database tools' equivalent choice — the rank default — is spread flat as `multi_host`/`rank` via `RankScope._asdict()`, which also makes NamedTuple field names the wire contract silently.
- Four words for "total before the cap": `op_count`, `match_count`, `operation_count`, `input_count`/`producer_count`.
[...]

PR #2014, opened 2026-09-15
Answer memory and operation questions over MCP

### LINKED ISSUE #2012: Expose memory and operation questions to an agent over MCP (second slice)
The first slice ([#1995](https://github.com/tenstorrent/ttnn-visualizer/issues/1995) / [#1996](https://github.com/tenstorrent/ttnn-visualizer/pull/1996)) answers time questions only. An agent can ask which ops dominate, where the time went inside them, and whether a change helped — and nothing at all about memory, which is the other half of bring-up: an OOM or an unexpected DRAM spill is as common a bring-up failure as a slow op.

The gap is already named in the tool surface's own response. `load_report` returns `data_present_without_tools: ["operations_database"]`: the profiler `db.sqlite` is there, and no tool reads a table from it.

## Tools

- `find_operations(handle, name_contains?, limit)` → bounded `operation_id` / name / duration list, so an agent can locate an op before drilling in.
- `operation_detail(handle, operation_id)` → one op: name, duration, input and output tensors with shape, dtype, layout and memory config, and its allocations by buffer type. Bounded by construction.
- `memory_profile(handle, buffer_type?, limit)` → per-op allocated bytes by buffer type (`DRAM`, `L1`, `L1_SMALL`), the run peak, and the device's bank geometry alongside it so the number can be read as a fraction of capacity.
- `tensor_flow(handle, tensor_id)` → the producing op and the consuming ops. `query_producers_consumers` already exists.

Both round-one properties hold unchanged: bounded results only, and a number that can mislead carries its caveat in the response.

## The caveat this slice has to carry

`buffers.max_size_per_bank` is **per bank**, and the API layer already renames it to `size` (`serializers.py:309`), so the misreading is one rename away rather than hypothetical. On a local resnet50 capture, peak L1 sums to 487,424 bytes against an `l1_bank_size` of 1,370,848 with `l1_num_banks` 64 — an agent comparing that sum to device L1 capacity understates usage by up to the bank count. `total_l1_for_tensors` is `0` in the same capture, so it cannot serve as the denominator; `l1_bank_size` is the figure to report against.

## Out of this slice

`buffer_pages` — 5,850,688 rows in that one capture. Page-level questions (fragmentation, the memory plot's own source) need a different shape than a tool response and are not attempted here.

## Wiring note

`LocalQueryRunner` connects to `instance.profiler_path` as the SQLite *file* (`queries.py:104`); the agent's `load_report` takes a *directory* and joins `db.sqlite`. `DatabaseQueries(connection=...)` is already supported, so the agent layer opens its own read-only connection rather than making either meaning of the field serve both.

aidemsined:
## Summary

The [first slice](https://github.com/tenstorrent/ttnn-visualizer/pull/1996) reads the performance CSVs, so an agent could ask where the time went and nothing about what the run allocated. That is the other half of bring-up — an out-of-memory failure or an unintended DRAM spill is as common as a slow op — and the data was already sitting there unread. `load_report` said so in its own response, returning `data_present_without_tools: ["operations_database"]`.

Four tools over the profiler report's SQLite database, bringing the surface to eight:

| Tool | Answers |
|---|---|
| `find_operations` | operations matching a name substring, with the ids `operation_detail` takes |
| `operation_detail` | one operation: its input and output tensors with shape, dtype and layout, and what it had allocated |
| `memory_profile` | footprint per operation, keyed by buffer type and ranked within each, with each type's peak and the device's L1 geometry |
| `tensor_flow` | which operation produced a tensor and which ones consumed it |

Closes #2012

### Nothing adds one memory type to another

`buffers.max_size_per_bank` is divided by the bank count of its own memory type, and those counts differ — so a DRAM figure plus an L1 figure is two denominators in one integer rather than a quantity. The report carries no DRAM bank count at all, so one of them is not even knowable. `memory_profile` is therefore keyed by buffer type and ranks operations within a type. This is the arithmetic round one already refused for rates and core counts (`tools.ADDITIVE_METRICS`), applied to allocations.

A per-bank L1 figure is comparable to the `l1_bank_size` returned beside it, which is a conservative bound. **A device-wide total cannot be derived from the response**, and it says so rather than offering a formula: multiplying by the bank count is right only for a buffer interleaved across every bank, and most are not — the operation holding resnet50's L1 peak uses 56 of 64 banks, so multiplying overstates it by 14%, and 16-bank operations in the same report by 4x. Which banks a buffer occupies lives in `buffer_pages`, which no tool exposes.

### A unit that depends on the report is reported, not asserted

`tensors.size` is a whole-tensor byte count — where the column exists. Where it doesn't, `query_tensors` substitutes `b.max_size_per_bank AS size`, which is per bank: the same figure as an allocation total. **Of 86 local captures only 5 carry the column**, so the fallback is the common case. Responses report `tensor_size_unit` rather than claiming a unit, and `DatabaseQueries.report_has_tensor_size_column` decides it, next to `report_has_rank_column` for the same reason.

### A peak is a maximum, and usually a plateau

`buffers` records what was live *at* each operation rather than what it allocated, so a per-operation sum is the footprint at that point and the largest is that type's peak. Peaks are maxima, never sums across the run: buffers persist, so adding them reports one 1000-byte allocation as a 3000-byte peak.

Because resident memory barely moves, a peak is usually shared — 201 of resnet50's operations hold the DRAM peak, 18 hold the L1 peak, while minicpm's L1 peak is a genuine single spike. `operations_at_peak` says how many, which is the difference between one operation you can go and fix and a plateau across the whole run.

### Aggregated in SQL, for the reason `query_zone_summary` is

`DatabaseQueries.query_buffer_totals_by_operation` sums per operation and buffer type in the query. Grouping in Python cost ~390 bytes of process memory per row with no bound — 2.07 s and 580 MB on a 1.17M-buffer capture, and 20 s and 4.4 GB on a report ten times that, inside a long-lived stdio server, to produce a 2 KB answer. After: **0.84 s and 144 MB against a 138 MB baseline**, flat in report size, and byte-identical to SQL ground truth for every buffer type.

### Multi-host reports are read one rank at a time

Operation ids restart at 1 per rank (#1842), so an unfiltered read unions the ranks and collides operations that merely share an id. These tools default to rank 0 — matching the app's own routes — take a `rank` argument, and name it in every response with a caveat. Every filter goes through `merge_rank_filter`, per table: `report_has_rank_column` probes `operations` alone, and a report with `rank` there but not on `buffers` otherwise raised `no such column: rank`. A rank the report has no rows for is refused rather than answered empty, as `views._rank_query_param` does.

### `_inventory` probes rather than trusts

A zero-byte `db.sqlite` opens as a valid empty database, so checking that the file exists advertised all four tools and then failed each with `no such table` — the same broken promise as naming a tool that was never registered, which is what that function exists to prevent.

### Not in this slice

`buffer_pages` — 5,850,688 rows on one local capture. Page-level questions need a different shape than a tool response. Structural follow-ups from the review are in #2013.

## Testing

33 tests in `test_agent_operations.py`, built against SQLite reports written to disk rather than the `make_report` fixture: that fixture registers an instance through the Flask app, and this surface takes neither an app nor a database session, so using it would pin a coupling the tools deliberately do not have. Fixtures cover a report without `tensors.size`, one without `devices`, one with `rank` on some tables only, integer and text `buffer_type`, and two devices.

Every case is verified by mutation. That caught three of my own tests being weaker than they looked — the size-unit test could not fail because its fixture always carried the column, the first cap test held trivially on a three-operation report (capping was removable from `find_operations` with the suite green), and the rank-scoped-device test passed regardless of the code. The explicit tie-break is kept and documented as unobservable: SQLite's `GROUP BY` already orders by its leading key, so no test can pin it while that holds.

Also driven end to end over stdio against real captures throughout, which is how the empty-database, unit and per-bank problems surfaced.

1357 passed; mypy clean over 92 files; black, isort and flake8 clean.


aidemsined:
All four addressed in e80823a2. I reproduced each against real captures before changing anything.

**Partial database advertised as answering everything** — correct, and it's the same failure the probe was added to prevent; I fixed the zero-byte case and missed this one. A database holding only `operations` offered all four tools and then failed three with `no such table`. One `sqlite_master` read now answers per tool against the tables it reads:

```
partial DB -> answerable:   ['find_operations']
               unanswerable: ['top_ops', 'zone_timings', 'memory_profile', 'operation_detail', 'tensor_flow']
```

**Explicit rank ignored on a report with no rank column** — confirmed: `rank=7` returned rank-0 rows labelled `rank: null`. Honest about what it did, but serving different data than was asked for, where `views._rank_query_param` answers 400. Now refused, while `rank=0` still reads — that being the one rank such a report holds.

**`operation_detail` dropped `memory_config`** — right on both counts: #2012 asks for it, and `Tensor.__post_init__` had already parsed it. It earns its place beyond scope compliance, because `memory_layout` is what makes the per-bank caveat actionable — the response that says a device-wide total can't be derived now also shows why. The operation this tool names as resnet50's L1 peak comes back `HEIGHT_SHARDED`, and `shard_spec` carries the grid.

Worth noting for anyone reading the diff: my first fixture for this asserted against a JSON `memory_config` and silently got `None` back, since `parse_memory_config` parses TTNN's `MemoryConfig(...)` string form. The fixture now carries that format verbatim from a capture.

**`write_report` missing its underscore** — correct per `AGENTS.md:227`; renamed.

The three behavioural fixes are each covered by a test that fails if the fix is reverted. 1360 passed, mypy clean over 92 files.

smountenay-tt (commented):
Four bounded tools over the report database, aggregation pushed into the query class where `query_zone_summary` already put it, and the per-bank unit problem handled by reporting the unit rather than asserting one — the design is sound and the reasoning in the PR body is the reasoning I'd want. No startup-requirements surface is touched and the MCP server stays a separate console script, so there's no deployment-start or hosted-exposure risk here. Verified locally at `e80823a2`: 1348 passed, mypy clean over 92 files, black and isort clean.

One blocker, though, and it's the same collision the rank scope exists to prevent, reappearing inside the allocation figures. Everything below is reproduced against a report built on disk, not read off the diff.

## Blocker

**The rank caveat asserts the opposite of what the numbers are, on a report with `rank` on `operations` but not on `buffers`** — `agent/operations.py:422`, and `:307-313`/`:356` for the same thing in `operation_detail`.

`merge_rank_filter` no-ops when the table has no `rank` column (`queries.py:275-276`), which is what stops the `no such column: rank` crash. But the no-op is silent, so on that schema mix the buffer rows union every rank and collide on `operation_id` — while `report_has_rank_column()` is `True`, so the response still reports `rank: N`, `multi_host: true`, and the caveat "figures here describe that rank rather than the job".

Two ranks each holding an operation 1, allocating 1000 and 9000 bytes, `buffers` with no rank column:

```
memory_profile(rank=0) -> L1 peak=10000, attributed to 'op_rank0'   (truth: 1000)
memory_profile(rank=1) -> L1 peak=10000, attributed to 'op_rank1'   (truth: 9000)
  both carry: "...read at rank N only. Operation ids restart per rank, so
               figures here describe that rank rather than the job."

operation_detail(op=1, rank=0) -> allocations={'L1': {'buffers': 2, 'size': 10000}}
operation_detail(op=1, rank=1) -> allocations={'L1': {'buffers': 2, 'size': 10000}}
```

That is #1842 in the memory numbers, and by your own standard in `operations.py:44` it's the worse kind: a caveat pointing the wrong way rather than none. `tensor_flow` has the same exposure through `:529` and `:535` when `tensors` lacks the column.

`_MIXED_RANK_SQL` can't catch this — it holds a single operation on a single rank, so it pins "doesn't crash" and nothing about attribution. Adding a second rank to that fixture fails the current code.

The buffer rows genuinely aren't attributable on that schema, so refusing is the honest answer rather than picking a number. Roughly:

```python
# in _scoped / a new guard, with a public `table_has_rank_column` on DatabaseQueries
if scope.multi_host and not queries.table_has_rank_column("buffers"):
    if len(known_ranks) > 1:
        raise ValueError(
            "this report carries `rank` on `operations` but not on `buffers`, "
            "so allocations cannot be attributed to a rank"
        )
```

Worth saying which way the real captures fall — you've got the 86 locally and I don't. If the mix only ever occurs on single-rank reports, the `len(known_ranks) > 1` guard costs nothing and documents why; if it occurs on genuine multi-host captures, this is live wrong data.

## Should fix

**The device caveat is computed before the `buffer_type` filter** — `agent/operations.py:434`, filter at `:444-446`.

`devices` is taken from every `grouped` row, so a filtered call inherits devices that only the excluded types touched. On a report where DRAM spans devices 0 and 1 but L1 is on device 0 alone:

```
memory_profile(buffer_type='L1') -> caveat: "These buffers span 2 devices (0, 1),
                                    and the totals add them together."
```

The L1 totals add nothing together. Same rule as above — derive it from the rows that survived the filter:

```python
kept = [row for row in grouped if wanted is None or _buffer_type_name(row[1]) == wanted]
devices = {row[2] for row in kept}
```

**An empty id list means "every row", not "no rows"** — `agent/operations.py:318-326` and `:553`.

`_query_table` and `query_tensors` both `continue` past a zero-length list (`queries.py:238-239`, `:722-723`), so the filter is dropped rather than matching nothing. An operation with no `input_tensors` and no `output_tensors` rows therefore reads the whole `tensors` table — precisely what the comment at `:315-317` says it avoids, since every other tensor's `memory_config` is parsed on the way past. Your own `_REPORT_SQL` operation 3 takes that path:

```
operation_detail(op=2) -> query_tensors(tensor_id=[10, 11, 12]) ->   3 Tensors built
operation_detail(op=3) -> query_tensors(tensor_id=[])          -> 503 Tensors built
```

`tensor_flow` on a tensor with no producers and no consumers does the same to `operations` via `operation_id=[]`. Results stay correct, so this is latency and memory rather than wrong data — but it's unbounded in report size on the surface whose whole premise is bounded reads, and it lands on the same `operation_detail` you measured at 16 s. A `if not wanted_ids:` short-circuit covers both.

## Nit

**`_refuse_absent_rank` materialises every operation to compute a distinct-rank set** — `agent/operations.py:154-161`. `query_operations` with an unfiltered rank builds an `Operation` dataclass per row across all ranks, to keep a set of ints. #2013 lists two queries that scan more than they need; this is a third, and the only one that pays dataclass construction for it. `SELECT DISTINCT rank FROM operations` on the query class would do it.

**`docs/src/agent-tools.md`** — "The first four read the performance CSVs" is now slightly off for `load_report`, which also opens `db.sqlite` to build the per-tool inventory.

## On #2013

Agreed on all of it, and filing it separately was the right call. Two of the items above are arguably in that set — I'd keep the caveat-scope and empty-list ones here rather than deferred, since both change what an agent is told rather than only the shape it's told in.

aidemsined:
All of it addressed in 4bca9668. I reproduced each before changing anything.

## Blocker — confirmed and refused

Your repro reproduces exactly:

```
rank=0 -> peak=10000 attributed to 'op_rank0'   (truth: 1000)
rank=1 -> peak=10000 attributed to 'op_rank1'   (truth: 9000)
```

You asked which way the real captures fall, and I have the 86 — **none carries the mix**. 84 have no `rank` column at all; the other two (`multihost_poc_jun19_2043`, `multihost_poc_jun24_2321`) have it on `devices`, `operations`, `tensors`, `buffers`, `input_tensors` and `output_tensors`. So this is insurance against schema drift rather than live wrong data — which makes the guard free, so it is in.

Refused rather than answered, as you suggested, and only when the report genuinely holds more than one rank: a single-rank mixed report cannot misattribute anything and still answers. `find_operations` also still answers, since it reads only `operations`. `table_has_rank_column` is public now, because a caller that reports a rank beside the rows needs to know whether the filter reached them.

## Should fix — both taken

**Device caveat scope.** Derived from the rows that survive the filter now. On your split-device shape: the L1 call carries no caveat, the DRAM call still says "span 2 devices (0, 1)".

**Empty id list.** Short-circuited in `operation_detail` and `tensor_flow`. Your `_REPORT_SQL` operation 3 was the case — it now builds zero `Tensor`s instead of the whole table, pinned by a test that counts `Tensor.__post_init__` calls.

## Nit — taken

`query_operation_ranks` does a `SELECT DISTINCT rank ... ORDER BY rank`, and the rank-existence check folds into `_rank_scope` where the ranks are already in hand, so nothing materialises an `Operation` per row to collect integers. `RankScope` carries them and hands responses an explicit `as_response()` rather than `_asdict()`, so a field added to the tuple can't leak onto the wire.

Docs line fixed — `load_report` opens `db.sqlite` too, which is the whole point of it.

## On #2013

Agreed, and I moved both of the ones you named out of it into this PR for the reason you gave: they change what an agent is told, not just its shape. #2013 keeps the shared read-only open, the tool descriptors and the response-shape consistency.

Four tests, each verified by mutation: the refusal per tool, the single-rank mix still answering, the filtered device caveat, and the zero-tensor read. 1364 passed; mypy clean over 92 files.

smountenay-tt (approved):
Approving. All four Copilot comments and everything from my round are addressed, and I verified each one against the code at `4bca9668` rather than off the summaries — built on disk, with the same shapes I used to raise them.

## Blocker — refuses, and the numbers are no longer wrong

My original repro on a report with `rank` on `operations` but not on `buffers`, two ranks each holding an operation 1:

```
memory_profile(rank=0)             -> REFUSED: carries `rank` on `operations` but not on
                                      `buffers`, so its rows span ranks 0, 1 and cannot
                                      be attributed to rank 0
memory_profile(rank=1)             -> REFUSED (same, rank 1)
operation_detail(op=1, rank=0/1)   -> REFUSED, naming buffers, tensors, input_tensors,
                                      output_tensors
tensor_flow(tensor_id=10, rank=0)  -> REFUSED, naming the three tensor tables
find_operations(rank=0)            -> [{'operation_id': 1, 'name': 'op_rank0'}]
```

Before, both ranks answered `peak=10000` against truths of 1000 and 9000, with a caveat pointing the wrong way. `find_operations` still answering is right — it reads only `operations` — and the single-rank version of the same schema mix still answers `peak=10000` for the one rank it holds, so the guard costs nothing where nothing can be misattributed. Scoping it with `len(scope.ranks) > 1` and putting the ranks on `RankScope` is a better shape than the guard I sketched.

Thanks for checking the 86 captures and saying which way they fall — 84 with no rank column and two with it on every table settles this as schema-drift insurance rather than live wrong data, which is what I wanted to know and couldn't check myself.

## Should fix — both confirmed

**Device caveat scope.** On the split-device shape (L1 on device 0, DRAM across 0 and 1):

```
buffer_type='L1'   -> no caveat
buffer_type='DRAM' -> "These buffers span 2 devices (0, 1)..."
buffer_type=None   -> "These buffers span 2 devices (0, 1)..."
```

**Empty id list.** Counted `Tensor.__post_init__` calls on a report with 500 unrelated tensors: `operation_detail` on the operation with no input or output rows builds 0. Reverting the short-circuit fails your test at `assert 3 == 0`.

## Nits — taken

`query_operation_ranks` is a `SELECT DISTINCT`, so nothing builds an `Operation` per row to collect integers, and folding the existence check into `_rank_scope` where the ranks are already in hand is cleaner than a separate pass. `as_response()` over `_asdict()` is the right instinct — `ranks` would otherwise have landed on the wire the moment it was added. Docs line fixed.

## Verification

Mutation-tested all six fixes rather than trusting a green suite. Each reverts to exactly one failing test, which is the part I actually care about:

| Reverted | Fails |
|---|---|
| `_refuse_unattributable` → no-op | `test_figures_that_cannot_be_attributed_to_a_rank_are_refused` |
| device caveat before the filter | `test_a_filtered_call_does_not_inherit_another_type_s_device_caveat` |
| empty-id short-circuit | `test_an_operation_with_no_tensors_does_not_read_the_whole_table` |
| legacy nonzero-rank refusal | `test_a_rank_is_refused_on_a_report_with_no_rank_column` |
| `memory_config` in `described()` | `test_it_reports_each_tensor_memory_layout` |
| per-tool inventory probe | `test_a_partial_database_answers_only_what_it_can` |

1352 passed locally, mypy clean over 92 files, black and isort clean, all 14 CI checks green. Scope is tight — the fix commits touch nothing outside the surface under review.

## Not blocking, for #2013

`_device_capacity` reads `devices` through `_scoped`, but the guard for `memory_profile` names only `buffers`. On a report with `rank` on `operations` and `buffers` but *not* on `devices`, the allocation figures come out correctly attributed and the device block does not:

```
rank=0 -> peak=1000 as 'op_r0'  (correct)   device={'devices': 2, 'geometry_from_device': 0}
rank=1 -> peak=9000 as 'op_r1'  (correct)   device={'devices': 2, 'geometry_from_device': 0}
```

`devices: 2` is a statement about the rank and it unions both, and rank 1's buffers sit on device 1 while the geometry comes from device 0. Much narrower than the blocker — the load-bearing numbers are right, `geometry_from_device` labels which device it took rather than asserting they match, and no real capture has this mix either. Adding `"devices"` to the guard would refuse a whole call over a descriptive count, which is the wrong trade; scoping or qualifying the count is the better fix. The comment at `_device_capacity` says it is rank-scoped "like the figures beside it", which is the bit that is true only when the column is there — worth a word either way.

Also trivial: the `**A multi-host report is read one rank at a time**, and refuses rather than guess.` line in `agent-tools.md` runs past the file's wrap and reads a little off grammatically.

Merging.

PR #2016, opened 2026-09-15
Keep the graph's fold state across a deallocate toggle

### LINKED ISSUE #2015: Operations graph: grouping collapse state is lost on any rebuild, and the active grouping button cannot re-apply it
Reported on `feature/2007-graph-viewport-policy`, but the grouping state predates that branch.

## What happens

With **Layers** selected and the blocks collapsed, toggling **deallocate ops** leaves the graph with nothing collapsed. Clicking **Layers** again does not re-collapse it — you have to click **Repeats** and then **Layers** to get back to collapsed layers.

## Expected

The collapse state belongs to the selected grouping and should survive a rebuild that has nothing to do with grouping. Toggling deallocate ops should leave Layers collapsed.

## Likely mechanism, not yet verified

`handleGroupingChange` (`OperationGraphReactFlow.tsx:925`) applies the fold *at the moment of the click* — it drops the outgoing mode's ids from `expandedBlockIds` so the incoming mode's blocks read as collapsed. Nothing re-applies that for the mode already selected.

A `hideDeallocate` toggle changes `buildOptions`, so the graph rebuilds and the detector produces a fresh set of block instance ids. The ids the earlier click folded no longer name anything, and absence means *unrolled* for a grouping block (the two opposite defaults noted at `:474`), so everything comes back expanded.

That also explains the round trip: the handler early-returns when `next === grouping` (deliberately — re-picking the active mode is not a second ask, #1977), so clicking **Layers** while Layers is active cannot re-fold. Going through **Repeats** makes the next Layers click a real mode change, which re-applies the fold.

If that is right, the fix is for the fold to be a property of the grouping rather than a one-shot applied on click — so a rebuild that re-detects blocks folds them per the mode's current state, without the user re-asking. Worth checking whether the same thing happens on the other rebuild triggers (operation range, weight-load collapse, device subgraphs) before settling on a shape.

aidemsined:
## Summary

Collapsing the operations graph by **Layers** and then toggling **Hide deallocate ops** left nothing collapsed, and the Layers button could not put it back — re-picking the mode already in use is deliberately not a second ask ([#1977](https://github.com/tenstorrent/ttnn-visualizer/issues/1977)), so recovering meant a round trip through Repeats.

Closes #2015

## Why the reset was there, and why it goes

`handleHideDeallocateChange` dropped the fold decision, on the grounds that re-running block detection would leave the held instance ids naming instances that no longer exist. An existing test asserted that by name, so **this PR reverses a documented decision** — worth being explicit about.

The reset was the wrong instrument. It cost every reader their collapse, and their open device operations, on an unrelated filter. A report change *does* still reset, correctly: those ids belong to a different report.

**Correction from review:** I originally justified this partly by `collapseWeightLoads` — "the other filter that re-runs detection" — never having dropped anything. That is wrong. The layout worker keys detection on `${sourceVersion}:${hideDeallocate}:${grouping}` and passes only `hideDeallocate` to `getKeptOperations`, so weight-load collapsing never reaches the detector and cannot rename an instance (measured: identical ids either side of it). The two filters were never in the same class, and the same check closes the third trigger #2015 asked about — device subgraphs are not in the key either. `hideDeallocate` is the only control that can rename an instance.

But the premise was not simply wrong either. Ids genuinely can shift, in two ways I initially missed and review caught:

- a deallocate **first** in a repeating unit changes each span's first member — `block:3,6,9` hidden vs `block:2,5,8` shown
- a run detected only when deallocates are hidden **renumbers** every later run, because `runIndex` is the index into runs sorted by start

Layers shift the same way when a deallocate follows a delimiter (`layer_norm`, then deallocate the pre-norm tensor).

### The consequence was the mirror of the bug

`expandedBlockIds` means *these unrolled, everything else folded*. So when every held id misses, an unrolled graph comes back **fully folded** — silently overriding an unroll, on exactly the reports where the filter matters. Keeping the decision without addressing this would have traded one direction of the bug for the other.

Two changes, at different levels:

**A repeat instance is keyed on its first member alone** (`block:<first member>`). This is what the layer detector already does, and it says why: *"Keyed on the first member rather than the index, so folding one instance cannot rename the others when detection shifts by a span."* Runs are disjoint and consumed, so the first member is unique and the index was redundant. That removes the renumbering class outright.

**`carryUnrollAll` re-expresses the one decision whose ids are incidental.** "Fold all" is `[]`, which names no instance and cannot go stale; "unroll all" was a list, and that asymmetry *was* the bug. When every instance the previous detection found was unrolled, the newly detected ids are unioned in. Every other set names instances the reader chose, and comes back by reference unchanged — so a partial fold cannot even provoke a spurious layout.

### Known residual

A reader who unrolled *some* instances, on a report where the filter renames them, gets those folded rather than unrolled. The ids that survive keep their state; specific choices are not reinvented for instances that no longer exist. Recoverable in one click, and the opposite of today's failure where everything springs open.

Second, raised in review: the carry-forward runs in `onBuilt`, *after* the build carrying the stale ids has been committed — so on a renaming report the reader gets one frame of the folded graph before the corrected one lands, and pays two layouts. It converges, and it is strictly better than the permanent loss this PR fixes, but it is not free.

Making unroll-all nameless is the right fix and it cannot be done on the current wire format: `expandedBlockIds` carries grouping blocks, which fold only `if (hasFoldDecision && !expandedBlocks.has(id))`, *and* weight fans, which fold on `!expandedBlocks.has(id)` with no such guard (`opGraphBuilder.ts:135` vs `:164`). So passing `undefined` would re-fold every fan the reader had opened — the bug `expandAllBlocks` unions to avoid. Separating the two sets is a protocol change across the component, worker and builder, filed as #2017. A test pins the 5-node frame and the extra build, and flips when that lands.

## Not changed

That clicking the *active* grouping button cannot re-apply a fold. It only mattered as a recovery path from the lost state, and #1977 is sound on it: folding is the Fold button's job, and that button is right there. Flagged rather than quietly extended.

## Testing

The existing test asserted the old behaviour by name and now asserts the new one.

The sharpest review finding was that my first round of tests never exercised what they argued about: `runBuild` is mocked and no fixture in the suite held a `ttnn.deallocate` op, so flipping the switch never re-ran detection. There is now a fixture whose every instance the filter renames, and two tests over it — an unrolled graph stays unrolled, and a partial fold is not promoted to unroll-all. Two detector tests pin id stability and uniqueness directly, rather than through the mocked harness.

The range test asserted an option that would have held even if `operationList` were ignored entirely; it now pins the re-derived source too. The weight-load test is reframed as the regression pin it is rather than as evidence.

Every test fails if its fix is reverted — verified by mutation, including restoring the old reset, restoring `runIndex` in the id, removing the carry-forward, and applying the carry-forward to every decision instead of just unroll-all.

2661 frontend tests pass; tsc and eslint clean.


smountenay-tt (commented):
Frontend-only change, so this is correctness only — no style notes.

The direction is right, and the detector change is the strongest part of it. `block:<first member>` really is unique: `markConsumed`/`isRangeConsumed` guard a run's whole span before it is pushed, so runs are disjoint, and instances within a run start at distinct offsets. Nothing in `src/` parses the id format, and block ids never reach `sessionStorage` or the URL, so the reshape is safe end to end. Dropping the `withoutBlockMembers`/`setRevealedNodeIds` cleanup on this path is also safe — `deviceSubgraphs` already filters folded members through `collapsedMemberIds` and unknown ops through `operationById`, and the revealed ring is documented to persist for as long as the blocks stay open. 2661 tests, `tsc` and `eslint` clean on my run.

One real regression in the path this PR fixes, and two factual corrections.

## Should fix — the carry-forward renders the wrong graph first

`OperationGraphReactFlow.tsx:635` re-expresses unroll-all *after* the build it is meant to correct has already landed. Traced on `RENAMING_OPERATION_LIST`:

1. Toggle click → **one** build, carrying the stale ids `['block:3','block:6','block:9']`.
2. The worker's reply to that build is a **5-node, fully-folded** graph — every held id misses — and `onBuilt` commits it to React Flow.
3. *Then* `carryUnrollAll` unions in `block:2,5,8`, `buildOptions` changes, and a **second** build produces the 11-node unrolled graph.

It converges — no third build — but the reader gets a committed frame of the exact state this PR set out to prevent ("an unrolled graph comes back fully folded"), and pays two full layout passes. Per the header comment on `OperationGraphRebuildDiscipline.spec.tsx`, a layout is the most expensive thing this view can do. The fold-all path is clean by comparison: 1 build, 0 rebuilds.

The asymmetry the PR body identifies is also the fix. "Fold all" is `[]` because it names no instance — so make "unroll all" nameless too, rather than re-deriving it:

```ts
const [expandedBlockIds, setExpandedBlockIds] = useState<ReadonlySet<string> | 'all' | null>(null);
```

`'all'` survives any rebuild by itself, the first post-toggle build is already correct, and both `carryUnrollAll` and `detectedBlocksRef` go away. It is also the shape #2015 proposed — "the fold [as] a property of the grouping rather than a one-shot applied on click" — and it removes the PR body's "Known residual" for the all-unrolled case.

Neither new test can see this: both read `runBuild.mock.calls.at(-1)` and hand-deliver the graph themselves. `expect(runBuild).toHaveBeenCalledTimes(1)` after the click, or an assertion on the *first* delivered graph's node count, would pin it.

## Minor — `collapseWeightLoads` does not re-run detection

`OperationGraphReactFlow.tsx:890-891` calls it "the other filter that re-runs detection", and the PR body rests its case on the two filters having disagreed. It doesn't: `opGraphLayoutWorker.ts:95-107` keys detection on `${sourceVersion}:${hideDeallocate}:${grouping}` and calls `getKeptOperations(operations, hideDeallocate, candidatesOf())`. Weight-load collapsing is applied later, inside `buildOpGraph`, and never reaches the detector. So `hideDeallocate` is the only filter that can rename an instance, and the two were never in the same class. The new test at `tests/OperationGraphRebuildDiscipline.spec.tsx:1805` is still a worthwhile pin; the reasoning attached to it at `:1808` isn't.

The same check closes the third rebuild trigger #2015 asked about: a device-subgraph change cannot rename anything either, for the same reason — `deviceSubgraphs` is not part of the detection key.

## Minor — stale comment contradicts the diff

`tests/OperationGraphRebuildDiscipline.spec.tsx:1684`: "instance ids key on the run index and first member, which hiding deallocates leaves alone." Both halves are false as of this PR — the run index is removed in this very diff, and `RENAMING_OPERATION_LIST` two tests below exists precisely to show that hiding deallocates *does* move the first member. Reads like a leftover from the pre-review revision.

aidemsined:
Both corrections are right and are in fb495ac2. The third I've traced, pinned, and filed rather than fixed — reasoning below, because your suggested shape doesn't survive one case.

## `collapseWeightLoads` does not re-run detection — you're right, and I was wrong

Checked: `opGraphLayoutWorker.ts:99` keys detection on `${sourceVersion}:${hideDeallocate}:${grouping}` and passes only `hideDeallocate` to `getKeptOperations`. Measured either side of the switch:

```
collapseWeightLoads=true  -> ["block:3","block:6","block:9"]
collapseWeightLoads=false -> ["block:3","block:6","block:9"]
```

So the two filters were never in the same class, and the consistency argument I leant on is void. The behaviour fix stands on its own — the reset was unnecessary and harmful — but the reasoning was wrong in a code comment, a test comment and the PR body, and all three now say that `hideDeallocate` is the only control that can rename an instance. Your device-subgraph observation closes the third trigger #2015 asked about for the same reason, and that's recorded too.

## Stale comment — fixed

Leftover from the pre-review revision, exactly as you read it.

## The wrong first frame — reproduced, but the sentinel folds every weight fan

Confirmed: after the toggle the delivered graph is the 5-node folded one, then a second build converges on the 11-node unrolled one.

Where I came out differently is the fix. `expandedBlockIds` carries two kinds of id with **opposite defaults**:

- grouping blocks — `opGraphBuilder.ts:135` — fold only `if (hasFoldDecision && !expandedBlocks.has(id))`, so `undefined` renders them unrolled
- weight fans — `opGraphBuilder.ts:164` — fold on `!expandedBlocks.has(fan.instanceId)`, with **no** `hasFoldDecision` guard, so `undefined` folds every one of them

So `'all'` mapping to `undefined` on the wire would silently re-fold every fan the reader had opened — the bug `expandAllBlocks` unions to avoid, and #1980 before that. There's also nowhere for a fan id to live once the field is a scalar.

The fix you want is right, it just needs the two sets separated in `OpGraphBuildOptions` first, across the component, the worker and the builder. Filed as #2017 with the analysis, including the question of whether the fan half wants `hasFoldDecision` semantics of its own — "no decision yet" and "deliberately all folded" are currently indistinguishable for fans.

Meanwhile the defect is pinned rather than implied: a test asserts the 5-node frame and the extra build, labelled as documenting a limitation, and it flips when #2017 lands. The PR body now carries both the correction and this residual.

Happy to do #2017 first and rebase this on it if you'd rather not merge the transient frame — it's a bigger change but it deletes `carryUnrollAll` and `detectedBlocksRef` outright, so the end state is smaller than what's here.

smountenay-tt (approved):

PR #2022, opened 2026-09-17
MCP: answer where an operation came from, and what it was called with

### LINKED ISSUE #2021: Expose operation provenance to an agent over MCP: arguments and call site
Round one ([#1995](https://github.com/tenstorrent/ttnn-visualizer/issues/1995)) answered where the time went. Round two ([#2012](https://github.com/tenstorrent/ttnn-visualizer/issues/2012)) answered what the run allocated. Both hand back an **operation id**, and nothing turns that id into something an agent can act on in the model code it is editing.

`top_ops` says operation 15 dominates. `memory_profile` says operation 15 holds the L1 peak. Neither can say what it was called with, or which line of which file called it.

## Tools

- `operation_provenance(handle, operation_id)` → the operation's arguments as name/value pairs, and its call site: the innermost frame's file, line, function and source line, plus the frame chain outward from it.
- `find_operations(..., called_from=...)` → the inverse. Which operations came from a given file or function, which is how a bring-up session actually proceeds once one of them is slow ("every conv2d from `run_downsample_if_req`").

## Why these two, measured

Across the 86 local captures:

| table | present | non-empty |
|---|---|---|
| `operation_arguments` | 86 | **84** |
| `stack_traces` | 86 | **78** |
| `captured_graph` | 84 | 80 |
| `edges` / `nodes` | 86 | 29 / 30 |
| `errors` | 40 | 4 |
| `global_tensor_comparison_records` | 86 | **0** |
| `local_tensor_comparison_records` | 86 | **0** |

Arguments and stack traces are the two unread tables that are populated nearly everywhere, so the slice is buildable and testable on real reports rather than on a fixture.

**An accuracy slice is not buildable.** Both tensor-comparison tables exist in all 86 captures and are empty in all 86, so PCC questions would be the Blaze join again — a tool over a join key that does not exist. Recorded here so it is not proposed a third time.

## Shape

Traces are standard Python tracebacks, innermost frame first — the `ttnn.conv2d` call site is frame 0 — so the call site needs no heuristic about which frame is "the user's". Median trace is 1,231 chars and the longest local one is 1,446, so the frame chain is bounded by construction but the raw text is not returned wholesale.

Argument values are mostly short (median 6 chars) with a tail: p95 835, max 844, because a `Conv2dConfig` or a tensor repr lands in one. Values are truncated at a stated cap with a flag when truncated, which is the same rule as every other bounded field on this surface.

Both tools carry the rank scope and its caveat like the rest of `agent/operations.py`, and `load_report` reports them per the tables they read.

## Note

`operation_detail` keeps its current shape. Folding 1.2 KB of trace and twenty arguments into the response that already carries tensors and allocations would make the largest payload on the surface larger still, against the bounded-response rule the package exists to hold.

aidemsined:
## Summary

Round one ([#1996](https://github.com/tenstorrent/ttnn-visualizer/pull/1996)) answered where the time went. Round two ([#2014](https://github.com/tenstorrent/ttnn-visualizer/pull/2014)) answered what the run allocated. Both hand back an **operation id**, and nothing turned that id into something to act on in the model code being edited.

`top_ops` says operation 15 dominates. `memory_profile` says operation 15 holds the L1 peak. Neither could say it is a 64→256 1×1 convolution called from `run_downsample_if_req` at `ttnn_functional_resnet50.py:135`.

Closes #2021

| Tool | Answers |
|---|---|
| `operation_provenance` | one operation's arguments, and the innermost stack frame's file, line, function and source line, plus the chain outward |
| `find_operations(called_from=…)` | the inverse: which operations came from a given file or function |

**The two halves of the surface do not share an id space**, which review caught me asserting otherwise. `top_ops` and `diff_reports` return row numbers of the performance CSV; the database tools take `operations.operation_id`. Both are small integers in overlapping ranges, so passing one where the other belongs does not fail — it answers about a different operation. The docs and the tool schema now say so and name `find_operations` as the way across.

Driven end to end on a real capture, this is the loop:

```
memory_profile L1 peak  -> op 15 (ttnn.conv2d) at 487,424 bytes/bank
operation_provenance    -> run_downsample_if_req() at ttnn_functional_resnet50.py:135
                           in=64 out=256 k=(1,1) stride=(1,1) batch=16
find_operations(called_from='run_downsample_if_req') -> 12 ops
```

The third step is the one that matters most in practice: once an operation is slow, the question is never about that operation but about every operation from the same helper. `called_from` matches anywhere in the stack, so a layer module finds the operations it called *through* helpers rather than only those it called directly.

### The slice was chosen by measurement

Of the tables no tool reads, across the 86 local captures:

| table | present | non-empty |
|---|---|---|
| `operation_arguments` | 86 | **84** |
| `stack_traces` | 86 | **78** |
| `captured_graph` | 84 | 80 |
| `edges` / `nodes` | 86 | 29 / 30 |
| `errors` | 40 | 4 |
| `global_tensor_comparison_records` | 86 | **0** |
| `local_tensor_comparison_records` | 86 | **0** |

Arguments and traces are the two that are populated nearly everywhere, so the slice is testable against real reports rather than only a fixture.

**An accuracy slice is not buildable, and that is recorded.** Both tensor-comparison tables exist in all 86 captures and are empty in all 86, so PCC questions would have been the Blaze join again — a tool over a join key that does not exist. It is stated in the issue and in `docs/src/agent-tools.md` so it is not proposed a third time.

### The call site needs no heuristic

Frames are recorded innermost first, so frame 0 is the `ttnn.<op>` call in the model code. `_parse_frames` deliberately agrees with `src/functions/stackTraceSource.ts`, which takes the first `File "..."` and the first `line N` for the operation details panel: two answers to "where is this operation from" that disagreed would be worse than one.

Checked rather than assumed — all 302 traces in a local capture parsed through both implementations, **zero mismatches**.

### Bounds

`called_from` filters in SQL. Reading the traces into Python to substring-match them cost +80 MB on a capture holding 70 MB of trace text and 808 MB on a report ten times that — which is verbatim what `query_buffer_totals_by_operation` was written to avoid. Measured after: 701 ms and +8 MB on the same capture.

`operation_provenance` takes a `limit` on the outward chain, defaulting to 10. Without it the median response was 10.5 KB against `operation_detail`'s 1 KB, and on one capture 31 of 43 frames were the pytest harness that invoked the model — 61% of the payload, for a tool whose job is finding the model code. The same response is now 2,986 bytes, with `chain_length` reporting the untruncated depth so the cap is detectable.

Argument values are capped at 500 characters with `truncated`, `full_length`, and `truncated_count` for how many were cut. Truncation is the normal case, not the exception: across the 204,636 argument values in the local captures, 28% run past the cap and the longest is 177,717 characters, with one report truncating 60% of its arguments.

**Two different absences are reported apart.** A capture that records no trace says so. A capture that records a trace this parser cannot read says the call site is *unknown* rather than absent — and that is live, not hypothetical: 1,666 of 60,673 local trace rows are native C++ backtraces, and 8 of 86 captures write them for every operation, disjoint from the 8 that write no trace at all. Half of those 16 used to get a bare `null` with no explanation.

## Testing

Nineteen tests in `test_agent_operations.py`, each verified by mutation.

`TestFramePariy` pins the agreement with `src/functions/stackTraceSource.ts` — which previously existed only as three prose claims and a shell command I had run, with nothing enforcing it. It records the divergences too: the frontend takes its file and its line number from two separate regexes over the whole trace, so they need not come from the same frame, and a quote inside a path makes its greedy `(.*)` report a location this parser declines to guess at. Review confirmed no difference across 15,374 real traces.

Four of my tests were weaker than they looked and were strengthened: two were unfalsifiable until the fixture gained a second trace origin (with one traced operation, "match the needle" and "match every operation with a trace" return the same set, and case-insensitivity held only because every needle was lowercase), and two asserted only an empty result or only an exclusion, both of which pass for an implementation that matches nothing at all.

Two existing tests in `test_agent_tools.py` needed the ninth tool in the inventory and the transport list.

`1481 passed`; mypy clean over 96 files; black, isort and flake8 clean on the files this touches.

## Review round

Three reviewers; the findings that mattered were all mine to own. Beyond the id space, the SQL filter, the payload and the two absences above: the rank guard named `operations`, the one table `report_has_rank_column` inspects, so it could never fire — reproduced serving a rank-1 call site under a `rank: 0` caveat; a NULL `stack_trace` reached `splitlines` and gave the agent an `AttributeError` while the `called_from` path six lines away guarded it; `query_stack_traces` let a missing table surface as a raw `no such table`; `find_operations`' `TOOL_TABLES` entry did not list `stack_traces` although `called_from` reads it; a whitespace-only `called_from` returned everything while echoing the filter; the frame window could borrow a neighbouring frame's source line, which on frame 0 points at another file's code; and a NULL argument value read as an empty string, collapsing "not recorded" into "recorded as empty".

## Follow-ups

`operations.py` is now ~900 lines, and this tool needed its private helpers (`_profiler_db`, `_rank_scope`, `_scoped`, `_caveats`), which is why it lives there rather than in its own module. Adding a ninth tool also meant edits in three files again. Both are the case #2013 already makes for a shared read-only open and tool descriptors.

Stack traces carry absolute paths from the capture host. Fine for a local developer tool, and worth knowing before this surface is exposed anywhere hosted.

Two implementations of the call-site parse remain, in Python and TypeScript. The real fix is the backend owning it and exposing the parsed call site to the frontend so those regexes can go; there is also no index on `stack_traces.operation_id`, so a single-operation lookup is a full scan of the largest text column in the report. Both belong with #2013.


smountenay-tt (commented):
Solid slice, and the measurement behind the choice of tables plus the SQL-side `called_from` filter are the right calls. The design is sound: provenance reads the two tables real captures actually populate, the call site needs no heuristic, and `_parse_frames`' agreement with `stackTraceSource.ts` is now pinned rather than asserted.

Two things must change before merge, both in the "number that can mislead" class this surface exists to prevent — one reports a count that is wrong whenever the cap binds, the other answers `0` where the honest answer is "this capture cannot say". I verified both against the branch rather than reading them off the diff.

## Blockers

**1. `returned_frames` reports the untruncated chain, so the cap it exists to expose is invisible** — `backend/ttnn_visualizer/agent/operations.py:404-405`

The two fields are the same expression:

```python
"frames": frames[1 : cap + 1],
"returned_frames": max(len(frames) - 1, 0),
"chain_length": max(len(frames) - 1, 0),
```

Run on the branch, with a 30-frame trace and `limit=5`:

```
len(frames) in payload: 5
returned_frames: 29
chain_length: 29
```

The comment above it says it "counts the chain this response returned, so comparing it against `len(frames)` detects the cap" — it does the opposite: an agent reading `returned_frames: 29` beside five frames is told the chain it did not get. This is the same defect shape as the `rank: 0` caveat pointing the wrong way, and AGENTS.md § Agent-facing tools / [CONVENTIONS.md#agent-facing-tools](https://github.com/tenstorrent/ttnn-visualizer/blob/dev/CONVENTIONS.md#agent-facing-tools) (#1883) is the rule it breaks.

Compute the slice once:

```python
chain = frames[1 : cap + 1]
...
"frames": chain,
"returned_frames": len(chain),
"chain_length": max(len(frames) - 1, 0),
```

`test_the_outward_chain_is_capped` (`test_agent_operations.py:1005`) asserts `chain_length == 29` and `len(result["frames"]) == 5` but never `returned_frames`, which is why nothing failed — and `test_the_call_site_is_the_innermost_frame` asserts `returned_frames == 2` in the one case where the cap does not bind, so both values agree there. Adding `assert result["returned_frames"] == 5` to the capped test is the case that pins it.

**2. `called_from` answers `match_count: 0` with no caveat when the capture has no usable `stack_traces`** — `operations.py:457-462`

Both the empty-table and absent-table paths collapse into "no operations came from that helper", on the branch:

```
no stack_traces table  -> {'called_from': 'resnet.py', 'match_count': 0, 'returned': 0}  caveat: None
empty stack_traces     -> {'called_from': 'resnet.py', 'match_count': 0, 'returned': 0}  caveat: None
```

The empty case is the live one — the 8 of 86 captures the PR body counts. An agent asked "which operations came from `run_downsample_if_req`" gets a confident `0` back with its filter echoed, and the fact that the capture records no call sites at all never reaches it. `operation_provenance` gets this exactly right in the function above (`operations.py:411-417`, "recorded no stack trace ... Older reports omit them"), so the two halves of the same feature disagree about the same absence.

Cheapest fix that matches the neighbour: when `origin` is set, say so in the caveat rather than implying a narrowing —

```python
if origin:
    _refuse_unattributable(queries, scope, "stack_traces")
    ids = queries.query_operation_ids_by_stack_trace(origin, rank=scope.rank)
    from_origin = set(ids)
    # plus a cheap "does this capture hold any trace at all" probe, e.g. a
    # `SELECT EXISTS(SELECT 1 FROM stack_traces)` helper beside the query above
```

and append "This capture recorded no stack traces, so `called_from` can match nothing — the result is not evidence that no operation came from there." Refusing the filter outright would also be defensible; a bare `0` is not.

## Should fix

**3. `called_from` is a `LIKE` pattern, not the substring the schema advertises** — `backend/ttnn_visualizer/queries.py:368-370`

`f"%{needle.lower()}%"` goes in unescaped, so `%` and `_` are wildcards. Underscores are pervasive in the exact names this tool asks for. On the branch, a needle of `run_downsample_if_req` also matches an operation whose helper is `runXdownsampleYifZreq`:

```
needle 'run_downsample_if_req' -> [7, 2]     # op 7 is a different helper
needle '%'                     -> [1, 2]     # every operation, called_from echoed
```

`name_contains` beside it is a real Python `in` substring, so one tool now has two filters with different matching rules, and the schema calls both "case-insensitive substring". False positives are unlikely to collide often, but a `%` reaching the needle turns the filter into "everything" while the response still names it.

```python
escaped = needle.lower().replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
query = ("SELECT DISTINCT operation_id FROM stack_traces "
         "WHERE lower(stack_trace) LIKE ? ESCAPE '\\'")
params: List[Any] = [f"%{escaped}%"]
```

**4. Requiring `stack_traces` for `find_operations` withdraws the name search from captures that can answer it** — `backend/ttnn_visualizer/agent/handles.py:105`

The comment justifies it as avoiding "a raw `no such table`" — but that can no longer happen: this same commit gives `query_operation_ids_by_stack_trace` its own existence guard (`queries.py:363-364`) and `query_stack_traces` another (`queries.py:432-433`). What the entry costs is real, and the PR's own test change records it: a capture holding only `operations` now reports `answerable == []` (`test_agent_tools.py:148-157`), where it used to be told it could search by name. That is also the opposite of the call made two entries below for `operation_provenance` — "requiring `stack_traces` would report the tool unanswerable on a report it can partly answer" — which is the better argument, and it applies here too.

Suggest reverting to `frozenset({"operations"})` and letting finding 2's caveat carry the absence.

**5. Two sentences in the docs are now false** — `docs/src/agent-tools.md:47`, `:55-58`

- "The last four read the profiler report's SQLite database" — there are five now, and the four counted from the end exclude `find_operations`.
- "a truncated capture that holds `operations` but not `buffers` is reported as answering `find_operations` and nothing else, rather than advertising four tools and failing three of them" — with the `TOOL_TABLES` change in this PR that capture answers nothing and five tools are advertised. If finding 4 is taken, the sentence needs only "five"/"four" corrected; if not, it needs rewriting to describe the new behaviour.

## Nits

**6.** `class TestFramePariy` → `TestFrameParity` (`test_agent_operations.py:1047`).

**7.** `MAX_ARGUMENT_VALUE_CHARS` in `__all__` lands between the two `BUFFER_*` entries (`operations.py:883`); after `BUFFER_TYPE_NAMES` reads better.

**8.** `called_from` matches the whole column, which includes the indented source lines, not only the `File "..."` frames — so a needle like `ttnn.conv2d` matches operations whose traces merely *show* that call. Defensible, but the schema says "substring of the call stack: a file name or a function", which is narrower than what it does.

## Optional follow-up, not this PR's debt

`docs/src/agent-tools.md` has no parity test, which is why both stale sentences in finding 5 survived a green run — `event_logging` and `startup_requirements` each have one. A test pinning the tool table to `_tool_table`'s keys would have caught the missing row and, with a count, the "last four" clause too. Worth folding into #2013 alongside the shared read-only open and the tool descriptors.

aidemsined:
All of it addressed in bd4c389e. I reproduced each against the branch before changing anything, and both blockers were exactly as you described.

## Blockers

**1. `returned_frames` reported the chain it did not send.** Confirmed — `limit=5` on a 30-frame trace gave five frames and `returned_frames: 29`, because it and `chain_length` were the same expression. Worse than a wrong number: the comment above it said the pair existed to make the cap detectable, which is the opposite of what it did. `returned_frames` counts the slice now, and the capped test asserts all three fields — you were right that asserting `chain_length` and `len(frames)` but never `returned_frames` is why nothing failed.

**2. `called_from` answered `0` with the filter echoed back.** Confirmed on both the empty-table and absent-table paths. `report_records_stack_traces` asks before searching and the caveat says an empty result is not evidence that no operation came from there, matching what `operation_provenance` already said for a single operation.

## Should fix — all three taken

**3. The needle was a `LIKE` pattern.** This one I'd introduced moving the filter into SQL, and it is worse than the false-positive rate suggests, because `_` is in *every* Python function name:

```
before:  'run_downsample_if_req' -> [2, 7]   # 7 is runXdownsampleYifZreq
         '%'                     -> [2, 7]
after:   'run_downsample_if_req' -> [2]
         '%'                     -> []
```

Escaped with an explicit `ESCAPE`, so `%` now matches nothing — correct for a substring no trace contains.

**4. `find_operations` requiring `stack_traces`.** Agreed, and your argument is the one that settles it: the raw `no such table` that justified it cannot happen any more, since this branch gave both trace queries their own guard — and it contradicted the call made for `operation_provenance` two entries below, which is the better one. Reverted to `operations`; the capture that holds only `operations` is told it can search by name again.

**5. Both stale doc sentences.** Fixed, and your optional follow-up caught why they survived: there is no parity test for that page. I've put it on #2013 with the shared read-only open and the tool descriptors.

## Nits

**6.** `TestFrameParity`, in the test and in the README line that points at it. **7.** `__all__` reordered. **8.** The schema now says `called_from` matches the whole trace text — frames and the source lines under them — rather than implying only file and function names.

## From Copilot, one you'd both have wanted

`_refuse_unattributable` treated an *absent* table as an unranked one, so on a ranked report with no `stack_traces` table `operation_provenance` was advertised and then refused every call, instead of returning the arguments plus the documented no-trace caveat:

```
answerable: ['find_operations', 'operation_provenance']
provenance REFUSED -> this report carries `rank` on `operations` but not on `stack_traces` …
```

Only rows that exist can be misattributed. Copilot also found the `top_ops`-id implication still sitting in the `operation_provenance` docstring and in a docs paragraph that told readers to pass that id here — the correction had reached the schema and the overview but not those two.

Four new tests; every one of the six behaviour changes fails the suite when reverted, including the absent-table guard and the two different "no traces" paths. 1485 passed, mypy clean over 96 files.

smountenay-tt (commented):
Re-reviewed at `bd4c389e`. Both blockers are cleared and all five findings are fixed properly rather than papered over — the `returned_frames`/`chain_length` pair now says two different things, and the `called_from` caveat draws the same distinction `operation_provenance` already drew. Verified on the branch rather than from the diff:

```
limit=None: len(frames)=10 returned_frames=10 chain_length=29
limit=5:    len(frames)=5  returned_frames=5  chain_length=29
limit=100:  len(frames)=29 returned_frames=29 chain_length=29

absent stack_traces -> match_count 0 + "records no stack traces ... not evidence"
empty  stack_traces -> match_count 0 + same caveat
real miss (capture has traces, needle matches nothing) -> match_count 0, no caveat  ← the important half

called_from 'run_downsample_if_req' -> [2]        (no longer matches runXdownsampleYifZreq)
called_from '%'                     -> [7]        (the trace holding a literal 50%)
called_from '50%' -> [7]   called_from '\' -> []
```

The caveat not firing on a genuine miss is the part that was easy to get wrong, and the `%` needle now matching a literal percent rather than nothing is the right reading of "substring". `1473 passed` here, black/isort/mypy clean over the changed files.

Taking Copilot's `_refuse_unattributable` finding was the right call, and "only rows that exist can be misattributed" is the correct one-line statement of why — an absent table cannot misattribute anything, and refusing a tool the inventory advertised was the worse failure.

Nothing blocking left. Four small things, all optional:

**Should fix — a docs cross-reference to a note that isn't there.** `docs/src/agent-tools.md:149` sends the reader to "the id-space note above", and there is no such note in the file: `:114` is about operation ids restarting per rank, and the sentence at `:148-149` is itself the only place the CSV-row-vs-database-id distinction appears. Since that sentence already states it ("their `id` is a row of the performance CSV and does not belong here"), dropping the four words is the whole fix — or add the note it points at, given how much of this round was about that id space.

**Nit — the rank half of the "cannot say" fix is still open.** `report_records_stack_traces()` (`queries.py:384`) takes no rank, so it answers for the whole capture. On a ranked report whose `stack_traces` carries `rank` with rows for only some of them, the misleading zero survives at the ranks that have none:

```
rank=0 (no traces at this rank) -> match_count: 0, caveat: only the generic multi-host one
rank=1 (traces here)            -> match_count: 1
```

As narrow as the case `_refuse_unattributable` insures against — it needs the 2-of-86 schema plus per-rank trace coverage — but it is the same defect class, and `merge_rank_filter` already no-ops for a table without the column, so `report_records_stack_traces(rank=scope.rank)` threaded from `find_operations` covers it without a new guard.

**Nit** — `test_an_absent_table_does_not_refuse_a_ranked_report` (`test_agent_operations.py:1228`) exercises `operation_provenance` but sits in `TestFindOperations`; it reads as a provenance case.

**Nit** — the two rewrapped docs paragraphs (`agent-tools.md:47-50`, `:148-150`) now carry ~200-character lines where the rest of the file wraps at about ninety. `*.md` is in `.prettierignore`, so nothing enforces it either way.

One thing I raised with myself and dropped: `agent/operations.py:189` calls `queries._check_table_exists`, which is private to `DatabaseQueries` — but `views.py` already does that in four places, so it is house style here, not something this PR introduced.

smountenay-tt (approved):

Issue #1987, opened 2026-09-08
Weight fans are second-class in the block model


aidemsined:
A weight fan (#1980) is emitted as a `RepeatBlockInstance` and folded by the same machinery as a repeat or a layer, but it is deliberately kept out of `graph.blocks` so the toolbar's grouping count keeps its meaning. Everything downstream that reads `blocks` therefore does not know fans exist, and four separate behaviours fall out of that one omission.

- **The info panel describes the wrong thing.** `blockByMemberOperationId` is built from `graph.blocks`, so `selectedBlock` is `null` for a fan node and the panel renders the single-operation view for the fan's *first member* — one `ttnn.from_torch`, with its inputs, outputs and "Recenter on…" — while the canvas says "N weight loads". `selectedPerfDeviceTimeNs` reports that one member's device time rather than the fan's sum, contradicting the block node's own meta line.
- **An unrolled fan cannot be re-folded.** The expander pill lives on the block node, which no longer exists once the fan is unrolled, and the other route — double-clicking a member — goes through the same member map. The only ways back are Fold all, a grouping switch, a hide-deallocate toggle or a report swap, all of which destroy unrelated state; and when the active detector found nothing, Unroll/Fold are not rendered at all, so there is no route back.
- **A URL-named operation buried in a fan is not unrolled.** The reveal path only inspects grouping blocks, so the "I'm going to click op 49, where has it gone" case #1944 fixed for the other two kinds is still open for fans.
- **Folded fan members keep shipping their device subgraphs.** `collapsedMemberIds` walks `detectedBlocks` only, so an op whose device operations were expanded before a fan folded it keeps its subgraph in the array handed to the worker. Nothing renders it — the fan branch wins in the builder — but one operation's frame stream runs to thousands of nodes and it varies the layout cache key.

Underneath all four: `isBlockExpanded` decides "this is a fan, so default it folded" from *absence* from `detectedBlockIds`. That is the negative-derivation `OpGraphBlockKind` was introduced to avoid — the enum's own comment says a prefix check "would silently mislabel the next detector someone adds", and absence from an array does exactly that. Any future block kind the builder does not put in `blocks` silently inherits weight-fan semantics.

[...]

PR #2024, opened 2026-09-21
Key a weight fan on its membership, and carry an unroll across a fold

### LINKED ISSUE #1988: A weight fan's id still moves when a grouping fold merges two fans

aidemsined:
## Summary

A weight fan's members are grouped by the node they feed, and which node that is depends on what grouping just folded — so a fold that merges two consumers into one block node merges their fans. Keyed on its first member, the survivor of `[2,3]` plus `[1]` silently renamed itself, the reader's unrolled fan re-folded under them, and the id they had opened was stranded in a set nothing prunes.

That is the defect the original re-key in #1980 was meant to end, reached by a different route.

Closes #1988

Reproduced against the real builder before changing anything — sources 2 and 3 feed op 12, source 1 feeds op 11, and both consumers sit in the same repeating unit:

```
grouping unrolled : weights:2-3      two consumers, so two fans
grouping folded   : weights:1        one node, so one fan of [1,2,3]
```

## The fix

**The id spells out its membership, ascending.** It now changes exactly when the membership changes, so it can never silently name a different fan — and the members are *recoverable from it*, which is the part that makes the rest work.

**The builder treats a fan as unrolled when the expansion set holds its id, or one whose membership contains this fan's or is contained by it.** Because the id decodes, a fold can match a remembered decision against a fan that did not exist when the decision was made, with no history of the previous build kept anywhere.

Containment in either direction, because that is how a fold reshapes a fan: a merge absorbs the remembered one into a larger fan (`{2,3}` into `{1,2,3}`), a split breaks it into smaller ones (`{1,4}` out of `{1,2,3,4}`), and both are still what the reader opened.

I first wrote this as a bare intersection and review showed the difference is reachable: a decision about `{1,2}` would unroll a later `{1,7,8}` sharing only operation 1, showing weight loads the reader had deliberately left folded. What containment does keep, deliberately, is the merge of two fans with opposite decisions — they become one node and the open one wins. One node cannot be half open.

Spelling the members out rather than hashing them is affordable because fans are small — the largest across the 86 local captures holds three members, and #1980's own example is six per attention span. A graph folded to a single block is the pathological case, where every source in the report merges into one fan and the id grows with it.

## What this deliberately does not do

Folding a fan ought to drop the remembered ids that cover it, or the next build re-opens it. Nothing here does that, because there is no route to that state: `blockInstanceId` is set only on the collapsed block node, and `blockByMemberOperationId` is built from `detectedBlocks`, which holds no fans — so an unrolled fan's members link back to nothing and double-clicking one rebuilds nothing.

That was a claim in a comment; it is a test now. When #1987 makes fans first-class the test fails and the cleanup has to arrive with it.

**Worth knowing before merge:** this means a reader who opens a fan cannot close it again short of a report change, and `Collapse weight loads` is not the escape either — it is wired straight to its setter and touches no ids. That predates this change, but this change removes the one accidental escape: previously a grouping fold renamed the fan so the stale id missed and it re-folded itself. Trading a lost decision for a decision that cannot be revised is a real trade, and I would rather it were a decision than a side effect. Recorded on #1987.

Two more observations for that issue, found while doing this: fans never appear in `graph.blocks` at all, so `blockKind` reads `undefined` on the summary, and the expansion set has no route to shed a fan id once added.

## Testing

Builder tests for each direction the carry has to get right: a fan that survives a merge, one that survives a split, one whose remembered id merely shares a member and must stay folded, and one whose reader opened nothing — so the carry can be neither "always unrolled" nor a bare intersection. Plus unit coverage of the decode, including the malformed ids it used to guess at.

Every part is mutation-checked: removing the carry, treating every fan as unrolled, reverting to the first-member id, widening containment back to intersection, dropping the split direction, and restoring the loose prefix parse all fail the suite.

Two assertions needed strengthening before they could tell the difference. The predicate test passed for both "any" and "all" while the remembered fan was an exact subset, and the decoder's rejection path was exercised only through the prefix check — which is the one part of it that was right.

Eight hardcoded ids in existing specs moved with the format, including the #1980 test whose premise this narrows.

2724 passed; tsc and eslint clean.


aidemsined:
Good catch, and correct — fixed in 61518a4b.

The hoist I added in the previous round validated each remembered id and then stored the *string*, so `weightFanIdCovers` re-parsed it and allocated two sets for every (fan, remembered id) pair. The work I had claimed to move out of the loop was still in it.

Measured before changing anything, 200 fans against 500 remembered ids:

```
per-pair parse: 6.08 ms | decode once: 1.22 ms | same answer: true
```

Essentially all of the 6 ms was parsing.

The containment rule still has one implementation rather than two. `weightFanMembersCover` takes decoded memberships and the fan's own member set, and `weightFanIdCovers` is that with a decode in front — so the hot path never touches the string, callers holding an id keep the convenience, and the rule cannot drift between them. This file has been bitten by a rule living in two places before, so that seemed worth the small interface.

Three mutations fail: widening containment back to a bare intersection, dropping the split direction, and stopping the id form delegating to the shared rule. 2724 passed, tsc and eslint clean.

smountenay-tt (commented):
Frontend-only change, no API contract touched, so I held this to bugs, security and release-blockers. I found none, and the design reads sound: keying on the whole membership makes the id a function of the thing it names, and the decode is what lets a merge be recognised without keeping a previous build around.

I re-ran the claims rather than taking them:

- `tsc --noEmit` clean, eslint clean on the touched files, 2724 passed — matches the body exactly.
- All five mutations fail as stated: intersection instead of containment (2), dropping the split direction (2), dropping the carry in the builder (2), reverting to the first-member id (16), and loosening the decoder back to `startsWith` + `Number` (1).

And I checked the three properties the fix actually rests on, since two of them are claims in comments rather than code:

- `blockByMemberOperationId` is built from `detectedBlocks` (`OperationGraphReactFlow.tsx:565`) and the reveal path is its only other reader (`:1042`), so `buried` can never be a fan. The omitted cleanup really is unreachable, and the new test pins it at the right place.
- **The one thing containment would have widened dangerously is already closed.** A remembered membership that unrolls any fan contained by it would be actively harmful across a report swap — operation ids restart per report, which is #1674's lesson — and a low-id remembered fan like `weights:1-2-3` would land on exactly where `to_device` sources sit in every other report. `setExpandedBlockIds(null)` at `:449` resets it on a scope change, for that stated reason. Worth knowing that this is now load-bearing for the carry, not just for op expansion.
- `Collapse weight loads` is wired straight to its setter (`:1658`) and, per the note at `:1033`, never reaches the detector, so it cannot invalidate an id. Your "not the escape either" is accurate.

The decoder is tight enough: `^weights:\d+(?:-\d+)*$` rejects the empty, double-hyphen, `1e3` and `0x10` cases, `fanIdOf` never emits a form it cannot read back, and `expandedBlockIds` is component state with no URL or storage backing — so there is no untrusted id path and no persisted key shape to migrate. A single-member `weights:N`, which would cover any fan holding that op, has no route in either, since `MIN_FAN_MEMBERS` is 2 and every id entering the set comes from a rendered fan node.

One thing to state plainly rather than object to, since the body already discloses it: the net effect is that #1987's "an unrolled fan cannot be re-folded" goes from *usually escapable by accident* to *not escapable short of a report swap*. Fold all and a grouping switch both keep fan ids through `withoutGroupingBlocks`, and the id no longer moves under a fold, so the three routes back that #1987 lists are closed. That is the right trade — an unpredictable re-fold was the bug — and you have recorded it on #1987. Flagging only to confirm it is understood as a trade and not a side effect, which is how you put it.

No blockers from me.

smountenay-tt (approved):

Issue #2035, opened 2026-09-22
Add an MCP page to the navigation rail


aidemsined:
Add an MCP item to the navigation rail, opening a page that explains how to run the agent server and lists the tools it exposes.

The server ships today (`ttnn-visualizer-mcp`, #1995) but is only discoverable by reading the docs site, so nobody using the app knows it is there.

## Page content

- How to run it and register it with a client (`claude mcp add ttnn-visualizer -- ttnn-visualizer-mcp`).
- The tool table: what each tool answers.
- The caveats that change how an answer is read — per-bank figures, one rank at a time, totals on a partitioned run.

All of this is already written in [`docs/src/agent-tools.md`](docs/src/agent-tools.md). Render that file rather than restating it: a second copy will drift, and the drift is silent. `operation_provenance` is the live example — it is on `dev` and in no release, so any hand-written list is already wrong for whoever is reading it.

## What is in the way

Two things, neither solved:

- **`docs/` is not in the wheel.** `[tool.setuptools.packages.find]` takes `backend` only, and `package-data` covers `alembic` alone, so reading the file from disk at runtime works in a source checkout and 404s for every PyPI install.
- **No markdown renderer on the frontend.** Nothing in `package.json`, and nothing in `src/` imports a `.md`.

The option that avoids both is a build-time `?raw` import so Vite bakes the file into the bundle, plus a renderer. The doc stays the single source, and it works offline. The alternatives — shipping `docs/` as package data behind an endpoint, or linking out to the docs site the way `NPEProcessingStatus.tsx:110` does — either move docs into the Python package or reintroduce version drift.

## Wiring

- `NAVIGATION_ITEMS` in `src/definitions/NavigationItems.ts`, `ROUTES` in `src/definitions/Routes.ts`, route in `src/routes/routeObjectList.tsx`.
- `NavRequirement.NONE` — the page needs no report.
- `className` needs a matching selector in `scss/mixins/_navigation.scss`.

## Open

- **Server mode.** When the app is hosted, "run `ttnn-visualizer-mcp`" is about the reader's own machine, not the server they are looking at. Either the page says so or the item takes `hiddenInServerMode`.
- Where it sits in the rail.

## Related

[...]

Issue #2034, opened 2026-09-22
memory_profile tells agents its peak answers the OOM question; it excludes circular buffers and intra-op tensors


aidemsined:
## Problem

`memory_profile`'s docstring tells an agent its per-operation peak is "the question an out-of-memory failure actually asks". It is not. The tool reads the `buffers` table, which records what is live *at* each operation — so circular buffers and intra-op intermediate tensors are absent from it entirely, and both are large.

Measured on `resnet50_may08_1841 2`:

| | L1 peak | at operation |
|---|---|---|
| `memory_profile` (per-operation sum over `buffers`) | **487,424** | a plateau of 18; it names 15 first |
| replaying the captured graph, reconciled against the snapshot | **1,159,840** | 29 |

2.4x out, and not the same operations. The missing 672,416 B is 601,664 of circular buffers and 281,856 of intra-op intermediates, less what the snapshot already held.

Across the local corpus circular buffers are **52–100%** of peak L1 wherever a peak has any, and two reports (`bge_m3`, `test_ttnn_moe`) have essentially no L1 *tensors* at all — their L1 is almost entirely circular buffers, so `memory_profile` reports a peak near zero for a model that is saturating L1.

This is the same defect as #2026, in the agent surface rather than the perf table.

## Why it matters more here than in a chart

`CONVENTIONS.md` § Agent-facing tools requires that **a number that can mislead carries its caveat in the response**. This one carries the opposite: a docstring asserting it answers the OOM question. An agent has no way to discover the gap — there is no second number on screen to disagree with, and the tool's own prose tells it to stop looking.

`docs/src/agent-tools.md:43` and `:80-84` describe the same figure and are equally silent about it. The existing caveats there are good and specific (the per-bank denominator argument), which makes the omission read as "we thought about the units and this is what's left".

## Ask

Scope this to the caveat only — say what the number excludes, in the docstring, in the tool's response, and in `docs/src/agent-tools.md`. Something like: the figure counts tensor allocations recorded in `buffers`; circular buffers and tensors allocated and freed inside one operation are not in that table, and on the local corpus they are 52–100% and 24–38% of the real peak respectively, so this is a floor rather than the peak.

[...]

PR #2038, opened 2026-09-22
Make documenting a tool part of adding one

### LINKED ISSUE #2036: Require new MCP tools to be documented in agent-tools.md
State in `AGENTS.md` that adding an MCP tool means documenting it in [`docs/src/agent-tools.md`](docs/src/agent-tools.md) in the same change.

Nothing says this today. The closest thing is the agent README's "**Adding a tool means adding it to `_tool_table`**", which stops at registration — the user-facing doc is linked from `AGENTS.md:157` as reference material, never as an obligation.

The two are in sync right now: all nine tools appear in `_tool_table` and in the doc's table. This is to keep it that way, not to fix a gap.

## Why it is worth a rule

The doc is not a description of the surface, it is the only description. An agent picks a tool from the table and reads the caveats — per-bank figures, one rank at a time, totals on a partitioned run — to know how to read what comes back. A tool that is registered but undocumented is invisible to the client that would use it, and a caveat that never reaches the doc is one an agent acts without.

#2034 is the adjacent failure already on the books: a tool's own docstring describing behaviour it does not have. The surface drifts in one place already.

If #2035 lands, the doc is rendered in the app and the drift becomes user-visible.

## Where

- `AGENTS.md:157`, the Agent-facing tools bullet — add the obligation next to the existing links.
- `CONVENTIONS.md:1129` — "New tools go in that package and follow both" extends to the doc.
- `backend/ttnn_visualizer/agent/README.md` — the `_tool_table` line is where a developer adding a tool actually looks.

## Worth considering

A parity test would enforce this rather than ask for it, and the repo has the pattern twice — `test_event_logging_docs_parity.py` and `test_startup_requirements_docs_parity.py`. Pinning the doc's tool table against `server._tool_table` is a small test and would catch the omission the wording only discourages. Out of scope here, but the wording should not be mistaken for enforcement.

aidemsined:
Makes documenting a tool part of adding one.

`docs/src/agent-tools.md` was linked from the guidance as reference material and never as an obligation. The nearest thing to a rule was the agent README's "**Adding a tool means adding it to `_tool_table`**", which stops at registration.

## Why it needs saying

That page is not *a* description of the MCP surface, it is the only one an agent's reader gets — no UI beside it, no schema browser, nothing else that lists the tools. A client picks from the table and reads the prose under it to know how to interpret what comes back.

So a tool registered but undocumented is invisible to the client that would have chosen it, and a caveat that reaches only the response is one an agent has already acted without by the time it reads it.

## Three places, three readers

| File | What it gets |
|---|---|
| `AGENTS.md` | The summary bullet, in the "touches N places on purpose" form the Startup requirements section already uses |
| `CONVENTIONS.md` | A subsection under Agent-facing tools with the reasoning |
| `backend/ttnn_visualizer/agent/README.md` | One sentence on the `_tool_table` line, where someone adding a tool is actually looking |

## What is and is not out of step today

All nine registered tools already appear in the table, so nothing is *missing*. That is not the same as the page being right: #2034 is a caveat on this very page that is wrong today — `agent-tools.md` tells an agent `memory_profile`'s peak answers the out-of-memory question, and it does not. An earlier draft of this PR said "nothing is out of step", which was true about presence and misleading about accuracy.

So this codifies a rule the page does not yet fully satisfy, and the rule is accuracy rather than presence. #2034 is the work of making it true.

## This asks; it does not enforce

The parity test that would enforce it is in #2035's branch (#2037), not here, so this PR adds no test. That is deliberate: duplicating `test_agent_tools_docs_parity.py` across both branches would conflict on merge.

I checked the two are compatible rather than assuming it — ran #2037's parity test against this branch's `AGENTS.md` and all four cases pass, so the merge order does not matter. Full backend suite green here too (1485).

One follow-up for whichever lands second: #2037's `test_agents_md_requires_new_tools_to_be_documented` currently asserts only that the Agent-facing tools section mentions the page. Once both are on `dev` it could assert the obligation wording itself, and the AGENTS.md bullet could gain the "(a parity test enforces it)" parenthetical that the Startup requirements bullet carries. I left both out rather than claim enforcement that is not on `dev` yet — same reason the references to #2035 rendering the page are written as pending rather than done.

Closes #2036


smountenay-tt (commented):
Docs-only, and the rule it codifies is the right one: registration and documentation are one change, and the three placements match the maintenance contract at `CONVENTIONS.md:11` — one-liner in `AGENTS.md`, detail in `CONVENTIONS.md`, plus the line someone adding a tool actually reads. I verified the parity claim rather than taking it: `_tool_table` and the page's table are an exact nine-name set match. I also ran #2037's `test_agent_tools_docs_parity.py` against this branch — all four cases pass, so merge order really is free. One wording fix before merge.

## Should fix

**`CONVENTIONS.md:1137`** — "#2034 is a tool whose own docstring describes behaviour it does not have" understates the precedent, and understates it in the direction that weakens the rule.

#2034 names `docs/src/agent-tools.md:43` and `:80-84` as carrying the same claim. On `dev` today that prose sits at `docs/src/agent-tools.md:107-108`:

> the largest of them is that type's peak — which is the question an out-of-memory failure asks

That is exactly the sentence #2034 says is false: circular buffers and intra-op intermediates are absent from the `buffers` table, measured 2.4x out on `resnet50_may08_1841 2`.

So the page this section makes canonical currently carries a **wrong** caveat, not a missing one. As written, a reader takes the defect to be confined to `server.py` and the page to be trustworthy today — which the PR body's "nothing is out of step" reinforces. Naming it also makes the rule cover more ground: the obligation is an accurate row and caveat, not merely a present one.

Suggested:

> And the surface has drifted here before: #2034 is a caveat that is wrong on the page itself — `agent-tools.md` tells an agent that `memory_profile`'s peak answers the out-of-memory question, which it does not.

## Nit

**`AGENTS.md:158`** — trailing bare `#2036`. The other bullets that cite an issue at the end of a line close with `See #NNNN.` (`:191`, `:192`, `:249`).

## Checked, and fine

- `_tool_table` is in `server.py` as stated.
- "Two places" is accurate — `test_agent_tools.py:127` asserts a subset, not a pinned name set, so adding a tool does not touch it.
- The README's `../../../docs/src/agent-tools.md` resolves to the repository root.
- The new `###` needs no TOC entry; `CONVENTIONS.md`'s table of contents lists `##` only.
- Existing docs-parity suites stay green on this branch (23 passed).

## Follow-up (yours, already flagged)

Once #2037 is on `dev`, the `AGENTS.md` bullet can take the "(a parity test enforces it)" parenthetical that the Startup requirements bullet carries. Worth not losing: until then the wording asks rather than enforces, which is the caveat #2036 itself closes with.

smountenay-tt (changes requested):

smountenay-tt (approved):

aidemsined:
Thanks Scott — you're right, and the understatement was doing more damage than I realised.

Fixed in 593b015e. I verified the claim rather than taking it: #2034 names `docs/src/agent-tools.md:43` and `:80-84`, and on `dev` that prose is at 107-108 and still says what you quote.

I took your sentence and added the mechanism, so a reader does not have to open #2034 to see why it is false:

> And the surface has drifted here before: #2034 is a caveat that is *wrong on the page itself* — `agent-tools.md` tells an agent that `memory_profile`'s peak answers the out-of-memory question, which it does not, because the table it reads holds neither circular buffers nor tensors allocated and freed inside one operation.

Your second point is the one I would have missed on my own — that understating the precedent understates the *rule*. If the example is a missing row, the obligation reads as presence. So that is now explicit:

> That sets the bar. What is owed is an accurate row and an accurate caveat, not merely a present one — a missing tool is one an agent never calls, while a wrong caveat is one it acts on.

And you were right that the PR body reinforced it. "Nothing is out of step today" was true about presence and misleading about accuracy, on a page that carries a wrong caveat right now. I have rewritten that section to separate the two and to say this codifies a rule the page does not yet fully satisfy.

Nit taken too — `See #2036.`, which leaves no bare trailing citations in AGENTS.md.

On the follow-up: agreed, and it is noted on both PRs. Once #2037 is on `dev` the parenthetical goes in and #2037's parity test can assert the obligation's wording instead of just the reference.

PR #2041, opened 2026-09-22
Fix globally-allocated CBs indicator tooltip

### LINKED ISSUE #2032: Globally-allocated CB icon has no tooltip in the Device Operations legend
## Problem

In the **CBs** legend in Operation Details → Device Operations, the link icon that marks a globally-allocated circular buffer has no tooltip. There is nothing on hover and no accessible name, so the icon is unexplained — a reader has to already know what it means.

It marks a CB that is a kernel-side view onto an existing L1 tensor rather than a fresh allocation (`globally_allocated: '1'`, #1651). That is exactly the thing worth saying on hover, because it is also why those rows do not add to the CB total.

## Ask

Give the icon a tooltip and an accessible name. Wording along the lines of "View onto an existing L1 tensor — not a separate allocation".

Not investigated — filed from a screenshot while looking at something else.

aidemsined:
image

The link icon marking a globally-allocated circular buffer had two problems, both in Operation Details.

**The tooltip did not open in the Device Operations CBs list**, and the pointer flickered over the icon. The same icon in the L1 legend at the top of the page was fine.

Device Operations places the marker in a reserved slot on the right of the size column. That rule positioned the icon itself, and because the icon is the only child of the wrapper Blueprint measures to place a tooltip, the wrapper collapsed to nothing — the tooltip was anchored to a zero-sized point twelve pixels from the icon. The slot rule now moves the tooltip's anchor instead, so the anchor and the icon stay together. The row's layout is unchanged.

**Clicking one of those rows appeared to do nothing.** In fact it was worse: the row faded itself out while a toast announced it was selected.

A legend row takes its colour from the tensor found at its address when the row itself carries no tensor id — which is how Device Operations builds every CB row. The click did not do the same, so the highlight was looking for a colour derived from the address while the row was painted from the tensor, and never matched. On operation 13 of `resnet50_may08_1841 2` that was 16 of the 20 globally-allocated rows. Clicking now highlights the tensor everywhere it appears, and the toast names it rather than saying "Buffer".

The tooltip text is unchanged. Its wording and the screen-reader label are now built from one string rather than two that could disagree.

## How to check it

1. Load a report with globally-allocated CBs — `resnet50_may08_1841 2`, `resnet50_jul28_1759` or `sentence_bert_aug19_1726` — and open an operation that has them (13 on the resnet50 captures).
2. Turn on **Show Circular Buffers Details**.
3. Hover a link icon in the **CBs** list under Device Operations: the tooltip should open and stay open.
4. Click that row: it should highlight, the matching rows elsewhere should highlight with it, and the toast should read `Tensor N at <address> selected`.

## Follow-up, not in this PR

`.outside-click`, the invisible click-catcher that dismisses a focused buffer, is positioned wrongly: it has no `left`, so it starts at the navigation rail's width and hangs the same distance off the right edge. It covers these rows and accepts pointer events. It did not intercept anything in the window sizes I tried, so I have left it alone, but it is a hazard worth closing separately.

Closes #2032


smountenay-tt (commented):
Frontend-only change (component + SCSS + spec, no API contract touched), so this is a bugs/security/release-blockers pass rather than a style one. **No blockers.** All three fixes hold up against the surrounding code:

- **The anchor fix is real, not incidental.** `Tooltip` spreads `className` through `restProps` into `Popover`, which applies it to the target wrapper alongside `bp6-popover-target` (`@blueprintjs/core@6.6.1`, `popover.js:214`) — so `.globally-allocated-anchor` moves the element Popper actually measures. Nothing else in `src/scss` positions `.globally-allocated-marker`, and the containing block is still `.format-numbers:nth-child(3)`, so the "row layout is unchanged" claim checks out.
- **The click fix is correct, and correct to be ungated.** `resolvedColour` falls back to `getTensorColor(derivedTensor.id)` for *any* row carrying no `tensorId`, not only aliased ones, so the mismatch fired wherever that fallback did. `chunk.tensorId ?? derivedTensor?.id` restores the invariant `isMatchingBufferColour` depends on, and since `getTensorColor` is a per-id cache both sides now resolve to the identical string.
- **Blast radius is contained.** `selectedBufferColourAtom` is read only in `MemoryLegendElement`, so the one further effect is `selectedTensorId` now being set for these rows — which is what makes the tensor highlight in the tensor list and the toast name it, and is what every other click path (`onTensorClick`, the plot handlers) already did.

### Optional, non-blocking

`src/components/operation-details/MemoryLegendElement.tsx:205` — the shared string drops the "Globally allocated" prefix from the `aria-label`, and because the label sits on the same span as the sr-only `.globally-allocated-marker-label`, it *replaces* that text in the button's accessible name rather than adding to it (same as before the change — this isn't new behaviour, just newly shorter text). Net effect: AT hears "Aliased to Tensor 188 …" where it previously heard the concept name too. If you want the tooltip short and the label complete from one source:

```tsx
aria-label={`Globally allocated — ${globallyAllocatedDescription}`}
```

Take it or leave it — the tooltip wording itself reads better than what was there.

smountenay-tt (approved):

